PR #27, opened 2021-10-29
Optimize the way of executing io

DragonDriver:
In this pr, I changed the way of executing io to what [fio](https://github.com/axboe/fio/blob/master/engines/libaio.c) does.

The basic design:
- an io queue specified by the io_context_t has a `q-depth`(`MAX_EVENTS`);
- we should always keep the queue full to increase the IOPS;
- when the queue is not full, we can continously submit io-request to the queue;
- when there are any io events being done, we can use `io_getevents` to get the number (`r_done`);
- the number of events which we can submit to the queue next time is equal to the `r_done`.

Signed-off-by: dragondriver <jiquan.long@zilliz.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -356,3 +356,5 @@ MigrationBackup/
 /.vs
 /out/build/x64-Debug
 cscope*
+
+*.out
--- a/src/linux_aligned_file_reader.cpp
+++ b/src/linux_aligned_file_reader.cpp
@@ -14,6 +14,12 @@ namespace {
   typedef struct io_event io_event_t;
   typedef struct iocb     iocb_t;
 
+  // execute_io read file using the way fio does,
+  // iops tested by fio with num-jobs=4 can reach the peak,
+  // we can also create 4 io_context_t, and then parallel the execute_io.
+  // Note:
+  //    1, file should be opened with `O_DIRECT` flag;
+  //    2, the bufs in read_reqs must be aligned to 512 bytes.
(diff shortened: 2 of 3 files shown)


harsha-simhadri:
> In this pr, I changed the way of executing io to what [fio](https://github.com/axboe/fio/blob/master/engines/libaio.c) does.
> 
> The basic design:
> 
> * an io queue specified by the io_context_t has a `q-depth`(`MAX_EVENTS`);
> * we should always keep the queue full to increase the IOPS;
> * when the queue is not full, we can continously submit io-request to the queue;
> * when there are any io events being done, we can use `io_getevents` to get the number (`r_done`);
> * the number of events which we can submit to the queue next time is equal to the `r_done`.
> 
> Signed-off-by: dragondriver [jiquan.long@zilliz.com](mailto:jiquan.long@zilliz.com)

@DragonDriver  Thanks a lot for the contribution. Did you measure the performance impact of this change?

DragonDriver:
@harsha-simhadri 
hi, i've added a test for `execute_io`. But I don't know if I should broke the function structure of linux file reader. So I will remove this test file later.
What a shame is that now I don't have the NVME disk mounted on my local pc. But I'm sure `execute_io` now with 4 threads can reach the peak of IOPS. Can you try this in a env with NVME?

DragonDriver:
Need more evidence to prove that this pr can improve the performance of io.
So close this now.

PR #37, opened 2022-02-25
Support Ubuntu 20.04 & Initial CI

daxpryce:
This commit adds a Github actions workflow to continuously test that building is successful on Ubuntu 20.04 and Ubuntu 18.04 on every push.

Changes made to CMakeList.txt now support Intel MKL's header file
location installation from apt (on Ubuntu 20.04 and, presumably, later
versions), as well as manual installs using today's current Intel MKL
download and installation through their installer.

However, today's Intel MKL library may not be what you want, and it may
be that the path will change if we use an older version of the library,
but I didn't want to dig that up while I was working on this. 

I also don't know the first thing about building for Windows, so I didn't even try


CMakeLists.txt
```diff
@@ -44,26 +44,33 @@ function(checkEnvAndSetLocalVar env_var msg local_var)
...
+include_directories(include)
 #MKL Config
 if (MSVC)
 	checkEnvAndSetLocalVar("INTEL_ROOT" "Please install Intel MKL libraries and set the env variable INTEL_ROOT to the intel software directory. Should be similar to: C:\\Program Files (x86)\\IntelSWTools\\compilers_and_libraries\\windows\\. " "INTEL_ROOT")
 	set(MKL_ROOT ${INTEL_ROOT}/mkl)
 	add_compile_options(/arch:AVX2 /Qpar)
 	link_libraries("${INTEL_ROOT}/mkl/lib/intel64/mkl_core_dll.lib" "${INTEL_ROOT}/mkl/lib/intel64/mkl_rt.lib"  "${INTEL_ROOT}/mkl/lib/intel64/mkl_intel_thread_dll.lib" "${INTEL_ROOT}/compiler/lib/intel64/libiomp5md.lib"  "${INTEL_ROOT}/mkl/lib/intel64/mkl_intel_ilp64_dll.lib" "${INTEL_ROOT}/mkl/lib/intel64/mkl_sequential_dll.lib")
 	checkEnvAndSetLocalVar("BOOST_ROOT" "Please install Boost (1.71 or greater) from www.boost.org and set the env var BOOST_ROOT to the boost directory." "BOOST_ROOT")
-
+	include_directories(${INTEL_ROOT}/include)
```
daxpryce, line 55: I'm not sure how this got here, I'll remove it

PR #42, opened 2022-03-08
makeing cmake and CI built test for Ubuntu 18 consistent

harsha-simhadri:


CMakeLists.txt
```diff
@@ -56,8 +56,9 @@ if (MSVC)
 	include_directories(${MKL_ROOT}/include)
 else()
 	# expected path for manual intel mkl installs
-	set(MKL_ROOT /opt/intel/compilers_and_libraries/mkl/latest CACHE PATH "Intel mkl libraries and includes root path")
-	link_directories(${MKL_ROOT}/lib/intel64)
+	set(OMP_PATH /opt/intel/oneapi/compiler/2022.0.2/linux/compiler/lib/intel64_lin/)
```
daxpryce, line 59: so the CACHE part lets people actually override this with their own install location using -Dflags later.  I'd consider keeping the CACHE PATH stuff for that reason (and I guess the PATH bit lets the cmake GUI tools give them a file explorer dropdown type deal to find and specify it, for those who use GUI tools for whatever reason)

harsha-simhadri: @daxpryce  Thanks for pointing this out. Hope this PR addresses the problem: https://github.com/microsoft/DiskANN/pull/47

PR #55, opened 2022-04-25
Merging index.cpp from kannexpts

harsha-simhadri:


rakri:
Changes here have been incorporated using a different (and better) PR #64

PR #57, opened 2022-04-28
made the changes to get disk index to work for static index. 

rakri:
streaming SSD is pending.

### DIFF
--- a/include/pq_flash_index.h
+++ b/include/pq_flash_index.h
@@ -20,7 +20,7 @@
 
 #define MAX_GRAPH_DEGREE 512
 #define MAX_N_CMPS 16384
-#define SECTOR_LEN 4096
+#define SECTOR_LEN (_u64) 4096
 #define MAX_N_SECTOR_READS 128
 #define MAX_PQ_CHUNKS 256
 
@@ -86,6 +86,8 @@ namespace diskann {
 
     DISKANN_DLLEXPORT void load_cache_list(std::vector<uint32_t> &node_list);
 
+    DISKANN_DLLEXPORT _u64 return_nd();
+
 #ifdef EXEC_ENV_OLS
     DISKANN_DLLEXPORT void generate_cache_list_from_sample_queries(
         MemoryMappedFiles &files, std::string sample_bin, _u64 l_search,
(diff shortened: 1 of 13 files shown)


include/pq_flash_index.h
```diff
@@ -168,9 +172,10 @@ namespace diskann {
 
     // for very large datasets: we use PQ even for the disk resident index
     bool              use_disk_index_pq = false;
-    _u64              disk_pq_n_chunks;
+    _u64              disk_pq_n_chunks = 0;
```
harsha-simhadri, line 175: do we want to move this initialization to the constructor?

include/pq_table.h
```diff
@@ -5,6 +5,8 @@
 
 #include "utils.h"
 
+#define NUM_PQ_CENTROIDS 256
+
```
harsha-simhadri, line 9: Is this an upper bound on the number of PQ centroids? Then should we name it MAX_PQ_CENTROIDS?

src/aux_utils.cpp
```diff
@@ -425,7 +425,8 @@ namespace diskann {
             } */
     }
 
-    size_t merged_index_size = 16;
+    size_t merged_index_size = 24;
```
harsha-simhadri, line 428: What are these hard coded numbers?

src/aux_utils.cpp
```diff
@@ -425,7 +425,8 @@ namespace diskann {
...
+    size_t merged_index_frozen = 0;
     // create cached vamana writers
     cached_ofstream diskann_writer(output_vamana, 1024 * 1048576);
```
harsha-simhadri, line 431: move these numbers to macros?

src/aux_utils.cpp
```diff
@@ -777,19 +793,25 @@ namespace diskann {
         assert(nnbrs <= width_u32);
 
         // read node's nhood
-        vamana_reader.read((char *) nhood_buf, nnbrs * sizeof(unsigned));
+                vamana_reader.read((char *) nhood_buf,
```
harsha-simhadri, line 796: clang-format here?

PR #71, opened 2022-06-15
Enable building and Git workflows on Windows.

theantony:
Rework CMakefiles for Windows and use NuGet to download dependencies such as
MKL and optionally Boost.


CMakeLists.txt
```diff
@@ -76,24 +147,77 @@ endif()
...
+endif()
+
+if (NOT MSVC)
+    set(DISKANN_ASYNC_LIB aio)
+endif()
+
 #Main compiler/linker settings 
 if(MSVC)
 	#language options
-	add_compile_options(/permissive- /openmp:experimental /Zc:wchar_t /Zc:twoPhase- /Zc:forScope /Zc:inline /WX- /std:c++14 /Gd /W3 /MP /Zi /FC /nologo /diagnostics:classic)
```
theantony, line 82: Most of the removed options were either the defaults or not applicable. The only one worth calling out is /Qpar which auto-parallelizes some for loops. The Linux version doesn't do this and I think this is undesirable as we cannot control/know about the number of threads used by the compiler.

For CMAKE_EXECUTABLE_CXX_LINK_FLAGS, DiskANN is the only thing on the Internet that refers to that (IOW, the option doesn't exist) :)

include/pq_flash_index.h
```diff
@@ -78,8 +78,6 @@ namespace diskann {
 
     DISKANN_DLLEXPORT void load_cache_list(std::vector<uint32_t> &node_list);
 
-    DISKANN_DLLEXPORT _u64 return_nd();
```
gopalrs, line 81: Why is return_nd() removed?

include/windows_aligned_file_reader.h
```diff
@@ -30,18 +30,22 @@ class WindowsAlignedFileReader : public AlignedFileReader {
...
-  DISKANN_DLLEXPORT virtual void open(const std::string &fname);
-  DISKANN_DLLEXPORT virtual void close();
+  DISKANN_DLLEXPORT virtual void open(const std::string &fname) override;
+  DISKANN_DLLEXPORT virtual void close() override;
 
-  DISKANN_DLLEXPORT virtual void register_thread();
-  DISKANN_DLLEXPORT virtual void deregister_thread() {
+  DISKANN_DLLEXPORT virtual void register_thread() override;
+  DISKANN_DLLEXPORT virtual void deregister_thread() override {
+      // TODO: Needs implementation.
```
gopalrs, line 38: There should be a windows_aligned_file_reader.cpp that has the implementations?

include/windows_customizations.h
```diff
@@ -4,9 +4,13 @@
 #pragma once
 
 #ifdef _WINDOWS
+
+#ifdef _WINDLL
 #define DISKANN_DLLEXPORT __declspec(dllexport)
-#define DISKANN_DLLIMPORT __declspec(dllimport)
+#else
+#define DISKANN_DLLEXPORT __declspec(dllimport)
```
gopalrs, line 11: Why are we defining DISKANN_DLLEXPORT to be (dllimport)?

gopalrs (commented):
Thanks for doing this.  This is wonderful to see.

include/pq_flash_index.h
theantony: It's just a declaration without definition, AFAIR Intellisense issued a warning to that effect.

include/windows_aligned_file_reader.h
theantony: It may be in some other branch, it wasn't here.

include/windows_customizations.h
theantony: When compiling the DLL itself, it's declared as `dllexport` to make it part of the DLL interface.

When it's being used by DLL clients, it's declared as `dllimport` so that the client can access the data reference in the DLL (for data access) and that a compiler can emit a slightly more efficient function calling code (for method invocation).

Ideally, this would be called either `DISKANN_DLL_DECL` or `DISKANN_DLLEXPORT_IMPORT` or similar. Then it wouldn't be confusing why is `DISKANN_DLLEXPORT` defined as `dllimport`, but I didn't want to make too many changes with this PR, it can be done separately if needed.

(The MSDN documentation on this is [here](https://docs.microsoft.com/en-us/cpp/cpp/dllexport-dllimport) BTW)

PR #76, opened 2022-06-30
Introduce a set tailored for storing available index locations

theantony:
so that reserve_location (which runs under a lock) can be executed very quickly.

Before the change, calling enable_delete would populate the _empty_slots
robin_set with all available locations. This would cause insert_point
to slow to a crawl (nearly order of magnitude slower than expected) since
reserve_location would run relatively long to find an available slot
while holding a lock.

With the change, reserve_location gets the location almost immediately,
restoring the expected speed.

### DIFF
--- a/src/natural_number_set.cpp
+++ b/src/natural_number_set.cpp
@@ -1,0 +1,78 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+#include <boost/dynamic_bitset.hpp>
+
+#include "ann_exception.h"
+#include "natural_number_set.h"
+
+namespace diskann {
+  template<typename T>
+  natural_number_set<T>::natural_number_set()
+      : _values_bitset(std::make_unique<boost::dynamic_bitset<>>())
+  {
+  }
+
+  template<typename T>
+  bool natural_number_set<T>::is_empty() const
+  {
+      return _values_vector.empty();
+  }
+
+  template<typename T>
+  void natural_number_set<T>::reserve(size_t count)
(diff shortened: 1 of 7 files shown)


src/natural_number_set.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+  template<typename T>
+  void natural_number_set<T>::reserve(size_t count)
+  {
+      _values_vector.reserve(count);
+      _values_bitset->reserve(count);
+  }
+
+  template<typename T>
+  void natural_number_set<T>::insert(T id)
```
harsha-simhadri, line 30: Is this data structure thread safe? Can multiple threads insert and pop safely at the same time?

src/natural_number_set.cpp
```diff
@@ -0,0 +1,78 @@
...
+  {
+      _values_vector.emplace_back(id);
+
+      if (id >= _values_bitset->size())
+        _values_bitset->resize(static_cast<size_t>(id) + 1);
+
+      _values_bitset->set(id, true);
```
harsha-simhadri, line 37: should we do CAS or is a simple read/write suffucient?

src/natural_number_set.cpp
theantony: It's not, same as the `robin_set` it replaced for `_empty_slots`. The general rule for `std` containers is that multiple read-only operations are OK, but there can be only one writer at one point.

I'll add an explicit comment about thread-safety.

theantony:
Synced offline with Harsha, it's OK to merge.

PR #78, opened 2022-07-12
Uncaught Exceptions

daxpryce:
# This isn't the actual fix for our uncaught exceptions

But I assert it's a reasonable fallback if something exceptional happens.

-----

For some reason the current catch block in our handle_post method isn't firing when parseJson throws an std::invalid_argument exception.

We _do_ still need to properly fix it so that that known case can be handled and we can give a better error message, but having a catch-all that doesn't instantly
terminate the process when an unexpected error happens in the handle_post seems reasonable too. It may mean that we don't have a good "die due to a critical,
unrecoverable error" mechanism, but it also means a posted request off the happy path doesn't terminate our server too.


harsha-simhadri (approved):
OK with this temporary fix. Lets later investigate why exceptions are not being caught properly.

PR #83, opened 2022-07-25
updated unsigned int to _s64 for windows compilation

NeelamMahapatro:
Change in the pragma loop to include _s64 instead of unsigned as the unsigned int gives compilation error in windows


NeelamMahapatro:
duplicate PR

PR #87, opened 2022-08-02
Origin/add sector reordering

NeelamMahapatro:
modified disk layout for sector reordering.

1. The lorder and porder files are loaded into memory using load_bin function.
2. reordered the sector based on the new ordering.
3. New neighbors are modified based on the reordering.
4. Does medoid need to be changed in the metadata? **[Please check]** - 848 to 852 lines

### DIFF
--- a/src/aux_utils.cpp
+++ b/src/aux_utils.cpp
@@ -702,7 +702,9 @@ namespace diskann {
   void create_disk_layout(const std::string base_file,
                           const std::string mem_index_file,
                           const std::string output_file,
-                          const std::string reorder_data_file) {
+                          const std::string reorder_data_file,
+                          const std::string lorder_file,
+                          const std::string porder_file) {
     unsigned npts, ndims;
 
     // amount to read or write in one shot
@@ -716,6 +718,8 @@ namespace diskann {
     npts_64 = npts;
     ndims_64 = ndims;
 
+    // Check if we need to the sector re-ordering
+    bool reorder_data = false;
     // Check if we need to append data for re-ordering
     bool          append_reorder_data = false;
     std::ifstream reorder_data_reader;
@@ -739,7 +743,7 @@ namespace diskann {
         if (reorder_data_file_size != 8 + sizeof(float) *
                                               (size_t) npts_reorder_file *
                                               (size_t) ndims_reorder_file)
-          throw ANNException("Discrepancy in reorder data file size ", -1,
+          throw diskann::ANNException("Discrepancy in reorder data file size ", -1,
                              __FUNCSIG__, __FILE__, __LINE__);
       } catch (std::system_error &e) {
         throw FileException(reorder_data_file, e, __FUNCSIG__, __FILE__,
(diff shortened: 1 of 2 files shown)


src/aux_utils.cpp
```diff
@@ -739,15 +743,42 @@ namespace diskann {
...
         throw FileException(reorder_data_file, e, __FUNCSIG__, __FILE__,
                             __LINE__);
       }
     }
 
-    // create cached reader + writer
+    //lorder -> Location to point/Id file
+    //porder -> Id to location file
+    unsigned npts_lorder, ndims_lorder, npts_porder, ndims_porder;
+    std::unique_ptr<char[]> lorder_data = std::make_unique<char[]>(npts_64);
```
rakri, line 757: for load_bin, we dont need to initialize the array. the function will do it.. so just keep it as unique_ptr

src/aux_utils.cpp
```diff
@@ -739,15 +743,42 @@ namespace diskann {
...
+    std::unique_ptr<char[]> porder_data = std::make_unique<char[]>(npts_64);
+    if(lorder_file != std::string("") && porder_file != std::string("")){
+      reorder_data = true;
+      try {
+        diskann::load_bin<T>(lorder_file, lorder_data, npts_lorder, ndims_lorder);
```
rakri, line 762: T for this load_bin should be uint32_t as we need to load a binary file containing uint32_t data. same for next load_bin also

src/aux_utils.cpp
```diff
@@ -808,12 +839,17 @@ namespace diskann {
...
     _u64 disk_index_file_size =
-        (n_sectors + n_reorder_sectors + 1) * SECTOR_LEN;
+        (n_sectors + n_reorder_sectors + 1) * SECTOR_LEN;   
 
     // write first sector with metadata
     *(_u64 *) (sector_buf.get() + 0 * sizeof(_u64)) = disk_index_file_size;
     *(_u64 *) (sector_buf.get() + 1 * sizeof(_u64)) = npts_64;
     *(_u64 *) (sector_buf.get() + 2 * sizeof(_u64)) = medoid;
+    /*if(reorder_data){
+      *(_u64 *) (sector_buf.get() + 2 * sizeof(_u64)) = medoid;
```
rakri, line 849: excellent point, we need to enter the reordered medoid here.. why is this commented?

src/aux_utils.cpp
```diff
@@ -828,6 +864,15 @@ namespace diskann {
 
     diskann_writer.write(sector_buf.get(), SECTOR_LEN);
 
+    
+    unsigned  dummyUnsigned, loc_id;
+    std::unique_ptr<char[]> dummy_buf = std::make_unique<char[]>(max_node_len);
```
rakri, line 869: if this dummy_buf is only there for moving the pointer, we can replace this with seekg in the loop, no?

src/aux_utils.cpp
```diff
@@ -828,6 +864,15 @@ namespace diskann {
...
+    for (_u64 i = 0; i < npts_64; i++) {
+      pos_vamana_reader[i] = vamana_reader.tellg();
+      vamana_reader.read((char *) &dummyUnsigned, sizeof(unsigned));
```
rakri, line 872: replace 872 and 873 with a seekg to advance

src/aux_utils.cpp
```diff
@@ -837,36 +882,54 @@ namespace diskann {
...
-           sector_node_id++) {
+          sector_node_id < nnodes_per_sector && cur_node_id < npts_64;
+          sector_node_id++) {
         memset(node_buf.get(), 0, max_node_len);
-        // read cur node's nnbrs
+        /* read cur node's nnbrs.
+        if sector reordering is carried out, vamana reader need 
+        to point to the particular location*/
+        if(reorder_data){
+          loc_id = *(unsigned *) (lorder_data.get() + cur_node_id * sizeof(unsigned));
```
rakri, line 892: i think you can use lorder_data[cur_node_id] also, should work

src/aux_utils.cpp
```diff
@@ -837,36 +882,54 @@ namespace diskann {
...
-        vamana_reader.read((char *) nhood_buf,
-                           (std::min)(nnbrs, width_u32) * sizeof(unsigned));
+        vamana_reader.read((char *) nhood_buf, (std::min)(nnbrs, width_u32) * sizeof(unsigned));
+        
+        //Fetch corresponding nhood from new list after reordering
+        if(reorder_data){
+          unsigned old_neighbor, new_neighbor;
+          for(unsigned i=0; i< (unsigned)(std::min)(nnbrs, width_u32); i++ ){
+            old_neighbor = *(unsigned *) (nhood_buf + i);
+            new_neighbor = *(unsigned *) (porder_data.get() +  old_neighbor * sizeof(unsigned));
```
rakri, line 909: same, you can use porder_data[blah]

rakri (changes requested):

src/aux_utils.cpp
NeelamMahapatro: added unsigned pointer and passed it to the load_bin function

src/aux_utils.cpp
NeelamMahapatro: Commented this code earlier to verify. Now I have added the reordered medoid.

src/aux_utils.cpp
NeelamMahapatro: dummy_buf variable has been removed as it is not necessary.

PR #98, opened 2022-08-13
Add filter support for dynamic indices

darvg:


src/index.cpp
```diff
@@ -3017,6 +3027,10 @@ namespace diskann {
 
       _tag_to_location[tag] = location;
       _location_to_tag[location] = tag;
+      _pts_to_labels[location] = labels;
+    } else if (_filtered_index){
```
darvg, line 3031: put this if inside tags if

src/index.cpp
```diff
@@ -3041,10 +3055,17 @@ namespace diskann {
       pool.clear();
       visited.clear();
       std::vector<unsigned> init_ids;
-      get_expanded_nodes(location, Lindex, init_ids, pool, visited,
-                         scratch.des(), scratch.best_l_nodes(),
-                         scratch.inserted_into_pool_rs(),
-                         scratch.inserted_into_pool_bs());
+      if (_filtered_index){
```
darvg, line 3058: need to support filtered/unfiltered search in a single index

PR #105, opened 2022-08-16
Harshasi/unify pq files into one file rebased

harsha-simhadri:


rakri (commented):
Looks ok to me code-wise. Will run some offline tests and update.

CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ if (MSVC)
 
     message(STATUS "Invoking nuget to download Boost, OpenMP and MKL dependencies...")
     configure_file(${PROJECT_SOURCE_DIR}/windows/packages.config.in ${DISKANN_MSVC_PACKAGES_CONFIG})
-    exec_program(${NUGET_EXE} ARGS install ${DISKANN_MSVC_PACKAGES_CONFIG} -ExcludeVersion -OutputDirectory ${DISKANN_MSVC_PACKAGES})
+    exec_program(${NUGET_EXE} ARGS install \"${DISKANN_MSVC_PACKAGES_CONFIG}\" -ExcludeVersion -OutputDirectory \"${DISKANN_MSVC_PACKAGES}\")
```
harsha-simhadri, line 45: What is the reason for this change?

rakri: if the build folder has spaces like "ANNS Project" it the nuget will treat it as separate args instead of one folder argument if we didn't have quotes

rakri: learned this the hard way :)

rakri (approved):
Looks good to me.

PR #110, opened 2022-08-27
logger abstraction and few code fixes for making diskann.dll buildable with EXEC_ENV_OLS turned on

ltan1ms:
# Changes
All changes only effects when "EXEC_ENV_OLS" is defined:
- Expose a `SetCustomLogger `function to pass in custom logging function so we can invoke it, this replaced the old undefined `DISKANNLOGGING `micro and removed dependency on some unnecessary headers, like `IANNIndex.h, ANNLogging.h` etc.
- Start with supporting log level `Info` and `Error `only as that's what `diskann::cout `or `diskan::cerr` supports.
- Added missing declarations for overloads that takes `AlighedFileReader` in `index.h`
- Added missing implementations for following function overloads that takes `AlighedFileReader` in `utils.cpp `(copied from DLSV repo): `get_bin_metadata, load_bin, copy_aligned_data_from_file, read_array, read_value`

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -746,7 +746,7 @@ namespace diskann {
 #ifdef EXEC_ENV_OLS
     _u32 nodes_read = 0;
     _u64 cc = 0;
-    _u64 graph_offset = file_offset + header_size;
+    _u64 graph_offset = header_size;
     while (nodes_read < expected_num_points) {
       _u32 k;
       read_value(reader, k, graph_offset);
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -418,7 +418,7 @@ namespace diskann {
       compact_data();
       compact_frozen_point();
     } else {
-      if (not _data_compacted) {
+      if (!_data_compacted) {
         throw ANNException(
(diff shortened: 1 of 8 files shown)


src/index.cpp
```diff
@@ -746,7 +746,7 @@ namespace diskann {
 #ifdef EXEC_ENV_OLS
     _u32 nodes_read = 0;
     _u64 cc = 0;
-    _u64 graph_offset = file_offset + header_size;
```
ltan1ms, line 749: this `file_offset `was undefined, and causes build error. I guess graph_offset starts right after header so removed it.

gopalrs (approved):

src/index.cpp
gopalrs: The file offset parameter only makes sense when we have a single file index. This won't apply here, so the change is good.

harsha-simhadri:
@ltanms1 can we close this PR and work on 114 instead?

ltan1ms:
> @ltanms1 can we close this PR and work on 114 instead?

Yes, let me close this RP.

PR #114, opened 2022-09-19
logger abstraction and few code fixes for making diskann.dll buildabl…

ltan1ms:
The changes were ported from the previous [PR ](https://github.com/microsoft/DiskANN/pull/110) from my forked repo.
But send this new PR after I got permissions to push to this repo directly, and will discard the previous one.

# Changes
All changes only effects when "EXEC_ENV_OLS" is defined:
- Expose a `SetCustomLogger `function to pass in custom logging function so we can invoke it, this replaced the old undefined `DISKANNLOGGING `micro and removed dependency on some unnecessary headers, like `IANNIndex.h, ANNLogging.h` etc.
- Start with supporting log level `Info` and `Error `only as that's what `diskann::cout `or `diskan::cerr` supports.
- Added missing declarations for overloads that takes `AlighedFileReader` in `index.h`
- Added missing implementations for following function overloads that takes `AlighedFileReader` in `utils.cpp `(copied from DLSV repo): `get_bin_metadata, load_bin, copy_aligned_data_from_file, read_array, read_value`

### DIFF
--- a/include/index.h
+++ b/include/index.h
@@ -14,6 +14,10 @@
 #include "tsl/robin_map.h"
 #include "tsl/sparse_map.h"
 
+#ifdef EXEC_ENV_OLS
+#include "aligned_file_reader.h"
+#endif
+
 #include "boost_dynamic_bitset_fwd.h"
 #include "distance.h"
 #include "locking.h"
@@ -136,8 +140,13 @@ namespace diskann {
                                 bool        compact_before_save = false);
 
     // Load functions
+#ifdef EXEC_ENV_OLS
+    DISKANN_DLLEXPORT void load(AlignedFileReader &reader, uint32_t num_threads,
+                                uint32_t search_l);
+#else
     DISKANN_DLLEXPORT void load(const char *index_file, uint32_t num_threads,
                                 uint32_t search_l);
+#endif
 
     // get some private variables
     DISKANN_DLLEXPORT size_t get_num_points();
@@ -365,11 +374,19 @@ namespace diskann {
     DISKANN_DLLEXPORT _u64   save_data(std::string filename);
     DISKANN_DLLEXPORT _u64   save_tags(std::string filename);
     DISKANN_DLLEXPORT _u64   save_delete_list(const std::string &filename);
+#ifdef EXEC_ENV_OLS
+    DISKANN_DLLEXPORT size_t load_graph(AlignedFileReader &reade,
+                                        size_t             expected_num_points);
+    DISKANN_DLLEXPORT size_t load_data(AlignedFileReader &reader);
+    DISKANN_DLLEXPORT size_t load_tags(AlignedFileReader &reader);
+    DISKANN_DLLEXPORT size_t load_delete_set(AlignedFileReader &reader);
+#else
     DISKANN_DLLEXPORT size_t load_graph(const std::string filename,
                                         size_t            expected_num_points);
     DISKANN_DLLEXPORT size_t load_data(std::string filename0);
     DISKANN_DLLEXPORT size_t load_tags(const std::string tag_file_name);
     DISKANN_DLLEXPORT size_t load_delete_set(const std::string &filename);
+#endif
 
    private:
     // Distance functions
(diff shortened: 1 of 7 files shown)


include/index.h
```diff
@@ -365,11 +374,19 @@ namespace diskann {
     DISKANN_DLLEXPORT _u64   save_data(std::string filename);
     DISKANN_DLLEXPORT _u64   save_tags(std::string filename);
     DISKANN_DLLEXPORT _u64   save_delete_list(const std::string &filename);
+#ifdef EXEC_ENV_OLS
+    DISKANN_DLLEXPORT size_t load_graph(AlignedFileReader &reade,
```
harsha-simhadri, line 378: reader instead or reade?

include/logger_impl.h
```diff
@@ -46,7 +28,7 @@ namespace diskann {
     char*              _buf;
     int                _bufIndex;
     std::mutex         _mutex;
-    ANNIndex::LogLevel _logLevel;
+    LogLevel _logLevel;
```
harsha-simhadri, line 31: please run clang-format on files being checked in, or use VS built-in formatter that recognizes clang-format directives

src/utils.cpp
```diff
@@ -226,4 +230,257 @@ namespace diskann {
                   << std::endl;
   }
 
+ #ifdef EXEC_ENV_OLS
+  void get_bin_metadata(AlignedFileReader& reader, size_t& npts, size_t& ndim,
```
harsha-simhadri, line 234: Do you prefer to replicate most of the load code between EXEC_ENV_OLS and other environments? Or is there a way to share most of the code and avoid replication?

src/utils.cpp
```diff
@@ -226,4 +230,257 @@ namespace diskann {
...
+                                                    float&             value,
+                                                    size_t             offset);
+  template DISKANN_DLLEXPORT void read_value<uint32_t>(
+      AlignedFileReader& reader, uint32_t& value, size_t offset);
+  template DISKANN_DLLEXPORT void read_value<uint64_t>(
+      AlignedFileReader& reader, uint64_t& value, size_t offset);
+
+#endif
+
 }  // namespace diskann
```
harsha-simhadri, line 486: add new line at the end. Clang-format might fix this.

harsha-simhadri (changes requested):
Li, looks good. I suggested some minor changes and formatting files inline. Would be good to get Gopal's review too

src/utils.cpp
gopal-msr: Apart from a few assignments, there won't be much code duplication

gopal-msr (approved):
I had to add the load() functions that take AlignedFileReader in the Index class because of deadline pressure. It might be worth investigating if we can remove them completely and simply use FileBlobs. But this is not necessary.

include/index.h
```diff
@@ -14,6 +14,10 @@
 #include "tsl/robin_map.h"
 #include "tsl/sparse_map.h"
 
+#ifdef EXEC_ENV_OLS
```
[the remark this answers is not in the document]
ltan1ms: Yes, you are right, but it's only needed when EXEC_ENV_OLS is defined, so we can reduce the file size by not including it for normal case (EXEC_ENV_OLS not defined).

include/index.h
ltan1ms: Fixed.

src/utils.cpp
ltan1ms: yes. fixed.

include/logger_impl.h
ltan1ms: yes, i ran it on all the files. The index.cpp have some more lines of changes in formatting.

harsha-simhadri (commented):
In src/utils.cpp, do you expect to maintain the loading and storing of index files for your environment seperate from the rest of the software?

ltan1ms:
> In src/utils.cpp, do you expect to maintain the loading and storing of index files for your environment separate from the rest of the software?

@harsha-simhadri, if i understand correctly, you mean whether we will need to maintain all those overloads that take the AlignedFileReader class for our online serving environment, right? Are these stable/fundamental utility functions that are not expected to change often? If so, the effort for maintaining it should be low.

harsha-simhadri (approved):

PR #123, opened 2022-10-12
move filter  logic before distance compute

Sanhaoji2:


harsha-simhadri:
@Sanhaoji2 Do you have latency/QPS improvement measurements on any public dataset that you could post here for reference? The changes look good to me. Given the importance of this piece of code, I would for @rakri @ltan1ms @hliu18  to comment before mergning.

src/pq_flash_index.cpp
```diff
@@ -157,6 +157,8 @@ namespace diskann {
         diskann::alloc_aligned((void **) &scratch.aligned_query_float,
```
hliu18, line 157: Curious, can the allocations be combined into a single bigger one? It may reduce fragmentation, and also contention (probably not significant though). Not critical for this PR.

harsha-simhadri: @hliu18 What contention would they reduce? One advantage of merging this mallocs and making them continuous could be to reduce TLB cache misses (since all this data would be contiguous)

hliu18: No rush. Each alloc/free operation takes lock (critical section). large number of alloc/free operation can cause lock contention.

harsha-simhadri: Huisheng, this is only on the setup phase that is run once, not on the search path. Would you still suggest merging the allocs -- that would make the code messy, isn't it?

hliu18: If it only runs once, it's fine.

PR #127, opened 2022-10-19
Harshasi/reorg pq and partition

harsha-simhadri:
Consolidated all code relevant to PQ to files pq.h/pq.cpp and removed it from partition_and_pq which is now renamed to partition.h/cpp. This could be useful for using PQ in other places in the code such as index build or in-mem indices.


src/pq.cpp
```diff
@@ -0,0 +1,920 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+#include "mkl.h"
+
+#include "pq.h"
+#include "math_utils.h"
+#include "tsl/robin_map.h"
+
+// block size for reading/ processing large files and matrices in blocks
```
ltan1ms, line 10: super nit: remove extra space after "reading/"?

ltan1ms (approved):

PR #128, opened 2022-10-20
print only errors in msbuild

harsha-simhadri:


theantony (approved):

harsha-simhadri:
See PR129 which replaces this.

PR #129, opened 2022-10-20
print only errors in msbuild

harsha-simhadri:


harsha-simhadri:
This is effectively the same as https://github.com/microsoft/DiskANN/pull/128 which was reviewed and approved. It differs in that commits are consolidated and rebased. So skipping review to merge.

PR #131, opened 2022-10-20
move vector space reservation from cached_beam_search to thread-speci…

harsha-simhadri:
…fic scratch space pre-allocation


src/pq_flash_index.cpp
```diff
@@ -139,6 +139,8 @@ namespace diskann {
         memset(scratch.aligned_query_float, 0,
                this->aligned_dim * sizeof(float));
         memset(scratch.rotated_query, 0, this->aligned_dim * sizeof(float));
+        
+        scratch.full_retset.reserve(4096);
```
ltan1ms, line 143: nit: I saw 4096 is also used as the initial size of the robin_set `scratch.visited`, may consider defining a common constant.

ltan1ms (approved):

PR #134, opened 2022-10-24
concurrent queue handles a pointer to ThreadData rather than cloning

harsha-simhadri:
concurrent queue handles a pointer to ThreadData rather than cloning

### DIFF
--- a/src/pq_flash_index.cpp
+++ b/src/pq_flash_index.cpp
@@ -53,7 +53,6 @@
 #define VECTOR_SECTOR_OFFSET(id) \
   ((((_u64)(id)) % nvecs_per_sector) * data_dim * sizeof(float))
 
-
 namespace diskann {
   template<typename T>
   PQFlashIndex<T>::PQFlashIndex(std::shared_ptr<AlignedFileReader> &fileReader,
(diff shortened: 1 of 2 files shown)


src/pq_flash_index.cpp
```diff
@@ -156,22 +180,12 @@ namespace diskann {
...
     assert(this->thread_data.size() == this->max_nthreads);
     while (this->thread_data.size() > 0) {
-      ThreadData<T> data = this->thread_data.pop();
-      while (data.scratch.sector_scratch == nullptr) {
+      auto data = this->thread_data.pop();
+      while (data->scratch.sector_scratch == nullptr) {
         this->thread_data.wait_for_push_notify();
         data = this->thread_data.pop();
       }
-      auto &scratch = data.scratch;
```
hliu18, line 164: Minor: these lines of code can be made into a method.

hliu18: LGTM.

src/pq_flash_index.cpp
```diff
@@ -895,16 +908,17 @@ namespace diskann {
...
       diskann::pq_dist_lookup(pq_coord_scratch, n_ids, this->n_chunks, pq_dists,
-                       dists_out);
+                              dists_out);
     };
-    Timer                 query_timer, io_timer, cpu_timer;
-    std::vector<Neighbor> retset(l_search + 1);
-    tsl::robin_set<_u64> &visited = *(query_scratch->visited);
+    Timer query_timer, io_timer, cpu_timer;
+
+    tsl::robin_set<_u64> & visited = query_scratch->visited;
```
ltan1ms, line 917: nit: extra space before 'v'?

ltan1ms (approved):

src/pq_flash_index.cpp
harsha-simhadri: @hliu18, I am thinking of changing the pop method to return nullptr for data when all scratch spaces are taken. I can encapsulate as a method along with that change. can we delay your request until then or would you prefer to close before that?

src/pq_flash_index.cpp
harsha-simhadri: that's how clang format is resolving it, so let's leave it as is?

src/pq_flash_index.cpp
hliu18: We can delay the suggested change. You can check in this PR ASAP so that Philip can update BANN package with your change.

src/pq_flash_index.cpp
ltan1ms: ah, then let's leave as is.

PR #135, opened 2022-10-25
added util to do single thread BFS

harsha-simhadri:


src/index.cpp
```diff
@@ -2740,6 +2740,44 @@ namespace diskann {
...
+
+    size_t MAX_BFS_LEVELS = 32;
+    auto   bfs_sets = new tsl::robin_set<unsigned>[MAX_BFS_LEVELS];
+
+    if (_dynamic_index)
+      for (unsigned i = _max_points; i < _max_points + _num_frozen_pts; ++i) {
+        for (auto node : _final_graph[i]) {
+          bfs_sets[0].insert(node);
+          visited.set(node);
+        }
```
rakri, line 2755: dont we want to include just start / frozen point in level 0, and not their neighbors? a) might be cleaner, and b) these points might otherwise get added at later levels since they are not added to visited here

rakri (changes requested):

rakri (approved):

rakri (approved):

PR #137, opened 2022-10-26
Backward-compatible index load

PhilipBAdams:


src/pq.cpp
```diff
@@ -51,25 +51,39 @@ namespace diskann {
...
 #endif
 
-    if (nr != 4) {
+    bool use_old_filetype = false;
+
+    if (nr != 4 && nr != 5) {
       diskann::cout << "Error reading pq_pivots file " << pq_table_file
                     << ". Offsets dont contain correct metadata, # offsets = "
-                    << nr << ", but expecting " << 4;
+                    << nr << ", but expecting " << 4 << "or" << 5;
```
rakri, line 59: maybe add spacing here?

src/pq.cpp
```diff
@@ -101,13 +115,26 @@ namespace diskann {
           __FILE__, __LINE__);
     }
 
+    if (use_old_filetype)
```
rakri, line 118: maybe cleaner to say something like if(old_format) chunk_offsets_location = 3 else chunk_offset_location = 2, and use it when load_bin?

rakri (approved):
Good to go, thanks Philip!

rakri:
Merging and closing

PR #138, opened 2022-10-26
disallow insertion of point with tag 0; ignore tag 0 in GT calculation

harsha-simhadri:
update tools to insert points with non-zero tags;
GT ignores location with 0 tags


harsha-simhadri:
replaced with https://github.com/microsoft/DiskANN/pull/140, which is this branch rebased with main

PR #139, opened 2022-10-27
fix GT calculation when npts%PARTSIZE<k. 

harsha-simhadri:
In compute_grpundtruth, when npts%PARTSIZE<k, exact_knn generates more than k results in the last partition which is wrong


rakri (approved):
Good to go, thanks for identifying this bug!

PR #140, opened 2022-10-28
disallow insertion of point with tag 0;

harsha-simhadri:
update tools to insert non-zero tag;
GT ignores location with 0 tags

### DIFF
--- a/include/index.h
+++ b/include/index.h
@@ -48,16 +48,16 @@ namespace diskann {
 
   template<typename T>
   struct InMemQueryScratch {
-    std::vector<Neighbor> *   _pool = nullptr;
+    std::vector<Neighbor>    *_pool = nullptr;
     tsl::robin_set<unsigned> *_visited = nullptr;
-    std::vector<unsigned> *   _des = nullptr;
-    std::vector<Neighbor> *   _best_l_nodes = nullptr;
+    std::vector<unsigned>    *_des = nullptr;
+    std::vector<Neighbor>    *_best_l_nodes = nullptr;
     tsl::robin_set<unsigned> *_inserted_into_pool_rs = nullptr;
-    boost::dynamic_bitset<> * _inserted_into_pool_bs = nullptr;
+    boost::dynamic_bitset<>  *_inserted_into_pool_bs = nullptr;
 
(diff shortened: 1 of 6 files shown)


rakri (approved):
Good to go.

include/index.h
```diff
@@ -192,11 +192,11 @@ namespace diskann {
...
     DISKANN_DLLEXPORT size_t search_with_tags(const T *query, const uint64_t K,
                                               const unsigned L, TagT *tags,
-                                              float *           distances,
+                                              float            *distances,
                                               std::vector<T *> &res_vectors);
 
-    // Will fail if tag already in the index
-    DISKANN_DLLEXPORT int insert_point(const T *point, const TagT tag);
+    // Will fail if tag already in the index or if tag=0.
+    DISKANN_DLLEXPORT int insert_point(const T *point, const TagT &tag);
```
theantony, line 199: I think we're limiting `TagT` to primitive types (there are asserts to that effect in e.g. `natural_number_map.h` and `index` template is instantiated only for a few primitive types). Given that, reference for tag just adds unnecessary level of indirection.

tests/utils/compute_groundtruth.cpp
```diff
@@ -278,8 +279,7 @@ inline void load_bin_as_float(const char *filename, float *&data, size_t &npts_u
   uint64_t end_id = (std::min)(start_id + PARTSIZE, (uint64_t) npts_i32);
   npts_u64 = (uint64_t) end_id - (uint64_t) start_id;
```
theantony, line 280: (From the previous PR) Are the two casts here necessary, the two variables already have `uint64_t` type?

theantony (approved):
My suggestions are minor, feel free to merge as-is.

include/index.h
harsha-simhadri: I can remove the reference for now. Is it easy to generalize the nnmap.h to string tag types?

theantony: AFAIR the main thing would be to make sure that primitive types are passed by value and `std::string` etc. by reference. Should be relatively straightforward, but I didn't want to spent time on it when creating the class as it wasn't necessary.

PR #141, opened 2022-10-31
Harshasi/reorg pq aux utils

harsha-simhadri:
Re-oraganizing code to enable use of PQ for in-memory index.


rakri (approved):
Looks great! Compiled in windows and it works, haven't run any build/search yet. will also do one or two just to make sure.

PR #142, opened 2022-11-01
move random data slice to generate (disk) quantized data

harsha-simhadri:


harsha-simhadri:
replace with https://github.com/microsoft/DiskANN/pull/148

PR #143, opened 2022-11-03
[fix]fix buf alloc，the size shoud be MAX_N_CMPS * aligned_dim * sizeof(T)

plutolove:
[fix]fix buf malloc，the size shoud be MAX_N_CMPS * aligned_dim * sizeof(T)

### DIFF
--- a/src/pq_flash_index.cpp
+++ b/src/pq_flash_index.cpp
@@ -65,7 +65,7 @@ namespace diskann {
 
   template<typename T>
   QueryScratch<T>::QueryScratch(size_t aligned_dim, size_t visited_reserve) {
-    _u64 coord_alloc_size = ROUND_UP(MAX_N_CMPS * aligned_dim, 256);
+    _u64 coord_alloc_size = ROUND_UP(MAX_N_CMPS * aligned_dim * sizeof(T), 256);
 
     diskann::alloc_aligned((void **) &coord_scratch, coord_alloc_size, 256);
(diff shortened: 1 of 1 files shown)


src/pq_flash_index.cpp
```diff
@@ -65,7 +65,7 @@ namespace diskann {
 
   template<typename T>
   QueryScratch<T>::QueryScratch(size_t aligned_dim, size_t visited_reserve) {
-    _u64 coord_alloc_size = ROUND_UP(MAX_N_CMPS * aligned_dim, 256);
+    _u64 coord_alloc_size = ROUND_UP(MAX_N_CMPS * aligned_dim * sizeof(T), 256);
```
gopalrs, line 68: In the DLVS code, we removed the MAX_N_CMPS from the expression to save space. IIRC, coord_scratch only holds one point at a time. Maybe you want to do the same here? 
In the DLVS code, it is simply aligned_dim * sizeof(T).

gopalrs (changes requested):

src/pq_flash_index.cpp
plutolove: That is a Bug, If buf_idx > 4069 will coredump

harsha-simhadri: @gopalrs Let's take the reduction of the MAX_N_CMPS as a separate issue.

harsha-simhadri (approved):

harsha-simhadri:
Replaced with PR #220

PR #146, opened 2022-11-08
A (GROSS!) first stab at executing Java code from within C++.

daxpryce:
Next steps would be to actually implement the AlignedFileReader interface for the jni spec, of which the direct byte buffer access from https://docs.oracle.com/en/java/javase/16/docs/specs/jni/functions.html#newdirectbytebuffer might be useful.

This is not actually intended for merge, but it is a way of showing the differing files and steps necessary to get the barest bones implementation into the mix.

### DIFF
--- a/java/FileInputStreamMediator.java
+++ b/java/FileInputStreamMediator.java
@@ -1,0 +1,40 @@
+// package com.microsoft.diskann;
+
+import java.io.Closeable;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+
+public class FileInputStreamMediator implements Closeable {
+
+    private File file;
+    private FileInputStream stream;
+
+    public FileInputStreamMediator(final String canonicalPath) throws FileNotFoundException {
+        // TODO: add your preconditions here, make sure the file exists, is readable, etc
+        // we're not doing that because we're early days yet.
+        this.stream = new FileInputStream(new File(canonicalPath));
+    }
+
+    public boolean read(final int[] readFroms, final int[] readUntils, final long[] readIntos) throws Exception {
+        // these 3 arrays must be the same length
+        // the readIntos are currently just the memory we're supposed to read into, however one does that across JNI
+        // I'm going to punt on that and instead just print out all the readIntos and untils and we'll figure out
+        // how to actually read, using Java IO, <into> some pre-allocated C++ buffers
+        // in an absolute worst case scenario we can create our own return list of buffers and do a memcopy
+        // from one into another, but I'd prefer not to do that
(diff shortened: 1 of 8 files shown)


java/FileInputStreamMediator.java
```diff
@@ -0,0 +1,40 @@
...
+        // these 3 arrays must be the same length
+        // the readIntos are currently just the memory we're supposed to read into, however one does that across JNI
+        // I'm going to punt on that and instead just print out all the readIntos and untils and we'll figure out
+        // how to actually read, using Java IO, <into> some pre-allocated C++ buffers
+        // in an absolute worst case scenario we can create our own return list of buffers and do a memcopy
+        // from one into another, but I'd prefer not to do that
+        if (readFroms.length != readUntils.length || readFroms.length != readIntos.length) {
+            throw new Exception("The 3 parallel arrays must be the same size"); // this is a terrible error messaging pattern, but I don't know what control I have over *exceptions*
+        }
+        for (int i = 0; i < readFroms.length; i++) {
```
daxpryce, line 30: Sample output:

```bash
./diskann_with_java_io 
I would have read from 1024 until 5120 and placed it into 5120000
```

src/jni/basic_file_reader.cpp
```diff
@@ -0,0 +1,68 @@
...
+  this->javaConstructor = this->env.GetMethodID(this->javaClass, "<init>", "(Ljava/lang/String;)V");
+  // this would be a great place to verify the methods and arguments we expect to pass in are meeting our criteria
+  if (this->javaConstructor == nullptr) {
+    std::cerr << "ERROR: Java class FileInputStreamMediator does not comply with expected interface (ctor)" << std::endl;
+  }
+  this->javaClose = this->env.GetMethodID(this->javaClass, "close", "()V");
+  if (this->javaClose == nullptr) {
+    std::cerr << "ERROR: Java class FileInputStreamMediator does not comply with expected interface (close)" << std::endl;
+  }
+  // https://docs.oracle.com/javase/1.5.0/docs/guide/jni/spec/types.html#wp276 you'll need this. trust me.
```
daxpryce, line 20: https://docs.oracle.com/en/java/javase/16/docs/specs/jni/types.html#type-signatures actually for the JNI version we're targeting

xjtushilei:
I would like to ask politely when will the java interface be supported? Expect to use jni to achieve

daxpryce:
> I would like to ask politely when will the java interface be supported? Expect to use jni to achieve

I'm so sorry, but this isn't on our roadmap at all. This branch was intended to show some other Microsoft teams that JNI could be used, but it was only ever meant to be an exemplar toy for how they *could* proceed with it if they so chose to.

PR #149, opened 2022-11-14
Fixed windows debug cmake script and issues that caused asserts when running under the debugger

shanewil:
The commit message covers most of the changes with the addition of assigning a small buffer to the logging calls instead of zero length. A zero-length buffer allows a 1 char overwrite to the buffer. As I see it std::basic_streambuf<char> isn't intended to work with a zero-length buffer.


harsha-simhadri (approved):

rakri (approved):
not tested with the logger in windows (especially there seems to be a BUFFER_SIZE vs BUFFER_SIZE -1bugfix); 

The insertintopool changes seem to be good, thanks! btw, are they related to the topic of this PR?

[shane] Once the debug build worked in windows the other fixes in this PR were asserts that surfaced when running the debug bins.

PR #151, opened 2022-11-15
make neighbor compare stable

ltan1ms:
## Motivation
Make neighbor compare stable.

We've observed that search results can have vectors with same distance value return in different order when search the same query vector multiple times, which makes the search results not deterministic in corner cases. This fixed that.

### DIFF
--- a/include/neighbor.h
+++ b/include/neighbor.h
@@ -21,8 +21,10 @@ namespace diskann {
     }
 
     inline bool operator<(const Neighbor &other) const {
-      return distance < other.distance;
+      return distance < other.distance ||
+          (distance == other.distance && id < other.id);
     }
+
     inline bool operator==(const Neighbor &other) const {
       return (id == other.id);
     }


include/neighbor.h
```diff
@@ -21,8 +21,10 @@ namespace diskann {
     }
 
     inline bool operator<(const Neighbor &other) const {
```
theantony, line 23: Should we do the same change for `SimpleNeighbor` class below?

theantony (approved):
This change definitely helps to consistently return the same documents on the same machine for the same query.

As a side-note (@harsha-simhadri), in streaming scenarios different replicas will have different locations for vectors due to ordering and/or consolidation, so further work may be needed on that side if we frequently observe such instability.

harsha-simhadri:
> This change definitely helps to consistently return the same documents on the same machine for the same query.
> 
> As a side-note (@harsha-simhadri), in streaming scenarios different replicas will have different locations for vectors due to ordering and/or consolidation, so further work may be needed on that side if we frequently observe such instability.

Andrija, thanks! so you think it helps your experimentation too to make this change?

theantony:
> Andrija, thanks! so you think it helps your experimentation too to make this change?

Yes, I think that compared to the previous state the results from a single machine would be more consistent.

include/neighbor.h
ltan1ms: yes, we can add it too. I saw it is referenced in some test code and then only referenced once in source code below:
https://github.com/microsoft/DiskANN/blob/2e1e328e9362196de70d9f25ca7fc542b9438d5c/src/pq_flash_index.cpp#L924
And the variable `medoid_dists` is not referenced anywhere, we can remove that line @harsha-simhadri ?

harsha-simhadri: Is this related to this PR? Or a separate issue?

ltan1ms: Probably not related. So you suggest not to touch it?

PR #153, opened 2022-11-22
Fix neighbor pool insertion

hhy3:
The neighbor pool doesn't limit the size currently which I think is not a feature but a bug, and the bisection code is too cumbersome.


hhy3:
@microsoft-github-policy-service agree

harsha-simhadri:
> The neighbor pool doesn't limit the size currently which I think is not a feature but a bug, and the bisection code is too cumbersome.

Thanks for the contribution. The CI tests are failing, could you please take a look?

hhy3:
> > The neighbor pool doesn't limit the size currently which I think is not a feature but a bug, and the bisection code is too cumbersome.
> 
> Thanks for the contribution. The CI tests are failing, could you please take a look?

The bug that made CI fail has been fixed.

include/neighbor.h
```diff
@@ -49,37 +50,14 @@ namespace diskann {
...
 
   static inline unsigned InsertIntoPool(std::vector<Neighbor> &neighbors,
-                                        unsigned K, Neighbor nn) {
+                                        unsigned K, unsigned cap, Neighbor nn) {
     // find the location to insert
-    unsigned left = 0, right = K - 1;
-    if (neighbors[left].distance > nn.distance) {
-      neighbors.insert(neighbors.begin() + left, nn);
-      return left;
+    auto it = std::lower_bound(neighbors.begin(), neighbors.begin() + K, nn);
```
shanewil, line 55: In general your changes are a vast improvement. It would be good if the indexer K could be removed from the function such that this could be done:
auto it = std::lower_bound(neighbors.begin(), neighbors.end(), nn);
instead of
auto it = std::lower_bound(neighbors.begin(), neighbors.begin() + K, nn);
The current calling code isn't' structured in a way that makes this a simple change. In some cases there's an assumption that the vector is prepopulated (etc).
I'm heading out on vacation until late next week so I don't have time to dig into this. If you have a tidy way to change this that would be great otherwise I'll revisit this when I'm back. thanks -

hhy3: OK. I'll give it a try.

hhy3:
@shanewil I've refactored the neighbor insertion logic in #163. Could you please take a look

PR #157, opened 2022-11-23
reverted to older insertintopool to handle slow builds

rakri:
Reverted to older InsertIntoPool in neighbor.h to fix some slow build latecy issue.
Will investigate more why the vector-based code is slow, meanwhile we can use the pointer method.

### DIFF
--- a/include/neighbor.h
+++ b/include/neighbor.h
@@ -48,21 +48,22 @@ namespace diskann {
     std::vector<SimpleNeighbor> pool;
   };
 
-  static inline unsigned InsertIntoPool(std::vector<Neighbor> &neighbors,
-                                        unsigned K, Neighbor nn) {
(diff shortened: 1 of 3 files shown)


shanewil (approved):

shanewil:
This is very odd. The change that was undone had been tested by running multiple index builds by multiple people. I suspect a later commit changed something that either exposed the problem in PR 149 or has introduced its own problem.

harsha-simhadri:
> This is very odd. The change that was undone had been tested by running multiple index builds by multiple people. I suspect a later commit changed something that either exposed the problem in PR 149 or has introduced its own problem.

You are probably right. Lets just get this fix in some that other PRs can be checked and merge. And move things back to "vector&" again once other PRs are checked in.

PR #158, opened 2022-11-29
Support for using PQ distance comparisons for index build. Initial co…

harsha-simhadri:
…mmit, bugs still exist

### DIFF
--- a/tests/build_disk_index.cpp
+++ b/tests/build_disk_index.cpp
@@ -14,7 +14,7 @@ namespace po = boost::program_options;
 
 int main(int argc, char** argv) {
   std::string data_type, dist_fn, data_path, index_path_prefix;
-  unsigned    num_threads, R, L, disk_PQ;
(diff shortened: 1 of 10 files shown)


tests/build_disk_index.cpp
```diff
@@ -57,6 +57,9 @@ int main(int argc, char** argv) {
                        po::bool_switch()->default_value(false),
                        "Include full precision data in the index. Use only in "
                        "conjuction with compressed data on SSD.");
+    desc.add_options()(
+        "build_PQ_bytes", po::value<uint32_t>(&build_PQ)->default_value(0),
```
ltan1ms, line 61: @harsha-simhadri , for this parameter should it be the same as the PQ bytes per vector at search time? Like 48d for our 384d experimentation

harsha-simhadri: It can be different. I will write in the markdown documentation file

harsha-simhadri:
Replacing with https://github.com/microsoft/DiskANN/pull/177

PR #161, opened 2022-11-30
Clang-Format Enhancements

daxpryce:
This commit uses CMake to set up a pair of custom targets: `format` and `checkformat`, invoked via `make format` or `make checkformat`.

`make format` will format all .cpp or .h files in source as per our _clang-format specifications. Linux developers should be sure to use this prior to pushing. Windows developers will need to execute clang-format through Visual Studio (see: https://learn.microsoft.com/en-us/visualstudio/ide/reference/options-text-editor-c-cpp-formatting?view=vs-2022 ).

Any code that is not formatted will be called out as part of our CICD process when `make checkformat` is run. Any detected incompabitibilities with our `_clang-format` rules will cause `make checkformat` - a custom target wrapping `clang-format --dry-run -Werror <filelist>` - to fail, and thus fail the build.

This first commit adds the custom targets, removes the old clang-format rules predicated on building `test/utils/gen_random_slice`, and stubs out new steps in our github workflows.

`make checkformat` is left commented out, because a dozen or more files have not been clang-formatted in quite some time. I want to review these changes prior to muddying up the PR with dozens of formatting changes left as a result of us following these defined rules.

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -109,14 +109,14 @@ typedef uint8_t  _u8;
 typedef int8_t   _s8;
 inline void      open_file_to_write(std::ofstream&     writer,
                                     const std::string& filename) {
-  writer.exceptions(std::ofstream::failbit | std::ofstream::badbit);
-  if (!file_exists(filename))
+       writer.exceptions(std::ofstream::failbit | std::ofstream::badbit);
+       if (!file_exists(filename))
     writer.open(filename, std::ios::binary | std::ios::out);
   else
     writer.open(filename, std::ios::binary | std::ios::in | std::ios::out);
 
   if (writer.fail()) {
-    char buff[1024];
+         char buff[1024];
 #ifdef _WINDOWS
     strerror_s(buff, 1024, errno);
 #else
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -30,10 +30,14 @@ jobs:
             wget https://registrationcenter-download.intel.com/akdlm/irc_nas/18487/l_BaseKit_p_2022.1.2.146.sh
             sudo sh l_BaseKit_p_2022.1.2.146.sh -a --components intel.oneapi.lin.mkl.devel --action install --eula accept -s
         fi
-    - name: Clang Format Check
+    - name: Ubuntu CMake Configure
       if: runner.os != 'Windows'
       run: |
         mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release ..
+
+    - name: Clang Format Check
+      if: ${{ matrix.os == 'ubuntu-latest' }}
+      run: |
         make checkformat
 
     - name: build
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -26,7 +26,7 @@ jobs:
         if [ "${{ matrix.os }}" != "ubuntu-18.04" ]; then
             sudo apt install cmake g++ libaio-dev libgoogle-perftools-dev clang-format libboost-dev libboost-program-options-dev libmkl-full-dev libunwind-dev-
         else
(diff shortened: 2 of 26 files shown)


harsha-simhadri (commented):
Thanks for this welcome change. Have you ever checked if clang-format instructions are interprested by visual studio/vscode exactly as they are on linux. Do you want me to merge now or after you turn on the make checkformat?

daxpryce:
> Have you ever checked if clang-format instructions are interprested by visual studio/vscode exactly as they are on linux. 

I have not.  I would be distinctly shocked if that ended up being the case though - the only thing worse than coding standard fights amongst humans is coding standard fights amongst standardized tools on different operating systems. 🤮 

> Do you want me to merge now or after you turn on the make checkformat?

If you're good with this conceptual approach, and it seems like you are, I'll run the `make format` command to actually change all the files, add them to this PR, and enable the `make checkformat` in our github actions.  And also change this from a draft to a not-draft :)

include/utils.h
```diff
@@ -109,14 +109,14 @@ typedef uint8_t  _u8;
...
-  if (!file_exists(filename))
+       writer.exceptions(std::ofstream::failbit | std::ofstream::badbit);
+       if (!file_exists(filename))
     writer.open(filename, std::ios::binary | std::ios::out);
   else
     writer.open(filename, std::ios::binary | std::ios::in | std::ios::out);
 
   if (writer.fail()) {
-    char buff[1024];
+         char buff[1024];
```
harsha-simhadri, line 119: This indentation looks odd!

harsha-simhadri (commented):
Do we want to format include/tsl files we included from elsewhere?

daxpryce:
> Do we want to format include/tsl files we included from elsewhere?

So, I would think no, since the formatting is to keep all contributors consistent and so we don't end up with a bunch of "distinction without a difference" stylistic commits that make PRs harder to review than they should be. If we're including them as sources in our git repo then absolutely, but if it's just included as part of the build process then there's no real benefit to formatting someone else's code. 

_However_, if what you're really saying is I missed some files and folders, then I should fix that!  I presumed the CMake variable `ALL_SOURCE_FILES` would actually have all source files in it, but I must admit I didn't try to verify the contents were the same as the label on the tin :)

include/utils.h
daxpryce: I absolutely agree! Is it worth revisiting our _clang-format rules and try to track down just what might have caused it to (seemingly) triple indent?

harsha-simhadri: I don't know if this is a rule issue, or a bug in the IDE's interpretation of the rule. What happens if you clang-format this file from the command line in linux? Do you see this behavior?

daxpryce: @harsha-simhadri I only use linux for DiskANN, I haven't even gotten it to build on Windows yet - so I would absolutely suspect this is the _clang-format rule (and/or whatever "reasonable default" it has, if our rule file says nothing about it)

daxpryce:
> > _However_, if what you're really saying is I missed some files and folders, then I should fix that! I presumed the CMake variable `ALL_SOURCE_FILES` would actually have all source files in it, but I must admit I didn't try to verify the contents were the same as the label on the tin :)
> 
> The current branch has reformatted a bunch of files in the include/tsl/ directory. Maybe he is saying that those should not be formatted by our clang formatter?

Ah, I see!  I'll revert those changes and exclude them from `make format` and `make checkformat` then :)

daxpryce:
> Do we want to format include/tsl files we included from elsewhere?

In the latest commits, these have been excluded from the formatting rules.

Everything else is 100% command line driven clang-format - Linux only. Windows users will have to rely on clang-format in their IDE, but I'm not a good source for anything around that, as I am not a windows developer

harsha-simhadri (approved):

PR #163, opened 2022-12-03
Optimize neighbor insertion

hhy3:
This PR encapsulates search logic which is rather confusing to a data structure `NeighborSet` . It greatly simplifies the search logic.

### DIFF
--- a/include/neighbor.h
+++ b/include/neighbor.h
@@ -13,16 +13,16 @@ namespace diskann {
   struct Neighbor {
     unsigned id;
     float    distance;
-    bool     flag;
+    bool     checked;
 
     Neighbor() = default;
-    Neighbor(unsigned id, float distance, bool f)
-        : id{id}, distance{distance}, flag(f) {
+    Neighbor(unsigned id, float distance)
+        : id{id}, distance{distance}, checked(false) {
     }
 
     inline bool operator<(const Neighbor &other) const {
       return distance < other.distance ||
-          (distance == other.distance && id < other.id);
+             (distance == other.distance && id < other.id);
     }
 
     inline bool operator==(const Neighbor &other) const {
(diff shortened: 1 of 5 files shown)


include/neighbor.h
```diff
@@ -13,57 +13,100 @@ namespace diskann {
   struct Neighbor {
     unsigned id;
     float    distance;
-    bool     flag;
+    bool     checked;
 
     Neighbor() = default;
```
shanewil, line 18: This default constructor isn't initializing variables in the class. I was seeing crashes during index build because cur_ was set to a random value.

shanewil (commented):
Thanks for your effort. I haven't looking through all your changes yet. I did notice a performance drop between this pr and the main branch index-build although the results seem inconsistent, so I am wondering if there is something else going on. I'll spend some more time on this when I get a chance.

src/index.cpp
```diff
@@ -728,108 +729,88 @@ namespace diskann {
...
     }
 
     // sort best_L_nodes based on distance of each point to node_coords
-    std::sort(best_L_nodes.begin(), best_L_nodes.begin() + l);
-    unsigned best_unchanged = 0;
     uint32_t hops = 0;
     uint32_t cmps = 0;
 
-    while (best_unchanged < l) {
-      unsigned best_inserted_position = l;
```
shanewil, line 755: Before the first call to has_next() I think the NeighborSet object needs to be set to the first unchecked index. something like:
    void reset_cursor() {
      cur_ = 0;
      while (cur_ < size_ && data_[cur_].checked) {
        cur_++;
      }
    }
This would make it consistent with the way the calling code currently works. Note that calls to NeighborSet::insert() can change cur_  to a position after a checked item and has_next() could miss this item. This applies to a couple of other places in the code that walks though the NeighborSet items.

shanewil:
I finished reviewing the changes and it looks good once you address the 2 issues I mention in other comments. (thanks)

hhy3:
> I finished reviewing the changes and it looks good once you address the 2 issues I mention in other comments. (thanks)

Thanks for your review! I'll fix these issues soon.

include/neighbor.h
```diff
@@ -13,57 +13,100 @@ namespace diskann {
...
+  class NeighborSet {
+   public:
+    NeighborSet() = default;
+
+    explicit NeighborSet(size_t capacity)
+        : size_(0), capacity_(capacity), data_(capacity_ + 1) {
+    }
+
+    void insert(const Neighbor &nbr) {
+      if (size_ == capacity_ && nbr.distance >= data_[size_ - 1].distance) {
```
shanewil, line 44: Neighbor comparisons should use the bool operator< defined on the Neighbor class. This comparison functions includes using the id field to create a stable sort.

src/index.cpp
shanewil: On second look I don't think my previous comment about adding a reset_cursor() function is not needed. Your insert() function will always insert a new Neighbor object with the field checked == false and then set the 'cur_' index value to the inserted index if it's less than its current 'cur_' value. This looks OK to me.

shanewil:
Wang Zihao, if you don't think you have time to make these changes by end of week, please let me know. We are trying to burn down the PR list before we start new work items.  I could pull your branch and make the changes on a new PR. Thanks!

shanewil:
This has been ported to pr #190 with some small updates. Thanks for your effort

PR #170, opened 2022-12-16
Fix undefined symbol errors during python setup.py install by renaming variables

LakshyAAAgrawal:
Observed the following errors on running
```
python3 setup.py install
```
```
python/src/diskann_bindings.cpp: In member function ‘int DiskANNIndex<T>::_build_disk_index_float(const char*, const char*, const char*, diskann::Metric)’:
python/src/diskann_bindings.cpp:107:45: error: ‘data_file_path’ was not declared in this scope;
```

Further, ```diskannpy``` was referenced in python/tests/test_build_disk_index.py but not imported. Kindly let me know if it needs to go as a separate PR

### DIFF
--- a/python/src/diskann_bindings.cpp
+++ b/python/src/diskann_bindings.cpp
@@ -100,14 +100,6 @@ struct DiskANNIndex {
     return 0;
   }
 
-  int _build_disk_index_float(const char     *dataFilePath,
-                              const char     *indexFilePath,
-                              const char     *indexBuildParameters,
-                              diskann::Metric compareMetric) {
-    return diskann::build_disk_index<float>(dataFilePath, indexFilePath,
-                                            indexBuildParameters, compareMetric);
-  }
-
   void search(std::vector<T> &query, const _u64 query_idx, const _u64 dim,
               const _u64 num_queries, const _u64 knn, const _u64 l_search,
               const _u64 beam_width, std::vector<unsigned> &ids,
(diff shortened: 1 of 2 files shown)


python/src/diskann_bindings.cpp
```diff
@@ -104,8 +104,8 @@ struct DiskANNIndex {
                               const char     *indexFilePath,
```
daxpryce, line 104: You can actually remove this entire function - it somehow crept in and never should have merged in.

The latest working build is in my fork @ https://github.com/daxpryce/DiskANN/tree/daxpryce/pybind_diskann - I've been wrestling with getting our build to work via github actions across the OS / Python3.7-3.11 matrix over there and have also made this change

python/tests/test_build_disk_index.py
```diff
@@ -3,6 +3,7 @@
 
 import time
 import argparse
+import diskannpy
```
daxpryce, line 6: what is your intent of importing the module namespace?  I'm not against it, it just seems unnecessary given the `from diskannpy import ...` down below.

LakshyAAAgrawal: Line 22 is ```index = DiskANNFloatIndex(diskannpy.L2)``` which uses the module import. Alternative would be do a ```from diskannpy import L2```

daxpryce: ah ok!  thanks for clarifying, I appreciate it!

python/src/diskann_bindings.cpp
LakshyAAAgrawal: Removed the function in the latest commit.

Kindly let me know if there is any specific issue related to python build on Github Actions you would like me to look into.

PR #172, opened 2022-12-22
Use stable comparison through Neighbor class when inserting to result set

ltan1ms:
# Motivation
Under `USE_BING_INFRA` complier flag, the code path is using async reads, so when BW > 1, the multiple IOs may return in arbitrary order, and because the insertion of the neighbors to the results set can be unstable as we didn't handle ties in distance, it can lead to final search results being unstable. So, instead, delegate the comparisons to Neighbor classes's `<` operator which handles that already by doing tie breaking based on node id.

### DIFF
--- a/src/pq_flash_index.cpp
+++ b/src/pq_flash_index.cpp
@@ -963,7 +963,8 @@ namespace diskann {
             cmps++;
             float dist = dist_scratch[m];
             Neighbor nn(id, dist, true);
-            if (retset[cur_list_size - 1] < nn &&
+            Neighbor &lastResult = retset[cur_list_size - 1];
+            if ((lastResult < nn || lastResult == nn) &&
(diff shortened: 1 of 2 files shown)


src/pq_flash_index.cpp
```diff
@@ -959,16 +959,13 @@ namespace diskann {
         // process prefetched nhood
         for (_u64 m = 0; m < nnbrs; ++m) {
           unsigned id = node_nbrs[m];
-          if (visited.find(id) != visited.end()) {
-            continue;
-          } else {
-            visited.insert(id);
+          if (visited.insert(id).second) {
```
ltan1ms, line 962: piggypacked a minor optimization here to avoid double look up in the hash set.

harsha-simhadri (approved):

PR #177, opened 2022-12-26
Support for using PQ distance comparisons for index build.

harsha-simhadri:
create pq file

label pq pivots file with chunk size

Use full precision distance comparisons for pruning

added PQ based build option to build memory and disk index

fixing sampling probability calculation in PQ training data generation

documented the PQ build option in workflows


tests/build_memory_index.cpp
```diff
@@ -85,6 +88,12 @@ int main(int argc, char** argv) {
         po::value<uint32_t>(&num_threads)->default_value(omp_get_num_procs()),
         "Number of threads used for building index (defaults to "
         "omp_get_num_procs())");
+    desc.add_options()("num_pq_bytes",
+                       po::value<uint32_t>(&build_PQ_bytes)->default_value(0),
+                       "Number of PQ bytes to compress vectors to");
```
ltan1ms, line 93: nit: should this description clarifies that the "PQ bytes" refers to size of compressed vectors used for distance comparisons during index build (i.e., not to confuse with the size of compressed vectors during serving time, which is derived from -B option)?

include/scratch.h
```diff
@@ -34,50 +34,66 @@ namespace diskann {
...
     tsl::robin_set<unsigned> _inserted_into_pool_rs;
     boost::dynamic_bitset<> *_inserted_into_pool_bs;
+    std::vector<unsigned>    _id_scratch;
+    float                   *_dist_scratch = nullptr;
 
     T        *_aligned_query = nullptr;
     uint32_t *_indices = nullptr;
     float    *_interim_dists = nullptr;
+
+    std::vector<float> _occlude_factor;
```
ltan1ms, line 94: curious, what does this variable mean?

ltan1ms (approved):

include/scratch.h
harsha-simhadri: This is used in occlude_list function to prune candidates. This vector helps dynamically adjust alpha parameter to the dataset

tests/build_memory_index.cpp
harsha-simhadri: fixed

PR #187, opened 2023-01-05
remove ubuntu 18 from CI build and run OS matrix

harsha-simhadri:

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -22,13 +22,7 @@ jobs:
     - name: Install dependencies
       if: runner.os != 'Windows'
       run: |
-        if [ "${{ matrix.os }}" != "ubuntu-18.04" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-test.yml
```diff
@@ -7,7 +7,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-18.04, ubuntu-latest, windows-2019, windows-latest]
+        os: [ubuntu-latest, windows-2019, windows-latest]
```
theantony, line 10: We can now simplify lines 25-31 to just do `sudo apt install` from the `if` branch, we don't need the `else` branch any more.

theantony (approved):

.github/workflows/pr-test.yml
harsha-simhadri: I thought we might just have it around so that if anyway is looking, they know to download mkl separately. But happy to remove too.

theantony: I think the `README.md` has that part covered. It's OK to keep it here as well.

harsha-simhadri: removed. thanks!

PR #206, opened 2023-01-11
Support data files loading in separated paths

Sanhaoji2:
This change is adding a new load() function to allow customer to input index_file, pivots_file and compressed_file from different paths. 
Since the SSD write & delete operation impact the read performance significantly, to limit the regression during data refresh, the server may only copy the necessary file to SSD drive, that make the index file place in different path from other data files in this scenario.

### DIFF
--- a/include/neighbor.h
+++ b/include/neighbor.h
@@ -13,11 +13,12 @@ namespace diskann {
   struct Neighbor {
     unsigned id;
     float    distance;
-    bool     flag;
+    bool     expanded;
 
     Neighbor() = default;
-    Neighbor(unsigned id, float distance, bool f)
-        : id{id}, distance{distance}, flag(f) {
+
+    Neighbor(unsigned id, float distance)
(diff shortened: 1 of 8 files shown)


harsha-simhadri:
@Sanhaoji2  Could you please add a bit more descriptive text on what scenario or use case this is helping? Thanks!

harsha-simhadri (changes requested):
Please also seperate changes for the new load function from the vc16 compiler error fixes you mentioned and raise seperate PRs. Thanks!

Sanhaoji2:
> @Sanhaoji2 Could you please add a bit more descriptive text on what scenario or use case this is helping? Thanks!

Added description for the processing logic of our server.

Sanhaoji2:
> 

OK

> Please also seperate changes for the new load function from the vc16 compiler error fixes you mentioned and raise seperate PRs. Thanks!

OK.

harsha-simhadri (approved):

PR #211, opened 2023-01-18
Harshasi/add checks to search interface

harsha-simhadri:
 
Check if user asks for K > L results, where L is the search parameter. Throw an exception if that happens. Also check that no more than L entries from best_L_nodes are copied to returned results.

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -1520,11 +1520,11 @@ namespace diskann {
     auto &best_L_nodes = scratch->best_l_nodes();
 
     size_t pos = 0;
-    for (int i = 0; i < best_L_nodes.size() && i < L; ++i) {
+    for (int i = 0; i < best_L_nodes.size(); ++i) {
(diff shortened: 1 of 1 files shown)


src/index.cpp
```diff
@@ -1483,6 +1483,10 @@ namespace diskann {
                                                        const unsigned L,
                                                        IdType        *indices,
                                                        float *distances) {
+    if (K > (uint64_t) L)
```
dkaczynski, line 1486: I typically advise using curly braces `{...}` even on single-line blocks of code to avoid future bugs if additional lines are added to the block.

dkaczynski: I understand if there's an established precedent in this project of not putting curly braces on single-line blocks. Possibly leave this to a future refactoring.

harsha-simhadri: added {}

dkaczynski (approved):

PR #212, opened 2023-01-18
Harshasi/rewrite inmem scratch

harsha-simhadri:
 
In-memory scratch is not designed well. This PR fixes issues around allocation and resizing of member variables.

### DIFF
--- a/include/scratch.h
+++ b/include/scratch.h
@@ -16,6 +16,11 @@
 #include "pq.h"
 #include "aligned_file_reader.h"
 
+
+// In-mem index related limits
+#define GRAPH_SLACK_FACTOR 1.3
+
+// SSD Index related limits
 #define MAX_GRAPH_DEGREE 512
 #define MAX_N_CMPS 16384
 #define SECTOR_LEN (_u64) 4096
@@ -28,8 +33,6 @@ namespace diskann {
   template<typename T>
   class InMemQueryScratch {
    public:
-
-
     ~InMemQueryScratch();
     InMemQueryScratch(uint32_t search_l, uint32_t indexing_l, uint32_t r,
                       uint32_t maxc, size_t dim, bool init_pq_scratch = false);
@@ -87,23 +90,40 @@ namespace diskann {
     uint32_t _L;
     uint32_t _R;
     uint32_t _maxc;
-    
+
     T *_aligned_query = nullptr;
 
     PQScratch<T> *_pq_scratch = nullptr;
 
-    std::vector<Neighbor>    _pool;
-    NeighborPriorityQueue    _best_l_nodes;
-    std::vector<float>       _occlude_factor;
+    // _pool stores all neighbors explored from best_L_nodes.
+    // Usually around L+R, but could be higher.
+    // Initialized to 10L, but expands as needed.
+    std::vector<Neighbor> _pool;
+
+    // _best_l_nodes is reserved for storing best L entries
+    // Underlying storage is L+1 to support inserts
+    NeighborPriorityQueue _best_l_nodes;
+
+    // _occlude_factor.size() >= pool.size() in occlude_list function
+    // _pool is clipped to maxc in occlude_list before affecting _occlude_factor
+    // _occlude_factor is initialized to maxc size
+    std::vector<float> _occlude_factor;
 
     tsl::robin_set<unsigned> _inserted_into_pool_rs;
     boost::dynamic_bitset<> *_inserted_into_pool_bs;
 
+    // _id_scratch.size() must be > R*GRAPH_SLACK_FACTOR for iterate_to_fp
     std::vector<unsigned> _id_scratch;
-    std::vector<float>    _dist_scratch;
 
-    std::vector<uint32_t> _indices;        // only used by search
-    std::vector<float>    _interim_dists;  // only used by search
+    // _dist_scratch must be > R*GRAPH_SLACK_FACTOR for iterate_to_fp
+    // _dist_scratch should be at least the size of id_scratch
+    std::vector<float> _dist_scratch;
+
+    // Used as a buffer for search results, must be at least L in size
+    std::vector<uint32_t> _indices;
+
+    // Used as a buffer for search results, must be at least L in size
+    std::vector<float> _interim_dists;
   };
 
   //
--- a/include/scratch.h
+++ b/include/scratch.h
@@ -28,15 +28,30 @@ namespace diskann {
   template<typename T>
   class InMemQueryScratch {
    public:
-    uint32_t search_l;
-    uint32_t indexing_l;
-    uint32_t r;
+
 
     ~InMemQueryScratch();
     InMemQueryScratch(uint32_t search_l, uint32_t indexing_l, uint32_t r,
                       uint32_t maxc, size_t dim, bool init_pq_scratch = false);
-    void resize_for_query(uint32_t new_search_l);
+    void resize_for_new_L(uint32_t new_search_l);
     void clear();
+
+    inline uint32_t get_L() {
+      return _L;
+    }
+    inline uint32_t get_R() {
+      return _R;
+    }
+    inline uint32_t get_maxc() {
+      return _maxc;
+    }
+
+    inline T *aligned_query() {
+      return _aligned_query;
+    }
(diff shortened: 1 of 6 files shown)


src/index.cpp
```diff
@@ -852,9 +855,10 @@ namespace diskann {
 
       // Compute distances to unvisited nodes in the expansion
       if (_pq_dist) {
-        compute_dists(id_scratch.data(), id_scratch.size(), dist_scratch);
-
+        assert(dist_scratch.capacity() >= id_scratch.size());
+        compute_dists(id_scratch.data(), id_scratch.size(), dist_scratch.data());
```
theantony, line 859: In this code path, `dist_scratch` is never resized, so `data()` can return a `nullptr` and result in a crash. One way to address this is to change `compute_dists` to take a reference to the `float` vector and then do `resize()` before calling `diskann::pq_dist_lookup` passing it `data()`.

Since `dist_scratch.clear()` was called in line 832, the resize would also have an effect of initializing all values in the vector to 0. The `diskann::pq_dist_lookup` method also initializes the buffer to 0, so there's double work. There are several ways to avoid that, one with minimum changes is to add a flag to `diskann::pq_dist_lookup` whether to initialize distances.

src/index.cpp
```diff
@@ -1486,11 +1493,13 @@ namespace diskann {
...
-      scratch->resize_for_query(L);
-      diskann::cout << "Expanding query scratch_space. Was created with Lsize: "
-                    << scratch->search_l << " but search L is: " << L
-                    << std::endl;
+    if (L > scratch->get_L()) {
+      diskann::cout << "Attempting to expand query scratch_space. Was created "
+                    << "with Lsize: " << scratch->get_L()
+                    << " but search L is: " << L << std::endl;
+      scratch->resize_for_new_L(L);
+      diskann::cout << "Resize completed. Let scratch->L is "
```
theantony, line 1501: `Let scratch->L is` is probably meant to be something like `New search L is` or similar.

theantony: Same below in line 1562.

src/index.cpp
```diff
@@ -1545,37 +1554,43 @@ namespace diskann {
...
+    NeighborPriorityQueue &best_L_nodes = scratch->best_l_nodes();
+    assert(best_L_nodes.size() <= L);
+
     std::shared_lock<std::shared_timed_mutex> tl(_tag_lock);
-    size_t                                    pos = 0;
 
+    size_t pos = 0;
     for (int i = 0; i < (int) L; ++i) {
       TagT tag;
+      auto node = best_L_nodes[i];
```
theantony, line 1578: We don't know that there are `L` entries in `best_L_nodes`, could be less. And we seem to be using `i` just for accessing that vector.

How about changing the `for` loop to say:
```
for (const auto& node : best_L_nodes) {
```

That way we only iterate over what's actually inside `best_L_nodes`.

src/index.cpp
```diff
@@ -947,7 +951,10 @@ namespace diskann {
     assert(std::is_sorted(pool.begin(), pool.end()));
     if (pool.size() > maxc)
       pool.resize(maxc);
-    auto occlude_factor = scratch->occlude_factor();
+    auto &occlude_factor = scratch->occlude_factor();
+    // occlude_list can be called with the same scratch more than once by
+    // search_for_point_and_add_link through inter_insert. 
+    occlude_factor.clear();  
     occlude_factor.insert(occlude_factor.end(), pool.size(), 0);
```
theantony, line 958: Another way to write the line with `insert` is to say `occlude_factor.resize(pool.size());`

src/index.cpp
```diff
@@ -1486,11 +1493,13 @@ namespace diskann {
...
+    if (L > scratch->get_L()) {
+      diskann::cout << "Attempting to expand query scratch_space. Was created "
+                    << "with Lsize: " << scratch->get_L()
+                    << " but search L is: " << L << std::endl;
+      scratch->resize_for_new_L(L);
+      diskann::cout << "Resize completed. Let scratch->L is "
+                    << scratch->get_L() << std::endl;
     }
 
     return search_impl(query, K, L, indices, distances, scratch);
```
theantony, line 1505: `search_impl` has two issues (don't remember if it was fixed in some other branch, so commenting just in case):

1. `auto best_L_nodes = scratch->best_l_nodes();` is missing a reference after `auto`.
2. The `EXEC_ENV_OLS` path refers to `it.distance`, where `it` variable doesn't exist.

include/scratch.h
```diff
@@ -56,44 +75,42 @@ namespace diskann {
...
-    std::vector<Neighbor>    _pool;
-    tsl::robin_set<unsigned> _visited;
-    NeighborPriorityQueue    _best_l_nodes;
-    tsl::robin_set<unsigned> _inserted_into_pool_rs;
-    boost::dynamic_bitset<> *_inserted_into_pool_bs;
-    std::vector<unsigned>    _id_scratch;
-    float                   *_dist_scratch = nullptr;
+    // _pool stores all neighbors explored from best_L_nodes.
+    // Usually around L+R, but could be higher.
+    // Initialized to 10L, but expands as needed.
```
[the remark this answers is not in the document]
harsha-simhadri: I think the actual space needed is about L+R. So 10L is too much. Reducing to 3L+R. Will update the comments. The vector capacity grows with insertions anyway in case we underestimate.

src/index.cpp
harsha-simhadri: We need to initialize this vector to pool.size() many 0.0f for correctness.

src/index.cpp
harsha-simhadri: fixed

src/index.cpp
harsha-simhadri: Yes, that is the other commit in a separate PR I was going to make to fix exactly this. Do you want it in this PR? I can add it here too

src/scratch.cpp
```diff
@@ -24,75 +24,56 @@ namespace diskann {
...
-      if (_interim_dists != nullptr) {
-        delete[] _interim_dists;
-      }
-      _interim_dists = new float[new_search_l];
-      search_l = new_search_l;
+  void InMemQueryScratch<T>::resize_for_new_L(uint32_t new_l) {
+    if (new_l > _L) {
+      _L = new_l;
+      _pool.reserve(3 * _L + _R);
+      _best_l_nodes.reserve(_L);
```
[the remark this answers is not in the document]
harsha-simhadri: The NeighborPriorityQueue allocate L+1 slots when we reserve L

src/index.cpp
harsha-simhadri: You are right. Changed it to stop at best_L_nodes.size()

src/index.cpp
```diff
@@ -1522,7 +1522,8 @@ namespace diskann {
                                  // and IDType will be uint32_t or uint64_t
         if (distances != nullptr) {
 #ifdef EXEC_ENV_OLS
-          distances[pos] = it.distance;  // DLVS expects negative distances
+          distances[pos] =
+              bes_L_nodes[i].distance;  // DLVS expects negative distances
```
theantony, line 1526: Minor: missing `t` in `best_L_nodes`.

src/pq.cpp
```diff
@@ -262,6 +262,36 @@ namespace diskann {
...
+                        const _u64 ndims, _u8* out) {
+    for (_u64 i = 0; i < ids.size(); i++) {
+      memcpy(out + i * ndims, all_coords + ids[i] * ndims, ndims * sizeof(_u8));
+    }
+  }
+
+   void pq_dist_lookup(const _u8* pq_ids, const _u64 n_pts,
+                      const _u64 pq_nchunks, const float* pq_dists,
+                      std::vector<float> &dists_out) {
+    //_mm_prefetch((char*) dists_out, _MM_HINT_T0);
```
theantony, line 275: Interesting that the original method was prefetching `dists_out`, but then was clearing the array. I agree that it seems unnecessary.

src/scratch.cpp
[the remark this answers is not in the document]
theantony: Ah, I see, thanks

theantony (approved):

PR #217, opened 2023-01-26
default shuffle in cache_bfs_levels to false

harsha-simhadri:
 
Disables shuffle by default in cache_bfs_levels, which coupled with io_limit, leads to non-deterministic behavior in cached_beam_Search.

### DIFF
--- a/src/pq_flash_index.cpp
+++ b/src/pq_flash_index.cpp
@@ -367,6 +367,7 @@ namespace diskann {
            cur_level->size() == 0);
 
     node_list.clear();
+    node_list.reserve(node_set.size() + cur_level->size());
     for (auto node : node_set)
       node_list.push_back(node);
     for (auto node : *cur_level)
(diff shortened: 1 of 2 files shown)


src/pq_flash_index.cpp
```diff
@@ -343,37 +346,34 @@ namespace diskann {
...
-
-    std::shuffle(cur_level_node_list.begin(), cur_level_node_list.end(), urng);
-    size_t residual = num_nodes_to_cache - node_list.size();
+    assert(node_set.size() + cur_level->size() == num_nodes_to_cache ||
+           cur_level->size() == 0);
 
-    for (size_t i = 0; i < (std::min)(residual, cur_level_node_list.size());
-         i++)
-      node_list.push_back(cur_level_node_list[i]);
+    node_list.clear();
```
ltan1ms, line 369: nit: since size is known, can we reserve the size on the `node_list `first?

ltan1ms (approved):
thanks for the change!

src/pq_flash_index.cpp
harsha-simhadri: done

PR #228, opened 2023-02-02
Convert labels from strings to ints for in-memory build

varunsivashankar:
 
For in-memory build, this PR changes how labels are processed internally from strings to ints. This increases the speed of checking label set intersections between two points, resulting in faster build time and higher QPS at search time. 

#### Any other comments?
Code also written by Siddharth Gollapudi


varunsivashankar:

PR #229, opened 2023-02-02
labels from strings to ints for in-memory build

varunsivashankar:
 
For in-memory build, this PR changes how labels are processed internally from strings to ints. This increases the speed of checking label set intersections between two points, resulting in faster build time and higher QPS at search time.

#### Any other comments?
Code also written by Siddharth Gollapudi

### DIFF
--- a/include/index.h
+++ b/include/index.h
@@ -25,7 +25,6 @@
 #define DEFAULT_MAXC 750
 
 namespace diskann {
-    typedef unsigned label;
 
   inline double estimate_ram_usage(_u64 size, _u32 dim, _u32 datasize,
                                    _u32 degree) {
@@ -123,15 +122,13 @@ namespace diskann {
                                  Parameters              &parameters,
                                  const std::vector<TagT> &tags);
 
-
-// Filtered Support
+    // Filtered Support
     DISKANN_DLLEXPORT void build_filtered_index(
         const char *filename, const std::string &label_file,
         const size_t num_points_to_load, Parameters &parameters,
         const std::vector<TagT> &tags = std::vector<TagT>());
 
     DISKANN_DLLEXPORT void set_universal_label(const label &label);
-
 
     // Set starting point of an index before inserting any points incrementally
     DISKANN_DLLEXPORT void set_start_point(T *data);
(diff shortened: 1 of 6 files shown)


include/index.h
```diff
@@ -25,6 +25,8 @@
 #define DEFAULT_MAXC 750
 
 namespace diskann {
+    typedef unsigned label;
```
darvg, line 28: I recommend moving this typedef into `utils.h`: we may want to use it in places where `index.h` is not imported

include/index.h
```diff
@@ -239,17 +241,17 @@ namespace diskann {
                                               float                *distances,
                                               InMemQueryScratch<T> *scratch,
         bool               use_filters = false,
-        const std::string &filter_label = std::string());
+        const label &filter_label = -1);
```
darvg, line 244: `label` is an unsigned: maybe use `UINT_MAX` instead of `-1`?

tests/search_memory_index.cpp
```diff
@@ -56,8 +56,11 @@ int search_memory_index(diskann::Metric& metric, const std::string& index_path,
   }
 
   bool filtered_search = false;
-  if (filter_label != "")
+  diskann::label filter_label_as_num = -1;
```
darvg, line 59: same issue here: `diskann::label` is an unsigned: use `UINT_MAX` or some other appropriate value

darvg:
also run a clang format, there seems to be a few indentation issues

src/index.cpp
```diff
@@ -489,16 +489,17 @@ namespace diskann {
           std::istringstream iss(line);
           _u32               cnt = 0;
           _u32               medoid = 0;
-          std::string        label;
+          label        label;
```
darvg, line 492: A small indentation issue: run `clang-format` on all files included in the PR

darvg (approved):
lgtm

src/index.cpp
```diff
@@ -511,7 +512,7 @@ namespace diskann {
       if (file_exists(universal_label_file)) {
         std::ifstream universal_label_reader(universal_label_file);
         universal_label_reader >> _universal_label;
-        _use_universal_label = true;
+        _use_universal_label = true; // to check
```
dkaczynski: What is `// to check` referring to here?

varunsivashankar: This was to try debugging something with the disk build. Not relevant to in-memory

dkaczynski: I'd recommend either being more explicit in the comment or removing it if it's no longer relevant

varunsivashankar: done

dkaczynski (approved):

PR #231, opened 2023-02-05
Add `stitched_vamana` code

darvg:
 
Introduces the `stitched_vamana` algorithm as described in the WWW paper. Key changes include a new test for building such an index, and a small backwards-compatible API addition to `src/index.cpp`.

#### Any other comments?
Due to me cherry-picking, the commit history is abbreviated. Refer to [here](https://github.com/darvg/DiskANN/commits/ravi/stitched_rewrite_2) for a detailed history.

### DIFF
--- a/include/index.h
+++ b/include/index.h
@@ -192,6 +192,8 @@ namespace diskann {
     DISKANN_DLLEXPORT consolidation_report
     consolidate_deletes(const Parameters &parameters);
 
+		DISKANN_DLLEXPORT void prune_all_nbrs(const Parameters &parameters);
+
     DISKANN_DLLEXPORT bool is_index_saved();
 
     // repositions frozen points to the end of _data - if they have been moved
--- a/include/index.h
+++ b/include/index.h
@@ -192,7 +192,7 @@ namespace diskann {
     DISKANN_DLLEXPORT consolidation_report
     consolidate_deletes(const Parameters &parameters);
 
-		DISKANN_DLLEXPORT void prune_all_nbrs(const Parameters &parameters);
+    DISKANN_DLLEXPORT void prune_all_nbrs(const Parameters &parameters);
 
     DISKANN_DLLEXPORT bool is_index_saved();
 
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -1405,7 +1405,7 @@ namespace diskann {
                   _distance->compare(_data + _aligned_dim * (size_t) node,
                                      _data + _aligned_dim * (size_t) cur_nbr,
                                      (unsigned) _aligned_dim);
-              dummy_pool.emplace_back(Neighbor(cur_nbr, dist, true));
+              dummy_pool.emplace_back(Neighbor(cur_nbr, dist));
               dummy_visited.insert(cur_nbr);
             }
           }
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -483,6 +483,14 @@ namespace diskann {
     std::string labels_file = mem_index_file + "_labels.txt";
     std::string labels_to_medoids = mem_index_file + "_labels_to_medoids.txt";
     if (file_exists(labels_file)) {
+      std::string universal_label_file(filename);
+      universal_label_file += "_universal_label.txt";
+      if (file_exists(universal_label_file)) {
+        std::ifstream universal_label_reader(universal_label_file);
+        universal_label_reader >> _universal_label;
+        _use_universal_label = true;
+        universal_label_reader.close();
+      }
       parse_label_file(labels_file);
       if (file_exists(labels_to_medoids)) {
         std::ifstream medoid_stream(labels_to_medoids);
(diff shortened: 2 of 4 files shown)


tests/build_stitched_index.cpp
```diff
@@ -0,0 +1,855 @@
+// author: Siddharth Gollapudi
```
harsha-simhadri, line 1: Contributions are indicated by commit author. Please replace this header with Microsoft copyright and license info

include/index.h
```diff
@@ -192,6 +192,8 @@ namespace diskann {
     DISKANN_DLLEXPORT consolidation_report
     consolidate_deletes(const Parameters &parameters);
 
+		DISKANN_DLLEXPORT void prune_all_nbrs(const Parameters &parameters);
```
harsha-simhadri, line 195: not formatted

tests/build_stitched_index.cpp
```diff
@@ -0,0 +1,855 @@
...
+        "1.2 or 1.4 for denser graphs with lower diameter");
+    desc.add_options()(
+        "num_threads,T",
+        po::value<uint32_t>(&num_threads)->default_value(omp_get_num_procs()),
+        "Number of threads used for building index (defaults to "
+        "omp_get_num_procs())");
+    desc.add_options()("label_file",
+                       po::value<path>(&label_data_path)->default_value(""),
+                       "Input label file in txt format if present");
+    desc.add_options()(
```
harsha-simhadri, line 106: Some comments on universal lable would be helpful. 
Also, whenever you add a test tool, please document usage in ROOT/workflows/ directory. Maybe add a file that says "build_and_search_filtered_index.md". You would want to make it similar to the other readmes in there. You could write a small script to generate random labels for SIFT for testing.

tests/build_stitched_index.cpp
```diff
@@ -0,0 +1,855 @@
...
+/*
+ * Parses the label datafile, which has comma-separated labels on
+ * each line. Line i corresponds to point id i.
+ *
+ * Returns three objects via std::tuple:
+ * 1. map: key is point id, value is vector of labels said point has
+ * 2. map: key is label, value is number of points with the label
+ * 3. the label universe as a set
+ */
+parse_label_file_return_values parse_label_file(path  label_data_path,
```
harsha-simhadri, line 134: Isn't there a parsing function in david/filter branch already? Can that be reused?

tests/build_stitched_index.cpp
```diff
@@ -0,0 +1,855 @@
...
+
+/*
+ * Custom index save to write the in-memory index to disk.
+ * Also writes required files for diskANN API -
+ *  1. labels_to_medoids
+ *  2. universal_label
+ *  3. data (redundant for static indices)
+ *  4. labels (redundant for static indices)
+ */
+void save_full_index(path final_index_path_prefix, path input_data_path,
```
harsha-simhadri, line 497: Would we want this here or in the Index class as a save_filtered_index method?

tests/build_stitched_index.cpp
```diff
@@ -0,0 +1,855 @@
...
+}
+
+/*
+ * Unions the per-label graph indices together via the following policy:
+ *  - any two nodes can only have at most one edge between them -
+ *
+ * Returns the "stitched" graph and its expected file size.
+ */
+template<typename T>
+stitch_indices_return_values stitch_label_indices(
```
harsha-simhadri, line 607: Could many of these helper functions, not just this one, be moved to common utils files so that they can be reused by future code? Maybe have a src/filter_utils.cpp. Placing them here means they will not get reused in the future.

src/index.cpp
```diff
@@ -482,6 +482,14 @@ namespace diskann {
     std::string labels_file = mem_index_file + "_labels.txt";
     std::string labels_to_medoids = mem_index_file + "_labels_to_medoids.txt";
     if (file_exists(labels_file)) {
+      std::string universal_label_file(filename);
```
harsha-simhadri, line 485: What is the reason for moving up this code block here?

harsha-simhadri (changes requested):

tests/build_stitched_index.cpp
darvg: It looks redundant, but the `parse_label_file` in `index.cpp` differs in its treatment of the universal label. For the stitched algorithm, we need to expand the universal label into all labels for bookkeeping purposes; we don't want to do that in the case of membranne, due to performance costs with `std::set_intersection`.

tests/build_stitched_index.cpp
darvg: The index class already essentially has a "save_filtered_index" method. However, since the graph stitching occurs outside of the index API (via this file's `stitch_label_indices` function), the final result cannot be saved using that function. So, we simply redefine it here.

tests/build_stitched_index.cpp
darvg: While I'm not convinced as to the reusability of this code, I think that the level of refactoring required would take much more time and should constitute a subsequent PR on top of this one.

darvg:

tests/build_stitched_index.cpp
darvg: Resolved the comment for the universal labels. Working on an md for building/searching filtered indices to go in the workflows directory.

dkaczynski: I'm going to create an issue to track the need for a matching workflow doc, to be addressed before merging w/ main

PR #236, opened 2023-02-06
Clang-format now errors on push and PR if formatting is incorrect

### LINKED ISSUE #235: [BUG] CI format check disabled
## Expected Behavior
CI test should fail PRs with inconsistent formatting

## Actual Behavior
The check seems disabled.

## Example Code
Please see [How to create a Minimal, Reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) for some guidance on creating the best possible example of the problem
```bash

```

## Dataset Description
Please tell us about the shape and datatype of your data, (e.g. 128 dimensions, 12.3 billion points, floats)
- Dimensions: 
- Number of Points:
- Data type: 

## Error 
```
Paste the full error, with any sensitive information minimally redacted and marked $$REDACTED$$

```

## Your Environment
* Operating system (e.g. Windows 11 Pro, Ubuntu 22.04.1 LTS)
* DiskANN version (or commit built from)

## Additional Details
Any other contextual information you might feel is important.

daxpryce:
Rather than sift through all the *.cpp and *.h in the root directory, we're looking for only the sources in our main repository for formatting. Git submodules are excluded.

I had `--Werrors` off prior because "we" had a ton of errors in our source code coming from third parties (namely, CMake generate files - these are not files in version control, so we don't care if they're formatted as to our specifications, nor should we be modifying them directly anyway.  Let them be different if they want to be different.)

Now they're on, and the build will break until I format all of the source and commit it once again.

Closes #235


harsha-simhadri (approved):
looks good to me. When do we merge this? Right away or after formatting all files and fixing main?

daxpryce:
> looks good to me. When do we merge this? Right away or after formatting all files and fixing main?

We'll wait for #225, since it's probably going to fix most/all of the formatting files.  I'll pull main back into here, do any final formatting that may have to happen, then you can do final approval and merge :)

PR #239, opened 2023-02-09
Turn prune_neighbors into a no-op if pool input is empty

darvg:
 
Previously, `prune_neighbors` would do nothing if the input pool consisted solely of the node itself. However, the logic for verifying if the pool was a singleton was moved to be right before `prune_neighbors`: this would cause the pool passed to `prune_neighbors` to be empty, leading to an exception. 

We remove the exception from prune_neighbors, turning it into a noop in the case of an empty pool.

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -1157,8 +1157,9 @@ namespace diskann {
       const _u32 max_candidate_size, const float alpha,
       std::vector<unsigned> &pruned_list, InMemQueryScratch<T> *scratch) {
     if (pool.size() == 0) {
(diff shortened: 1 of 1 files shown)


harsha-simhadri:
@darvg does this now mean we are adding nodes to the graph without any out or in edges?

darvg:
@harsha-simhadri Yes. For the turing2m dataset, there are quite a few filters that only occur once, leading to singleton nodes. Because the dataset doesn't have a universal label, the nodes always end up isolated. However, the search for such nodes is trivial, because we store a medoid for every filter.

dkaczynski:
I have confirmed this fixes the empty pool exception as expected.

PR #244, opened 2023-02-14
Adding filter disk readme

NeelamMahapatro:


harsha-simhadri (changes requested):
Can we suggest that these changes are optional? 

Also we need to add command line example for building, computing GT and searching the SSD index.

dkaczynski:
I believe this PR has been deprecated by the addition of the dedicated workflow for filtered search: https://github.com/microsoft/DiskANN/blob/main/workflows/filtered_in_memory.md

However, I think there's still room for a couple improvements:
* Workflows for both inmemory and disk search could provide a link to the filtered search workflow for more info on label/filter-specific args
* The filter workflow could be updated to include disk indices for building/searching labeled (currently, the workflow only documents inmemory build/search)

PR #248, opened 2023-02-16
Converting Label type as Template parameter

NeelamMahapatro:
 
- This change allow user to set label_type either uint32_t or uint16_t format
- template parameter of Index and PQFlashIndex class has been changed to include new template parameter LabelT and accordingly in all their function definitions.
- All the private variables which stores label values have been modified as LabelT type


dkaczynski:
This change is working as expected. Thank you, @NeelamMahapatro!

dkaczynski (approved):

Issue #240, opened 2023-02-10
Converting string labels to disk labels within DiskANN


varunsivashankar:
## Is your feature request related to a problem? Please describe.
Currently the filter support code assumes that the label file contains the labels as integers. Changes need to be made to build to ensure that label files with string labels also work.

## Describe the solution you'd like
At build time, we simply convert the input label file to one with integers, save it along with all the other build files and use the new file for the build. Will need to maintain a map from the strings to integers to support dynamic inserts of points with text labels.

## Provide references (if applicable)
If your feature request is related to a published algorithm/idea, please provide links to 
any relevant articles or webpages.

PR #252, opened 2023-02-21
Internal Conversion from String to Integer Labels

NeelamMahapatro:
- Converting string to integer labels internally

String to Int Internal Conversion - 
https://github.com/microsoft/DiskANN/issues/240

Fixing comments - 

- Removed commented code
- Added file validation for label file and label to medoid file
- Shifted parse label logic after loading base file for comparing number of points (So that we have the number of points in the input file before parsing label file. Then we compare the number of lines in label.txt file with number of points in input file with assert logic)
- std::cout replaced with diskann::cout

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -31,6 +31,10 @@ typedef int FileHandle;
 #include "memory_mapped_files.h"
 #endif
 
+#include <unordered_map>
+#include <sstream>
+#include <iostream> 
+
 // taken from
 // https://github.com/Microsoft/BLAS-on-flash/blob/master/include/utils.h
 // round up X to the nearest multiple of Y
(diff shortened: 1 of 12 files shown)


include/utils.h
```diff
@@ -145,6 +149,48 @@ inline int delete_file(const std::string& fileName) {
...
+    string_int_map[unv_label] = 0;
+  std::string line, token;
+  while (std::getline(label_reader, line)) {
+    std::istringstream new_iss(line);
+    std::vector<_u32>lbls;
+    while (getline(new_iss, token, ',')) {
+        token.erase(std::remove(token.begin(), token.end(), '\n'), token.end());
+        token.erase(std::remove(token.begin(), token.end(), '\r'), token.end());
+        if (string_int_map.find(token) == string_int_map.end()) {
+            int nextId = string_int_map.size()+1;
```
varunsivashankar, line 169: Question: when universal label is set, will any label have value 1? Because size of string_int_map will be 1, so label numbers will start from 2.

tests/build_memory_index.cpp
```diff
@@ -48,18 +51,21 @@ int build_in_memory_index(const diskann::Metric& metric,
   if (label_file == "") {
     index.build(data_path.c_str(), data_num, paras);
   } else {
+    convert_labels_string_to_int(label_file, labels_file_to_use, 
+                                  mem_labels_int_map_file, universal_label);
     if (universal_label != "") {
       LabelT unv_label_as_num = std::stoul(universal_label);
```
varunsivashankar, line 57: Can we assume universal_label is the string representation of a number here, or do we have to use the get_converted_label function here instead?

darvg:
Is stitched vamana not in scope for the PR? I don't see any changes for it.

tests/build_memory_index.cpp
NeelamMahapatro: We can use get_converted_label function here

include/utils.h
NeelamMahapatro: yes it will start from 2. But its internal map anyways (does it have any side effect on building and searching ?)

dkaczynski:
@NeelamMahapatro What do you think about adding the string-to-int internal conversion to building an index with stitched vamana?

dkaczynski:
@NeelamMahapatro @darvg 

I'm going to opt to address the stitched vamana str-to-int label conversion in a separate branch/PR and consider the scope of this PR as it stands

dkaczynski:
Sorry, I made a mistake when validating this on Linux. I rolled back `dkaczynski/filter` and created a new PR for this branch, #255

PR #256, opened 2023-02-28
Refactored Parameters class.

daxpryce:
The existing Parameters class tried to make for a generic, type-agnostic configuration class.

After finding a double free bug rooted in the destructor of Parameters and not being able to satisfactorily rectify it as is, we decided a simple refactor and a very simple set of configuration objects would help sanitize things and unblock us from using this in Python, which invariably resulted in either a munmap: invalid pointer, free: invalid size, double free (!prev), or segfault

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -118,7 +118,6 @@ jobs:
 
     - name: Generate 10K random int8 index vectors, 1K query vectors, in 10 dims and compute GT
       run: |
-      run: |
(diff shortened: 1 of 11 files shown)


bryantower:
I just want to note that the issue we saw was with `new` being called to allocate memory and `free` was being called to release memory. This was happening because we lost the type information by casting things through (void*).

bryantower (approved):

harsha-simhadri:
@dkaczynski Dax suggests this refactoring of the parameter class. I requested him to wait until your PR225 till he pushes this through. Please leave your comments on this PR, and if you like it, please consider using the same organization for the filtered build/search parameters too.

daxpryce:
Superseded by #300

PR #257, opened 2023-03-03
Fix BANN build failure after adding streamingDiskANN.

NingyuanChen:
When EXEC_ENV_OLS is defined, build will fail because of load_pq_centroid_bin function signature mismatch.

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -1427,9 +1427,9 @@ namespace diskann {
 #ifdef EXEC_ENV_OLS
       throw ANNException("load_pq_centroid_bin should not be called when EXEC_ENV_OLS is defined.",
                          -1, __FUNCSIG__, __FILE__, __LINE__);
-      //pq_table.load_pq_centroid_bin(files, pq_pivots_file.c_str(), _num_pq_chunks);
+      //_pq_table.load_pq_centroid_bin(files, pq_pivots_file.c_str(), _num_pq_chunks);
 #else
-      pq_table.load_pq_centroid_bin(pq_pivots_file.c_str(), _num_pq_chunks);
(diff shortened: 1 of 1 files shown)


src/index.cpp
```diff
@@ -1424,7 +1424,11 @@ namespace diskann {
       copy_aligned_data_from_file<_u8>(pq_compressed_file.c_str(), _pq_data,
                                        file_num_points, _num_pq_chunks,
                                        _num_pq_chunks);
-      _pq_table.load_pq_centroid_bin(pq_pivots_file.c_str(), _num_pq_chunks);
+#ifdef EXEC_ENV_OLS
+      //pq_table.load_pq_centroid_bin(files, pq_pivots_file.c_str(), _num_pq_chunks);
```
PhilipBAdams, line 1428: Can we throw an exception here for now, I think the behavior will be unexpected if pq_table is not  loaded

PhilipBAdams (approved):

src/index.cpp
```diff
@@ -1424,7 +1424,13 @@ namespace diskann {
       copy_aligned_data_from_file<_u8>(pq_compressed_file.c_str(), _pq_data,
                                        file_num_points, _num_pq_chunks,
                                        _num_pq_chunks);
+#ifdef EXEC_ENV_OLS
+      throw ANNException("load_pq_centroid_bin should not be called when EXEC_ENV_OLS is defined.",
+                         -1, __FUNCSIG__, __FILE__, __LINE__);
+      //_pq_table.load_pq_centroid_bin(files, pq_pivots_file.c_str(), _num_pq_chunks);
```
harsha-simhadri, line 1430: remove comment?

harsha-simhadri (approved):

PR #282, opened 2023-03-22
Some fixes to pass internal building pipeline

yiylin:
- [ ] Does this PR add any new dependencies? No
- [ ] Does this PR modify any existing APIs? No
- [ ] Should this result in any changes to our documentation, either updating existing docs or adding new ones? No
 
Fix some compile errors in internal building pipeline


include/utils.h
```diff
@@ -31,6 +31,9 @@ typedef int FileHandle;
 #include "memory_mapped_files.h"
 #endif
 
+#undef max
```
harsha-simhadri, line 34: Why is this needed?

src/pq_flash_index.cpp
```diff
@@ -1178,7 +1178,6 @@ void PQFlashIndex<T, LabelT>::cached_beam_search(const T *query1, const _u64 k_s
     unsigned cmps = 0;
     unsigned hops = 0;
     unsigned num_ios = 0;
-    unsigned k = 0;
```
harsha-simhadri, line 1181: where is this extra variable used?

include/utils.h
yiylin: it could conflict with some macro

yiylin: https://github.com/microsoft/DiskANN/blob/main/CMakeLists.txt . here you also configure "NOMINMAX". but I could not use this way in our project, or else it will cause other compile errors

src/pq_flash_index.cpp
yiylin: no code uses this variable. our internal building pipeline catches this

src/distance.cpp
```diff
@@ -79,8 +79,6 @@ float SlowDistanceCosineUInt8::compare(const uint8_t *a, const uint8_t *b, uint3
 
 float DistanceL2Int8::compare(const int8_t *a, const int8_t *b, uint32_t size) const
 {
-    int32_t result = 0;
```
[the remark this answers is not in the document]
yiylin: In the some case(Window), this variable is defined, but no place uses it. so it causes the compiler error in our internal building pipeline

include/utils.h
harsha-simhadri: @theantony Does this change affect you? Asking since you included the -DNOMINMAX in windows cmake options.

theantony: Thanks @harsha-simhadri ! @yiylin What is the scenario where you're encountering the issue? Are you building DiskANN as part of your pipeline and the build breaks? Or are you using pre-built DiskANN as nuget package or similar and are getting build breaks when using DiskANN headers?

yiylin: I use this repo as submodule, then include this header to our project. Do you have the pre-built DiskANN nuget package?

theantony: We build DiskANN as part of another nuget, it's not an "official" nuget.

I think these #undef calls are not the right place in DiskANN, they may cause issues with projects using DiskANN.

If you're building DiskANN as submodule, won't you get `NOMINMAX` defined so you won't have this issue? Or if you're using your custom build setting, can you add `NOMINMAX` just for DiskANN?

Then if you really need to #undef min or max, you can do it before including DiskANN headers and not impact other users of DiskANN.

yiylin: fine. if you have concerns, then I could remove this. I will try other ways to fix the issue in our side.

theantony: Thanks! Looks good.

theantony (approved):

PR #285, opened 2023-03-24
GT computation with individual labels for queries

varunsivashankar:
 
Based on the queries that appear in the query label file, both the base bin file and the query bin file are split into label specific ones. Then individual searches are run for each label, and then finally the label specific results are combined. Some bookkeeping was required to combine the label specific ground truth files.


yashpatel007:
This has been merged to https://github.com/microsoft/DiskANN/tree/varunsivashankar/support_multiple_filterlabels.

PR #288, opened 2023-03-29
add codebook passing and pq/opq dim overwrite.

jinwei14:
- [X] Does this PR have a descriptive title that could go in our release notes?
- [X] Should this result in any changes to our documentation, either updating existing docs or adding new ones?
 
1. this change will pass in a pretrained codebook prefix. for pq and OPQ part it will load the pretrained roatation matrix and pivots.
2. changed the default max PQ dim from 256 to 384 to adopt richer embeddings.
3. we will need to update readme for new parameters if this PR is accepted.


src/disk_utils.cpp
```diff
@@ -1047,21 +1050,22 @@ int build_disk_index(const char *dataFilePath, const char *indexFilePath, const
...
+             "B (RAM limit of final index in GB)\n"
+             "M (memory limit while indexing)\n"
+             "T (number of threads for indexing)\n"
+             "B' (PQ bytes for disk index: optional parameter for "
+             "very large dimensional data)\n"
+             "reorder (set true to include full precision in data file"
+             ": optional paramter, use only when using disk PQ\n"
+             "build_PQ_byte (number of PQ bytes for inde build; set 0 to use "
+             "full precision vectors)\n"
+             "QD Quantized Dimension"
```
PhilipBAdams, line 1066: need some logic and documentation around how it interacts with the memory limit, `B` -- will quantized dimension only apply if it fits inside the memory limit, or does it override it?

tests/build_disk_index.cpp
```diff
@@ -70,24 +74,23 @@ int main(int argc, char **argv)
...
+    po::notify(vm);
+    if (vm["append_reorder_data"].as<bool>())
+      append_reorder_data = true;
+    if (vm["use_opq"].as<bool>())
+      use_opq = true;
+  } catch (const std::exception& ex) {
+    std::cerr << ex.what() << '\n';
+    return -1;
+  }
+  codebook_prefix = codebook_prefix == "0"? "" : codebook_prefix;
```
PhilipBAdams, line 93: Why is the default value 0 here?

PhilipBAdams:
One other thing to consider, do you want to support overriding codebook for disk PQ too? We don't currently use disk PQ in production

tests/build_disk_index.cpp
jinwei14: it's easier for the loading script in PowerShell to pass in "0" rather than empty. maybe I will rewrite it to be easier to understand.

src/disk_utils.cpp
jinwei14: sure, can do.

jinwei14:
> One other thing to consider, do you want to support overriding codebook for disk PQ too? We don't currently use disk PQ in production

Idk, what do you think? shall I also do it in diskpq?

src/disk_utils.cpp
```diff
@@ -1036,8 +1036,9 @@ void create_disk_layout(const std::string base_file, const std::string mem_index
 
 template <typename T, typename LabelT>
 int build_disk_index(const char *dataFilePath, const char *indexFilePath, const char *indexBuildParameters,
-                     diskann::Metric compareMetric, bool use_opq, bool use_filters, const std::string &label_file,
-                     const std::string &universal_label, const _u32 filter_threshold, const _u32 Lf)
+                     diskann::Metric compareMetric, bool use_opq, const std::string &codebook_prefix, bool use_filters,
```
gopalrs, line 1039: Should we bite the bullet and make all the const char*'s as std::string&?

src/disk_utils.cpp
```diff
@@ -1036,8 +1036,9 @@ void create_disk_layout(const std::string base_file, const std::string mem_index
...
+                     const std::string &label_file, const std::string &universal_label, const _u32 filter_threshold,
+                     const _u32 Lf)
```
gopalrs, line 1041: This is probably fixed in 290, but _u32

src/disk_utils.cpp
```diff
@@ -1047,7 +1048,7 @@ int build_disk_index(const char *dataFilePath, const char *indexFilePath, const
     {
         param_list.push_back(cur_param);
     }
-    if (param_list.size() < 5 || param_list.size() > 8)
+    if (param_list.size() < 5 || param_list.size() > 9)
```
gopalrs, line 1051: I suppose the new parameters' classes will change this, right?

src/disk_utils.cpp
```diff
@@ -1107,7 +1109,9 @@ int build_disk_index(const char *dataFilePath, const char *indexFilePath, const
     std::string labels_file_original = label_file;
     std::string index_prefix_path(indexFilePath);
     std::string labels_file_to_use = index_prefix_path + "_label_formatted.txt";
```
gopalrs, line 1111: Should we make all the suffixes as const strings in some header file?

src/disk_utils.cpp
```diff
@@ -1333,43 +1342,31 @@ template DISKANN_DLLEXPORT uint32_t optimize_beamwidth<float, uint16_t>(
...
-                                                                   bool use_filters, const std::string &label_file,
-                                                                   const std::string &universal_label,
-                                                                   const _u32 filter_threshold, const _u32 Lf);
-template DISKANN_DLLEXPORT int build_disk_index<float, uint32_t>(const char *dataFilePath, const char *indexFilePath,
-                                                                 const char *indexBuildParameters,
-                                                                 diskann::Metric compareMetric, bool use_opq,
-                                                                 bool use_filters, const std::string &label_file,
-                                                                 const std::string &universal_label,
-                                                                 const _u32 filter_threshold, const _u32 Lf);
+template DISKANN_DLLEXPORT int build_disk_index<int8_t, uint32_t>(
```
gopalrs, line 1345: Some _u32 variables here, but I suppose they will be fixed by the other PR.

tests/build_disk_index.cpp
```diff
@@ -41,6 +42,9 @@ int main(int argc, char **argv)
         desc.add_options()("num_threads,T", po::value<uint32_t>(&num_threads)->default_value(omp_get_num_procs()),
                            "Number of threads used for building index (defaults to "
                            "omp_get_num_procs())");
+        desc.add_options()("QD", po::value<uint32_t>(&QD)->default_value(0), " Quantized Dimension for compression");
+        desc.add_options()("codebook_prefix", po::value<std::string>(&codebook_prefix)->default_value(""),
```
gopalrs, line 46: Shouldn't this be a full path to a file?

gopalrs (changes requested):

harsha-simhadri:
> > One other thing to consider, do you want to support overriding codebook for disk PQ too? We don't currently use disk PQ in production
> 
> Idk, what do you think? shall I also do it in diskpq?

I am OK with leaving this out unless you need this too.

src/disk_utils.cpp
harsha-simhadri: we should

src/disk_utils.cpp
harsha-simhadri: am OK, will DLVS or some other wrapper break.

include/pq.h
```diff
@@ -10,7 +10,7 @@
 #define MAX_OPQ_ITERS 20
 #define NUM_KMEANS_REPS_PQ 12
 #define MAX_PQ_TRAINING_SET_SIZE 256000
-#define MAX_PQ_CHUNKS 256
+#define MAX_PQ_CHUNKS 384
```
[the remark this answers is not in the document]
harsha-simhadri: lets just bump it up to 512.

src/disk_utils.cpp
gopalrs: Oh, BANN might. The old DLVS still uses the "randnsg" sources, so it should be OK.

include/pq.h
```diff
@@ -118,5 +118,6 @@ void generate_disk_quantized_data(const std::string data_file_to_use, const std:
 template <typename T>
 void generate_quantized_data(const std::string data_file_to_use, const std::string pq_pivots_path,
                              const std::string pq_compressed_vectors_path, const diskann::Metric compareMetric,
-                             const double p_val, const size_t num_pq_chunks, const bool use_opq);
+                             const double p_val, const size_t num_pq_chunks, const bool use_opq,
+                             const std::string codebook_path);
```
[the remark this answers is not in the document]
jinwei14: can we do that in a separate PR, this need to modify existing parameters.

tests/build_disk_index.cpp
jinwei14: no, because OPQ need rotation matrix as well, it's better to pass in a prefix..

src/disk_utils.cpp
jinwei14: > Should we make all the suffixes as const strings in some header file?

same suggestion as above :)

src/pq.cpp
```diff
@@ -961,35 +961,41 @@ void generate_disk_quantized_data(const std::string data_file_to_use, const std:
 template <typename T>
 void generate_quantized_data(const std::string data_file_to_use, const std::string pq_pivots_path,
                              const std::string pq_compressed_vectors_path, diskann::Metric compareMetric,
-                             const double p_val, const size_t num_pq_chunks, const bool use_opq)
+                             const double p_val, const size_t num_pq_chunks, const bool use_opq,
+                             const std::string codebook_prefix)
```
harsha-simhadri, line 965: use a default parameter "" for codebook_prefix ?

src/index.cpp
```diff
@@ -1773,7 +1773,7 @@ void Index<T, TagT, LabelT>::build(const char *filename, const size_t num_points
         auto pq_pivots_file = std::string(filename) + suffix + "_pivots.bin";
         auto pq_compressed_file = std::string(filename) + suffix + "_compressed.bin";
         generate_quantized_data<T>(std::string(filename), pq_pivots_file, pq_compressed_file, _dist_metric, p_val,
-                                   _num_pq_chunks, _use_opq);
+                                   _num_pq_chunks, _use_opq, "");
```
harsha-simhadri, line 1776: let's remove this change and make "" default parameter for generate_quantized_data.

tests/build_disk_index.cpp
```diff
@@ -41,6 +42,9 @@ int main(int argc, char **argv)
         desc.add_options()("num_threads,T", po::value<uint32_t>(&num_threads)->default_value(omp_get_num_procs()),
                            "Number of threads used for building index (defaults to "
                            "omp_get_num_procs())");
+        desc.add_options()("QD", po::value<uint32_t>(&QD)->default_value(0), " Quantized Dimension for compression");
```
harsha-simhadri, line 45: let's allow only one of QD or B parameters?

src/disk_utils.cpp
```diff
@@ -1061,7 +1062,8 @@ int build_disk_index(const char *dataFilePath, const char *indexFilePath, const
                          "reorder (set true to include full precision in data file"
                          ": optional paramter, use only when using disk PQ\n"
                          "build_PQ_byte (number of PQ bytes for inde build; set 0 to use "
-                         "full precision vectors)"
+                         "full precision vectors)\n"
+                         "QD Quantized Dimension to overwrite the derived dim from B "
```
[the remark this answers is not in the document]
jinwei14: > should we make this one or the other? Either you can specify memory budget or you can just #quantized bytes but not both?

what I intend to do here is to use -QD (quantized dim) to overwrite the existing derived pq dimension from-B, this is more backswords compactable. and do not need to force the prod DLVS diskann index build to pass in just one of the params

src/pq.cpp
jinwei14: > use a default parameter "" for codebook_prefix ?

good suggestion, done

src/index.cpp
jinwei14: > let's remove this change and make "" default parameter for generate_quantized_data.

good suggestion, done

tests/build_disk_index.cpp
jinwei14: > let's allow only one of QD or B parameters?

same reason above.

src/disk_utils.cpp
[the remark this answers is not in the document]
harsha-simhadri: The existing pipelines will be fine since we still allow B-only parameter.  For new pipelines, allowing both QD and B is confusing. If QD is greater than what B allows, will quantization be QD or B. I prefer that we allow user to specify B or QD and not both together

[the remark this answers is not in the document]
jinwei14: > The existing pipelines will be fine since we still allow B-only parameter. For new pipelines, allowing both QD and B is confusing. If QD is greater than what B allows, will quantization be QD or B. I prefer that we allow user to specify B or QD and not both together

added check to disable both params pass in

include/pq.h
```diff
@@ -120,5 +120,6 @@ void generate_disk_quantized_data(const std::string data_file_to_use, const std:
 template <typename T>
 void generate_quantized_data(const std::string data_file_to_use, const std::string pq_pivots_path,
                              const std::string pq_compressed_vectors_path, const diskann::Metric compareMetric,
-                             const double p_val, const uint64_t num_pq_chunks, const bool use_opq);
+                             const double p_val, const uint64_t num_pq_chunks, const bool use_opq,
+                             const std::string codebook_prefix = "");
```
gopalrs, line 124: const std::string& ?

src/disk_utils.cpp
gopalrs, line 1066: I think it makes sense to have QD override the derived dimension.

gopalrs (approved):

PR #298, opened 2023-04-04
Patelyash/index refactor

yashpatel007:


include/index_factory.h
```diff
@@ -0,0 +1,190 @@
...
+    bool load_on_build = false;
+
+    std::string data_type;
+    std::string label_type;
+    std::string tag_type;
+};
+
+struct BuildParams
+{
+    uint32_t R;
```
daxpryce, line 49: opening discussion here: what say everyone to using more descriptive names for these parameters?  I took a stab at naming them in #300 for the old way of providing params, but I'm open to name changes from the ones I picked. 

I like single char variables in whitepapers, but I have a strong dislike for them when I'm trying to process the code I'm reading, especially if I'm not super familiar.

include/index_factory.h
```diff
@@ -0,0 +1,190 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
daxpryce, line 1: Is the idea to have an IndexFactory and a DynamicIndexFactory, or should we figure out how the dynamic indices come into here?  Is it two enums, or a singular set of enums like `DYNAMIC_DISK` and `STATIC_MEMORY`.  Also, how does Disk and SSD differ?

include/index_factory.h
yashpatel007: I agree, and this was created with just POC. There is so much that we can improve here. First off reading those vars from a yml config file would be great.  Then as your #300 PR can be very helpful here.  I am planning to add yml-cpp library to do the config reading.

include/index_factory.h
yashpatel007: I was thinking more like, we have a flag if the build is dynamic or not, while creating the Instance of Disk or Memory Index we can use this flag to appropriately instantiate the class.  Let me look into this, thanks for pointing this out.

harsha-simhadri:
replaced with #311

PR #299, opened 2023-04-04
updated dockerfile

harsha-simhadri:
 
#### Reference Issues/PRs
Addresses #109

Adds a dockerfile to clone, install deps and build the core C++ code

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,4 +14,4 @@ RUN git clone https://github.com/microsoft/DiskANN.git
 WORKDIR /app/DiskANN
 RUN mkdir build
 RUN cmake -S . -B build  -DCMAKE_BUILD_TYPE=Release
-RUN cmake --build build
+RUN cmake --build build -- -j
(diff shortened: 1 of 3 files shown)


harsha-simhadri:
@daxpryce, do we want to add CI test for docker in this PR?

Dockerfile
```diff
@@ -1,13 +1,17 @@
...
-
-RUN mkdir -p build && cd build && \
-    cmake -DCMAKE_BUILD_TYPE=Release .. && \
-    make -j $(nproc)
+WORKDIR /app
+RUN git clone https://github.com/microsoft/DiskANN.git 
+WORKDIR /app/DiskANN
+RUN mkdir build
+RUN cmake -S . -B build  -DCMAKE_BUILD_TYPE=Release
+RUN cmake --build build
```
daxpryce, line 17: I just found out this actually should be:

`RUN cmake --build build -- -j` for parallel builds.  I wasted so much time until I found this out.

daxpryce:
> @daxpryce, do we want to add CI test for docker in this PR?

Yes, though we could just as arguably move the entire test battery we run for the ubuntu-latest in this container too and kill 2 birds with one stone.  For the short term I'd just add step for a `docker build` and presume that if the build works and the tests outside the container work then they also work in container too.  It's up to you on how fancy you want to make this change

Dockerfile
harsha-simhadri: added

PR #300, opened 2023-04-04
Python API and Test Suite

### LINKED ISSUE #194: Pythonic wrapper for python extension module

### LINKED ISSUE #195: Python Unit Tests

daxpryce:
  - dax here: imo the title should be "First Python Release (Linux)", but that's a bad PR title
 
#### Reference Issues/PRs
Closes #195 and closes #194

This is a major refactor/rewrite of the existing PQ Flash index python module.  It includes:
- A python unit test suite
- Restricts only to Linux builds (windows are still not creating a dll for packaging for some reason)
- Has a more pythonic api to disk indices based on numpy array type
- pydoc docstrings that we might be able to publish documentation for immediately (still work to be done on that)
- Fixes a bug when the Parameters class was being deleted, but only showed up when running within a Python module


pyproject.toml
```diff
@@ -3,7 +3,7 @@ requires = [
     "setuptools>=59.6",
     "pybind11>=2.10.0",
     "cmake>=3.22",
-    "numpy>=1.21", 
+    "numpy>=1.21",
     "wheel",
 ]
 build-backend = "setuptools.build_meta"
```
harsha-simhadri, line 9: bump up version to 0.5

include/index.h
```diff
@@ -79,7 +79,7 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
 
     // Constructor for incremental index
     DISKANN_DLLEXPORT Index(Metric m, const size_t dim, const size_t max_points, const bool dynamic_index,
-                            const Parameters &indexParameters, const Parameters &searchParameters,
+                            const IndexWriteParameters &indexParameters, const IndexReadParameters &searchParameters,
```
harsha-simhadri, line 82: Shall we call these classes IndexBuildParameters and IndexSearchParameters?
Also, do we really need IndexSearchParameters in the ctor? Can we not just pass the number of threads to create scratch space for? The Ls parameter is passed directly in search anyway.

daxpryce: I would prefer to pass fewer of these params objects around (if you notice I did make an executive decision to do that on one of the method/function calls, `prune_all_neighbors` I think.)  I'd much prefer not to have extra objects if we don't need them

python/tests/test_search_disk_index.py
```diff
@@ -1,105 +0,0 @@
-# Copyright (c) Microsoft Corporation. All rights reserved.
-# Licensed under the MIT license.
```
harsha-simhadri, line 2: can we move these files that you are deleting to ROOT/python/apps/  or ROOT/apps/ folder instead with suitable modifications.
These would be good examples for people to use our python interface. We would want to document such a workflow in workflows/python_interface.md as well.

src/disk_utils.cpp
```diff
@@ -628,34 +628,26 @@ int build_merged_vamana_index(std::string base_file, diskann::Metric compareMetr
...
+        diskann::IndexWriteParameters paras = diskann::IndexWriteParametersBuilder(L, R)
+                                                  .with_filter_list_size(Lf)
+                                                  .with_saturate_graph(!use_filters)
+                                                  .build();
         using TagT = uint32_t;
-        paras.Set<std::string>("save_path", mem_index_path);
-        std::unique_ptr<diskann::Index<T, TagT, LabelT>> _pvamanaIndex =
-            std::unique_ptr<diskann::Index<T, TagT, LabelT>>(new diskann::Index<T, TagT, LabelT>(
-                compareMetric, base_dim, base_num, false, false, false, build_pq_bytes > 0, build_pq_bytes, use_opq));
+        diskann::Index<T, TagT, LabelT> _pvamanaIndex(compareMetric, base_dim, base_num, false, false, false,
```
harsha-simhadri, line 637: Let's use lower case and call it _index?

tests/test_insert_deletes_consolidate.cpp
```diff
@@ -142,20 +142,16 @@ void build_incremental_index(const std::string &data_path, const uint32_t L, con
...
-                              // value to make it backward compatible.
-    params.Set<uint32_t>("num_frozen_pts", num_start_pts);
+
+    diskann::IndexWriteParameters params = diskann::IndexWriteParametersBuilder(L, R)
+                                               .with_max_occlusion_size(500) // C = 500
+                                               .with_alpha(alpha)
+                                               .with_num_rounds(1)
+                                               .with_num_threads(thread_count)
+                                               .with_num_frozen_points(num_start_pts)
+                                               .build();
```
harsha-simhadri, line 152: I like this!

harsha-simhadri (changes requested):
Mostly looks good, requesting a few minor changes. Let me know if you want me to do these

python/tests/test_search_disk_index.py
daxpryce: originally I had a folder named "examples" inside of python but it was using the native interface, not the python wrapper.  I can certainly come up with some good examples though!

include/index.h
daxpryce: > Shall we call these classes IndexBuildParameters ...

I went with IndexWriteParameters because it's for both the major build stuff and any incremental/dynamic changes too.  I didn't want to imply it was only used for building the whole index, and IndexMutationParameters was too dang long.  I don't really have strong feelings about this though

python/tests/test_search_disk_index.py
daxpryce: #303 will make sure we get those examples and documentation added prior to actual release

include/index.h
harsha-simhadri: I think of index building as both a batch and incremental operation. Your call on this.

Also, can we rename indexParameters to write_params or build_params ?

daxpryce: yeah, I'll rename it in a future PR - I like write_params

harsha-simhadri (approved):

PR #308, opened 2023-04-07
Add InMemory indice support to python

daxpryce:


python/src/py.typed
harsha-simhadri, line 1: IS this empty file necessary?

daxpryce: yep, it's a tombstone file that informs python that we have type hints in our code and that you should use them rather than checking the typeshed

harsha-simhadri (approved):

PR #309, opened 2023-04-07
Fix StaticIndex class and add examples

harsha-simhadri:
 
Add index.build to StaticIndex python binding ctor

### DIFF
--- a/python/apps/in-mem-dynamic.py
+++ b/python/apps/in-mem-dynamic.py
@@ -7,7 +7,10 @@ import numpy as np
 import utils
 
 
-def insert_and_search(dtype_str, indexdata_file, querydata_file, Lb, graph_degree, K, Ls, num_threads):
+def insert_and_search(dtype_str, indexdata_file, querydata_file,
+                      Lb, graph_degree,
+                      K, Ls, num_threads,
+                      gt_file):
     npts, ndims = utils.get_bin_metadata(indexdata_file)
 
     if dtype_str == "float":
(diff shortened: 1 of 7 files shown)


python/src/_wrapper.py
```diff
@@ -438,7 +438,7 @@ def batch_search(
 
 class StaticMemoryIndex:
     def __init__(
-        self, metric: Literal["l2", "mips"], vector_dtype: VectorDType, index_path: str
+        self, metric: Literal["l2", "mips"], vector_dtype: VectorDType, index_path: str, l_build: int, graph_degree: int
```
daxpryce, line 441: We've been using list_size and max_degree. 

FWIW, I'm not happy with either of them, as they don't actually tell me what they do or are for.  With `list_size` I can actually know it's the size of a list of some sort, but I don't know why or how it's being used.  `l_build` means nothing to me unless I also have the paper open at the same time.  `graph_degree` | `max_degree` have about the same problem.  

Regardless of what we pick, I want us to be consistent, and I want us to be descriptive about intent in our naming, with the code being self descriptive enough it won't need a whitepaper to accompany it for it to make sense.  So, I'm open to anything that might hit all of these!

python/apps/in-mem-dynamic.py
```diff
@@ -0,0 +1,58 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
daxpryce, line 1: Couple things here:

1: `fire` is a way better library than `argparse` to make things simple and runnable.  I'll update these later to use it, but I'm a huge fan in how easy it makes stuff!
2:  Eventually I'd like to move these out to like `examples/python/` in the top level, and we can throw in some jupyter notebooks too.  it doens't need to go there right now, just wanted to comment on my intent for these kinds of py files :)

python/apps/utils.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
daxpryce, line 1: A file similar to this should be packaged up with our python at some point.  Before we made the push for the inmem indices last week it was on my radar to do!  I want writers and readers and quality of life utility methods right in the library.

PR #311, opened 2023-04-11
Patelyash/refactor

yashpatel007:


include/index_factory.h
```diff
@@ -0,0 +1,154 @@
...
+#include "index.h"
+#include "common_includes.h"
+
+namespace diskann
+{
+template <typename T, typename TagT, typename LabelT> class MemoryIndex;
+template <typename T, typename LabelT> class DiskIndex;
+enum BuildType
+{
+    MEMORY,
```
harsha-simhadri, line 15: I dont think MEMORY/DISK is the appropriate hierarchy here. The index consists of multiple objects such as 1 or 2 data stores, graph stores, filter stores. Each of those objects can be on disk, memory, etc. So the config is the choice of data and graph stores.

include/index_factory.h
```diff
@@ -0,0 +1,154 @@
...
+    DISK
+};
+
+enum LoadStoreStratagy
+{
+    GraphStore,
+    DataStore
```
harsha-simhadri, line 22: what is the enumeration over?

harsha-simhadri (changes requested):

include/index_factory.h
yashpatel007: ok, so the primary choice should be load_store_stratagy. But that part is being provided in the index.cpp via  a factory. In order to achieve that we can just pass an arg into the index constructor, and possibly initiate an index constructor by a config. So what would be the point of having an index factory? That part I am missing a bit.

include/index_factory.h
yashpatel007: This was intended to provide an argument to the index constructor so that we can tie Ravi and Gopl's work to mine. The basic idea was to take this and pass it as an argument so that it can be used to factory instances of Abstract Data Store vs Abstract Graph Store.

PR #316, opened 2023-04-13
Handle Invalid universal label and remove some duplicate code

NeelamMahapatro:
- Handle invalid label value in the filtered search
- Removed repetitive function for filter_num conversion
- Assigned a fixed value for universal label i.e. 0

- [ ] Does this PR have a descriptive title that could go in our release notes?
- [ ] Does this PR add any new dependencies?
- [ ] Does this PR modify any existing APIs?
   - [ ] Is the change to the API backwards compatible?
- [ ] Should this result in any changes to our documentation, either updating existing docs or adding new ones?

PR #326, opened 2023-04-18
Fix Issue - Filter DiskANN

NeelamMahapatro:

### DIFF
--- a/include/index.h
+++ b/include/index.h
@@ -120,7 +120,7 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
                                                 const size_t num_points_to_load, IndexWriteParameters &parameters,
                                                 const std::vector<TagT> &tags = std::vector<TagT>());
 
-    DISKANN_DLLEXPORT void set_universal_label(const LabelT &label);
+    DISKANN_DLLEXPORT void set_universal_label();
 
     // Get converted integer label from string to int map (_label_map)
     DISKANN_DLLEXPORT LabelT get_converted_label(const std::string &raw_label);
@@ -150,6 +150,10 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
                                               float *distances, std::vector<T *> &res_vectors);
 
     // Filter support search
+    template <typename IndexType>
+    DISKANN_DLLEXPORT std::pair<uint32_t, uint32_t> search_with_filters(const T *query, const std::string &filter_label,
+                                                                        const size_t K, const uint32_t L,
+                                                                        IndexType *indices, float *distances);
     template <typename IndexType>
     DISKANN_DLLEXPORT std::pair<uint32_t, uint32_t> search_with_filters(const T *query, const LabelT &filter_label,
                                                                         const size_t K, const uint32_t L,
--- a/src/disk_utils.cpp
+++ b/src/disk_utils.cpp
@@ -642,9 +642,8 @@ int build_merged_vamana_index(std::string base_file, diskann::Metric compareMetr
         else
         {
             if (universal_label != "")
-            { //  indicates no universal label
-                LabelT unv_label_as_num = 0;
-                _index.set_universal_label(unv_label_as_num);
+            { //  indicates universal label
+                _index.set_universal_label();
             }
             _index.build_filtered_index(base_file.c_str(), label_file, base_num, paras);
         }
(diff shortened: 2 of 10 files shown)


src/disk_utils.cpp
```diff
@@ -642,9 +642,8 @@ int build_merged_vamana_index(std::string base_file, diskann::Metric compareMetr
         else
         {
             if (universal_label != "")
-            { //  indicates no universal label
-                LabelT unv_label_as_num = 0;
-                _index.set_universal_label(unv_label_as_num);
+            { //  indicates universal label
```
Sanhaoji2, line 645: universal_label is only use as bool?

include/index.h
```diff
@@ -151,6 +151,10 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
 
     // Filter support search
     template <typename IndexType>
+    DISKANN_DLLEXPORT std::pair<uint32_t, uint32_t> search_with_filters(const T *query, const std::string &filter_label,
```
Sanhaoji2, line 154: what is the usage of the return value? I saw the first value of std::pair is always 0.

src/pq_flash_index.cpp
```diff
@@ -582,13 +567,8 @@ void PQFlashIndex<T, LabelT>::parse_label_file(const std::string &label_file, si
...
             }
-            int32_t filter_num = get_filter_number(token_as_num);
-            if (filter_num == -1)
-            {
-                diskann::cout << "Error!! " << std::endl;
-                exit(-1);
-            }
-            _pts_to_labels[counter++] = filter_num;
+            _pts_to_labels[counter++] = token_as_num;
+            ;
```
Sanhaoji2, line 571: remove

src/index.cpp
```diff
@@ -1877,10 +1877,12 @@ LabelT Index<T, TagT, LabelT>::get_converted_label(const std::string &raw_label)
...
-    std::stringstream stream;
-    stream << "Unable to find label in the Label Map";
-    diskann::cerr << stream.str() << std::endl;
-    throw diskann::ANNException(stream.str(), -1, __FUNCSIG__, __FILE__, __LINE__);
+    if (_use_universal_label)
+    {
+        return _universal_label;
+    }
+    else
+        throw ANNException("Invalid Label provided for filter search.", 4);
```
Sanhaoji2, line 1885: can we return error num instead of throw exception?

harsha-simhadri:
replaces #316 according to @NeelamMahapatro

src/disk_utils.cpp
dkaczynski: If a universal label exists, it will always be represented internally as the value 0. The existence of a universal label in a dataset is a boolean value.

include/index.h
NeelamMahapatro: first is supposed to count number of hop, currently not in use as confirmed from Ravi

Sanhaoji2: can we only return error code? since there may be more metrics in the future, it's better to move it to an object and transfer as input params.

src/disk_utils.cpp
Sanhaoji2: then we don't need to transfer std::string here? we can just use bool enable_universal_label?

include/index.h
```diff
@@ -149,11 +150,21 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
     DISKANN_DLLEXPORT size_t search_with_tags(const T *query, const uint64_t K, const uint32_t L, TagT *tags,
                                               float *distances, std::vector<T *> &res_vectors);
 
+    // Filter support search String interface for converting to integer
+    template <typename IndexType>
+    DISKANN_DLLEXPORT std::unique_ptr<ANNReturnCode> search_with_filters(const T *query,
```
Sanhaoji2, line 155: ANNReturnCode only contain enum member, you can remove std::unique_ptr

include/ann_returncode.h
```diff
@@ -0,0 +1,48 @@
...
+class ANNReturnCode
+{
+  public:
+    enum class Value
+    {
+        SUCCESS,
+        INVALID_LABEL
+    };
+
+    DISKANN_DLLEXPORT const char *getErrorDescription()
```
Sanhaoji2, line 20: all implement is in header file, DISKANN_DLLEXPORT isn't necessary

include/percentile_stats.h
```diff
@@ -35,6 +35,12 @@ struct QueryStats
     unsigned n_hops = 0;       // # search hops
 };
 
+struct QueryStatsMemory
```
Sanhaoji2, line 38: QueryStats already contains m_cmps and n_hops, so we can just reuse it?

include/index.h
NeelamMahapatro: Its for better memory management, If I dont give unique ptr, it ends up with stack smashing

PR #331, opened 2023-04-21
Set compile warning as error for core projects

yiylin:
#### What does this implement/fix? Briefly explain your changes.
1. Set compile warning as error for core projects
2.  Fix warning related errors

PR #342, opened 2023-04-28
Force error on warnings and add casts to test directory

dkaczynski:
Following in the philosophy of #331, the tests directory has been included in the fail-on-warning compilation behavior, and appropriate casts have been added to remove warnings.


tests/range_search_disk_index.cpp
```diff
@@ -226,21 +226,22 @@ int search_disk_index(diskann::Metric &metric, const std::string &index_path_pre
         auto mean_ios = diskann::get_mean_stats<uint32_t>(stats, query_num,
                                                           [](const diskann::QueryStats &stats) { return stats.n_ios; });
 
-        float mean_cpuus = diskann::get_mean_stats<float>(
+        double mean_cpuus = diskann::get_mean_stats<float>(
```
rakri, line 229: why double?

tests/test_streaming_scenario.cpp
```diff
@@ -91,7 +91,7 @@ void insert_next_batch(diskann::Index<T, TagT, LabelT> &index, size_t start, siz
         std::cout << std::endl << "Inserting from " << start << " to " << end << std::endl;
 
         size_t num_failed = 0;
-#pragma omp parallel for num_threads(insert_threads) schedule(dynamic) reduction(+ : num_failed)
+#pragma omp parallel for num_threads((int32_t)insert_threads) schedule(dynamic) reduction(+ : num_failed)
```
rakri, line 94: why (signed) int32_t for num_threads?

tests/utils/compute_groundtruth_for_filters.cpp
```diff
@@ -215,11 +215,12 @@ void exact_knn(const size_t dim, const size_t k,
         {
             maxPQIFCS point_dist;
             for (uint64_t p = 0; p < k; p++)
-                point_dist.emplace(p, dist_matrix[(ptrdiff_t)p + (ptrdiff_t)(q - q_b) * (ptrdiff_t)npoints]);
+                point_dist.emplace((int32_t)p, dist_matrix[(ptrdiff_t)p + (ptrdiff_t)(q - q_b) * (ptrdiff_t)npoints]);
```
rakri, line 218: int32_t or uint32_t for p?

tests/utils/compute_groundtruth_for_filters.cpp
```diff
@@ -215,11 +215,12 @@ void exact_knn(const size_t dim, const size_t k,
...
             for (size_t p = k; p < npoints; p++)
             {
                 if (point_dist.top().second > dist_matrix[(ptrdiff_t)p + (ptrdiff_t)(q - q_b) * (ptrdiff_t)npoints])
-                    point_dist.emplace(p, dist_matrix[(ptrdiff_t)p + (ptrdiff_t)(q - q_b) * (ptrdiff_t)npoints]);
+                    point_dist.emplace((int32_t)p,
```
rakri, line 222: same as above/

tests/utils/generate_synthetic_labels.cpp
```diff
@@ -12,27 +12,27 @@ namespace po = boost::program_options;
...
   public:
-    ZipfDistribution(int num_points, int num_labels)
+    ZipfDistribution(uint64_t num_points, uint32_t num_labels)
         : num_labels(num_labels), num_points(num_points),
           uniform_zero_to_one(std::uniform_real_distribution<>(0.0, 1.0))
     {
     }
 
-    std::unordered_map<int, int> createDistributionMap()
+    std::unordered_map<uint32_t, uint32_t> createDistributionMap()
```
rakri, line 21: int has become uint32_t here. make sure this preserves correctness?

tests/range_search_disk_index.cpp
dkaczynski: This was simply to align with the function signature of `get_mean_stats`, which returns a double. Alternatively, we can change the function signature to return a float. If there is a performance consideration in either case, I'm fine with it. Since this code does not impact the performance of building or searching the index, I'm inclined to leave it as double.

tests/test_streaming_scenario.cpp
dkaczynski: GCC doesn't complain about it, but MSVC gives the following warning (which is treated as an error):

> ...\microsoft\DiskANN\tests\test_streaming_scenario.cpp(94): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data

I can't find the relevant OpenMP documentation that defines the expected integer type for the `num_threads` clause. These two sources seem reliable, but they don't go beyond describing the expected type as "num" or "integer":
* https://learn.microsoft.com/en-us/cpp/parallel/openmp/reference/openmp-clauses?view=msvc-170#num-threads
* https://www.openmp.org/spec-html/5.0/openmpse14.html#x54-800002.6

To appease the MSVC, I'm inclined to cast it to the signed int as warned. We should never be in a situation where we are declaring billions of threads, after all. And we can leave the input data type as size_t since that is more explicit as to the expected behavior.

tests/utils/generate_synthetic_labels.cpp
dkaczynski: The expected behavior is that both the key and value are non-negative integer values: the key is the integer value for the generated label, and the value is the number of occurrences of that label according to the distribution.

However, line 51 (`distribution_map[it->first] -= 1;`) would have some unwanted behavior if the number of expected occurrences were somehow set to 0. Refactoring this.

rakri (approved):
Looks good to me, thanks!

PR #343, opened 2023-05-08
Enable Windows python bindings

nicaurvi:
 
Windows python bindings were being created but were failing at runtime due to missing dependencies. This PR fixes these issues by moving the Windows build output into the python package.

#### Any other comments?
All python tests now pass on both windows and linux.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -371,3 +371,5 @@ wheelhouse/*
 dist/*
 venv*/**
 *.swp
+
+gperftools
--- a/python/src/diskann_bindings.cpp
+++ b/python/src/diskann_bindings.cpp
@@ -204,7 +204,7 @@ template <class T> struct DynamicInMemIndex
         py::array_t<int> insert_retvals(num_inserts);
 
 #pragma omp parallel for schedule(dynamic, 1)
-        for (size_t i = 0; i < num_inserts; i++)
+        for (int64_t i = 0; i < num_inserts; i++)
         {
             insert_retvals.mutable_data()[i] = _index->insert_point(vectors.data(i), *(ids.data(i)));
         }
--- a/setup.py
+++ b/setup.py
@@ -3,12 +3,14 @@
 
 import os
 import re
+import shutil
 import subprocess
 import sys
 from pathlib import Path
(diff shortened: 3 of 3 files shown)


python/src/diskann_bindings.cpp
```diff
@@ -204,7 +204,7 @@ template <class T> struct DynamicInMemIndex
         py::array_t<int> insert_retvals(num_inserts);
 
 #pragma omp parallel for schedule(dynamic, 1)
-        for (size_t i = 0; i < num_inserts; i++)
+        for (int64_t i = 0; i < num_inserts; i++)
```
nicaurvi, line 207: size_t causes a windows build error, using explicit type instead

setup.py
```diff
@@ -44,7 +46,7 @@ def build_extension(self, ext: CMakeExtension) -> None:
             f"-DCMAKE_LIBRARY_OUTPUT_DIRECTORY={extdir}{os.sep}",
             f"-DPYTHON_EXECUTABLE={sys.executable}",
             f"-DCMAKE_BUILD_TYPE={cfg}",  # not used on MSVC, but no harm
-            f"-DVERSION_INFO={self.distribution.get_version()}" # commented out because we want this set in the CMake file
+            f"-DVERSION_INFO={self.distribution.get_version()}"  # commented out, we want this set in the CMake file
```
nicaurvi, line 49: pep8

.gitignore
```diff
@@ -371,3 +371,5 @@ wheelhouse/*
 dist/*
 venv*/**
 *.swp
+
+gperftools
```
nicaurvi, line 375: Ignore the git submodule. In windows, this directory is created when updating the submodule. Adding to .gitignore to... ignore it.

python/src/diskann_bindings.cpp
harsha-simhadri: Then the argument passed to the function should be int64_t too

harsha-simhadri (changes requested):

harsha-simhadri (approved):

PR #344, opened 2023-05-08
Refine custom logger, support early-return and expose more stats(memory usage, scratching wating time .etc)

yiylin:
   - [ ] Add new APIs
 
1. Track major memory usage to load index
2. Use class IndexSearchContext to pass in new parameters or pass out results for searching index.
Using this way to avoid creating new "search" functions to support new features
3. Use new macro(ENABLE_CUSTOM_LOGGER) to turn on Custom logger since we also wants to enable the logger feature
4. Support to set searching time limit to return early when the searching is time out
5. Track the time to wait for the available scratch(pre-allocated objects)

### DIFF
--- a/include/index.h
+++ b/include/index.h
@@ -146,6 +146,11 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
     DISKANN_DLLEXPORT std::pair<uint32_t, uint32_t> search(const T *query, const size_t K, const uint32_t L,
                                                            IDType *indices, float *distances = nullptr);
 
+    template <typename IDType>
+    DISKANN_DLLEXPORT std::pair<uint32_t, uint32_t> search(const T *query, const size_t K, const uint32_t L,
+                                                           IDType *indices, float *distances,
+                                                           IndexSearchContext<LabelT> &context);
+
     // Initialize space for res_vectors before calling.
     DISKANN_DLLEXPORT size_t search_with_tags(const T *query, const uint64_t K, const uint32_t L, TagT *tags,
                                               float *distances, std::vector<T *> &res_vectors);
@@ -235,7 +240,8 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
     std::pair<uint32_t, uint32_t> iterate_to_fixed_point(const T *node_coords, const uint32_t Lindex,
                                                          const std::vector<uint32_t> &init_ids,
                                                          InMemQueryScratch<T> *scratch, bool use_filter,
-                                                         const std::vector<LabelT> &filters, bool search_invocation);
+                                                         const std::vector<LabelT> &filters, bool search_invocation,
+                                                         IndexSearchContext<LabelT> *context = nullptr);
 
     void search_for_point_and_prune(int location, uint32_t Lindex, std::vector<uint32_t> &pruned_list,
                                     InMemQueryScratch<T> *scratch, bool use_filter = false,
@@ -308,6 +314,11 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
     DISKANN_DLLEXPORT size_t load_tags(const std::string tag_file_name);
     DISKANN_DLLEXPORT size_t load_delete_set(const std::string &filename);
 #endif
+
+    template <typename IndexType>
+    std::pair<uint32_t, uint32_t> search_with_filters(const T *query, const size_t K, const uint32_t L,
+                                                      IndexType *indices, float *distances,
+                                                      IndexSearchContext<LabelT> &context);
 
   private:
     // Distance functions
--- a/include/index.h
+++ b/include/index.h
@@ -102,6 +102,7 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
     // get some private variables
     DISKANN_DLLEXPORT size_t get_num_points();
     DISKANN_DLLEXPORT size_t get_max_points();
+    DISKANN_DLLEXPORT uint64_t get_memory_in_bytes();
 
     // Batch build from a file. Optionally pass tags vector.
     DISKANN_DLLEXPORT void build(const char *filename, const size_t num_points_to_load,
@@ -377,6 +378,8 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
     uint8_t *_pq_data = nullptr;
     bool _pq_generated = false;
     FixedChunkPQTable _pq_table;
+
+    uint64_t _memory_in_bytes;
 
     //
     // Data structures, locks and flags for dynamic indexing and tags
--- a/include/index.h
+++ b/include/index.h
@@ -390,8 +390,6 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
     bool _pq_generated = false;
     FixedChunkPQTable _pq_table;
 
-    uint64_t _memory_in_bytes;
-
     //
     // Data structures, locks and flags for dynamic indexing and tags
     //
@@ -424,6 +422,8 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
 
     // Per node lock, cardinality=_max_points
     std::vector<non_recursive_mutex> _locks;
+
+    uint64_t _memory_in_bytes;
 
     static const float INDEX_GROWTH_FACTOR;
 };
--- a/include/parameters.h
+++ b/include/parameters.h
@@ -214,11 +214,11 @@ template <typename LabelT = uint32_t> class IndexSearchContext
 
   private:
     uint32_t _time_limit_in_microseconds;
-    State _result_state;
-    Timer _timer;
     uint32_t _io_limit;
-    LabelT _label;
+    State _result_state;
     bool _use_filter;
+    LabelT _label;
+    Timer _timer;
 };
 
 } // namespace diskann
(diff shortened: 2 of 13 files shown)


include/parameters.h
```diff
@@ -113,4 +117,94 @@ class IndexWriteParametersBuilder
...
+{
+  public:
+    IndexSearchContext(uint32_t time_limit_in_microseconds = 0u, uint32_t io_limit = UINT32_MAX)
+        : _time_limit_in_microseconds(time_limit_in_microseconds), _io_limit(io_limit), _result_state(State::Unknown)
+    {
+        _use_filter = false;
+        _label = (LabelT)0;
+    }
+
+    void SetLabel(LabelT label, bool use_filter)
```
harsha-simhadri, line 148: move implementation to src/*cpp file?

include/boost_dynamic_bitset_fwd.h
```diff
@@ -6,6 +6,7 @@
 namespace boost
 {
 #ifndef BOOST_DYNAMIC_BITSET_FWD_HPP
+#define BOOST_DYNAMIC_BITSET_FWD_HPP
 template <typename Block = unsigned long, typename Allocator = std::allocator<Block>> class dynamic_bitset;
```
harsha-simhadri, line 10: what is the addition for?

include/index.h
```diff
@@ -147,6 +147,11 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
     DISKANN_DLLEXPORT std::pair<uint32_t, uint32_t> search(const T *query, const size_t K, const uint32_t L,
                                                            IDType *indices, float *distances = nullptr);
 
+    template <typename IDType>
+    DISKANN_DLLEXPORT std::pair<uint32_t, uint32_t> search(const T *query, const size_t K, const uint32_t L,
+                                                           IDType *indices, float *distances,
```
harsha-simhadri, line 152: Did you mean for distances to be mandatory in this interface?

include/parameters.h
```diff
@@ -113,4 +117,94 @@ class IndexWriteParametersBuilder
...
+    FailureException = 4,    // Fail because of Exception
+    FailureInvalidLabel = 5, // Fail because of invalid label
+    StateCount = 6           // The number of state
+};
+
+/// <summary>
+/// Use this class to pass in searching parameters and pass out searching result
+/// </summary>
+
+template <typename LabelT = uint32_t> class IndexSearchContext
```
harsha-simhadri, line 138: why is this class in parameters file?

src/index.cpp
```diff
@@ -1080,7 +1094,24 @@ std::pair<uint32_t, uint32_t> Index<T, TagT, LabelT>::iterate_to_fixed_point(
...
+            return std::make_pair(hops, cmps);
+        }
+    }
+
+    if (context != nullptr && context->CheckTimeout())
+    {
+        return std::make_pair(hops, cmps);
+    }
+
+    if (context != nullptr)
```
harsha-simhadri, line 1109: why replicate stats here in this?

harsha-simhadri (commented):
Can you clarify what the new context class is for? Its seems to be a mix of many different purposes (early stop, filter/no-filter, stats). Can we keep each of these separate?
Also, in the context of current re-organization into separate graph and data stores, can you rethink this class?

include/boost_dynamic_bitset_fwd.h
yiylin: this header file is in the boost package. but somehow, someone added it to here. Then it could cause conflict with the header file in the boost package

include/index.h
yiylin: for our usage, yes. I also could make it optional if you want.

include/parameters.h
yiylin: I could move it to a separate file

include/parameters.h
yiylin: ok

yiylin:
> Can you clarify what the new context class is for? Its seems to be a mix of many different purposes (early stop, filter/no-filter, stats). Can we keep each of these separate? Also, in the context of current re-organization into separate graph and data stores, can you rethink this class?

I just want to have a class to allow users to pass in parameters and pass out results/stats so that we do not need to keep adding new interfaces when we want to support new features/scenarios

harsha-simhadri:
> > Can you clarify what the new context class is for? Its seems to be a mix of many different purposes (early stop, filter/no-filter, stats). Can we keep each of these separate? Also, in the context of current re-organization into separate graph and data stores, can you rethink this class?
> 
> I just want to have a class to allow users to pass in parameters and pass out results/stats so that we do not need to keep adding new interfaces when we want to support new features/scenarios. So that we could have a unified API

Got it. For this goal:
1. We need to also remove other "stats" and related classes. 
2. Collect all such functionality across use cases into your proposed class and get more buy in.
3. Think how this would exist in the new re-factored code.

include/parameters.h
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 #pragma once
+#include <omp.h>
```
daxpryce, line 5: What does including this in this file do?

include/percentile_stats.h
```diff
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include <algorithm>
```
daxpryce, line 6: Same question here - what does including this do?

src/pq_flash_index.cpp
```diff
@@ -1059,6 +1059,21 @@ void PQFlashIndex<T, LabelT>::cached_beam_search(const T *query1, const uint64_t
...
+                              stats);
+}
+
+template <typename T, typename LabelT>
+void PQFlashIndex<T, LabelT>::cached_beam_search(const T *query1, const uint64_t k_search, const uint64_t l_search,
+                                                 uint64_t *indices, float *distances, const uint64_t beam_width,
+                                                 IndexSearchContext<LabelT> &context, const bool use_reorder_data,
+                                                 QueryStats *stats)
+{
+    auto use_filter = context.UseFilter();
```
daxpryce, line 1074: As much as I hate to pick nits, we do seem to use lower-snake-case for all methods and functions.  We should probably stick with that convention and/or make everything Pascal

src/index.cpp
```diff
@@ -1971,13 +2002,30 @@ template <typename IdType>
 std::pair<uint32_t, uint32_t> Index<T, TagT, LabelT>::search(const T *query, const size_t K, const uint32_t L,
                                                              IdType *indices, float *distances)
 {
+    IndexSearchContext<LabelT> context;
+    return search(query, K, L, indices, distances, context);
+}
+
+template <typename T, typename TagT, typename LabelT>
+template <typename IdType>
+std::pair<uint32_t, uint32_t> Index<T, TagT, LabelT>::search(const T *query, const size_t K, const uint32_t L,
```
gopalrs, line 2011: An after thought. Can we change the return type of search to return an IndexSearchContext object and return the stats, the number of results returned, timing information and what ever else we need via the return object?

src/index.cpp
```diff
@@ -1971,13 +2002,30 @@ template <typename IdType>
...
+                                                             IdType *indices, float *distances,
+                                                             IndexSearchContext<LabelT> &context)
+{
+    if (context.UseFilter())
+    {
+        return search_with_filters(query, K, L, indices, distances, context);
+    }
+
     if (K > (uint64_t)L)
     {
```
gopalrs, line 2021: Why not simply increase L to K? It seems extreme to throw an exception for this.

include/index_search_context.h
```diff
@@ -0,0 +1,69 @@
...
+/// Search state
+/// </summary>
+enum State : uint8_t
+{
+    Unknown = 0,
+    Success = 1,
+    Failure = 2,
+    FailureTimeout = 3,      // Fail because of timeout
+    FailureException = 4,    // Fail because of Exception
+    FailureInvalidLabel = 5, // Fail because of invalid label
```
gopalrs, line 23: IT seems FailureInvalidLabel is not used anywhere in the code at present. Is it necessary?

src/index_search_context.cpp
```diff
@@ -0,0 +1,63 @@
...
+    return false;
+}
+
+template <typename LabelT> QueryStats &IndexSearchContext<LabelT>::GetStats()
+{
+    return _stats;
+}
+
+template DISKANN_DLLEXPORT class IndexSearchContext<uint16_t>;
+template DISKANN_DLLEXPORT class IndexSearchContext<uint32_t>;
```
gopalrs, line 61: Not defining this class for uint64_t?

src/pq_flash_index.cpp
```diff
@@ -1077,13 +1092,30 @@ void PQFlashIndex<T, LabelT>::cached_beam_search(const T *query1, const uint64_t
...
     ScratchStoreManager<SSDThreadData<T>> manager(this->thread_data);
     auto data = manager.scratch_space();
     IOContext &ctx = data->ctx;
     auto query_scratch = &(data->scratch);
     auto pq_query_scratch = query_scratch->_pq_scratch;
+    if (stats == nullptr)
+    {
+        stats = &context.GetStats();
+    }
+    if (stats != nullptr)
```
gopalrs, line 1110: I think we can get rid of the QueryStats and stick to the one inside IndexSearchContext .

gopalrs (changes requested):

gopalrs (changes requested):
Requesting some changes

PR #345, opened 2023-05-09
Use new macro(ENABLE_CUSTOM_LOGGER) to turn on Custom logger

yiylin:
 
Use new macro(ENABLE_CUSTOM_LOGGER) to turn on Custom logger since we also wants to enable the logger feature


src/logger.cpp
```diff
@@ -16,12 +16,13 @@ DISKANN_DLLEXPORT ANNStreamBuf cerrBuff(stderr);
...
 DISKANN_DLLEXPORT std::basic_ostream<char> cerr(&cerrBuff);
 
-#ifdef EXEC_ENV_OLS
+#ifdef ENABLE_CUSTOM_LOGGER
 std::function<void(LogLevel, const char *)> g_logger;
 
 void SetCustomLogger(std::function<void(LogLevel, const char *)> logger)
 {
     g_logger = logger;
+    cout << "Setted Custom Logger" << std::endl;
```
PhilipBAdams, line 25: If this is needed, please make it as debug log rather than to cout.

PR #346, opened 2023-05-09
Track major memory usage to load index

yiylin:
 
Track major memory usage to load index and add new API to get this value


include/pq.h
```diff
@@ -31,9 +31,9 @@ class FixedChunkPQTable
     virtual ~FixedChunkPQTable();
 
 #ifdef EXEC_ENV_OLS
-    void load_pq_centroid_bin(MemoryMappedFiles &files, const char *pq_table_file, size_t num_chunks);
+    uint64_t load_pq_centroid_bin(MemoryMappedFiles &files, const char *pq_table_file, size_t num_chunks);
```
PhilipBAdams, line 34: nit: Consistent interface might be preferred here e.g. pq_table.get_memory_in_bytes()

src/index.cpp
```diff
@@ -785,6 +788,7 @@ size_t Index<T, TagT, LabelT>::load_graph(std::string filename, size_t expected_
     }
 #endif
 
+    _memory_in_bytes += cc * sizeof(uint32_t);
```
harsha-simhadri, line 791: Are we sure that the allocation is exactly this at all times. Perhaps it starts with this, but it also changes as the graph is edited, right?

harsha-simhadri (changes requested):
If we want get_memory_in_bytes() to be accurate, we need a mechanism to force all malloc/freeto go through this bookkeeping. As it stands, people can add future code without adding the bookkeeping. Also, in your PR, you are not counting the scratch space allocation -- so your estimate is already off.

src/index.cpp
yiylin: for our case, our graph is fixed

yiylin:
> If we want get_memory_in_bytes() to be accurate, we need a mechanism to force all malloc/freeto go through this bookkeeping. As it stands, people can add future code without adding the bookkeeping. Also, in your PR, you are not counting the scratch space allocation -- so your estimate is already off.

My PR title says to track the "Major" memory usage, not all the memory usage. The size of scratch space is small compared with the index. We want to know the memory usage for each index. If we want to get the accurate value, then it will introduce much more changes.

PhilipBAdams:
> > If we want get_memory_in_bytes() to be accurate, we need a mechanism to force all malloc/freeto go through this bookkeeping. As it stands, people can add future code without adding the bookkeeping. Also, in your PR, you are not counting the scratch space allocation -- so your estimate is already off.
> 
> My PR title says to track the "Major" memory usage, not all the memory usage. The size of scratch space is small compared with the index. We want to know the memory usage for each index and not all the memory uses malloc/free to allocate/free directly. If we want to get the accurate value, then it will introduce much more changes.

Do you think it is feasible to add some tests with `::new`/`::new[]`/`malloc`/`alloc_aligned` overwritten to track memory usage? Then CI can enforce that your bookkeeping is at least within some margin of accuracy when people change the code.

Alternate approach is to wrap those functions to take a reference to your per-index counter when called, though it would require more code change as you said.

yiylin:
> bookkeeping

yiylin:
> > > If we want get_memory_in_bytes() to be accurate, we need a mechanism to force all malloc/freeto go through this bookkeeping. As it stands, people can add future code without adding the bookkeeping. Also, in your PR, you are not counting the scratch space allocation -- so your estimate is already off.
> > 
> > 
> > My PR title says to track the "Major" memory usage, not all the memory usage. The size of scratch space is small compared with the index. We want to know the memory usage for each index and not all the memory uses malloc/free to allocate/free directly. If we want to get the accurate value, then it will introduce much more changes.
> 
> Do you think it is feasible to add some tests with `::new`/`::new[]`/`malloc`/`alloc_aligned` overwritten to track memory usage? Then CI can enforce that your bookkeeping is at least within some margin of accuracy when people change the code.
> 
> Alternate approach is to wrap those functions to take a reference to your per-index counter when called, though it would require more code change as you said.

@harsha-simhadri  what do you think to add the tests to let CI to catch the changing of memory usage?

src/pq_flash_index.cpp
```diff
@@ -41,7 +41,7 @@ namespace diskann
 
 template <typename T, typename LabelT>
 PQFlashIndex<T, LabelT>::PQFlashIndex(std::shared_ptr<AlignedFileReader> &fileReader, diskann::Metric m)
-    : reader(fileReader), metric(m)
+    : reader(fileReader), metric(m), memory_in_bytes(0)
```
gopalrs, line 44: Are you considering scratch space allocated inside setup_thread_data()?

gopalrs (changes requested):

src/index.cpp
yiylin: we want to know the memory usage for each index

src/pq_flash_index.cpp
yiylin: yes. I also could add that. but even though I add that, @harsha-simhadri  may still have concern(get accurate value and count the new memory for the new code). So I may wait for the final conclusion to continue the change.

PR #347, opened 2023-05-11
Fix the disparity between disk and memory search for Universal label

NeelamMahapatro:
 
**Fix the disparity between disk and memory search for Universal label**
Issue - If query label is UNV, memory search currently do unfiltered search over the graph without checking what is the label of current node. However, disk search returns the points corresponding to UNV label only

- Used the search_invocation flag inside iterate_to_fixed_point method and changed the logic for filtered memory search to check the current point's label and push to the output result accordingly.

### DIFF
--- a/include/index.h
+++ b/include/index.h
@@ -104,8 +104,8 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
     DISKANN_DLLEXPORT size_t get_num_points();
     DISKANN_DLLEXPORT size_t get_max_points();
 
-    DISKANN_DLLEXPORT const std::vector<LabelT> find_common_filters(uint32_t point_id, bool search_invocation,
-                                                                    const std::vector<LabelT> &incoming_labels);
+    DISKANN_DLLEXPORT size_t find_common_filters(uint32_t point_id, bool search_invocation,
+                                                 const std::vector<LabelT> &incoming_labels);
 
     // Batch build from a file. Optionally pass tags vector.
     DISKANN_DLLEXPORT void build(const char *filename, const size_t num_points_to_load,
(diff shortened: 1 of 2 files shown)


nikitbegwani (commented):
- Check Line 1027-1033
- If same utility, build a wrapper

include/index.h
```diff
@@ -104,6 +104,9 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
     DISKANN_DLLEXPORT size_t get_num_points();
     DISKANN_DLLEXPORT size_t get_max_points();
 
+    DISKANN_DLLEXPORT const std::vector<LabelT> find_common_filters(uint32_t point_id, bool search_invocation,
```
gopalrs, line 107: Might be better to return the common filters by reference and return the count as return type.

src/index.cpp
```diff
@@ -837,6 +837,31 @@ template <typename T, typename TagT, typename LabelT> std::vector<uint32_t> Inde
     return init_ids;
 }
 
+template <typename T, typename TagT, typename LabelT>
+const std::vector<LabelT> Index<T, TagT, LabelT>::find_common_filters(uint32_t point_id, bool search_invocation,
+                                                                      const std::vector<LabelT> &incoming_labels)
+{
+    std::vector<LabelT> common_filters;
+    auto &curr_node_labels = _pts_to_labels[point_id];
+    std::set_intersection(incoming_labels.begin(), incoming_labels.end(), curr_node_labels.begin(),
```
gopalrs, line 846: Depending on the number of labels we expect to see, this might become very slow. The question is whether to change it upfront assuming that we might deal with large # of labels at some point in time, or deal with it at that time. Just a comment.

gopalrs (changes requested):

src/index.cpp
NeelamMahapatro: Siddharth and Ravi are already working on this to replace with bit wise and operation.
Closing this comment in the context of this PR.

include/index.h
NeelamMahapatro: Common filters vector is also not getting used further. so I have just initialized it inside the find_common_filters methos. Just returned the size of this vector

gopalrs (approved):

rakri (approved):

PR #349, opened 2023-05-16
Add convenience functions for parsing the PQ index

PhilipBAdams:

User may need to parse index graph information into different format for different applications. This PR gives an API for getting the neighborhood of a given vector, and that vector's PQ and full-dimensional data.

### DIFF
--- a/include/pq_flash_index.h
+++ b/include/pq_flash_index.h
@@ -105,7 +105,7 @@ template <typename T, typename LabelT = uint32_t> class PQFlashIndex
                                                    std::vector<T *> &coord_buffers,
                                                    std::vector<std::pair<uint32_t, uint32_t *>> &nbr_buffers);
 
-    DISKANN_DLLEXPORT std::vector<std::uint8_t> get_pq_vector(std::uint64_t VID);
+    DISKANN_DLLEXPORT std::vector<std::uint8_t> get_pq_vector(std::uint64_t vid);
     DISKANN_DLLEXPORT uint64_t get_num_points();
 
   protected:
--- a/include/pq_flash_index.h
+++ b/include/pq_flash_index.h
@@ -135,8 +135,6 @@ template <typename T, typename LabelT = uint32_t> class PQFlashIndex
     // returns region of `node_buf` containing [COORD(T)]
     DISKANN_DLLEXPORT T *offset_to_node_coords(char *node_buf);
 
-
-
     // index info for multi-node sectors
     // nhood of node `i` is in sector: [i / nnodes_per_sector]
     // offset in sector: [(i % nnodes_per_sector) * max_node_len]
--- a/include/pq_flash_index.h
+++ b/include/pq_flash_index.h
@@ -94,6 +94,20 @@ template <typename T, typename LabelT = uint32_t> class PQFlashIndex
 
     DISKANN_DLLEXPORT diskann::Metric get_metric();
 
+    //
+    // node_ids: input list of node_ids to be read
+    // coord_buffers: pointers to pre-allocated buffers that coords need to copied to. If null, dont copy.
+    // nbr_buffers: pre-allocated buffers to copy neighbors into
+    //
+    // returns a vector of bool one for each node_id: true if read is success, else false
+    //
+    DISKANN_DLLEXPORT std::vector<bool> read_nodes(const std::vector<uint32_t> &node_ids,
+                                                   std::vector<T *> &coord_buffers,
+                                                   std::vector<std::pair<uint32_t, uint32_t *>> &nbr_buffers);
+
+    DISKANN_DLLEXPORT std::vector<std::uint8_t> get_pq_vector(std::uint64_t VID);
+    DISKANN_DLLEXPORT uint64_t get_num_points();
+
   protected:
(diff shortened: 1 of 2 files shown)


harsha-simhadri:
@gopalrs is working on a PR to make distance class (including PQ) a member of data store. Can this wait until that PR? If we merge it now, we would have to redo a bit. I am open to both. @gopalrs?

PhilipBAdams:
If the interface is ok with you I would prefer to merge it now and update the implementation after Gopal's PR so it is simpler to add it to my tool. If you want to have interface change, we can wait for Gopal's PR

src/pq_flash_index.cpp
```diff
@@ -1536,6 +1536,85 @@ template <typename T, typename LabelT> diskann::Metric PQFlashIndex<T, LabelT>::
...
+    }
+#endif
+
+    char *node_buf = OFFSET_TO_NODE(buf, VID);
+    T *node_coords = OFFSET_TO_NODE_COORDS(node_buf);
+
+    std::vector<T> out(get_data_dim());
+    memcpy(out.data(), node_coords, sizeof(T) * get_data_dim());
+
+    return out;
```
gopalrs, line 1568: Should we need to specify std::move() here or will return value optimization take care of move semantics?

include/pq_flash_index.h
```diff
@@ -94,6 +94,10 @@ template <typename T, typename LabelT = uint32_t> class PQFlashIndex
 
     DISKANN_DLLEXPORT diskann::Metric get_metric();
 
+    DISKANN_DLLEXPORT std::vector<T> get_vector(std::uint64_t VID);
+    DISKANN_DLLEXPORT std::vector<std::uint8_t> get_pq_vector(std::uint64_t VID);
+    DISKANN_DLLEXPORT std::vector<std::uint64_t> get_neighbors(std::uint64_t VID);
```
gopalrs, line 99: Should get_neighbors return vector of uint32_t? Or are you trying to make this future-proof?

src/pq_flash_index.cpp
```diff
@@ -1536,6 +1536,85 @@ template <typename T, typename LabelT> diskann::Metric PQFlashIndex<T, LabelT>::
     return this->metric;
 }
 
+template <typename T, typename LabelT> std::vector<T> PQFlashIndex<T, LabelT>::get_vector(std::uint64_t VID)
```
gopalrs, line 1539: VID should be in small letters?

gopalrs (approved):
Agree with @harsha-simhadri that we could hold this off until the datastore/graphstore PR is done, if it is not urgent.

src/pq_flash_index.cpp
```diff
@@ -1536,6 +1536,85 @@ template <typename T, typename LabelT> diskann::Metric PQFlashIndex<T, LabelT>::
...
+        aligned_free(buf);
+        return std::vector<std::uint64_t>();
+    }
+#endif
+
+    char *node_buf = OFFSET_TO_NODE(buf, VID);
+    std::uint32_t *nbr_buf = OFFSET_TO_NODE_NHOOD(node_buf);
+    std::uint64_t nnbrs = (std::uint64_t)(*nbr_buf);
+
+    std::vector<std::uint64_t> out(nnbrs);
```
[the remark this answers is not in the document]
ltan1ms: yes, i think it might be more efficient than relying on return value optimization (as it depends on caller pattern and may not always be guaranteed to avoid copying)?

src/pq_flash_index.cpp
```diff
@@ -1536,6 +1536,85 @@ template <typename T, typename LabelT> diskann::Metric PQFlashIndex<T, LabelT>::
...
+    T *node_coords = OFFSET_TO_NODE_COORDS(node_buf);
+
+    std::vector<T> out(get_data_dim());
+    memcpy(out.data(), node_coords, sizeof(T) * get_data_dim());
+
+    return out;
+}
+
+template <typename T, typename LabelT>
+std::vector<std::uint8_t> PQFlashIndex<T, LabelT>::get_pq_vector(std::uint64_t VID)
```
ltan1ms, line 1572: Since the pq'ed vector is always in memory (without changing location), do you think we can just return a pointer to the underlying uint8_t array to save some efficiency?

ltan1ms (approved):

src/pq_flash_index.cpp
PhilipBAdams: I don't think it's possible to make a vector in-place on an existing array without having very tricky rules about what happens if the vector is resized or the edited.

I replace it with call to the copy constructor in rvalue position, so it is guaranteed to be return-optimized

PhilipBAdams:
Hi @harsha-simhadri, do you mind to re-review/merge? If I move your new `read_nodes` to be public, the change is much simpler than it was before

src/pq_flash_index.cpp
harsha-simhadri: @philipBAdams could you please make the change Gopal requests

PhilipBAdams: This method no longer exists since I use your `read_nodes` method to get the vector coordinates instead

PR #350, opened 2023-05-17
Nmahapatro/fix unv search

NeelamMahapatro:

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -846,6 +846,12 @@ size_t Index<T, TagT, LabelT>::find_common_filters(uint32_t point_id, bool searc
     std::vector<LabelT> common_filters;
     std::set_intersection(incoming_labels.begin(), incoming_labels.end(), curr_node_labels.begin(),
                           curr_node_labels.end(), std::back_inserter(common_filters));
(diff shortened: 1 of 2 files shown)


src/index.cpp
```diff
@@ -837,6 +837,32 @@ template <typename T, typename TagT, typename LabelT> std::vector<uint32_t> Inde
...
+    {
+        if (!search_invocation)
+        {
+            if (std::find(incoming_labels.begin(), incoming_labels.end(), _universal_label) != incoming_labels.end() ||
+                std::find(curr_node_labels.begin(), curr_node_labels.end(), _universal_label) != curr_node_labels.end())
+                common_filters.push_back(_universal_label);
+        }
+        else
+        {
+            if (std::find(curr_node_labels.begin(), curr_node_labels.end(), _universal_label) != curr_node_labels.end())
```
rakri, line 859: if curr_node_labels has UNV, and incoming_labels is also UNV, then set intersection alone would be enough right? This may be redundant?

On the other hand, how do we handle the case when say a base point has explicitly listed all labels as opposed to UNV? Are we assuming this will not happen?

PR #352, opened 2023-05-19
updting from std cpp 14 to cpp 17

yashpatel007:

Updating project to use C++ standard 17

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ project(diskann)
 
 cmake_minimum_required(VERSION 3.15)
 
-set(CMAKE_STANDARD 14)
+set(CMAKE_STANDARD 17)
 
 if(NOT MSVC)
 	set(CMAKE_CXX_COMPILER g++)
@@ -212,7 +212,7 @@ endif()
 #Main compiler/linker settings 
 if(MSVC)
 	#language options
-	add_compile_options(/permissive- /openmp:experimental /Zc:twoPhase- /Zc:inline /WX- /std:c++14 /Gd /W3 /MP /Zi /FC /nologo)
+	add_compile_options(/permissive- /openmp:experimental /Zc:twoPhase- /Zc:inline /WX- /std:c++17 /Gd /W3 /MP /Zi /FC /nologo)
 	#code generation options
 	add_compile_options(/arch:AVX2 /fp:fast /fp:except- /EHsc /GS- /Gy)
 	#optimization options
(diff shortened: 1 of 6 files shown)


src/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 #Copyright(c) Microsoft Corporation.All rights reserved.
 #Licensed under the MIT                        license.
 
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
```
yiylin, line 4: any comments to move to 20? we are using 20 internally

yashpatel007: I am open to do the same. @harsha-simhadri what do you think ? std:14 was lacking a lot of things, specifically **std::any** but if the other stakeholders are fine with 20. It should be fine

harsha-simhadri: I am OK. @theantony @PhilipBAdams @yiylin  @Sanhaoji2 @gopalrs @daxpryce  opinions?

yiylin: I prefer to use 20 if no one has concerns

theantony: It's fine for things within DiskANN, but it requires all consumers to enable C++20. AFAIK that's not the case everywhere (e.g. recently we wanted to use `std::span` in a codebase that uses DiskANN, but weren't able to since the codebase didn't enable the `/std:c++20` flag). I think C++17 should be OK.

CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ project(diskann)
 
 cmake_minimum_required(VERSION 3.15)
 
-set(CMAKE_STANDARD 14)
+set(CMAKE_STANDARD 17)
```
theantony, line 22: Does `CMAKE_STANDARD` even do anything? I couldn't find any documentation on it and I don't see any references in DiskANN itself.

From a search, it appears that the way to deal with C++ standard is to set `CMAKE_CXX_STANDARD` instead. If I'm reading it correctly, it should then automatically add the right compiler settings for the used compiler (`e.g. /std:c++17` for MSVC) rather than adding it explicitly like in the existing `CMakeLists.txt` files.

yashpatel007: Yes, I had that on my local, but because it compiled fine, i thought may be **CMAKE_STANDARD** does the trick.

PhilipBAdams:
https://learn.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-170

I think OS currently distributes vc140 or vc141, and it is hard for us to upgrade w/o impacting other teams. Not all c++20 features are available in vc141, it will cause us linking issues if those features are used. So would prefer to restrict to c++17 for now.

src/CMakeLists.txt
yiylin: some consumers are not ready to use C++20. I am fine to use 17 now.

PR #354, opened 2023-05-20
CICD Refactor

daxpryce:
This is a completely worked CICD flow for our project.

It also renames a lot of stuff, namely:
1. `tests/` is now `apps/`.  this is looking toward actually having unit tests in our tests directory.  documentation SHOULD be fully updated to reference `apps`, so if you notice any issues please raise them.
2. helper scripts (like installing ubuntu system dependencies) now have a home in `scripts/`

I tried to ensure everything common was done in precisely one place - with the exception of all the checkout ceremony.

I tried to optimize for speeding things up, as well as making the acceptance tests we run run in parallel so that we can re-trigger smaller sections.

One thing I was not able to accomplish: a relocatable build that actually works after being packaged up on one github action runner and then downloaded and used on another of the same platform.  I ended up with something like 200 workflow runs just for this single effort and I finally lost my temper trying to get a working, relocated distribution build.  Let's just say it's an opportunity for future improvement :)


daxpryce:
I have no idea what is causing the dynamic tests to fail like they are. If I'm missing some necessary files then we can fix that, but at the least it shouldn't error like this.  I'm a bit surprised to run into these issues, and I'm strapped for time to fix them before my time off next week, so if anyone can take a look into why the dynamic tests are failing I'd be the happiest little panda

daxpryce:
> so if anyone can take a look into why the dynamic tests are failing I'd be the happiest little panda

I may have fixed it - I hope so!

harsha-simhadri (approved):

PR #358, opened 2023-05-22
Add centralized class to allocate\deallocate memory

yiylin:
 
Add centralized class to allocate\deallocate memory, so that we could track memory usage more accurately.

### DIFF
--- a/.github/workflows/push-test.yml
+++ b/.github/workflows/push-test.yml
@@ -8,7 +8,7 @@ jobs:
       uses: actions/checkout@v2
     - name: Install deps
       run: |
-        sudo apt install cmake g++ libaio-dev libgoogle-perftools-dev libunwind-dev clang-format libboost-dev libboost-program-options-dev libmkl-full-dev libcpprest-dev
+        sudo apt install cmake g++ libaio-dev libgoogle-perftools-dev libunwind-dev clang-format libboost-dev libboost-program-options-dev libboost-unit-test-framework-dev libmkl-full-dev libcpprest-dev
     - name: Clang Format Check
       run: |
         mkdir build && cd build && cmake -DRESTAPI=True -DCMAKE_BUILD_TYPE=Release -DFORCE_TO_TRACK_MEMORY_IN_ALLOCATOR=True ..
--- a/.github/workflows/push-test.yml
+++ b/.github/workflows/push-test.yml
@@ -15,7 +15,7 @@ jobs:
         make checkformat
     - name: build
       run: |
-        cd build && make -j && ctest
+        cd build && make -j && ctest -C Release
 #     - uses: actions/setup-python@v3
 #       with:
 #         python-version: "3.10"
@@ -61,5 +61,5 @@ jobs:
 
     - name: Run configure and build
       run: |
-        mkdir build && cd build && cmake .. -DFORCE_TO_TRACK_MEMORY_IN_ALLOCATOR=True && msbuild diskann.sln /m /nologo /t:Build /p:Configuration="Release" /property:Platform="x64" -consoleloggerparameters:"ErrorsOnly;Summary" && ctest
+        mkdir build && cd build && cmake .. -DFORCE_TO_TRACK_MEMORY_IN_ALLOCATOR=True && msbuild diskann.sln /m /nologo /t:Build /p:Configuration="Release" /property:Platform="x64" -consoleloggerparameters:"ErrorsOnly;Summary" && ctest -C Release
       shell: cmd
--- a/.github/workflows/push-test.yml
+++ b/.github/workflows/push-test.yml
@@ -11,11 +11,11 @@ jobs:
         sudo apt install cmake g++ libaio-dev libgoogle-perftools-dev libunwind-dev clang-format libboost-dev libboost-program-options-dev libboost-test-dev libmkl-full-dev libcpprest-dev
     - name: Clang Format Check
       run: |
-        mkdir build && cd build && cmake -DRESTAPI=True -DCMAKE_BUILD_TYPE=Release -DFORCE_TO_TRACK_MEMORY_IN_ALLOCATOR=True ..
+        mkdir build && cd build && cmake -DRESTAPI=True -DCMAKE_BUILD_TYPE=Release ..
         make checkformat
     - name: build
       run: |
-        cd build && make -j && ctest -C Release
+        cd build && make -j
 #     - uses: actions/setup-python@v3
 #       with:
 #         python-version: "3.10"
@@ -61,5 +61,5 @@ jobs:
 
     - name: Run configure and build
       run: |
-        mkdir build && cd build && cmake .. -DFORCE_TO_TRACK_MEMORY_IN_ALLOCATOR=True && msbuild diskann.sln /m /nologo /t:Build /p:Configuration="Release" /property:Platform="x64" -consoleloggerparameters:"ErrorsOnly;Summary" && ctest -C Release
+        mkdir build && cd build && cmake .. && msbuild diskann.sln /m /nologo /t:Build /p:Configuration="Release" /property:Platform="x64" -consoleloggerparameters:"ErrorsOnly;Summary"
       shell: cmd
--- a/.github/workflows/push-test.yml
+++ b/.github/workflows/push-test.yml
@@ -13,9 +13,9 @@ jobs:
       run: |
         mkdir build && cd build && cmake -DRESTAPI=True -DCMAKE_BUILD_TYPE=Release ..
         make checkformat
-#     - name: build
-#       run: |
-#         cd build && make -j
+    - name: build
+      run: |
+        cd build && make -j && ctest
 #     - uses: actions/setup-python@v3
 #       with:
 #         python-version: "3.10"
@@ -61,5 +61,5 @@ jobs:
 
     - name: Run configure and build
       run: |
-        mkdir build && cd build && cmake .. && msbuild diskann.sln /m /nologo /t:Build /p:Configuration="Release" /property:Platform="x64" -consoleloggerparameters:"ErrorsOnly;Summary"
+        mkdir build && cd build && cmake .. && msbuild diskann.sln /m /nologo /t:Build /p:Configuration="Release" /p:FORCE_TO_TRACK_MEMORY_IN_ALLOCATOR="1" /property:Platform="x64" -consoleloggerparameters:"ErrorsOnly;Summary" && ctest
       shell: cmd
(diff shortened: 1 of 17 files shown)


tests/unittests/memory_manager_tests.cpp
```diff
@@ -0,0 +1,145 @@
+#include <boost/test/unit_test.hpp>
+
+#include "memory_manager.h"
+
+BOOST_AUTO_TEST_SUITE(memory_manager_tests)
```
PhilipBAdams, line 5: If we are going to introduce unit testing framework can we split it into separate PR for review

src/memory_manager.cpp
```diff
@@ -0,0 +1,204 @@
...
+#ifndef _WINDOWS
+    free(ptr);
+#else
+    ::_aligned_free(ptr);
+#endif
+
+    erase_block(ptr);
+}
+
+void MemoryManager::insert_block(void *ptr, size_t size)
```
PhilipBAdams, line 69: this separate map for tracking allocation sizes seems very heavyweight, especially since it requires a hash-calculation for lookup in the map and locking to update the map.

What do you think of an approach of storing this info in the header of the memory block, like
buf = malloc(size + 4)
*buf = size
return buf + 4

I guess it requires some extra logic around alignment, but if we are okay with the possibility of a bit more padding waste I think it will be more performant.

include/allocator.h
```diff
@@ -0,0 +1,67 @@
...
+
+    std::shared_ptr<std::atomic<size_t>> _memory_used_in_bytes_ptr;
+
+  private:
+    void update_memory_usage(std::size_t count, bool is_allocation);
+
+    std::allocator<T> _allocator;
+};
+
+template <class T> inline bool operator==(const Allocator<T> &a, const Allocator<T> &b)
```
PhilipBAdams, line 47: These operations should probably also call the underlying operator==/operator!= on _allocator

tests/unittests/memory_manager_tests.cpp
yiylin: actually, this will not break any existing things. and the changes under unittests folder is to add the unit testing framework. It is not a very complicated PR. I do not think we need to split it into a separate PR.

.github/workflows/push-test.yml
```diff
@@ -8,14 +8,14 @@ jobs:
       uses: actions/checkout@v2
     - name: Install deps
       run: |
-        sudo apt install cmake g++ libaio-dev libgoogle-perftools-dev libunwind-dev clang-format libboost-dev libboost-program-options-dev libmkl-full-dev libcpprest-dev
+        sudo apt install cmake g++ libaio-dev libgoogle-perftools-dev libunwind-dev clang-format libboost-dev libboost-program-options-dev libboost-test-dev libmkl-full-dev libcpprest-dev
```
harsha-simhadri, line 11: Is adding another dependence essential to this?

src/memory_manager.cpp
yiylin: this is only used by alloc_aligned and new_array. Ideally, they are only used during loading index, they should not be used during searching index. so I expect this should not cause too much issue. But I am also ok to change to use the padding memory to store the size

.github/workflows/push-test.yml
yiylin: yes. to use the boost unit test framework to write the unit test

CMakeLists.txt
```diff
@@ -246,10 +249,13 @@ else()
     endif()
 endif()
 
+enable_testing()
+
 add_subdirectory(src)
 if (NOT PYBIND)
     add_subdirectory(apps)
     add_subdirectory(apps/utils)
+    add_subdirectory(tests)
```
yiylin, line 258: @daxpryce hope this is the folder you plan to put the unit tests

Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN apt update
 RUN apt install -y software-properties-common
 RUN add-apt-repository -y ppa:git-core/ppa
 RUN apt update
-RUN DEBIAN_FRONTEND=noninteractive apt install -y git make cmake g++ libaio-dev libgoogle-perftools-dev libunwind-dev clang-format libboost-dev libboost-program-options-dev libmkl-full-dev libcpprest-dev python3.10
+RUN DEBIAN_FRONTEND=noninteractive apt install -y git make cmake g++ libaio-dev libgoogle-perftools-dev libunwind-dev clang-format libboost-dev libboost-program-options-dev libboost-test-dev libmkl-full-dev libcpprest-dev python3.10
```
harsha-simhadri, line 10: Do we need unit tests in the docker?

include/allocator.h
```diff
@@ -0,0 +1,73 @@
...
+template <class T> class Allocator
+{
+  public:
+    using value_type = T;
+    using propagate_on_container_move_assignment = std::true_type;
+    using is_always_equal = std::true_type;
+
+    Allocator(std::shared_ptr<std::atomic<size_t>> &memory_used_in_bytes_ptr = nullptr)
+    {
+        _memory_used_in_bytes_ptr = memory_used_in_bytes_ptr;
```
harsha-simhadri, line 27: move implementation to cpp file?

src/CMakeLists.txt
```diff
@@ -7,15 +7,10 @@ set(CMAKE_COMPILE_WARNING_AS_ERROR ON)
...
     add_subdirectory(dll)
 else()
-    #file(GLOB CPP_SOURCES *.cpp)
-    set(CPP_SOURCES abstract_data_store.cpp ann_exception.cpp disk_utils.cpp 
-        distance.cpp index.cpp in_mem_graph_store.cpp in_mem_data_store.cpp
-        linux_aligned_file_reader.cpp math_utils.cpp natural_number_map.cpp
-        in_mem_data_store.cpp in_mem_graph_store.cpp
-        natural_number_set.cpp memory_mapper.cpp partition.cpp pq.cpp
-        pq_flash_index.cpp scratch.cpp logger.cpp utils.cpp filter_utils.cpp)
+    file(GLOB CPP_SOURCES *.cpp)
```
harsha-simhadri, line 10: please revert to listing the files. This is not recommended: https://cmake.org/cmake/help/latest/command/file.html?highlight=glob#glob

src/allocator.cpp
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+#include "allocator.h"
+#include "utils.h"
+
+namespace diskann
+{
+
+template <class T> T *Allocator<T>::allocate(std::size_t count)
```
harsha-simhadri, line 10: why do need an allocator class? Can this be an internal detail of memory_manager class?

src/dll/CMakeLists.txt
```diff
@@ -1,10 +1,8 @@
 #Copyright(c) Microsoft Corporation.All rights reserved.
 #Licensed under the MIT                        license.
 
-add_library(${PROJECT_NAME} SHARED dllmain.cpp ../abstract_data_store.cpp ../partition.cpp ../pq.cpp ../pq_flash_index.cpp ../logger.cpp ../utils.cpp 
-    ../windows_aligned_file_reader.cpp ../distance.cpp ../memory_mapper.cpp ../index.cpp 
-    ../in_mem_data_store.cpp ../in_mem_graph_store.cpp ../math_utils.cpp ../disk_utils.cpp ../filter_utils.cpp 
-    ../ann_exception.cpp ../natural_number_set.cpp ../natural_number_map.cpp ../scratch.cpp)
+file(GLOB CPP_SOURCES ../*.cpp)
```
harsha-simhadri, line 4: please list files instead of GLOB

src/memory_manager.cpp
```diff
@@ -0,0 +1,204 @@
...
+#else
+    *ptr = ::_aligned_malloc(size, align); // note the swapped arguments!
+#endif
+    if (*ptr == nullptr)
+    {
+        report_memory_allocation_failure(size, align);
+        return;
+    }
+
+    insert_block(*ptr, size);
```
harsha-simhadri, line 31: can we make this optional? For example, by using it under "#ifdef TRACK_MEM_USAGE"

src/memory_manager.cpp
harsha-simhadri: We need to make sure we are not allocating anything on the heap in insert or search functions. Otherwise, perf problems might arise.

include/memory_manager.h
```diff
@@ -0,0 +1,47 @@
...
+
+  private:
+    // No copy/assign.
+    MemoryManager(const MemoryManager &) = delete;
+    MemoryManager &operator=(const MemoryManager &) = delete;
+
+    void insert_block(void *ptr, size_t size);
+    void erase_block(void *ptr);
+
+    std::shared_ptr<std::atomic<size_t>> _memory_used_in_bytes_ptr;
```
harsha-simhadri, line 42: Why a shared pointer as opposed to std::atomic<size_t>?

harsha-simhadri (changes requested):

src/allocator.cpp
yiylin: The Allocator class is planned to use in std::vector, std::unordered_map .etc to track their memory usage

Dockerfile
yiylin: yes. for me, I do not have linux, so I use the docker to build/test codes in linux

tests/unittests/memory_manager_tests.cpp
yiylin: to speed up the review process, I have created another PR which only adds the unit test project. https://github.com/microsoft/DiskANN/pull/365

include/allocator.h
yiylin: the implementation here is very simple. it is fine to keep it here?

.github/actions/build/action.yml
```diff
@@ -8,7 +8,7 @@ runs:
       if: ${{ runner.os == 'Linux' }}
       run: |
         sudo scripts/dev/install-dev-deps-ubuntu.bash
-        cmake -S . -B build -DCMAKE_BUILD_TYPE=Release
+        cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DFORCE_TO_TRACK_MEMORY_IN_ALLOCATOR=True
```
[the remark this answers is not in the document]
yiylin: changed

src/CMakeLists.txt
yiylin: changed.

src/dll/CMakeLists.txt
yiylin: changed

PR #360, opened 2023-05-23
Add WIN macro for non-win function

Sanhaoji2:
 
1. generate_label_specific_vector_files() use Unix system function in the implement which windows doesn't contain. In our CoreXT environment, that will cause compile error even this template isn't instanced.
2. Fix compile error of std::stringstream don't accept std::wstring.
The msbuild version is 16.11.3 in our environment.


src/windows_aligned_file_reader.cpp
```diff
@@ -43,9 +45,17 @@ void WindowsAlignedFileReader::register_thread()
         FILE_ATTRIBUTE_READONLY | FILE_FLAG_NO_BUFFERING | FILE_FLAG_OVERLAPPED | FILE_FLAG_RANDOM_ACCESS, NULL);
     if (ctx.fhandle == INVALID_HANDLE_VALUE)
     {
-        std::stringstream ss;
-        ss << m_filename.c_str();
-        diskann::cout << "Error opening " << ss.str() << " -- error=" << GetLastError() << std::endl;
+        const size_t c_max_filepath_len = 256;
+        size_t actual_len = 0;
+        char filePath[c_max_filepath_len];
```
yiylin, line 50: use vector<char> to avoid allocating big array in stack?

Sanhaoji2: I think 256 bytes is not so big?

src/windows_aligned_file_reader.cpp
```diff
@@ -6,6 +6,8 @@
 #include "windows_aligned_file_reader.h"
 #include <iostream>
 #include "utils.h"
+#include <locale>
```
dkaczynski, line 9: Is this included because of the use of `wcstomb_s`? (Go ahead and mark resolved if so)

src/windows_aligned_file_reader.cpp
```diff
@@ -43,9 +45,17 @@ void WindowsAlignedFileReader::register_thread()
         FILE_ATTRIBUTE_READONLY | FILE_FLAG_NO_BUFFERING | FILE_FLAG_OVERLAPPED | FILE_FLAG_RANDOM_ACCESS, NULL);
     if (ctx.fhandle == INVALID_HANDLE_VALUE)
     {
-        std::stringstream ss;
-        ss << m_filename.c_str();
-        diskann::cout << "Error opening " << ss.str() << " -- error=" << GetLastError() << std::endl;
+        const size_t c_max_filepath_len = 256;
```
dkaczynski, line 48: Should this be defined in `defaults.h` or some other location instead of being hardcoded here? Is there any chance of this value being reused in the codebase?

dkaczynski (approved):

src/windows_aligned_file_reader.cpp
harsha-simhadri: Why do we need an upper limit here?

Sanhaoji2: wcstombs_s() need pre-allocate char array as input parameter, so we have to allocate a buffer here, we can use vector<char>, but we still need resize() before using it.

Sanhaoji2: this function is convert wchar to char, and it only use for print error message, so maybe we can just same the original string as the member of WindowsAlignedFileReader (keep m_filename and m_original_filename).

src/windows_aligned_file_reader.cpp
Sanhaoji2: remove unuse header.

src/windows_aligned_file_reader.cpp
```diff
@@ -43,9 +44,17 @@ void WindowsAlignedFileReader::register_thread()
         FILE_ATTRIBUTE_READONLY | FILE_FLAG_NO_BUFFERING | FILE_FLAG_OVERLAPPED | FILE_FLAG_RANDOM_ACCESS, NULL);
     if (ctx.fhandle == INVALID_HANDLE_VALUE)
     {
```
yiylin, line 46: I define this "UNICODE" to avoid the changing in this file

yiylin (approved):

PR #362, opened 2023-05-27
Reduce the size of coord_scratch in SSDQueryScratch to reduce memory usage

yiylin:
 
According to current implement, we just need one embedding buffer(coord_scratch) with the size(sizeof(T) * aligned_dim) and the buffer could be re-used for different nodes.
Not need to create MAX_N_CMPS(16384) embedding buffer.

### DIFF
--- a/src/pq_flash_index.cpp
+++ b/src/pq_flash_index.cpp
@@ -1125,7 +1125,6 @@ void PQFlashIndex<T, LabelT>::cached_beam_search(const T *query1, const uint64_t
 
     // pointers to buffers for data
     T *data_buf = query_scratch->coord_scratch;
-    uint64_t &data_buf_idx = query_scratch->coord_idx;
     _mm_prefetch((char *)data_buf, _MM_HINT_T1);
 
     // sector scratch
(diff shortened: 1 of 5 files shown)


src/pq_flash_index.cpp
```diff
@@ -1333,24 +1328,18 @@ void PQFlashIndex<T, LabelT>::cached_beam_search(const T *query1, const uint64_t
...
-                data_buf_idx = 0;
 
-            T *node_fp_coords_copy = data_buf + (data_buf_idx * aligned_dim);
-            data_buf_idx++;
-            memcpy(node_fp_coords_copy, node_fp_coords, disk_bytes_per_point);
             float cur_expanded_dist;
             if (!use_disk_index_pq)
             {
-                cur_expanded_dist = dist_cmp->compare(aligned_query_T, node_fp_coords_copy, (uint32_t)aligned_dim);
+                cur_expanded_dist = dist_cmp->compare(aligned_query_T, node_fp_coords, (uint32_t)aligned_dim);
```
yiylin, line 1335: I do not think we need to copy the data to some buf, then compute the distance. We do not compute the distance in-parallel, so I do not think we need to use "coord_scratch". Maybe some old version uses this way to compute the distance in-parallel

harsha-simhadri:
Could you please complete the PR temaplate -- What does this implement/fix? Briefly explain your changes? Does it resolve any issues ?

yiylin:
> Could you please complete the PR temaplate -- What does this implement/fix? Briefly explain your changes? Does it resolve any issues ?

added

Sanhaoji2 (approved):

gopalrs (approved):

PR #364, opened 2023-05-31
Python Type Enhancements

daxpryce:
This is my last absurdly large PR. It is admittedly too large, and touches on way too many things.  I'm not sure how to untangle it in any meaningful way, so we'll do this as one last big thing and I'll keep all my future enhancements much more narrowly scoped.  As it turns out, "fix the python bindings" is quite large.

Notable changes:
- tcmalloc is disabled for python builds in linux.
  - windows still requires work on this specific task
 - we had to support different locations of installed system libraries so that we could build the python bindings in `cibuildwheel`, most importantly that it's `Alma Linux 8`, which uses `dnf` instead of apt and has very different package names and locations
 -  **adds a new metadata file** that only contains information pertaining to one of 3 valid vector dtypes, the distance metric used, number of vectors, and number of dimensions.
   - This really should be stored via the C++ library, and used by the C++ library.  I have a fallback currently written such that if you build an index (or already have one) without this metadata file, you *can* provide these values for loading, but ideally we can just capture what is needed for load and search in a metadata file and prefer that over any config values - or at least warn if they deviate?  The current implementation is mostly a proof of concept that should be made integral to the index building and loading process, again, if the file exists.

### DIFF
--- a/python/apps/in-mem-dynamic.py
+++ b/python/apps/in-mem-dynamic.py
@@ -106,7 +106,11 @@ if __name__ == "__main__":
         args.gt_file,
     )
 
-# An ingest optimized example with SIFT1M
-# python3 ~/DiskANN/python/apps/in-mem-dynamic.py -d float \
-# -i sift_base.fbin -q sift_query.fbin --gt_file  gt100_base \
-# -Lb 10 -R 30 -Ls 200
+"""
+An ingest optimized example with SIFT1M
+source venv/bin/activate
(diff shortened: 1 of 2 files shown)


bryantower (approved):
Looks reasonable. It is quite a large commit.

daxpryce:
The tcmalloc stuff still isn't fixed.  I should have known it wasn't just a simple thing.

I'll rectify it later, after I create an issue for it.  It's a good way to keep my changes small, eh? :D

PR #366, opened 2023-06-01
Upload data and binary files to artifact in CI workflows

yiylin:
… locally when the workflows fails

Upload data and binary files to artifact so that we could debug issues locally when the CI workflows fail.


daxpryce (approved):
These look good - I'm not sure we'll be able to use anything in the dist, but it also won't hurt to have it just in case we can get _something_ out of it!

PR #368, opened 2023-06-05
fixed a bug with loading medoids for sharded filtered index, and adde…

rakri:
…d better caching for filtered index

This PR does two things:
(i) if the index is a sharded build, there will be multiple medoids for each filter, so we have changed filter_to_medoid_id to filter_to_medoid_ids, one per shard. 
(ii) for filtered index, we now do smarter "filter-aware" caching by sampling filter labels according to base distribution and running filtered searches to identify most visited nodes to cache.

### DIFF
--- a/src/pq_flash_index.cpp
+++ b/src/pq_flash_index.cpp
@@ -1247,7 +1247,7 @@ void PQFlashIndex<T, LabelT>::cached_beam_search(const T *query1, const uint64_t
     {
         if (_filter_to_medoid_ids.find(filter_label) != _filter_to_medoid_ids.end())
         {
-            const auto& medoid_ids = _filter_to_medoid_ids[filter_label];
+            const auto &medoid_ids = _filter_to_medoid_ids[filter_label];
             for (uint64_t cur_m = 0; cur_m < medoid_ids.size(); cur_m++)
             {
                 // for filtered index, we dont store global centroid data as for unfiltered index, so we use PQ distance
--- a/src/pq_flash_index.cpp
+++ b/src/pq_flash_index.cpp
@@ -1247,15 +1247,16 @@ void PQFlashIndex<T, LabelT>::cached_beam_search(const T *query1, const uint64_t
     {
         if (_filter_to_medoid_ids.find(filter_label) != _filter_to_medoid_ids.end())
         {
-            for (uint64_t cur_m = 0; cur_m < _filter_to_medoid_ids[filter_label].size(); cur_m++)
+            const auto& medoid_ids = _filter_to_medoid_ids[filter_label];
+            for (uint64_t cur_m = 0; cur_m < medoid_ids.size(); cur_m++)
             {
                 // for filtered index, we dont store global centroid data as for unfiltered index, so we use PQ distance
                 // as approximation to decide closest medoid matching the query filter.
-                compute_dists(&_filter_to_medoid_ids[filter_label][cur_m], 1, dist_scratch);
+                compute_dists(&medoid_ids[cur_m], 1, dist_scratch);
                 float cur_expanded_dist = dist_scratch[0];
                 if (cur_expanded_dist < best_dist)
                 {
-                    best_medoid = _filter_to_medoid_ids[filter_label][cur_m];
+                    best_medoid = medoid_ids[cur_m];
                     best_dist = cur_expanded_dist;
                 }
             }
--- a/src/pq_flash_index.cpp
+++ b/src/pq_flash_index.cpp
@@ -252,11 +252,23 @@ void PQFlashIndex<T, LabelT>::generate_cache_list_from_sample_queries(std::strin
     std::vector<uint64_t> tmp_result_ids_64(sample_num, 0);
     std::vector<float> tmp_result_dists(sample_num, 0);
 
+    bool filtered_search = false;
+    std::vector<LabelT> random_query_filters(sample_num);
+    if (_filter_to_medoid_ids.size() != 0)
+    {
+        filtered_search = true;
+        generate_random_labels(random_query_filters, (uint32_t)sample_num, nthreads);
+    }
+
 #pragma omp parallel for schedule(dynamic, 1) num_threads(nthreads)
     for (int64_t i = 0; i < (int64_t)sample_num; i++)
     {
+        auto &label_for_search = random_query_filters[i];
+        // run a search on the sample query with a random label (sampled from base label distribution), and it will
+        // concurrently update the node_visit_counter to track most visited nodes. The last false is to not use the
+        // "use_reorder_data" option which enables a final reranking if the disk index itself contains only PQ data.
         cached_beam_search(samples + (i * sample_aligned_dim), 1, l_search, tmp_result_ids_64.data() + i,
-                           tmp_result_dists.data() + i, beamwidth);
+                           tmp_result_dists.data() + i, beamwidth, filtered_search, label_for_search, false);
     }
 
     std::sort(this->node_visit_counter.begin(), node_visit_counter.end(),
@@ -304,9 +316,24 @@ void PQFlashIndex<T, LabelT>::cache_bfs_levels(uint64_t num_nodes_to_cache, std:
     cur_level = std::make_unique<tsl::robin_set<uint32_t>>();
     prev_level = std::make_unique<tsl::robin_set<uint32_t>>();
 
-    for (uint64_t miter = 0; miter < num_medoids; miter++)
+    for (uint64_t miter = 0; miter < num_medoids && cur_level->size() < num_nodes_to_cache; miter++)
     {
         cur_level->insert(medoids[miter]);
+    }
+
+    if ((_filter_to_medoid_ids.size() > 0) && (cur_level->size() < num_nodes_to_cache))
+    {
+        for (auto &x : _filter_to_medoid_ids)
+        {
+            for (auto &y : x.second)
+            {
+                cur_level->insert(y);
+                if (cur_level->size() == num_nodes_to_cache)
+                    break;
+            }
+            if (cur_level->size() == num_nodes_to_cache)
+                break;
+        }
     }
 
     uint64_t lvl = 1;
@@ -396,7 +423,7 @@ void PQFlashIndex<T, LabelT>::cache_bfs_levels(uint64_t num_nodes_to_cache, std:
         }
 
         diskann::cout << ". #nodes: " << node_set.size() - prev_node_set_size
-                      << ", #nodes thus far: " << node_list.size() << std::endl;
+                      << ", #nodes thus far: " << node_set.size() << std::endl;
         prev_node_set_size = node_set.size();
         lvl++;
     }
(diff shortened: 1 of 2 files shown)


src/pq_flash_index.cpp
```diff
@@ -252,11 +252,20 @@ void PQFlashIndex<T, LabelT>::generate_cache_list_from_sample_queries(std::strin
...
+        generate_random_labels(random_query_filters, sample_num, nthreads);
+    }
+
 #pragma omp parallel for schedule(dynamic, 1) num_threads(nthreads)
     for (int64_t i = 0; i < (int64_t)sample_num; i++)
     {
-        cached_beam_search(samples + (i * sample_aligned_dim), 1, l_search, tmp_result_ids_64.data() + i,
-                           tmp_result_dists.data() + i, beamwidth);
+        auto &label_for_search = random_query_filters[i];
+        cached_beam_search(samples + (i * sample_aligned_dim), 1, l_search, tmp_result_ids_64.data() + (i * 1),
```
yiylin, line 267: remove "* 1" ?

src/pq_flash_index.cpp
```diff
@@ -252,11 +252,20 @@ void PQFlashIndex<T, LabelT>::generate_cache_list_from_sample_queries(std::strin
...
+                           tmp_result_dists.data() + (i * 1), beamwidth, filtered_search, label_for_search, false);
```
yiylin, line 268: remove " * 1"

yiylin, line 268: could you add a comment for "false", so that we know what it means?

src/pq_flash_index.cpp
```diff
@@ -717,24 +776,24 @@ int PQFlashIndex<T, LabelT>::load_from_separate_paths(uint32_t num_threads, cons
...
                     {
                         if (cnt == 0)
                             label = (LabelT)std::stoul(token);
                         else
-                            medoid = (uint32_t)stoul(token);
+                            medoids.push_back((uint32_t)stoul(token));
                         cnt++;
                     }
-                    _filter_to_medoid_id[label] = medoid;
+                    _filter_to_medoid_ids[label] = medoids;
```
yiylin, line 796: use "std::move(mediods)" to avoid copying vector elements

src/pq_flash_index.cpp
```diff
@@ -304,11 +313,26 @@ void PQFlashIndex<T, LabelT>::cache_bfs_levels(uint64_t num_nodes_to_cache, std:
     cur_level = std::make_unique<tsl::robin_set<uint32_t>>();
     prev_level = std::make_unique<tsl::robin_set<uint32_t>>();
 
-    for (uint64_t miter = 0; miter < num_medoids; miter++)
+    for (uint64_t miter = 0; miter < num_medoids && cur_level->size() < num_nodes_to_cache; miter++)
```
yiylin, line 316: Do you know why "cache_bfs_levels" is commented in https://github.com/microsoft/DiskANN/blob/0acff41968fa557b1099c056c78d8898326c1003/apps/search_disk_index.cpp#L123 ?

yiylin: Is it recommended to use this method? Because it is commented here, so I also do not use this method in our repo

src/pq_flash_index.cpp
```diff
@@ -1167,27 +1226,29 @@ void PQFlashIndex<T, LabelT>::cached_beam_search(const T *query1, const uint64_t
...
-    {
-        best_medoid = _filter_to_medoid_id[filter_label];
-    }
-    else
+    if (use_filter)
     {
-        throw ANNException("Cannot find medoid for specified filter.", -1, __FUNCSIG__, __FILE__, __LINE__);
+        if (_filter_to_medoid_ids.find(filter_label) != _filter_to_medoid_ids.end())
+        {
+            best_medoid = _filter_to_medoid_ids[filter_label][best_shard];
```
yiylin, line 1246: Can we guarantee that each shard contains all kinds of labels? if no, then here may have issue

yiylin: especially, can we guarantee that the best_shard contains this label "filter_label"

src/pq_flash_index.cpp
```diff
@@ -478,6 +502,41 @@ inline int32_t PQFlashIndex<T, LabelT>::get_filter_number(const LabelT &filter_l
     return idx;
 }
 
+template <typename T, typename LabelT>
```
yiylin, line 505: do you have experiment result to show the improvement?

src/pq_flash_index.cpp
```diff
@@ -478,6 +502,41 @@ inline int32_t PQFlashIndex<T, LabelT>::get_filter_number(const LabelT &filter_l
...
+    for (uint32_t i = 0; i < num_points; i++)
+    {
+        skip_locs.insert(_pts_to_label_offsets[i]);
+    }
+
+#pragma omp parallel for schedule(dynamic, 1) num_threads(nthreads)
+    for (uint32_t i = 0; i < num_labels; i++)
+    {
+        bool found_flag = false;
+        while (!found_flag)
```
yiylin, line 528: considering that the average of label number for each node is small, so this way of sampling may not be efficient. You could count the frequency of each label, then use std::discrete_distribution to generate the random value based on the label's frequency.

src/pq_flash_index.cpp
```diff
@@ -1167,27 +1226,29 @@ void PQFlashIndex<T, LabelT>::cached_beam_search(const T *query1, const uint64_t
     std::vector<Neighbor> &full_retset = query_scratch->full_retset;
 
     uint32_t best_medoid = 0;
+    uint32_t best_shard = 0;
     float best_dist = (std::numeric_limits<float>::max)();
-    if (!use_filter)
```
yiylin, line 1171: regarding to the searching parameter "use_filter", if the index is built for filter diskann, we do not expect users to search the index with "use_filter" == false; if the index is built for non-filter diskann, we do not expect users to search the index with "use_filter" == true. So we could remove this parameter or at least we need to do some checking in case uses mis-use the parameter.

src/pq_flash_index.cpp
yiylin: or you also could consider to save the label when we generate the sample data after building index

yiylin: and do you consider to generate the cache node list and save it to some file after building index. Actually, we do not like the extra time cost to search and generate the cache node list during loading index.

src/pq_flash_index.cpp
rakri: sure.

src/pq_flash_index.cpp
rakri: if we have sample query file, it is better to do caching based on most frequently visited nodes during search for sample queries, so we do that. if not, we can simply cache according to bfs from medoid(s). Both can be used, but former is usually better

src/pq_flash_index.cpp
```diff
@@ -304,11 +313,26 @@ void PQFlashIndex<T, LabelT>::cache_bfs_levels(uint64_t num_nodes_to_cache, std:
...
         cur_level->insert(medoids[miter]);
     }
 
+    if (_filter_to_medoid_ids.size() > 0)
+    {
+        for (auto &x : _filter_to_medoid_ids)
+        {
+            for (auto &y : x.second)
+            {
+                cur_level->insert(y);
```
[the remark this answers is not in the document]
rakri: will do.

src/pq_flash_index.cpp
rakri: yes, will add you to the thread.

src/pq_flash_index.cpp
rakri: trouble with generating cache list at build time is that we need to run sample searches to count the most frequent hits. For large index, the search will have to be a disk-index search, so we will add a dependency that build time also needs a good SSD. That is why we went with this design.

rakri: Hmm, counting label frequency and then sampling, is similar to sampling from the "expanded filter list", which is just a concatenated vector of all the labels right, except we need to discount the elements storing the offsets. If you can explain to me, or find examples where it is much slower, we can consider re-implementing.

src/pq_flash_index.cpp
rakri: done

src/pq_flash_index.cpp
rakri: good point, indeed this issue can arise. Let me think of a fix for this.

src/pq_flash_index.cpp
yiylin: we use the same hardwares as our prod environments. so we could search the index during building index. Not sure if you have concerns to add a parameter in the build_disk_index.cpp to get the cache list

yiylin: > Hmm, counting label frequency and then sampling, is similar to sampling from the "expanded filter list", which is just a concatenated vector of all the labels right, except we need to discount the elements storing the offsets. If you can explain to me, or find examples where it is much slower, we can consider re-implementing.

The final result is the same. I do not have the performance result. just feel that the number of unique labels is much smaller than the total number of labels. and the implementation is much simpler. I am fine if you want to keep current implementation.

src/pq_flash_index.cpp
```diff
@@ -1181,13 +1243,27 @@ void PQFlashIndex<T, LabelT>::cached_beam_search(const T *query1, const uint64_t
...
-    else if (_filter_to_medoid_id.find(filter_label) != _filter_to_medoid_id.end())
-    {
-        best_medoid = _filter_to_medoid_id[filter_label];
-    }
     else
     {
-        throw ANNException("Cannot find medoid for specified filter.", -1, __FUNCSIG__, __FILE__, __LINE__);
+        if (_filter_to_medoid_ids.find(filter_label) != _filter_to_medoid_ids.end())
+        {
+            for (uint64_t cur_m = 0; cur_m < _filter_to_medoid_ids[filter_label].size(); cur_m++)
```
yiylin, line 1250: use "const auto& medio_ids = _filter_to_medoid_ids[filter_label];" to refer to the vector to avoid lookuping _filter_to_medoid_ids map multiple times

yiylin: the same for line 1254 and line 1258

src/pq_flash_index.cpp
yiylin: yes. separate PR

src/pq_flash_index.cpp
```diff
@@ -252,11 +252,23 @@ void PQFlashIndex<T, LabelT>::generate_cache_list_from_sample_queries(std::strin
     std::vector<uint64_t> tmp_result_ids_64(sample_num, 0);
     std::vector<float> tmp_result_dists(sample_num, 0);
 
+    bool filtered_search = false;
+    std::vector<LabelT> random_query_filters(sample_num);
+    if (_filter_to_medoid_ids.size() != 0)
+    {
+        filtered_search = true;
+        generate_random_labels(random_query_filters, sample_num, nthreads);
```
yiylin, line 260: image

yiylin: you need to fix these warnings

gopalrs (approved):
Don't have any other comments

src/pq_flash_index.cpp
rakri: done

src/pq_flash_index.cpp
rakri: done

yiylin (approved):

PR #373, opened 2023-06-09
Fix inefficiency in constructing reverse label map

varunsivashankar:

In the function build_filtered_index, label_to_points was getting populated with a double loop over number of points and number of labels, which is very inefficient for a large number of labels. Instead, we can just do a single pass through the data to populate label_to_points, which makes the filtered build faster.

#### Any other comments?
It's a very local change, and I checked that build/search performance is not affected.


darvg (approved):
lgtm

rakri (approved):

gopalrs (approved):

Issue #317, opened 2023-04-14
Integrate building of incremental index with labeled data


dkaczynski:
Update the interface to the Index to support inserting labeled data points into a streaming index.

Update the `test_treaming_scenario.cpp` to support building an index with labeled data.

PR #376, opened 2023-06-13
Build streaming index of labeled data

dkaczynski:
 
#### Reference Issues/PRs
#317 

Add overloaded `insert_point` method to index class to support labeled data.

#### Any other comments?
Also adjusted the algorithm (both for streaming & regular batch index building) to fall back on unlabeled algo for populating `pruned_list` when the filtered algo results in it being empty. This prevents disconnected subgraphs when building an index of labeled data in which points are only connected if they share 1 or more labels. This behavior of creating an index of disconnected subgraphs is not unique to streaming indices and is present in the `main` branch, but encountering a point with a unique set of labels is a particularly common case when inserting points individually, like how the streaming index is built.


yashpatel007:
This branch is trying to add two features.
1. Adding function to perform unfiltered search on a batch built filtered index.
2. Adding Filtered Streaming Index support, which can be search with and without filters.

.github/workflows/labels.yml
```diff
@@ -55,6 +55,11 @@ jobs:
...
           dist/bin/search_memory_index --data_type uint8 --dist_fn l2 --filter_label 10 --fail_if_recall_below 70 --index_path_prefix data/index_l2_rand_uint8_10D_10K_norm50_wlabel --query_file data/rand_uint8_10D_1K_norm50.0.bin --recall_at 10 --result_path temp --gt_file data/l2_rand_uint8_10D_10K_norm50.0_10D_1K_norm50.0_gt100_wlabel -L  16 32
           dist/bin/search_memory_index --data_type uint8 --dist_fn cosine --filter_label 10 --fail_if_recall_below 70 --index_path_prefix data/index_cosine_rand_uint8_10D_10K_norm50_wlabel --query_file data/rand_uint8_10D_1K_norm50.0.bin --recall_at 10 --result_path temp --gt_file data/cosine_rand_uint8_10D_10K_norm50.0_10D_1K_norm50.0_gt100_wlabel -L  16 32
+
+          echo "Searching without filters"
+          dist/bin/search_memory_index --data_type uint8 --dist_fn l2 --fail_if_recall_below 50 --index_path_prefix data/index_l2_rand_uint8_10D_10K_norm50_wlabel --query_file data/rand_uint8_10D_1K_norm50.0.bin --recall_at 10 --result_path temp --gt_file data/l2_rand_uint8_10D_10K_norm50.0_10D_1K_norm50.0_gt100 -L  32 64
```
harsha-simhadri, line 60: why fail on 50 recall as opposed to 70?

apps/test_streaming_scenario.cpp
```diff
@@ -285,18 +317,16 @@ void build_incremental_index(const std::string &data_path, const uint32_t L, con
...
 
     diskann::aligned_free(data);
 }
 
 int main(int argc, char **argv)
 {
-    std::string data_type, dist_fn, data_path, index_path_prefix;
-    uint32_t insert_threads, consolidate_threads;
-    uint32_t R, L, num_start_pts;
+    std::string data_type, dist_fn, data_path, index_path_prefix, label_file, universal_label, label_type;
```
harsha-simhadri, line 328: Is an index built with filters backward compatible for searching without filters?

include/abstract_index.h
```diff
@@ -79,6 +79,9 @@ class AbstractIndex
                                                       const size_t K, const uint32_t L, IndexType *indices,
                                                       float *distances);
 
+    template <typename data_type, typename tag_type, typename label_type>
+    int insert_point(const data_type *point, const tag_type tag, const std::vector<label_type> &labels = {});
```
harsha-simhadri, line 83: Could we document why there are multiple "insert_point" functions and the scenarios that require them?

include/abstract_index.h
```diff
@@ -79,6 +79,9 @@ class AbstractIndex
...
+
```
harsha-simhadri, line 84: should the base class AbstractIndex have insert with labels? What if some one wants to write an index that does not support labels?

apps/test_insert_deletes_consolidate.cpp
```diff
@@ -160,6 +159,7 @@ void build_incremental_index(const std::string &data_path, diskann::IndexWritePa
                                             .with_index_write_params(params)
                                             .with_search_threads(params.num_threads)
                                             .with_initial_search_list_size(params.search_list_size)
+                                            .with_num_frozen_pts(params.num_frozen_points)
                                             .with_data_type(data_type)
```
harsha-simhadri, line 163: Did you test the impact of deletes and consolidate_deletes on recall?

include/index.h
```diff
@@ -377,6 +384,7 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
     std::string _labels_file;
     std::unordered_map<LabelT, uint32_t> _label_to_medoid_id;
     std::unordered_map<uint32_t, uint32_t> _medoid_counts;
+    std::unordered_map<LabelT, uint32_t> _label_counts;
```
harsha-simhadri, line 387: do we need to protect any of the label related data structures with a lock? If so, can we document it in index.h?

src/index.cpp
```diff
@@ -3098,7 +3235,6 @@ template <typename T, typename TagT, typename LabelT> int Index<T, TagT, LabelT>
     _delete_set->insert(location);
     _location_to_tag.erase(location);
     _tag_to_location.erase(tag);
-
     return 0;
 }
```
harsha-simhadri, line 3240: How are handling deletes and related clean up?

.github/workflows/labels.yml
yashpatel007: Actually, the accuracy is 95%+, this was initially added as anticipation that the performance will be somewhat poor, but it is clearly not.

include/abstract_index.h
yashpatel007: Sure, I would add this to appropriate Readme file.

include/abstract_index.h
yashpatel007: There should be a method to do just that, its right below.

include/index.h
yashpatel007: This i believe was added by David, I don't think its being used as such. I will remove it.

.github/workflows/labels.yml
yashpatel007: fail_if_recall_below bumped up to 70%

apps/test_streaming_scenario.cpp
yashpatel007: Yes, it is and I have added dynamic-labels.yml work flow to test just that.

src/index.cpp
yashpatel007: With current strategy, we don't need to do much after delete. We have made frozen_points as medoids, so deleting medoids is not possible. Also when we do compact_data after deleting and repositioning frozen_points, medoids are also repositioned accordingly as in filtered streaming scenario medoids = frozen_points.

include/index.h
```diff
@@ -283,6 +287,9 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
     // Acquire exclusive _tag_lock and _delete_lock before calling
     int reserve_location();
 
+    // reposition medoids/frozen points for dynamic filtered index.
+    void reposition_medoids(uint32_t old_location_start, uint32_t new_location_start);
```
rakri, line 291: wasn't there such a function even for unfiltered dynamic index?

src/index.cpp
```diff
@@ -362,6 +375,10 @@ void Index<T, TagT, LabelT>::save(const char *filename, bool compact_before_save
                 assert(label_writer.is_open());
                 for (uint32_t i = 0; i < _pts_to_labels.size(); i++)
                 {
+                    if (_pts_to_labels[i].empty())
```
rakri, line 378: what is this for?

src/index.cpp
```diff
@@ -603,7 +620,7 @@ void Index<T, TagT, LabelT>::load(const char *filename, uint32_t num_threads, ui
     {
         _label_map = load_label_map(labels_map_file);
         parse_label_file(labels_file, label_num_pts);
-        assert(label_num_pts == data_file_num_pts);
+        assert(label_num_pts == data_file_num_pts - _num_frozen_pts);
```
rakri, line 623: if we are loading an index, wont the labels file contain label info for the frozen points also?

src/index.cpp
rakri: essentially, _pts_to_labels will have size num_pts  + num_frozen_pts, right? if so , see the comment in load section

src/index.cpp
```diff
@@ -1170,9 +1189,34 @@ void Index<T, TagT, LabelT>::search_for_point_and_prune(int location, uint32_t L
         for (auto &x : _pts_to_labels[location])
             filter_specific_start_nodes.emplace_back(_label_to_medoid_id[x]);
 
+        std::set<Neighbor> best_candidate_pool;
```
rakri, line 1192: should we use robin_set, and should we push this into scratch, to avoid creating local variables and alloc-ing memory each time we call insert point?

src/index.cpp
```diff
@@ -1255,6 +1298,8 @@ void Index<T, TagT, LabelT>::occlude_list(const uint32_t location, std::vector<N
                 {
                     uint32_t a = iter->id;
                     uint32_t b = iter2->id;
+                    if (_pts_to_labels.size() < b || _pts_to_labels.size() < a)
```
rakri, line 1301: under what scenario will this occur? i guess unfiltered index will have size == 0. anything else?

src/index.cpp
```diff
@@ -2636,6 +2682,10 @@ template <typename T, typename TagT, typename LabelT> void Index<T, TagT, LabelT
         reposition_points((uint32_t)_max_points, (uint32_t)_nd, (uint32_t)_num_frozen_pts);
         _start = (uint32_t)_nd;
     }
+    if (_filtered_index && _dynamic_index)
```
rakri, line 2685: why should this be called AND line 2682 be called? wont it reposition twice?

src/index.cpp
```diff
@@ -2892,6 +2942,29 @@ template <typename T, typename TagT, typename LabelT> void Index<T, TagT, LabelT
...
     reposition_points((uint32_t)_nd, (uint32_t)_max_points, (uint32_t)_num_frozen_pts);
     _start = (uint32_t)_max_points;
+
+    if (_filtered_index && _dynamic_index)
+    {
+        reposition_medoids((uint32_t)_nd, (uint32_t)_max_points);
+    }
+}
+
+template <typename T, typename TagT, typename LabelT>
```
rakri, line 2952: comment on what this function does

include/index.h
yashpatel007: Yes, there is reposition_frozen_points(), but in this case actually we are nor repositioning anything we are just updating respective medoid id with already repositioned frozen points, so that they would align. We can rename the function as update_medoid_location or something like that.

src/index.cpp
yashpatel007: Yes, you are right, but in a case where total points is 12000 and we have max_pts_to_load is 10000, as we have in test_insert_delete_consolidate, the final points will have _pts_to_labels as empty. Notice the for loop is acting on **_pts_to_labels.size()** and not **_nd** which is number of active points.

src/index.cpp
yashpatel007: I believe this was added by David, I can ask him the reasoning behind this. What you said more sense.

src/index.cpp
yashpatel007: Shall we create something like combined_candidates_pool, or just don't check if the pool is empty() and insert the candidates at the very back?

src/index.cpp
yashpatel007: Added the comment.
/* updates medoid id's in case of filtered dynamic index as frozen_points acts as medoids.
 * is called after reposition_frozen_points().
 */
template <typename T, typename TagT, typename LabelT>
void Index<T, TagT, LabelT>::reposition_medoids(uint32_t old_location_start, uint32_t new_location_start){......}

src/index.cpp
yashpatel007: I see the confusion here, the function ```reposition_points``` on line 2682, actually reposition all points where as reposition_medoids just updates the id's of medoids as fz_points = medoids have been moved.

yashpatel007: the function has been removed and logic is now incorporated into reposition_points.

include/abstract_index.h
yashpatel007: Added docs to dynamic_index.md file, also added instructions to build and search dynamic filtered index.

src/index.cpp
yashpatel007: This was added a sanitary check, In case of Filtered Streaming Index pts_to_labels will have size num_points + num_frozen_points. In case where num_frozen_points > unique labels.  and the candidate has a frozen_point without filter. It results in program crash.

apps/test_insert_deletes_consolidate.cpp
yashpatel007: I am adding code to test that as well.

include/abstract_index.h
yashpatel007: Added instructions in dynamic_index.md for filtered streaming index.

src/index.cpp
rakri: but wont this skip lines, if an intermediate point is deleted? we should at least write out a blank line right?

src/index.cpp
rakri: we need to make sure all sizes check out correctly. _pts_to_labels is set to size of total internal points in resize operation. Here we are checking for data_file_num_pts - num_frozen_points, etc. Need to make these consistent.

src/index.cpp
rakri: can you elaborate bit more? if we continue, we wont update the occlude factor etc. right? wont that be incorrect behavior, especially for non-filtered index?

rakri: For example, it eems like the for loop below insists that a has all the labels that b has. So we can allow empty labels for b right?

src/index.cpp
yashpatel007: I agree. How would we avoid the overflow condition.

rakri (changes requested):

src/index.cpp
```diff
@@ -2877,6 +2923,17 @@ void Index<T, TagT, LabelT>::reposition_points(uint32_t old_location_start, uint
...
     }
     _data_store->move_vectors(old_location_start, new_location_start, num_locations);
+
+    // update medoid id's as frozen points are treated as medoid
+    if (_filtered_index && _dynamic_index)
+    {
+        for (auto &[label, medoid_id] : _label_to_medoid_id)
+        {
+            if (label == 0)
+                continue;
```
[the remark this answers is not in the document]
yashpatel007: We are not moving the medoid for universal_label that is why we do not move it while updating rest of fz_pts

src/index.cpp
```diff
@@ -2877,6 +2923,17 @@ void Index<T, TagT, LabelT>::reposition_points(uint32_t old_location_start, uint
         }
     }
     _data_store->move_vectors(old_location_start, new_location_start, num_locations);
+
```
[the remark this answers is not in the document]
yashpatel007: Updated compact_data to now update _pts_to_labels as we compact data. and also when we call compact_fz_points()

src/index.cpp
```diff
@@ -1458,19 +1503,20 @@ void Index<T, TagT, LabelT>::link(const IndexWriteParameters &parameters)
...
         {
-            search_for_point_and_prune(node, _indexingQueueSize, pruned_list, scratch, _filtered_index,
-                                       _filterIndexingQueueSize);
+            search_for_point_and_prune(node, _indexingQueueSize, pruned_list, scratch, true, _filterIndexingQueueSize);
         }
         else
         {
             search_for_point_and_prune(node, _indexingQueueSize, pruned_list, scratch);
         }
+        assert(pruned_list.size() > 0);
```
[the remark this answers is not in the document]
yashpatel007: I thought it was there already on main. May be David added it and it skipped my eye. Removed this.

src/index.cpp
```diff
@@ -995,6 +1004,31 @@ void Index<T, TagT, LabelT>::search_for_point_and_prune(int location, uint32_t L
...
+        // clear scratch for finding unfiltered candidates
+        scratch->clear();
+
+        _data_store->get_vector(location, scratch->aligned_query());
+        iterate_to_fixed_point(scratch->aligned_query(), Lindex, init_ids, scratch, false, unused_filter_label, false);
+
+        for (auto unfiltered_neighbour : scratch->pool())
+        {
+            // insert if this neighbour is not already in best_candidate_pool
+            if (best_candidate_pool.find(unfiltered_neighbour) == best_candidate_pool.end())
```
rakri, line 1024: we probably also have a robin_set in scratch which tracks all th e points inserted into pool. A lookup into that might be more efficient.?

apps/test_insert_deletes_consolidate.cpp
```diff
@@ -272,12 +298,21 @@ void build_incremental_index(const std::string &data_path, diskann::IndexWritePa
...
-        const auto save_path_inc = get_save_filename(save_path + ".after-concurrent-delete-", points_to_skip,
-                                                     points_to_delete_from_beginning, last_point_threshold);
         index->save(save_path_inc.c_str(), true);
     }
     else
     {
+        const auto save_path_inc = get_save_filename(save_path + ".after-delete-", points_to_skip,
+                                                     points_to_delete_from_beginning, last_point_threshold);
+        std::string labels_file_to_use = save_path_inc + "_label_formatted.txt";
+        std::string mem_labels_int_map_file = save_path_inc + "_labels_map.txt";
```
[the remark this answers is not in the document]
yashpatel007: Its the save path that is being changed ( on conditions ) and we have to generate the label file where the path looks like
`<save_path> + "_labels_formatted.txt"`

include/filter_utils.h
```diff
@@ -57,6 +57,10 @@ DISKANN_DLLEXPORT void generate_label_indices(path input_data_path, path final_i
 DISKANN_DLLEXPORT load_label_index_return_values load_label_index(path label_index_path,
                                                                   uint32_t label_number_of_points);
 
+template <typename LabelT>
+DISKANN_DLLEXPORT std::tuple<std::vector<std::vector<LabelT>>, tsl::robin_set<LabelT>> parse_formatted_label_file(
```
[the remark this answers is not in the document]
yashpatel007: To be honest, I looked at both and there is not much of a difference I see, David initially implemented this so I kept it untouched. 

I will revisit the function and eliminate it if necessary

src/index.cpp
yashpatel007: I tried using local RobinSet, but it had some errors. Can you suggest something from scratch that we may use here.

apps/test_streaming_scenario.cpp
```diff
@@ -167,23 +179,23 @@ void delete_and_consolidate(diskann::AbstractIndex &index, diskann::IndexWritePa
...
-    using LabelT = uint32_t;
+    bool has_labels = label_file != "";
 
     diskann::IndexWriteParameters params = diskann::IndexWriteParametersBuilder(L, R)
                                                .with_max_occlusion_size(C)
                                                .with_alpha(alpha)
                                                .with_saturate_graph(saturate_graph)
                                                .with_num_threads(insert_threads)
-                                               .with_num_frozen_points(num_start_pts)
+                                               .with_filter_list_size(Lf)
```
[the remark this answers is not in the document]
yashpatel007: Regarding universal label, can you please take a look  at this point
https://microsoft-my.sharepoint.com/:wb:/p/patelyash/Ef5XPJQRxGBMnryWV7o1rMEBAEb_U0zZesMJTHYJi_hvrA?e=2efnJV

include/filter_utils.h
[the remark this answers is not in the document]
yashpatel007: Upon close inspection i think we can eliminate `parse_formatted_label_file()` if we can make sure that universal label is always mapped to 0

apps/test_streaming_scenario.cpp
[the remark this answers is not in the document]
yashpatel007: We are now exposing, set_universal_label() through AbstractIndex, so we can set it for dynamic index.

include/filter_utils.h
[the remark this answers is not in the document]
yashpatel007: Also this returns a pts_to_labels in a `vector<vector<LabelT>>` format while as parse_label_file returns results in `vector<set<LabelT>>`  also this assumes that everything is mapped already.

We can remove this function as we support a set of universal labels.

.github/workflows/dynamic-labels.yml
```diff
@@ -0,0 +1,102 @@
...
+        run: |
+          echo "Generating synthetic labels and computing ground truth for filtered search with universal label"
+          dist/bin/generate_synthetic_labels  --num_labels 50 --num_points 10000  --output_file data/rand_labels_50_10K.txt --distribution_type random
+
+          echo "Generating synthetic labels with a zipf distribution and computing ground truth for filtered search with universal label"
+          dist/bin/generate_synthetic_labels  --num_labels 50 --num_points 10000  --output_file data/zipf_labels_50_10K.txt --distribution_type zipf
+
+      - name: Test a streaming index (float) with labels (Zipf distributed)
+        run: |
+          dist/bin/test_streaming_scenario --data_type float --dist_fn l2 --data_path data/rand_float_10D_10K_norm1.0.bin --universal_label 0 --label_file data/zipf_labels_50_10K.txt --index_path_prefix data/index_zipf_stream -R 64 --FilteredLbuild 200 -L 50 --alpha 1.2 --insert_threads 1 --consolidate_threads 1 --max_points_to_insert 10000 --active_window 4000 --consolidate_interval 2000 --start_point_norm 3.2 --unique_labels_supported 51
```
harsha-simhadri, line 42: shall we run the test with multiple threads to check for any concurrency bugs.  Also, can concurrent consolidate be true?

include/index_build_params.h
```diff
@@ -8,6 +8,7 @@ struct IndexFilterParams
   public:
     std::string save_path_prefix;
     std::string label_file;
+    std::string tags_file;
```
harsha-simhadri, line 11: start member var with underscore?

apps/utils/compute_groundtruth_for_filters.cpp
```diff
@@ -415,11 +415,11 @@ inline void parse_label_file_into_vec(size_t &line_cnt, const std::string &map_f
             lbls.push_back(token);
             labels.insert(token);
         }
-        if (lbls.size() <= 0)
```
harsha-simhadri, line 418: just remove the code?

workflows/dynamic_index.md
```diff
@@ -22,6 +22,15 @@ The program then simultaneously inserts newer points drawn from the file and del
 in chunks of `consolidate_interval` points so that the number of active points in the index is approximately `active_window`.
 It terminates when the end of data file is reached, and the final index has `active_window + consolidate_interval` number of points.
 
+The index also supports filters on steaming index, you can use `insert_point` function overloads to either insert points as before or insert points with labels.
+Additional options are added to support this in `apps/test_streaming_scenario` and `apps/test_streaming_scenario` please refer to program arguments for more details.
+
+---
+> Note
+* The index does not support mixed points, that is, either all points do not have filters or all points have filters.
```
harsha-simhadri, line 30: you mean all points have labels or no labels?

workflows/dynamic_index.md
```diff
@@ -63,7 +72,13 @@ The arguments are as follows:
 12. **--consolidate_interval**: Granularity at which insert and delete functions are called.
 13. **--start_point_norm**: Set the starting node to a random point on a sphere of this radius.  A reasonable choice is to set this to the average norm of the data stream.
 
+** to build with filters add these
 
+14. **--label_file**: Filter data for each point, in `.txt` format. Line `i` of the file consists of a comma-separated list of filters corresponding to point `i` in the file passed via `--data_file`.
```
harsha-simhadri, line 77: list of labels instead of list of filters? Also, are labels numeric or can they be arbitrary strings?

workflows/dynamic_index.md
```diff
@@ -63,7 +72,13 @@ The arguments are as follows:
 12. **--consolidate_interval**: Granularity at which insert and delete functions are called.
 13. **--start_point_norm**: Set the starting node to a random point on a sphere of this radius.  A reasonable choice is to set this to the average norm of the data stream.
 
+** to build with filters add these
```
harsha-simhadri, line 75: mark as optional?

workflows/dynamic_index.md
yashpatel007: I have made the chnage in wording. And yes it does support any type of label. Both Strings and Numbers are supported as we create an internal map irrespective of the raw label type.

workflows/dynamic_index.md
yashpatel007: Yes. I changed the wording here as well.

include/index_build_params.h
yashpatel007: May be you did not notice. These are public params and in code we make private variables with underscore.

.github/workflows/dynamic-labels.yml
yashpatel007: Ohh this remained out of sight, I have tested the code with 15 threads and it works I will update the workflow. I am updating the threads to 8, to match with the rest of code.

workflows/dynamic_index.md
```diff
@@ -63,7 +72,13 @@ The arguments are as follows:
...
 
+14. **--label_file**: Filter data for each point, in `.txt` format. Line `i` of the file consists of a comma-separated list of filters corresponding to point `i` in the file passed via `--data_file`.
+15. **--FilteredLbuild**: If building a filtered index, we maintain a separate search list from the one provided by `--Lbuild/-L`.
+16. **--num_start_points**: number of frozen points in this case should be more then number of unique labels. 
+17. **--universal_label**: Optionally, the the filter data may contain a "wild-card" filter corresponding to all filters. This is referred to as a universal label. Note that if a point has the universal label, then the filter data must only have the universal label on the line corresponding to said point.
```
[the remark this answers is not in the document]
yashpatel007: We may need to update at other places as well, because I copied it from other file.

apps/build_memory_index.cpp
```diff
@@ -127,7 +127,7 @@ int main(int argc, char **argv)
                                       .with_num_threads(num_threads)
                                       .build();
 
-        auto build_params = diskann::IndexBuildParamsBuilder(index_build_params)
+        auto build_params = diskann::IndexBuildParamsBuilder()
```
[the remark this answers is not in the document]
yashpatel007: It is not required here because our config and constructor, both takes the build params upfront.

workflows/dynamic_index.md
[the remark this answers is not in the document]
yashpatel007: The definition has been updated to:
```
Optionally, the label data may contain a special "universal" label. A point with the universal label can be matched against a query with any label. Note that if a point has the universal label, then the filter data must only have the universal label on the line corresponding.
```

.github/workflows/dynamic-labels.yml
```diff
@@ -0,0 +1,102 @@
...
+
+          echo "Computing groundtruth with filter"
+          dist/bin/compute_groundtruth_for_filters --data_type float --filter_label 5 --universal_label 0 --dist_fn l2 --base_file data/index_zipf_ins_del.after-concurrent-delete-del2500-7500.data --query_file data/rand_float_10D_1K_norm1.0.bin --K 100 --gt_file data/gt100_zipf_random10D_1K_wlabel_5 --label_file data/index_zipf_ins_del.after-concurrent-delete-del2500-7500_raw_labels.txt --tags_file data/index_zipf_ins_del.after-concurrent-delete-del2500-7500.tags
+          echo "Searching with filter"
+          dist/bin/search_memory_index --data_type float --dist_fn l2 --filter_label 5 --fail_if_recall_below 10 --index_path_prefix data/index_zipf_ins_del.after-concurrent-delete-del2500-7500 --result_path data/res_zipf_stream --query_file data/rand_float_10D_1K_norm1.0.bin --gt_file data/gt100_zipf_random10D_1K_wlabel_5 -K 10 -L 20 40 60 80 100 150 -T 64 --dynamic true --tags 1
+
+          echo "Computing groundtruth w/o filetr"
```
harsha-simhadri, line 77: typo: filetr-> filter

include/index.h
```diff
@@ -362,11 +369,14 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
     // Filter Support
 
     bool _filtered_index = false;
-    std::vector<std::vector<LabelT>> _pts_to_labels;
+    // Location to label is only updated during insert_point(), all other reads are protected by
+    // default as a location can only be released at end of consolidate deletes
+    std::vector<std::vector<LabelT>> _location_to_labels;
     tsl::robin_set<LabelT> _labels;
     std::string _labels_file;
     std::unordered_map<LabelT, uint32_t> _label_to_medoid_id;
```
harsha-simhadri, line 377: should we just call it _label_to_start_id? Medoid can be one way to decide start node, but there could be better ones, right?

apps/test_insert_deletes_consolidate.cpp
```diff
@@ -91,16 +92,22 @@ std::string get_save_filename(const std::string &save_path, size_t points_to_ski
     return final_path;
 }
 
-template <typename T, typename TagT>
+template <typename T, typename TagT, typename LabelT>
 void insert_till_next_checkpoint(diskann::AbstractIndex &index, size_t start, size_t end, int32_t thread_count, T *data,
-                                 size_t aligned_dim)
+                                 size_t aligned_dim, std::vector<std::vector<LabelT>> &pts_to_labels)
```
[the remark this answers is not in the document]
yashpatel007: updated the naming

workflows/dynamic_index.md
```diff
@@ -83,6 +100,9 @@ The arguments are as follows:
 10. **--dynamic** (default false): whether the index being searched is dynamic or not.
 11. **--tags** (default false): whether to search with tags. This should be used if point *i* in the ground truth file does not correspond the point in the *i*th position in the loaded index.
 
+** to search with filters add these
+
+12. **--filter_label**: Filter for each query. For each query, a search is performed with this filter.
```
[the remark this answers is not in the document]
yashpatel007: Yes we can add it as an issue. The compute GT works with having separate label for each query. May be the logic can be extended from there.

include/index.h
yashpatel007: renamed variable to _label_to_start_id

harsha-simhadri (approved):

workflows/dynamic_index.md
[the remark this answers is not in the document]
yashpatel007: created an issue

rakri (approved):

PR #385, opened 2023-06-29
Python Refactor

daxpryce:
The purpose of this PR was primarily to rewrite the bindings, hoping that we introduced an issue just at the pybind11 boundary that caused segfaults in CPython on linux when python terminates. While I can consistently replicate it, I can't say it's fixed.

Still, we can at least flight the api and work through any of the build issues in CICD or publication to PyPI with this rc release, and it does work most of the time.


bryantower (approved):
Looks good.

Issue #370, opened 2023-06-07
Validate required arguments in CLI tools


jonmclean:
## Is your feature request related to a problem? Please describe.
The CLI tools help documentation does not always make it clear which arguments are required.  This has caused me to have to spend a lot of time debugging an error just to find out that an argument was missing.

For example, in search_memory_index.cpp the "search_list" parameter is required but is missing "->required()" which would force the CLI to exit with a descriptive error message if the argument is not supplied.

## Describe the solution you'd like
For each argument in each CLI tool:
* Determine if the argument is required or optional.
* For all required arguments:
  * Verify that it has "->required()" to force the user to supply the argument
  * Verify that the Help documentation says that the argument is required.  Standardize the way required args are displayed.  Maybe "-K number of items to return [REQUIRED]" or something similar.

## Stretch goal
Validate that parameters are of the required type and in the required range.

PR #390, opened 2023-07-11
Improve help formatting in CLI tools

jonmclean:
 
#### Reference Issues/PRs
#370 - helps user identify required/optional arguments using the command-line help.

* Added "Required" and "Optional" sections to all CLI tool help documentation help users differentiate between required and optional arguments
* Centralized common argument strings so we can easily update them in all locations.  As we make the help more descriptive this will make sure we catch all instances of each argument.
* Standardized the "Usage: <cmd> [OPTIONS]" string for all CLI tools.


apps/build_disk_index.cpp
```diff
@@ -9,6 +9,7 @@
 #include "math_utils.h"
 #include "index.h"
 #include "partition.h"
+#include "../include/program_options_utils.hpp"
```
daxpryce, line 12: should just have to do `#include "program_option_utils.hpp" - the include directories are already pulled in as the root dir for includes

jonmclean: Thank you.  Removed the superfluous paths and spot-checked a few of the compiled tools to make sure they still launch.

include/program_options_utils.hpp
```diff
@@ -0,0 +1,71 @@
...
+        .append(description)
+        .append("\n\n")
+        .append("Usage: ")
+        .append(executable_name)
+        .append(" [OPTIONS]");
+}
+
+// Required parameters
+const char *DATA_TYPE_DESCRIPTION = "data type, one of {int8, uint8, float} - float is single precision (32 bit)";
+const char *DISTANCE_FUNCTION_DESCRIPTION =
```
daxpryce, line 23: let's change it to distance function one of `{l2, mips, fast_l2, cosine}` ; we should just use `{item1, item2}` to indicate it's a set of values, and keep it consistent

include/program_options_utils.hpp
```diff
@@ -0,0 +1,71 @@
...
+    "distance function <l2/mips/fast_l2/cosine>.  'fast l2' and 'mips' only support data_type float";
+const char *INDEX_PATH_PREFIX_DESCRIPTION = "Path prefix to the index, e.ge '/mnt/data/my_ann_index'";
```
daxpryce, line 25: `Path prefix to the index, e.g.`

include/program_options_utils.hpp
```diff
@@ -0,0 +1,71 @@
...
+const char *RESULT_PATH_DESCRIPTION =
+    "Path prefix for saving results of the queries, e.g. '/mnt/data/query_file_X.bin'";
+const char *QUERY_FILE_DESCRIPTION = "Query file in binary format, e.g. '/mnt/data/query_file_X.bin'";
+const char *NUMBER_OF_RESULTS_DESCRIPTION = "Number of neighbors to be returned";
+const char *SEARCH_LIST_DESCRIPTION =
```
daxpryce, line 30: `Size of search list to use. This value is the number of neighbor/distance pairs to keep in memory at the same time during a query. ... (number of neighbors to return).  Corresponds to L in the white paper.`

include/program_options_utils.hpp
```diff
@@ -0,0 +1,71 @@
...
+const char *INPUT_DATA_PATH = "Input data file in bin format.  This is the file you want to build the index over.";
+
+// Optional parameters
+const char *FILTER_LABEL_DESCRIPTION =
+    "Filter to use when running a query.  'filter_label' and 'query_filters_file' are mutually exclusive.";
+const char *FILTERS_FILE_DESCRIPTION =
+    "Filter file for Queries for Filtered Search.  File format is text with one filter per line.  File must "
+    "have exactly one filter OR the same number of filters as there are queries in the 'query_file'.";
+const char *LABEL_TYPE_DESCRIPTION =
+    "Storage type of Labels <uint/ushort>, default value is uint which will consume memory 4 bytes per filter";
```
daxpryce, line 43: `{uint...32_t?, uint8_t}` maybe?  I think it's uint32_t at least

include/program_options_utils.hpp
```diff
@@ -0,0 +1,71 @@
...
+const char *GROUND_TRUTH_FILE_DESCRIPTION =
+    "ground truth file for the queryset"; // what's the format, what's the requirements? does it need to include an
+                                          // entry for every item or just a small subset? I have so many questions about
+                                          // this file
+const char *NUMBER_THREADS_DESCRIPTION = "Number of threads used for building index.  Defaults to number of logical "
+                                         "processor cores on your this machine returned by omp_get_num_procs()";
+const char *FAIl_IF_RECALL_BELOW = "If set to a value >0 and <100%, program returns -1 if best recall found is below "
```
daxpryce, line 50: FAIL not FAIl

daxpryce: Also, maybe the description should be "Value between 0 and 100, indicating the recall tolerance percentage threshold before program fails with a non-zero exit code. "

Also, if we return the results even if it results in a non-zero exit code, we should note that you'll get the results written out as well, and the exit code is intended to be treated as a signal that it did not meet your recall requirements for use (presuming that is what we do, that is).

include/program_options_utils.hpp
```diff
@@ -0,0 +1,71 @@
...
+                                   "this threshold. "; // does it continue running or die immediately?  Will I still get
+                                                       // my results even if the return code is -1?
+const char *NUMBER_OF_NODES_TO_CACHE = "Number of BFS nodes around medoid(s) to cache.  Default value: 0";
+const char *BEAMWIDTH = "Beamwidth for search. Set 0 to optimize internally.  Default value: 2";
+const char *MAX_BUILD_DEGREE = "Maximum graph degree";
+const char *GRAPH_BUILD_COMPLEXITY = "Build complexity, higher value results in better graphs";
+const char *GRAPH_BUILD_ALPHA = "alpha controls density and diameter of graph, set 1 for sparse graph, 1.2 or 1.4 for "
```
daxpryce, line 57: Capital A

include/program_options_utils.hpp
```diff
@@ -0,0 +1,71 @@
...
+                         "separated filters for each node with each line corresponding to a graph node";
+const char *UNIVERSAL_LABEL =
+    "Universal label, Use only in conjunction with label file for filtered index build. If a "
+    "graph node has all the labels against it, we can assign a special universal filter to the "
+    "point instead of comma separated filters for that point";
+const char *FILTERED_LBUILD = "Build complexity for filtered points, higher value results in better graphs";
+
+} // namespace program_options_utils
+
+#endif // DISKANN_PROGRAM_OPTIONS_UTILS_CPP
```
daxpryce, line 71: If you do #pragma once, you can remove this line too

daxpryce (approved):
Comments for improvements inline, but I trust your judgement in what comments to include!

include/program_options_utils.hpp
jonmclean: I verified that is what we do :).

include/program_options_utils.hpp
```diff
@@ -0,0 +1,71 @@
...
+const char *INDEX_PATH_PREFIX_DESCRIPTION = "Path prefix to the index, e.ge '/mnt/data/my_ann_index'";
+const char *RESULT_PATH_DESCRIPTION =
+    "Path prefix for saving results of the queries, e.g. '/mnt/data/query_file_X.bin'";
+const char *QUERY_FILE_DESCRIPTION = "Query file in binary format, e.g. '/mnt/data/query_file_X.bin'";
+const char *NUMBER_OF_RESULTS_DESCRIPTION = "Number of neighbors to be returned";
+const char *SEARCH_LIST_DESCRIPTION =
+    "List of L values of search.  L value is the number of neighbor/distance pairs to keep in-memory at the same time "
+    "during a query.  The higher the value the more accurate the results and the slower the search.  This value MUST "
+    "be at least K (number of neighbors to return).";
+const char *INPUT_DATA_PATH = "Input data file in bin format.  This is the file you want to build the index over.";
```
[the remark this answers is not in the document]
jonmclean: I agree that the file format should be documented but think it is too complex to fully document here with examples (as you said).  I have added a short explanation in the help for now.

PR #391, opened 2023-07-12
Fixed inputs type-o

jonmclean:
Fixes a type-o in the GitHub wheel action file.


daxpryce (approved):
legit not sure how this ever worked if the input is wrong.

PR #392, opened 2023-07-13
Bypass errors in logging for non-msft-prod environments

daxpryce:
# First and Foremost, this does not actually fix whatever bug is in the logging areas. 

We have had consistent segfaults in python, and we finally (thanks to @pbourke !) isolated it to something in logger.h, logger_impl.h, or logger.cpp. 

Ultimately though, those *not* in production environments will just want std::cout and std::cerr.

In theory - and what I'm asking for your specific attention on review - I have maintained all the symbols and expectations for our internal partners while simply aliasing std::cout and std::cerr to bypass the segfault issues we have.  Now, when our python module is unloaded, we no longer end up freeing memory python needs to continue running.  Segfaults are gone!

This is a super naive and fast approach to fixing this issue. I may very well be missing context or requirements, and if so, please tell me so I can address them.


include/abstract_data_store.h
```diff
@@ -18,6 +18,8 @@ template <typename data_t> class AbstractDataStore
   public:
     AbstractDataStore(const location_t capacity, const size_t dim);
 
+    virtual ~AbstractDataStore() = default;
```
pbourke, line 21: this was an unrelated issue that address sanitizer picked up. Index holds a pointer of type AbstractDataStore but the destructor in that class is not virtual, so the InMemDataStore destructor is not invoked when the data store is freed

pbourke (approved):

pbourke:
Here are the two common errors that I observed:

`AddressSanitizer: heap-buffer-overflow (/home/pbourke/projects/DiskANN/python/src/_diskannpy.cpython-310-x86_64-linux-gnu.so+0x17313a8) in diskann::ANNStreamBuf::overflow(int)`

```
ERROR: AddressSanitizer: negative-size-param: (size=-1)
    #0 0x7fa78f56bb78 in __interceptor_memcpy ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:790
    #1 0x7fa788d6fb27 in std::basic_streambuf<char, std::char_traits<char> >::xsputn(char const*, long) (/lib/x86_64-linux-gnu/libstdc++.so.6+0x141b27)
    #2 0x7fa788d61823 in std::basic_ostream<char, std::char_traits<char> >& std::__ostream_insert<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*, long) (/lib/x86_64-linux-gnu/libstdc++.so.6+0x133823)
    #3 0x7fa6db8b7d9a in diskann::Index<unsigned char, unsigned int, unsigned int>::search_with_tags(unsigned char const*, unsigned long, unsigned int, unsigned int*, float*, std::vector<unsigned char*, std::allocator<unsigned char*> >&) (/home/pbourke/projects/DiskANN/python/src/_diskannpy.cpython-310-x86_64-linux-gnu.so+0xe98d9a)
```

include/logger.h
```diff
@@ -14,8 +14,13 @@
 
 namespace diskann
 {
+#ifdef ENABLE_CUSTOM_LOGGER
 DISKANN_DLLEXPORT extern std::basic_ostream<char> cout;
 DISKANN_DLLEXPORT extern std::basic_ostream<char> cerr;
+#else
+using std::cerr;
```
yiylin, line 21: I find that it will keep using more memory in linux even though I use the std::cerr and std::cout directly in multiple-threads for a long time. Will this avoid the issues in your side?

include/logger.h
```diff
@@ -14,8 +14,13 @@
 
 namespace diskann
 {
+#ifdef ENABLE_CUSTOM_LOGGER
```
yiylin, line 17: because of the log issue, so I also do not use this in our side now.

yiylin (approved):

src/logger.cpp
```diff
@@ -10,21 +10,19 @@
 namespace diskann
 {
 
+#ifdef ENABLE_CUSTOM_LOGGER
 DISKANN_DLLEXPORT ANNStreamBuf coutBuff(stdout);
 DISKANN_DLLEXPORT ANNStreamBuf cerrBuff(stderr);
 
 DISKANN_DLLEXPORT std::basic_ostream<char> cout(&coutBuff);
 DISKANN_DLLEXPORT std::basic_ostream<char> cerr(&cerrBuff);
```
ltan1ms, line 18: The intention of ENABLE_CUSTOM_LOGGER is to allow prod env set a custom logger in `g_logger.`

The default behavior (i.e., without ENABLE_CUSTOM_LOGGER) is to use stdout and stderror logging which is what the line 14-line18 does?

So those global variables seems to be an issue?

src/logger.cpp
```diff
@@ -10,21 +10,19 @@
...
-
-#ifdef ENABLE_CUSTOM_LOGGER
```
ltan1ms, line 19: The intention of ENABLE_CUSTOM_LOGGER is to allow prod env set a custom logger in `g_logger.`

The default behavior (i.e., without ENABLE_CUSTOM_LOGGER) is to use stdout and stderror logging which is what the line 14-line18 does?

So these global variables seems to be an issue?

src/logger.cpp
```diff
@@ -10,21 +10,19 @@
 namespace diskann
 {
 
+#ifdef ENABLE_CUSTOM_LOGGER
```
ltan1ms, line 13: The intention of ENABLE_CUSTOM_LOGGER is to allow prod env set a custom logger in `g_logger.`

The default behavior (i.e., without ENABLE_CUSTOM_LOGGER) is to use stdout and stderror logging which is what the line 14-line18 does?

So these global variables seems to be an issue?

src/logger.cpp
ltan1ms, line 18: The intention of ENABLE_CUSTOM_LOGGER is to allow prod env set a custom logger in `g_logger.`

The default behavior (i.e., without ENABLE_CUSTOM_LOGGER) is to use stdout and stderror logging which is what the line 14-line18 does?

So those global variables seems to be an issue?

ltan1ms (commented):
I think with this change, meaning of  `ENABLE_CUSTOM_LOGGER ` is slightly changed, left a comment about it. 
Do we know why the default `cout` and `cerr` global objects can cause segfault?

include/logger.h
daxpryce: Our very basic test in `python/apps/in-mem-dynamic.py` is not very long lived at all.  The problem we encounter is a segfault when we're finished executing it - and it all stems around *something* in our custom logging impl.  

I had (perhaps wrongly!) assumed the standard library would have a _correct_ implementation of cout/cerr and I would be unequivocally safe in using it directly.  If you're saying you noticed memory pressure by using it directly, it would mean my assumption was wrong!  Perhaps a better fix (or one we can work towards in another PR) is re-enabling the behavior prior to this commit but tracking down the heap buffer overflow @pbourke found in his comment here: https://github.com/microsoft/DiskANN/pull/392#issuecomment-1634981778

Ultimately, I want to ensure that when `diskannpy` is removed from python, it doesn't muddle python's heap internals as well, causing a runtime detected error that takes python down with it.

src/logger.cpp
daxpryce: Correct, the global/statically initialized diskann::{cout,cerr} impls we've created are the progenitor of issues that we *see* in python but probably also exist everywhere. 

As I mentioned in a response to @yiylin in this PR, I had assumed we could just safely rely on std::{cout,cerr} instead, but it would mean that the custom logger would most likely continue having the problem. I was very hesitant to update any code that existed in prod, though - especially without knowing why it was the way it was.

daxpryce:
> I think with this change, meaning of `ENABLE_CUSTOM_LOGGER ` is slightly changed, left a comment about it. Do we know why the default `cout` and `cerr` global objects can cause segfault?

No - it wasn't until mid week last week that we were even able to isolate what was causing it. Valgrind is useless and I was left grasping at straws until @pbourke came to the rescue with address sanitizer to even identify the logger as the culprit.

PR #395, opened 2023-07-19
In Memory Graph Store

yashpatel007:
 

#### What does this implement/fix? 
This now implements Abstract Graph Store as In Memory Graph Store and uses it in index class as injectable.


src/in_mem_graph_store.cpp
```diff
@@ -7,25 +7,286 @@
...
+    size_t file_offset = 0;
+    out.seekp(file_offset, out.beg);
+    size_t index_size = 24;
+    uint32_t max_degree = 0;
+    out.write((char *)&index_size, sizeof(uint64_t));
+    out.write((char *)&_max_observed_degree, sizeof(uint32_t));
+    uint32_t ep_u32 = _start;
+    out.write((char *)&ep_u32, sizeof(uint32_t));
+    out.write((char *)&_num_frozen_pts, sizeof(size_t));
+    // Note: at this point, either _nd == _max_points or any frozen points have
```
harsha-simhadri, line 232: rename _nd  since its not a member var?

src/index.cpp
```diff
@@ -3255,13 +3123,13 @@ template <typename T, typename TagT, typename LabelT> void Index<T, TagT, LabelT
...
+        uint32_t k = (uint32_t)_graph_store->get_neighbours(i).size();
         std::memcpy(cur_node_offset, &k, sizeof(uint32_t));
-        std::memcpy(cur_node_offset + sizeof(uint32_t), _final_graph[i].data(), k * sizeof(uint32_t));
-        std::vector<uint32_t>().swap(_final_graph[i]);
+        std::memcpy(cur_node_offset + sizeof(uint32_t), _graph_store->get_neighbours(i).data(), k * sizeof(uint32_t));
+        std::vector<uint32_t>().swap(_graph_store->get_neighbours(i));
     }
-    _final_graph.clear();
-    _final_graph.shrink_to_fit();
+    _graph_store->clear_graph();
```
harsha-simhadri, line 3131: instead of two calls to clear_Graph and shrink_to_fit, can we just have one function?

harsha-simhadri (commented):
did you check for any perf regressions?

src/in_mem_graph_store.cpp
```diff
@@ -7,25 +7,286 @@
...
+size_t InMemGraphStore::resize_graph(const size_t new_size)
+{
+    _final_graph.resize(new_size);
+    set_total_points(new_size);
+    return _final_graph.size();
+}
+
+void InMemGraphStore::clear_graph()
+{
+    _final_graph.clear();
```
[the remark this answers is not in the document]
yashpatel007: line 3131 of index.cpp under optimize_index_layout() clears the graph.

src/index.cpp
yashpatel007: I agree. but shrink to fit seems fundamentally different. Right now we are using adjacency list as  our graph data_structure. But lets say in future we have a matrix(or something else) to sore a graph. We need to be able to support shrink to fit and the respective graph store classes will know how to compact the underlying structures.

include/in_mem_graph_store.h
```diff
@@ -11,13 +11,40 @@ namespace diskann
 class InMemGraphStore : public AbstractGraphStore
 {
   public:
-    InMemGraphStore(const size_t max_pts);
+    InMemGraphStore(const size_t max_pts, const size_t frozen_points);
```
rakri, line 14: same as above, hide frozen points from this class?

include/in_mem_graph_store.h
```diff
@@ -11,13 +11,40 @@ namespace diskann
...
+  protected:
+    virtual location_t load_impl(const std::string &filename, size_t expected_num_points);
+#ifdef EXEC_ENV_OLS
+    virtual location_t load_impl(AlignedFileReader &reader, size_t expected_num_points);
+#endif
+
+    int save_graph(const std::string &index_path_prefix, const size_t active_points);
+
+  private:
+    size_t _max_range_of_loaded_graph = 0;
```
rakri, line 43: whats the difference between max_range and max_observed_degree.

include/index.h
```diff
@@ -349,17 +351,14 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
     // externally and won't be returned by search. At least 1 frozen point is
     // needed for a dynamic index. The frozen points have consecutive locations.
     // See also _start below.
-    size_t _num_frozen_pts = 0;
```
rakri, line 352: we'd still need this, in the design i proposed.

include/index.h
```diff
@@ -349,17 +351,14 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
...
-    size_t _max_range_of_loaded_graph = 0;
     size_t _node_size;
     size_t _data_len;
     size_t _neighbor_len;
 
-    uint32_t _max_observed_degree = 0;
-    // Start point of the search. When _num_frozen_pts is greater than zero,
-    // this is the location of the first frozen point. Otherwise, this is a
-    // location of one of the points in index.
-    uint32_t _start = 0;
```
rakri, line 362: again, start could also be a property of index, like frozen points

include/index_config.h
```diff
@@ -186,6 +188,12 @@ class IndexConfigBuilder
                 throw ANNException("Error: please pass initial_search_list_size for building dynamic index.", -1);
         }
 
+        // sanity check
+        if (_dynamic_index && _num_frozen_pts == 0)
+        {
+            _num_frozen_pts = 1;
```
rakri, line 194: add a cout with warning?

include/index_factory.h
```diff
@@ -17,7 +17,8 @@ class IndexFactory
     std::unique_ptr<AbstractDataStore<T>> construct_datastore(DataStoreStrategy stratagy, size_t num_points,
                                                               size_t dimension);
 
-    std::unique_ptr<AbstractGraphStore> construct_graphstore(GraphStoreStrategy stratagy, size_t size);
+    std::unique_ptr<AbstractGraphStore> construct_graphstore(GraphStoreStrategy stratagy, size_t size,
+                                                             size_t frozen_points);
```
rakri, line 21: perhaps frozen is not needed in new design?

src/in_mem_graph_store.cpp
```diff
@@ -7,25 +7,276 @@
...
+    expected_file_size = *((size_t *)header.get());
+    _max_observed_degree = *((uint32_t *)(header.get() + sizeof(size_t)));
+    _start = *((uint32_t *)(header.get() + sizeof(size_t) + sizeof(uint32_t)));
+    file_frozen_pts = *((size_t *)(header.get() + sizeof(size_t) + sizeof(uint32_t) + sizeof(uint32_t)));
+
+    diskann::cout << "From graph header, expected_file_size: " << expected_file_size
+                  << ", _max_observed_degree: " << _max_observed_degree << ", _start: " << _start
+                  << ", file_frozen_pts: " << file_frozen_pts << std::endl;
+
+    if (file_frozen_pts != _num_frozen_pts)
```
rakri, line 64: since we wont save num frozen points on file, we can remove these sanity checks

src/index.cpp
```diff
@@ -99,11 +100,10 @@ Index<T, TagT, LabelT>::Index(Metric m, const size_t dim, const size_t max_point
         std::memset(_pq_data, 0, total_internal_points * _num_pq_chunks * sizeof(char));
     }
 
-    _start = (uint32_t)_max_points;
```
rakri, line 102: add these back, since _Start should be a member of index class.

rakri (changes requested):

include/in_mem_graph_store.h
yashpatel007: there is no comment on main to explain this but this is for sure different from max_observed_degree

src/index.cpp
yashpatel007: in save_graph we have 
```
uint32_t ep_u32 = _start;
    out.write((char *)&ep_u32, sizeof(uint32_t));
```

Do you think we can omit writing_start as well ?

yashpatel007: I believe the medtadata relting to fz_points and are _start are very crucial to search. So we need a way to pass in those values while saving during build. so that we can use them while load.

include/abstract_graph_store.h
```diff
@@ -19,13 +18,45 @@ class AbstractGraphStore
...
+    virtual int store(const std::string &index_path_prefix, const size_t active_points) = 0;
+
+    virtual std::vector<location_t> &get_neighbours(const location_t i) = 0;
+    virtual void set_neighbours(const location_t i, std::vector<location_t> &neighbors) = 0;
+
+    virtual size_t resize_graph(const size_t new_size) = 0;
+    virtual void clear_graph() = 0;
+
+    virtual size_t get_num_frozen_points() = 0;
+    virtual size_t get_max_range_of_loaded_graph() = 0;
```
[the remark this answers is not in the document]
yashpatel007: It is already captured by _max_observed_degree

apps/build_memory_index.cpp
```diff
@@ -120,32 +120,34 @@ int main(int argc, char **argv)
...
         diskann::get_bin_metadata(data_path, data_num, data_dim);
 
+        auto index_build_params = diskann::IndexWriteParametersBuilder(L, R)
+                                      .with_filter_list_size(Lf)
+                                      .with_alpha(alpha)
+                                      .with_saturate_graph(false)
+                                      .with_num_threads(num_threads)
+                                      .build();
+
+        auto build_params = diskann::IndexBuildParamsBuilder(index_build_params)
```
rakri, line 130: whats the difference between index_build_params and build_params? if it is the filtering addons, then shouldnt Lf go into build_params?

include/abstract_graph_store.h
```diff
@@ -14,18 +13,55 @@ namespace diskann
...
 {
   public:
-    AbstractGraphStore(const size_t max_pts) : _capacity(max_pts)
+    AbstractGraphStore(const size_t total_pts, const size_t reserve_graph_degree)
+        : _capacity(total_pts), _reserve_graph_degree(reserve_graph_degree)
+    {
+    }
+
+    // returns tuple of <nodes_read, start, num_frozen_points>
+    virtual std::tuple<uint32_t, uint32_t, size_t> load(const std::string &index_path_prefix,
```
rakri, line 22: does load need to know num_points?

src/in_mem_graph_store.cpp
```diff
@@ -6,26 +6,238 @@
...
+const std::vector<location_t> &InMemGraphStore::get_neighbours(const location_t i) const
+{
+    return _graph.at(i);
+}
+
+void InMemGraphStore::add_neighbour(const location_t i, location_t neighbour_id)
+{
+    _graph[i].emplace_back(neighbour_id);
+    if (_graph[i].size() > _max_observed_degree)
+    {
```
rakri, line 38: how does this check affect perf?

rakri: i guess add_neighbor is not called often? Also, neighbor vs neighbour make it consistent

src/in_mem_graph_store.cpp
```diff
@@ -6,26 +6,238 @@
...
+        _max_observed_degree = (uint32_t)_graph.size();
```
rakri, line 39: you mean graph[i].size()?

src/in_mem_graph_store.cpp
```diff
@@ -6,26 +6,238 @@
...
+                  << ", file_frozen_pts: " << file_frozen_pts << std::endl;
+
+    diskann::cout << "Loading vamana graph from reader..." << std::flush;
+
+    // If user provides more points than max_points
+    // resize the _graph to the larger size.
+    if (get_total_points() < expected_num_points)
+    {
+        diskann::cout << "resizing graph to " << expected_num_points << std::endl;
+        _graph.resize(expected_num_points);
```
rakri, line 97: shouldn't we call the resize function of this class instead of _graph?

src/in_mem_graph_store.cpp
```diff
@@ -6,26 +6,238 @@
...
+                  << ", file_frozen_pts: " << file_frozen_pts << std::endl;
+
+    diskann::cout << "Loading vamana graph " << filename << "..." << std::flush;
+
+    // If user provides more points than max_points
+    // resize the _graph to the larger size.
+    if (get_total_points() < expected_num_points)
+    {
+        diskann::cout << "resizing graph to " << expected_num_points << std::endl;
+        _graph.resize(expected_num_points);
```
rakri, line 160: same, shouldnt we call resize function of in_mem_Graph_store class?

apps/build_memory_index.cpp
yashpatel007: index_build_params aka (Indexwriteparams) are tunable params for building index like (L, R... etc) the BuildParams in turn  contains the IndexWritePrams and files path's like label_file, data_file, tag_file etc. BuildParams is to hold all those fields, as we have consolidated build method.

include/abstract_graph_store.h
yashpatel007: LoadGraph Method in Index class takes expected num points and params, and if that is larger then num points in graph we kind of resize the graph.

src/in_mem_graph_store.cpp
yashpatel007: Yes, I will add that.

src/in_mem_graph_store.cpp
yashpatel007: Yes. thanks for the catch Ravi.

src/in_mem_graph_store.cpp
yashpatel007: Yes. good catch. I have updated the code.

yashpatel007:
With current changes here are the results

Checked perf regression. Here are the results of main vs this branch. 1M Random dataset (128D), 10K queries
------------------------------ Graph Store Branch--------------------------------
==> Build
Index built with degree: max:64  avg:64  min:64  count(deg<2):0
Indexing time: 247.491
Not saving tags as they are not enabled.
Time taken for save: 1.33113s.

==> Search
Using 20 threads to search
  Ls         QPS     Avg dist cmps  Mean Latency (mus)   99.9 Latency   Recall@10
=================================================================================
  16    23248.08           1549.80              849.69       26110.50        4.25
  32    13255.24           2766.66             1506.36       15802.40        7.67
  
  
  --------------------------------------Main Branch ----------------------------------------------
  ==> Build
Index built with degree: max:64  avg:64  min:64  count(deg<2):0
Indexing time: 242.544
Not saving tags as they are not enabled.
Time taken for save: 1.38595s.

==>Search
Using 20 threads to search
  Ls         QPS     Avg dist cmps  Mean Latency (mus)   99.9 Latency   Recall@10
=================================================================================
  16    24171.22           1550.64              812.30       13272.70        4.35
  32    13872.36           2763.71             1439.43        4261.40        7.89

include/index.h
```diff
@@ -344,15 +345,13 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
...
     size_t _neighbor_len;
 
-    uint32_t _max_observed_degree = 0;
-    // Start point of the search. When _num_frozen_pts is greater than zero,
-    // this is the location of the first frozen point. Otherwise, this is a
-    // location of one of the points in index.
+    //  Start point of the search. When _num_frozen_pts is greater than zero,
+    //  this is the location of the first frozen point. Otherwise, this is a
+    //  location of one of the points in index.
     uint32_t _start = 0;
```
harsha-simhadri, line 355: should this be a property of graph store?

src/index.cpp
```diff
@@ -2479,9 +2306,9 @@ inline void Index<T, TagT, LabelT>::process_delete(const tsl::robin_set<uint32_t
         if (expanded_nodes_set.size() <= range)
         {
             std::unique_lock<non_recursive_mutex> adj_list_lock(_locks[loc]);
-            _final_graph[loc].clear();
+            _graph_store->clear_neighbours((location_t)loc);
             for (auto &ngh : expanded_nodes_set)
-                _final_graph[loc].push_back(ngh);
+                _graph_store->add_neighbour((location_t)loc, ngh);
```
rakri, line 2311: instead of adding one by one, can't we use set_nbrs method?

rakri: same thing happens in 1380 also.

src/index.cpp
```diff
@@ -2671,8 +2498,8 @@ template <typename T, typename TagT, typename LabelT> void Index<T, TagT, LabelT
         if ((new_location[old] < _max_points) // If point continues to exist
             || (old >= _max_points && old < _max_points + _num_frozen_pts))
         {
-            new_adj_list.reserve(_final_graph[old].size());
-            for (auto ngh_iter : _final_graph[old])
+            new_adj_list.reserve(_graph_store->get_neighbours((location_t)old).size());
```
rakri, line 2501: instead of get_nbrs().size, is it useful to have a num_nbrs method in graph store?

src/index.cpp
```diff
@@ -2804,10 +2631,18 @@ void Index<T, TagT, LabelT>::reposition_points(uint32_t old_location_start, uint
     // integer arithmetic rules.
     const uint32_t location_delta = new_location_start - old_location_start;
 
+    std::vector<location_t> updated_neighbours_location;
     for (uint32_t i = 0; i < _max_points + _num_frozen_pts; i++)
-        for (auto &loc : _final_graph[i])
+    {
+        auto &i_neighbours = _graph_store->get_neighbours((location_t)i);
```
rakri, line 2637: why is all this equivalent code not there in main?

src/index.cpp
```diff
@@ -3218,7 +3059,7 @@ template <typename T, typename TagT, typename LabelT> void Index<T, TagT, LabelT
     float *cur_vec = new float[_data_store->get_aligned_dim()];
     std::memset(cur_vec, 0, _data_store->get_aligned_dim() * sizeof(float));
     _data_len = (_data_store->get_aligned_dim() + 1) * sizeof(float);
-    _neighbor_len = (_max_observed_degree + 1) * sizeof(uint32_t);
+    _neighbor_len = (_graph_store->get_max_observed_degree() + 1) * sizeof(uint32_t);
```
rakri, line 3062: where is the value of get_max_observed_degree set?

rakri (approved):

src/index.cpp
yashpatel007: This is a good catch.  I have updated all instances where we can directly call set neighbours()

src/index.cpp
yashpatel007: I think this should be fine as well because we are returning a const reference so there is no way this is modified.

src/index.cpp
yashpatel007: In main we were directly updating the neighbour_id's. But now we use get_neighbours() and its a const reference. So we have to copy it then augment it then set it.

PR #396, opened 2023-07-20
Python build with a far more portable wheel

daxpryce:
Identified the appropriate build flags to get a working python build that doesn't rely on `-march=native` or `-mtune=native`, which are severely problematic unless the consuming user's processor has identical instruction sets available that match the github action runner that builds it.

We've run benchmarks on multiple computers that indicate the only important flag other than `-mavx2 -msse2 -mfma` is `-funroll-loops`.  Optimization levels such as `-O1`, `-O2`, or `-O3` actually makes for less performant code. `-Ofast` is unavailble for use in Python, as it causes problems with floating point math in Python; see https://moyix.blogspot.com/2022/09/someones-been-messing-with-my-subnormals.html for more details on -Ofast, -ffast-math, and the impact on subnormals.  


CMakeLists.txt
```diff
@@ -279,17 +279,18 @@ if(MSVC)
...
-	set(ENV{TCMALLOC_LARGE_ALLOC_REPORT_THRESHOLD} 500000000000)
-    #	set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -DDEBUG -O0 -fsanitize=address -fsanitize=leak -fsanitize=undefined")
-    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -DDEBUG -Wall -Wextra")
+    set(ENV{TCMALLOC_LARGE_ALLOC_REPORT_THRESHOLD} 500000000000)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mavx2 -mfma -msse2 -ftree-vectorize -fno-builtin-malloc -fno-builtin-calloc -fno-builtin-realloc -fno-builtin-free -fopenmp -fopenmp-simd -funroll-loops -Wfatal-errors -DUSE_AVX2")
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -DDEBUG")
     if (NOT PYBIND)
-	    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -Ofast -DNDEBUG -march=native -mtune=native -ftree-vectorize")
+        set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -DNDEBUG -Ofast")
+        if (NOT PORTABLE)
```
daxpryce, line 287: Adding a new `-DPORTABLE=True` flag in case we want to make builds that are relocatable to other machines, such as in our CICD workflow. The default is to continue building with `-march=native -mtune=native` though.

CMakeLists.txt
```diff
@@ -279,17 +279,18 @@ if(MSVC)
...
+            set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -march=native -mtune=native")
+        endif()
     else()
-        #-Ofast is super problematic for python.  see: https://moyix.blogspot.com/2022/09/someones-been-messing-with-my-subnormals.html
-        set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -DNDEBUG -march=native -mtune=native -ftree-vectorize")
-        add_compile_options(-fPIC)
+        # -Ofast is not supported in a python extension module
+        set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -DNDEBUG -fPIC")
     endif()
-	add_compile_options(-march=native -Wall -fno-builtin-malloc -fno-builtin-calloc -fno-builtin-realloc -fno-builtin-free -fopenmp -fopenmp-simd -funroll-loops -Wfatal-errors -DUSE_AVX2)
```
daxpryce, line 292: A note on add_compile_options vs. set(CMAKE_CXX_FLAGS[_RELEASE, _DEBUG], ...):

add_compile_options actually sets these flags for *every* compiler. C, Fortran, you name it, it gets the flags. Now, we don't actually use any compiler other than the CXX compiler.  However, having them in both places actually had no real effect; the same flags were being passed in multiple times, and we were keeping them in sync for no real good reason.

I made a decision to rip out the `add_compile_options` calls and instead am standardizing us on the `CMAKE_CXX_FLAGS` family of flag specifications just to stop repeating ourself AND provide better control over the Release/Debug profiles that we're going to act upon

pyproject.toml
```diff
@@ -3,22 +3,22 @@ requires = [
...
+    "numpy==1.25", # this is important to keep fixed. It also means anyone using something other than 1.22 won't be able to use this library
     "wheel",
     "ninja"
 ]
 build-backend = "setuptools.build_meta"
 
 [project]
 name = "diskannpy"
-version = "0.5.0.rc2"
+version = "0.5.0.rc4"
```
daxpryce, line 14: we tagged 0.5.0.rc3 in github even though our pyproject.toml wasn't updated.  manually keeping this stuff in sync is a huge dang pain.

bryantower (approved):

.github/workflows/build-python.yml
```diff
@@ -6,7 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        cibw-identifier: ["cp38-manylinux_x86_64", "cp39-manylinux_x86_64", "cp310-manylinux_x86_64", "cp311-manylinux_x86_64"]
```
jonmclean, line 9: Good-bye Python 3.8 support ....

daxpryce: yep.  numpy 1.25 stopped supporting python 3.8, so we're going to too - at least here.

once we get a conda recipe together we can set it up to build for python3.8 and all the different versions of numpy as well, but that's a task for another day

PR #397, opened 2023-07-20
Jomclean/write timings

jonmclean:
Added timers that write out to a configured JSON file so we can easily analyze multiple runs.

### DIFF
--- a/python/apps/in-mem-dynamic.py
+++ b/python/apps/in-mem-dynamic.py
@@ -35,6 +35,8 @@ def insert_and_search(
     """
     timer_results: dict[str, float] = {}
 
+    method_timer: utils.Timer = utils.Timer()
+
     npts, ndims = utils.get_bin_metadata(indexdata_file)
 
     if dtype_str == "float":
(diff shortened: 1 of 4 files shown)


python/apps/in-mem-dynamic.py
```diff
@@ -18,7 +18,25 @@ def insert_and_search(
     num_insert_threads,
     num_search_threads,
     gt_file,
-):
+) -> dict[str, float]:
```
daxpryce, line 21: annoyingly, this annotation may or may not be valid wrt the version of python you're using.  I think it became standard in 3.10 but in 3.9 it might still be `from typing import Dict`.  I actually can't remember the specific version but it would be an issue - but since we're the ones using this for testing purposes, we can probably just ignore that little bit entirely

daxpryce: nevermind, it was 3.9 that changed it to builtins.dict and deprecated the typing.Dict

daxpryce (approved):

PR #402, opened 2023-07-25
[FIX] index fail during save in python

deepakdhull80:


daxpryce:
@deepakdhull80 I don't think this is quite the right way to fix this.  I just made some changes in #404 that should address the failure to save that I think are the correct way to do this for now, without impacting our Cxx-first users (not everyone is using this via `diskannpy` yet, and adding it in here may break the CLI utilities or libs already built). 

Regardless, thank you so much for submitting a PR for this; we really appreciate your contribution!

I hope to have rc5 released today or early tomorrow which should include a working save on dynamic index.

PR #407, opened 2023-07-31
Preparing for 0.6.0 diskannpy release

daxpryce:

Changes in 0.6.0 include:
- updated documentation
- consistent naming in utility functions
- simplification of api (trying to coerce lists of longs or doubles into strict np arrays of our supported dtypes is too prone to silently swallowing errors in the downcast, and would be completely opaque to the user).  we support numpy arrays *only*.

Docs should be generated via `pdoc` in the style of `pdoc -o build/docs diskannpy` after building and installing diskannpy and pdoc. For the 0.6.0 release I will manually push this to a gh-pages branch and folder, and work on the automation to do this in github actions in a later PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,8 +25,18 @@ authors = [
     {name = "Dax Pryce", email = "daxpryce@microsoft.com"}
 ]
 
+[project.optional-dependencies]
+dev = ["black", "isort", "mypy"]
+
 [tool.setuptools]
 package-dir = {"" = "python/src"}
+
+[tool.isort]
+profile = "black"
+multi_line_output = 3
+
+[tool.mypy]
+plugins = "numpy.typing.mypy_plugin"
 
 [tool.cibuildwheel]
 manylinux-x86_64-image = "manylinux_2_28"
@@ -34,7 +44,6 @@ test-requires = ["scikit-learn~=1.2"]
 build-frontend = "build"
 skip = ["pp*", "*-win32", "*-manylinux_i686", "*-musllinux*"]
 test-command = "python -m unittest discover {project}/python/tests"
-
 
 [tool.cibuildwheel.linux]
 before-build = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "diskannpy"
-version = "0.5.0.rc5"
+version = "0.6.0"
 
 description = "DiskANN Python extension module"
 readme = "python/README.md"
--- a/python/src/_dynamic_memory_index.py
+++ b/python/src/_dynamic_memory_index.py
@@ -81,9 +81,12 @@ class DynamicMemoryIndex:
         - **index_directory**: The directory containing the index files. This directory must contain the following
             files:
             - `{index_prefix}.data`
-            - `vectors.bin`: `diskannpy` builder functions create or copy this file to the `index_directory`
             - `{index_prefix}.tags`
             - `{index_prefix}`
+
+          It may also include the following optional files:
+            - `{index_prefix}_vectors.bin`: Optional. `diskannpy` builder functions may create this file in the
+              `index_directory` if the index was created from a numpy array
             - `{index_prefix}_metadata.bin`: Optional. `diskannpy` builder functions create this file to store metadata
             about the index, such as vector dtype, distance metric, number of vectors and vector dimensionality.
             If an index is built from the `diskann` cli tools, this file will not exist.
@@ -244,12 +247,13 @@ class DynamicMemoryIndex:
         )
         _assert_is_nonnegative_uint32(search_threads, "search_threads")
 
-        if vector_dtype == np.single:
-            _index = _native_dap.DynamicMemoryFloatIndex
-        elif vector_dtype == np.ubyte:
+        if vector_dtype == np.uint8:
             _index = _native_dap.DynamicMemoryUInt8Index
-        else:
+        elif vector_dtype == np.int8:
             _index = _native_dap.DynamicMemoryInt8Index
+        else:
+            _index = _native_dap.DynamicMemoryFloatIndex
+
         self._index = _index(
             distance_metric=dap_metric,
             dimensions=dimensions,
@@ -283,8 +287,7 @@ class DynamicMemoryIndex:
         """
         _query = _castable_dtype_or_raise(
             query,
-            expected=self._vector_dtype,
-            message=f"StaticMemoryIndex expected a query vector of dtype of {self._vector_dtype}"
+            expected=self._vector_dtype
         )
         _assert(len(_query.shape) == 1, "query vector must be 1-d")
         _assert(
@@ -319,7 +322,7 @@ class DynamicMemoryIndex:
           increases accuracy at the cost of latency. Must be at least k_neighbors in size.
         - **num_threads**: Number of threads to use when searching this index. (>= 0), 0 = num_threads in system
         """
-        _queries = _castable_dtype_or_raise(queries, expected=self._vector_dtype, message=f"DynamicMemoryIndex expected a query vector of dtype of {self._vector_dtype}")
+        _queries = _castable_dtype_or_raise(queries, expected=self._vector_dtype)
         _assert_2d(_queries, "queries")
         _assert(
             _queries.shape[1] == self._dimensions,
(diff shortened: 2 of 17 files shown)


pyproject.toml
```diff
@@ -25,17 +25,26 @@ authors = [
...
+dev = ["black", "isort", "mypy"]
+
 [tool.setuptools]
 package-dir = {"" = "python/src"}
 
+[tool.isort]
+profile = "black"
+multi_line_output = 3
+
+[tool.mypy]
```
daxpryce, line 38: mypy still doesn't work but I'll fix that later

daxpryce:
I should also note I ran isort and black over this for import ordering and code formatting, respectively. They're not part of CICD yet, but I'll get to that later.

python/tests/fixtures/build_memory_index.py
```diff
@@ -12,19 +11,15 @@
 
 
 def build_random_vectors_and_memory_index(
-        dtype,
-        metric,
-        with_tags: bool = False,
-        index_prefix: str = "ann",
-        seed: int = 12345
+    dtype, metric, with_tags: bool = False, index_prefix: str = "ann", seed: int = 12345
```
jonmclean, line 14: Why move this back to one-line from multi-line?  Did the formatter do this?  Not a big deal, I just am curious about the style.

python/tests/fixtures/create_test_data.py
```diff
@@ -11,18 +11,18 @@
...
             low=0, high=256, size=(rows, dimensions), dtype=dtype
         )  # low is inclusive, high is exclusive
-    elif dtype == np.byte:
+    elif dtype == np.int8:
         vectors = rng.integers(
             low=-128, high=128, size=(rows, dimensions), dtype=dtype
         )  # low is inclusive, high is exclusive
     else:
-        raise RuntimeError("Only np.single, np.byte, and np.ubyte are supported")
+        raise RuntimeError("Only np.float32, np.int8, and np.uint8 are supported")
```
jonmclean, line 25: I have always been bothered by scoping in Python .... every other language I use would not allow you to define "vectors" inside an if block and return it outside the if block.  I just forget about that rule.

python/src/_files.py
```diff
@@ -1,26 +1,111 @@
...
+    Utility function that writes a DiskANN binary vector formatted file to the location of your choosing.
+
+    ### Parameters
+    - **vector_file**: The path to the vector file to write the vectors to.
+    - **vectors**: A 2d array of dtype `numpy.float32`, `numpy.uint8`, or `numpy.int8`
+    """
+    _assert_dtype(vectors.dtype)
+    _assert_2d(vectors, "vectors")
+    with open(vector_file, "wb") as fh:
+        _write_bin(vectors, fh)
```
jonmclean, line 56: I am excited about having this function available.  It will make ingestion scripts much more straight forward.

jonmclean (approved):

python/tests/fixtures/build_memory_index.py
daxpryce: yep, black did this for us.  I'm not keen on it personally, but that's the price you pay when you pick a code formatter to do it for you; it's going to make someone unhappy about something :(

python/src/_files.py
daxpryce: it actually existed before as "numpy_to_diskann_file" or something but aside from me, nobody seemed to know it existed or, if they did, what it was for.  once I pulled in the vectors_from_file it became super obvious what it should be named :laughing:

python/src/_dynamic_memory_index.py
```diff
@@ -27,16 +28,28 @@
...
 class DynamicMemoryIndex:
+    """
+    A DynamicMemoryIndex instance is used to both search and mutate a `diskannpy` memory index. This index is unlike
+    either `diskannpy.StaticMemoryIndex` or `diskannpy.StaticDiskIndex` in the following ways:
+
+    - It requires an explicit vector identifier for each vector added to it.
+    - Insert and (lazy) deletion operations are provided for a flexible, living index
+
+    The mutable aspect of this index will absolutely impact search time performance as new vectors are added and
+    old deleted. `DynamicMemoryIndex.consolidate_deletes()` should be called periodically to restructure the index
```
harsha-simhadri, line 49: can we auto trigger the consolidate when we are about to run out of capacity?

python/src/_dynamic_memory_index.py
```diff
@@ -59,17 +72,79 @@ def from_file(
...
+        - **initial_search_complexity**: Should be set to the most common `complexity` expected to be used during the
+          life of this `diskannpy.DynamicMemoryIndex` object. The working scratch memory allocated is based off of
+          `initial_search_complexity` * `search_threads`. Note that it may be resized if a `search` or `batch_search`
+          operation requests a space larger than can be accommodated by these values.
+        - **search_threads**: Should be set to the most common `num_threads` expected to be used during the
+          life of this `diskannpy.DynamicMemoryIndex` object. The working scratch memory allocated is based off of
+          `initial_search_complexity` * `search_threads`. Note that it may be resized if a `batch_search`
+          operation requests a space larger than can be accommodated by these values.
+        - **concurrent_consolidation**: If True, the consolidation process will be parallelized. This will result in
+          faster consolidation times, but will also result in higher memory usage during the consolidation process.
```
harsha-simhadri, line 121: The concurrency here is not about parallelism. hat is controlled by another thread count parameter. This flag dictates whether consolidation can be run alongside inserts and deletes, or whether the index is locked down to changes while consolidation is ongoing

python/src/_dynamic_memory_index.py
```diff
@@ -59,17 +72,79 @@ def from_file(
...
+        - **index_prefix**: The prefix of the index files. Defaults to "ann".
```
harsha-simhadri, line 122: can we possibly default ANN_R{R}_L{L}? OK with either

python/src/_dynamic_memory_index.py
```diff
@@ -108,55 +183,46 @@ def __init__(
...
+        - **vector_dtype**: One of {`np.float32`, `np.int8`, `np.uint8`}. The dtype of the vectors this index will
+          be storing.
+        - **dimensions**: The vector dimensionality of this index. All new vectors inserted must be the same
+          dimensionality.
+        - **max_vectors**: Capacity of the data store including space for future insertions
+        - **graph_degree**: Graph degree (a.k.a. `R`) is the maximum degree allowed for a node in the index's graph
+          structure. This degree will be pruned throughout the course of the index build, but it will never grow beyond
+          this value. Higher `graph_degree` values require longer index build times, but may result in an index showing
+          excellent recall and latency characteristics.
+        - **saturate_graph**: If True, the adjacency list of each node will be saturated with neighbors to have exactly
```
harsha-simhadri, line 206: most of these args would be set to default rather than requiring the user to interpret, right? Can we place the default value in the doc here?

python/src/_dynamic_memory_index.py
```diff
@@ -59,17 +72,79 @@ def from_file(
...
+        - **graph_degree**: Graph degree (a.k.a. `R`) is the maximum degree allowed for a node in the index's graph
+          structure. This degree will be pruned throughout the course of the index build, but it will never grow beyond
+          this value. Higher R values require longer index build times, but may result in an index showing excellent
+          recall and latency characteristics.
+        - **saturate_graph**: If True, the adjacency list of each node will be saturated with neighbors to have exactly
+          `graph_degree` neighbors. If False, each node will have between 1 and `graph_degree` neighbors.
+        - **max_occlusion_size**: The maximum number of points that can be considered by occlude_list function.
+        - **alpha**: The alpha parameter (>=1) is used to control the nature and number of points that are added to the
+          graph. A higher alpha value (e.g., 1.4) will result in fewer hops (and IOs) to convergence, but probably
+          more distance comparisons.
```
harsha-simhadri, line 107: ... more distance comparisons compared to alpha=1 or 1.2

python/src/_dynamic_memory_index.py
daxpryce: They're all being sourced from a constants file so that we specifically kept them in sync across the builders and indices. Referencing their default value in the comment is just going to risk an incongruity between our documented values and our actual values being used.  I don't think it's worth it when the html docs render like 

image

And if you're reading the docs via the `help()` function or jump-to-symbol in an IDE, you're just as capable of jumping to the consts file too. It only really helps during PR, not when most people will be reading the documentation as part of the whole file.  

If you still think we need it, I can add it, but I think our chances of making it less confusing are balanced by the risk of even more confusion if we ever update a default out of sync with the C++ constants.

python/src/_dynamic_memory_index.py
daxpryce: Ugh. This actually makes for even more problems; if we do this when we have R and L, we need to force the user to know what it would be interpolated to on load.  I'd much rather keep it fixed as `ann` and the user can create their own convention, just so we don't make it easy to get out of sync.

The static search classes don't have any idea what R or L is used, since it's immaterial to searching.  Dynamic knows, but it just makes for a more nuanced api that is hard to keep track of

PR #411, opened 2023-08-02
[on-going] skip dup vector in graph pruning.

jinwei14:


src/index.cpp
```diff
@@ -1192,9 +1192,19 @@ void Index<T, TagT, LabelT>::search_for_point_and_prune(int location, uint32_t L
     }
 
     prune_neighbors(location, pool, pruned_list, scratch);
-
-    assert(!pruned_list.empty());
-    assert(_final_graph.size() == _max_points + _num_frozen_pts);
+    // skip full dupl vectors
+    NeighborPriorityQueue &best_L_nodes = scratch->best_l_nodes();
+    _full_duplicated =
+        best_L_nodes.size() > 0 && best_L_nodes.operator[](0).distance == 0 && best_L_nodes.operator[](0).expanded;
```
PhilipBAdams, line 1198: If I understand the pruning right, it would be better to prune to the max number of neighbors by some stable way like vector ID, otherwise you can have more than the legal number of neighbors

src/index.cpp
```diff
@@ -1192,9 +1192,19 @@ void Index<T, TagT, LabelT>::search_for_point_and_prune(int location, uint32_t L
     }
 
     prune_neighbors(location, pool, pruned_list, scratch);
-
-    assert(!pruned_list.empty());
-    assert(_final_graph.size() == _max_points + _num_frozen_pts);
+    // skip full dupl vectors
+    NeighborPriorityQueue &best_L_nodes = scratch->best_l_nodes();
+    _full_duplicated =
```
harsha-simhadri, line 1197: _full_duplicated seems to be a common shared variable amongst threads. 
Here you are modifying it concurrently in each thread. would that not lead to race?

src/index.cpp
```diff
@@ -1192,9 +1192,19 @@ void Index<T, TagT, LabelT>::search_for_point_and_prune(int location, uint32_t L
...
+        best_L_nodes.size() > 0 && best_L_nodes.operator[](0).distance == 0 && best_L_nodes.operator[](0).expanded;
+    if (_full_duplicated)
+    {
+        diskann::cout << "Skipping validating full dupl vector " << location << std::endl;
```
harsha-simhadri, line 1201: I don't follow what behavior is being modified due to this flag?

harsha-simhadri:
@jinwei14 I am not sure I follow the behavior you want to change here and the way you are doing it.

Also, could you please detail your intention on the PR template?

jinwei14:
> @jinwei14 I am not sure I follow the behavior you want to change here and the way you are doing it.
> 
> Also, could you please detail your intention on the PR template?

I am still investigating the root cause, basically we can not build duplicated vectors in debug mode, in release mode it will lead to search return inconsistant results

PR #412, opened 2023-08-02
Added PDoc workflow to publish github pages documentation

jonmclean:
 
Make GitHub publish pdoc documentation to gh-pages as part of the release.  Create documentation as a ZIP as part of each build.

### DIFF
--- a/.github/workflows/build-python-pdoc.yml
+++ b/.github/workflows/build-python-pdoc.yml
@@ -4,7 +4,10 @@ jobs:
   build-reference-documentation:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
       - name: Set up Python 3.9
         uses: actions/setup-python@v2
         with:
--- a/.github/workflows/build-python-pdoc.yml
+++ b/.github/workflows/build-python-pdoc.yml
@@ -17,12 +17,12 @@ jobs:
           pdoc -o docs/python/html diskannpy
       - name: Create version environment variable
         run: |
-          echo "DISKANN_VERSION=`$(python <<EOF
-            from importlib.metadata import version
-            v = version('diskannpy')
-            print(v)
-            EOF
-            )`" >> $GITHUB_ENV
+          echo "DISKANN_VERSION=$(python <<EOF
+          from importlib.metadata import version
+          v = version('diskannpy')
+          print(v)
+          EOF
+          )" >> $GITHUB_ENV
       - name: Archive documentation version artifact
         uses: actions/upload-artifact@v2
         with:
--- a/.github/workflows/build-python-pdoc.yml
+++ b/.github/workflows/build-python-pdoc.yml
@@ -12,12 +12,16 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: 3.9
-      - name: Build wheel for documentation
-        uses: ./.github/actions/build
+      - name: Install python build
+        run: python -m pip install build
+        shell: bash
+      - name: Building Python Wheel for documentation generation
+        run: python -m build --wheel --outdir documentation_dist
+        shell: bash
       - name: "Run Reference Documentation Generation"
         run: |
           pip install pdoc pipdeptree
-          pip install dist/*.whl 
+          pip install documentation_dist/*.whl 
           echo "documentation" > dependencies_documentation.txt
           pipdeptree >> dependencies_documentation.txt
           pdoc -o docs/python/html diskannpy
--- a/.github/workflows/build-python-pdoc.yml
+++ b/.github/workflows/build-python-pdoc.yml
@@ -20,7 +20,7 @@ jobs:
           echo "DISKANN_VERSION=`$(python <<EOF
             from importlib.metadata import version
             v = version('diskannpy')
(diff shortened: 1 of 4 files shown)


jonmclean:
NOTE:  To test the publish I disabled the "main branch" check and added "-test" to the version number.  Once this is fully setup we can delete the 0.6.0-test from gh-pages.

.github/workflows/push-test.yml
```diff
@@ -30,6 +30,11 @@ jobs:
           submodules: true
       - name: DiskANN Build CLI Applications
         uses: ./.github/actions/build
+  build-documentation:
+      strategy:
+        fail-fast: true
+      name: DiskANN Build Documentation
+      uses: ./.github/workflows/build-python-pdoc.yml
```
jonmclean, line 37: Run the documentation generation on every build so we can download the docs and review them if needed.

.github/workflows/build-python-pdoc.yml
```diff
@@ -0,0 +1,51 @@
...
+on: [workflow_call]
+jobs:
+  build-reference-documentation:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
```
jonmclean, line 11: I am not sure why we need to setup python but the file I used as a template did this.

.github/workflows/build-python-pdoc.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: "Run Reference Documentation Generation"
+        run: |
+          pip install diskannpy pdoc pipdeptree
+          echo "documentation" > dependencies_documentation.txt
+          pipdeptree >> dependencies_documentation.txt
```
jonmclean, line 16: Write out the dependencies (including the deps of pdoc) so we can see what deps were used in the build.

.github/workflows/build-python-pdoc.yml
```diff
@@ -0,0 +1,51 @@
...
+          pdoc -o docs/python/html diskannpy
+      - name: Create version environment variable
+        run: |
+          echo "DISKANN_VERSION=$(python <<EOF
+          from importlib.metadata import version
+          v = version('diskannpy')
+          print(v)
+          EOF
+          )" >> $GITHUB_ENV
```
jonmclean, line 25: For some reason "python setup.py --version" did not work on Linux.  It did work on Windows though.  This code will run in Bash.

.github/workflows/build-python-pdoc.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Archive documentation version artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: dependencies
+          path: |
+            dependencies_documentation.txt
```
jonmclean, line 31: Publish the build-time dependencies used by the Linux worker.

.github/workflows/build-python-pdoc.yml
daxpryce: setup python just means "make sure the right version of python is on the path so you can say `python` and get the one you expect".  I'd use 3.11

.github/workflows/build-python-pdoc.yml
daxpryce: my concern about `pip install diskannpy` is that it might not actually be showing on pypi.org yet (I don't know for sure, at least).  might be better to download the artifacts built in https://github.com/microsoft/DiskANN/blob/main/.github/workflows/python-release.yml, install the one meant for the python version you specified in the setup-python action, then get it from there - then we won't have to worry about publishing 0.6.0 docs for a 0.6.1 release

.github/workflows/build-python-pdoc.yml
daxpryce: In the python-release workflow, it's only triggered on a new release being published.  I bet we can get the version/tag from that, which would have the version number.

Alternatively, since we'll probably have the current release installed in the job due installing it from the wheel, we can also probably do the following: `python -c "import diskannpy; print(diskannpy._diskannpy.__version__);"` and capture it there

.github/workflows/build-python-pdoc.yml
jonmclean: Added code to build a wheel specifically for the documentation step.

.github/workflows/build-python-pdoc.yml
```diff
@@ -0,0 +1,78 @@
+name: DiskANN Build PDoc Documentation
+on: [workflow_call]
```
daxpryce, line 2: I think we want the trigger for this to be on workflow call and on push to main, and then remove it from the push test entirely.

Otherwise competing branch pushes could maybe end up doing undesired things if they both try to publish the docs at the same time (even ignoring the race conditions, now people pushing updates need to hope nobody else who *isn't* pushing updates to their branches aren't pushing, else they won't see their changes reflected.

So we should have something like:

```yaml
on:
  workflow_call:
  push:
    branches:
      - 'main'
```
And then remove it from the `push-test-yml` entirely.

daxpryce: Nevermind all of this, I see now you have the branch checks to keep it from publishing, instead people just download the documentation artifact to validate.  That works better.

daxpryce (approved):
Looks good to me!

PR #418, opened 2023-08-10
Consolidate Index Constructors

yashpatel007:
 
Exposing two constructors for index, one for advance Index class level API and one for Abstract Index Simplified API.

### DIFF
--- a/apps/build_memory_index.cpp
+++ b/apps/build_memory_index.cpp
@@ -40,8 +40,9 @@ int build_in_memory_index(const diskann::Metric &metric, const std::string &data
     size_t data_num, data_dim;
     diskann::get_bin_metadata(data_path, data_num, data_dim);
 
-    diskann::Index<T, TagT, LabelT> index(metric, data_dim, data_num, nullptr, 0, 0, false, false, false, use_pq_build,
-                                          num_pq_bytes, use_opq);
+    diskann::Index<T, TagT, LabelT> index(metric, data_dim, data_num,
+                                          std::make_shared<diskann::IndexWriteParameters>(paras), nullptr, 0, false,
+                                          false, false, use_pq_build, num_pq_bytes, use_opq);
     auto s = std::chrono::high_resolution_clock::now();
     if (label_file == "")
     {
--- a/apps/build_memory_index.cpp
+++ b/apps/build_memory_index.cpp
@@ -40,8 +40,8 @@ int build_in_memory_index(const diskann::Metric &metric, const std::string &data
     size_t data_num, data_dim;
     diskann::get_bin_metadata(data_path, data_num, data_dim);
 
-    diskann::Index<T, TagT, LabelT> index(metric, data_dim, data_num, false, false, false, use_pq_build, num_pq_bytes,
-                                          use_opq);
+    diskann::Index<T, TagT, LabelT> index(metric, data_dim, data_num, nullptr, 0, 0, false, false, false, use_pq_build,
+                                          num_pq_bytes, use_opq);
     auto s = std::chrono::high_resolution_clock::now();
     if (label_file == "")
     {
--- a/apps/build_memory_index.cpp
+++ b/apps/build_memory_index.cpp
@@ -22,51 +22,6 @@
 
 namespace po = boost::program_options;
 
-template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t>
-int build_in_memory_index(const diskann::Metric &metric, const std::string &data_path, const uint32_t R,
-                          const uint32_t L, const float alpha, const std::string &save_path, const uint32_t num_threads,
-                          const bool use_pq_build, const size_t num_pq_bytes, const bool use_opq,
-                          const std::string &label_file, const std::string &universal_label, const uint32_t Lf)
-{
-    diskann::IndexWriteParameters paras = diskann::IndexWriteParametersBuilder(L, R)
-                                              .with_filter_list_size(Lf)
-                                              .with_alpha(alpha)
-                                              .with_saturate_graph(false)
-                                              .with_num_threads(num_threads)
-                                              .build();
-    std::string labels_file_to_use = save_path + "_label_formatted.txt";
-    std::string mem_labels_int_map_file = save_path + "_labels_map.txt";
-
-    size_t data_num, data_dim;
-    diskann::get_bin_metadata(data_path, data_num, data_dim);
-
-    diskann::Index<T, TagT, LabelT> index(metric, data_dim, data_num,
-                                          std::make_shared<diskann::IndexWriteParameters>(paras), nullptr, 0, false,
-                                          false, false, use_pq_build, num_pq_bytes, use_opq);
-    auto s = std::chrono::high_resolution_clock::now();
-    if (label_file == "")
-    {
-        index.build(data_path.c_str(), data_num, paras);
-    }
-    else
-    {
-        convert_labels_string_to_int(label_file, labels_file_to_use, mem_labels_int_map_file, universal_label);
-        if (universal_label != "")
-        {
-            LabelT unv_label_as_num = 0;
-            index.set_universal_label(unv_label_as_num);
-        }
-        index.build_filtered_index(data_path.c_str(), labels_file_to_use, data_num, paras);
-    }
-    std::chrono::duration<double> diff = std::chrono::high_resolution_clock::now() - s;
-
-    std::cout << "Indexing time: " << diff.count() << "\n";
-    index.save(save_path.c_str());
-    if (label_file != "")
-        std::remove(labels_file_to_use.c_str());
-    return 0;
-}
-
 int main(int argc, char **argv)
 {
     std::string data_type, dist_fn, data_path, index_path_prefix, label_file, universal_label, label_type;
--- a/apps/build_stitched_index.cpp
+++ b/apps/build_stitched_index.cpp
@@ -285,7 +285,7 @@ void prune_and_save(path final_index_path_prefix, path full_index_path_prefix, p
     auto pruning_index_timer = std::chrono::high_resolution_clock::now();
 
     diskann::get_bin_metadata(input_data_path, number_of_label_points, dimension);
-    diskann::Index<T> index(diskann::Metric::L2, dimension, number_of_label_points, false, false);
+    diskann::Index<T> index(diskann::Metric::L2, dimension, number_of_label_points, nullptr, 0, 0, false, false);
 
     // not searching this index, set search_l to 0
     index.load(full_index_path_prefix.c_str(), num_threads, 1);
--- a/apps/build_stitched_index.cpp
+++ b/apps/build_stitched_index.cpp
@@ -285,7 +285,7 @@ void prune_and_save(path final_index_path_prefix, path full_index_path_prefix, p
     auto pruning_index_timer = std::chrono::high_resolution_clock::now();
 
     diskann::get_bin_metadata(input_data_path, number_of_label_points, dimension);
-    diskann::Index<T> index(diskann::Metric::L2, dimension, number_of_label_points, nullptr, 0, 0, false, false);
+    diskann::Index<T> index(diskann::Metric::L2, dimension, number_of_label_points, nullptr, nullptr, 0, false, false);
 
     // not searching this index, set search_l to 0
     index.load(full_index_path_prefix.c_str(), num_threads, 1);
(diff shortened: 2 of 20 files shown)


yashpatel007:
Index class has two constructors, providing access to Index class via AbstractIndex API(simplified) & Index API(full functionality) for internal use or advance users.

include/index.h
```diff
@@ -49,20 +49,14 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
...
   public:
-    // Constructor for Bulk operations and for creating the index object solely
-    // for loading a prexisting index.
-    DISKANN_DLLEXPORT Index(Metric m, const size_t dim, const size_t max_points = 1, const bool dynamic_index = false,
+    // For internal use - uses new constructor internally
+    DISKANN_DLLEXPORT Index(Metric m, const size_t dim, const size_t max_points,
+                            const std::shared_ptr<IndexWriteParameters> &indexParameters,
+                            const std::shared_ptr<IndexSearchParams> &indexSearchParams,
+                            const size_t num_frozen_pts = 0, const bool dynamic_index = false,
                             const bool enable_tags = false, const bool concurrent_consolidate = false,
```
harsha-simhadri, line 57: should we consolidate some of these params in write and search params?

yashpatel007: Its already done as you can see with Write Params (used when writing / building index) and search params.

daxpryce (approved):
Looks good to me

python/src/static_memory_index.cpp
```diff
@@ -17,15 +17,17 @@ diskann::Index<DT, StaticIdType, filterT> static_index_builder(const diskann::Me
     {
         throw std::runtime_error("initial_search_complexity must be a positive uint32_t");
     }
-
+    auto index_search_params = diskann::IndexSearchParams(initial_search_complexity, omp_get_num_threads());
```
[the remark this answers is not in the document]
yashpatel007: Sounds good. For our testing purposes we use num_threads in buils and use the same for search_threads. Its good if user have option to pass if they want them to be different.

I am marking this as resolved as its not a blocker :) Thanks Dax

apps/build_memory_index.cpp
```diff
@@ -40,8 +40,9 @@ int build_in_memory_index(const diskann::Metric &metric, const std::string &data
     size_t data_num, data_dim;
     diskann::get_bin_metadata(data_path, data_num, data_dim);
 
-    diskann::Index<T, TagT, LabelT> index(metric, data_dim, data_num, false, false, false, use_pq_build, num_pq_bytes,
-                                          use_opq);
+    diskann::Index<T, TagT, LabelT> index(metric, data_dim, data_num,
+                                          std::make_shared<diskann::IndexWriteParameters>(paras), nullptr, 0, false,
```
gopalrs, line 44: Why not simply use an object instead of a shared_ptr for IndexWriteParameters?

apps/build_memory_index.cpp
```diff
@@ -40,8 +40,9 @@ int build_in_memory_index(const diskann::Metric &metric, const std::string &data
...
+                                          false, false, use_pq_build, num_pq_bytes, use_opq);
```
gopalrs, line 45: Should this driver start using the IndexConfig constructor?

apps/build_stitched_index.cpp
```diff
@@ -285,7 +285,7 @@ void prune_and_save(path final_index_path_prefix, path full_index_path_prefix, p
     auto pruning_index_timer = std::chrono::high_resolution_clock::now();
 
     diskann::get_bin_metadata(input_data_path, number_of_label_points, dimension);
-    diskann::Index<T> index(diskann::Metric::L2, dimension, number_of_label_points, false, false);
+    diskann::Index<T> index(diskann::Metric::L2, dimension, number_of_label_points, nullptr, nullptr, 0, false, false);
```
gopalrs, line 288: As above, should we change this to start using the new constructor?

apps/test_streaming_scenario.cpp
```diff
@@ -186,6 +186,7 @@ void build_incremental_index(const std::string &data_path, const uint32_t L, con
                                                .with_num_frozen_points(num_start_pts)
                                                .build();
 
+    auto index_search_params = diskann::IndexSearchParams(L, insert_threads);
```
gopalrs, line 189: Why does search_params have an "insert_threads" parameter?

apps/build_stitched_index.cpp
gopalrs, line 288: Unrelated to this PR, but should we change this to accept any distance metric?

include/index.h
```diff
@@ -49,21 +49,15 @@ template <typename T, typename TagT = uint32_t, typename LabelT = uint32_t> clas
...
-                            const bool use_opq = false, const size_t num_frozen_pts = 0,
-                            const bool init_data_store = true);
-
-    // Constructor for incremental index
-    DISKANN_DLLEXPORT Index(Metric m, const size_t dim, const size_t max_points, const bool dynamic_index,
-                            const IndexWriteParameters &indexParameters, const uint32_t initial_search_list_size,
-                            const uint32_t search_threads, const bool enable_tags = false,
+    // Call this when creating and passing Index Config is inconvenient.
+    DISKANN_DLLEXPORT Index(Metric m, const size_t dim, const size_t max_points,
+                            const std::shared_ptr<IndexWriteParameters> indexParameters,
```
gopalrs, line 54: Can we stick to the snake case convention for variable names?

include/index_config.h
```diff
@@ -160,15 +157,31 @@ class IndexConfigBuilder
...
+            return *this;
+        }
+        this->_index_write_params = index_write_params_ptr;
+        return *this;
+    }
+
+    IndexConfigBuilder &with_index_search_params(IndexSearchParams &search_params)
     {
-        this->_search_threads = search_threads;
+        this->_index_search_params = std::make_shared<IndexSearchParams>(search_params);
```
gopalrs, line 173: Doesn't make_shared<> require a pointer?

include/index_config.h
```diff
@@ -160,15 +157,31 @@ class IndexConfigBuilder
...
         return *this;
     }
 
-    IndexConfigBuilder &with_initial_search_list_size(uint32_t search_list_size)
+    IndexConfigBuilder &with_index_search_params(std::shared_ptr<IndexSearchParams> search_params_ptr)
```
gopalrs, line 177: Do we need both the "with_index_search_params()" overloads?

include/index_config.h
```diff
@@ -177,19 +190,20 @@ class IndexConfigBuilder
...
         {
-            if (_search_threads == 0)
-                throw ANNException("Error: please pass search_threads for building dynamic index.", -1);
+            _num_frozen_pts = 1;
+        }
 
-            if (_initial_search_list_size == 0)
+        if (_dynamic_index && _index_search_params != nullptr)
+        {
+            if (_index_search_params->initial_search_list_size == 0)
```
gopalrs, line 200: Shouldn't this check be present for both _dynamic and _static indices?

include/parameters.h
```diff
@@ -38,6 +38,17 @@ class IndexWriteParameters
...
 };
 
+class IndexSearchParams
+{
+  public:
+    IndexSearchParams(const uint32_t initial_search_list_size, const uint32_t num_search_threads)
+        : initial_search_list_size(initial_search_list_size), num_search_threads(num_search_threads)
+    {
+    }
+    const uint32_t initial_search_list_size; // search L
```
gopalrs, line 48: Don't we support multiple values for search_list_size?

src/filter_utils.cpp
```diff
@@ -45,7 +45,8 @@ void generate_label_indices(path input_data_path, path final_index_path_prefix,
 
         size_t number_of_label_points, dimension;
         diskann::get_bin_metadata(curr_label_input_data_path, number_of_label_points, dimension);
-        diskann::Index<T> index(diskann::Metric::L2, dimension, number_of_label_points, false, false);
+        diskann::Index<T> index(diskann::Metric::L2, dimension, number_of_label_points, nullptr, nullptr, 0, false,
```
gopalrs, line 48: Should we assume L2?

src/index.cpp
```diff
@@ -2164,7 +2138,8 @@ std::pair<uint32_t, uint32_t> Index<T, TagT, LabelT>::search(const T *query, con
 
     std::shared_lock<std::shared_timed_mutex> lock(_update_lock);
 
-    _distance->preprocess_query(query, _data_store->get_dims(), scratch->aligned_query());
+    _data_store->get_dist_fn()->preprocess_query(query, _data_store->get_dims(), scratch->aligned_query());
```
gopalrs, line 2141: I think the _data_store() has a preprocess_query() so you can invoke it directly.

src/index.cpp
```diff
@@ -2266,7 +2241,7 @@ std::pair<uint32_t, uint32_t> Index<T, TagT, LabelT>::search_with_filters(const
     // REFACTOR
     // T *aligned_query = scratch->aligned_query();
     // memcpy(aligned_query, query, _dim * sizeof(T));
-    _distance->preprocess_query(query, _data_store->get_dims(), scratch->aligned_query());
+    _data_store->get_dist_fn()->preprocess_query(query, _data_store->get_dims(), scratch->aligned_query());
```
gopalrs, line 2244: I think the _data_store() has a preprocess_query() so you can invoke it directly.

src/index.cpp
```diff
@@ -2345,7 +2320,8 @@ size_t Index<T, TagT, LabelT>::search_with_tags(const T *query, const uint64_t K
     const std::vector<uint32_t> init_ids = get_init_ids();
     const std::vector<LabelT> unused_filter_label;
 
-    _distance->preprocess_query(query, _data_store->get_dims(), scratch->aligned_query());
+    //_distance->preprocess_query(query, _data_store->get_dims(), scratch->aligned_query());
+    _data_store->get_dist_fn()->preprocess_query(query, _data_store->get_dims(), scratch->aligned_query());
```
gopalrs, line 2324: I think the _data_store() has a preprocess_query() so you can invoke it directly.

gopalrs (changes requested):

src/index.cpp
yashpatel007: Sure, I will make the change.

src/index.cpp
yashpatel007: same comment as below. It is not exposed currently through data_store. Keeping this open in case we want to move it.

src/filter_utils.cpp
yashpatel007: It was there to begin with as you can see. The only change in this PR is calling new ctor with updated args.

include/parameters.h
yashpatel007: That is when we call search, This is when we initialize the index for dynamic index.

include/index.h
yashpatel007: sure, let  me change that.

include/index_config.h
yashpatel007: I see, i will add that.

include/index_config.h
yashpatel007: Yes and no. its a matter of convenience. Both of them are being used as required.

include/index_config.h
yashpatel007: It is a shared pointer, stored inside index_config

apps/build_stitched_index.cpp
yashpatel007: So, the distance metric can be different then index metric ??

apps/test_streaming_scenario.cpp
yashpatel007: Its just reusing the variable. It was like this before so i kept it the same, however if someone wants to pass a different value. they can.

apps/build_stitched_index.cpp
yashpatel007: It is indeed using the new constructor. Its a matter of convenience, we have 2 constructors, one that takes config and the other takes args directly. If we use the constructor with config, we have to write too many lines of code to achieve this. not idea i think.

apps/build_memory_index.cpp
yashpatel007: We need to check if it is nullable.  As this will be absent when we initialize the constructor for search only.

apps/build_stitched_index.cpp
harsha-simhadri: do you mean we should pass the wrapping method's metric instead of assuming L2?

PR #421, opened 2023-08-11
Add Performance Tests

daxpryce:
This PR has a custom docker container that can be used both by local devs trying to replicate performance timings as well as a scheduled task that runs daily and captures the times in json format for performance regression analysis.

Note: There's ample room for future improvements to this. 
- Ideally our tool would monitor it's own times and export operation times to a file in a json format, without having to rely on gnu time.
- We don't pin specific system library versions anywhere, so replication of specific libs is impossible. We could publish these to a docker registry which would solve that problem, but I don't know where we can freely publish these things
- We can (and should) write something that parses the json returned and emits them as events to application insights or something so we can monitor behaviors without having to inspect the artifacts manually (a jupyter notebook was fine for now, but it's not appropriate over the long term)

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -1,0 +1,26 @@
+name: DiskANN Nightly Performance Metrics
+on:
+  schedule:
+    - cron: "41 14 * * *"  # 14:41 UTC, 7:41 PDT, 8:41 PST, 08:11 IST
+jobs:
+  perf-test:
+    name: Run Perf Test from main
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+      - name: Build Perf Container
+        run: |
+          docker build --build-arg GIT_COMMIT_ISH="$GITHUB_SHA" -t perf -f scripts/perf/Dockerfile scripts
+      - name: Performance Tests
+        run: |
+          mkdir metrics
+          docker run -v ./metrics:/app/logs perf &> ./metrics/combined_stdouterr.log
+      - name: Upload Metrics Logs
+        uses: actions/upload-artifact@v3
+        with:
+          name: metrics
+          path: |
+            ./metrics/**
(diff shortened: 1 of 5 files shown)


scripts/perf/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN DEBIAN_FRONTEND=noninteractive apt install -y git time
+
+COPY dev/install-dev-deps-ubuntu.bash /app/fallback/install-dev-deps-ubuntu.bash
+WORKDIR /app
+RUN git clone https://github.com/microsoft/DiskANN.git
+WORKDIR /app/DiskANN
+RUN git checkout $GIT_COMMIT_ISH
+
+# we would prefer to use the deps requested at the same commit. if the script doesn't exist we'll use the current one.
+RUN bash scripts/dev/install-dev-deps-ubuntu.bash || bash /app/fallback/install-dev-deps-ubuntu.bash
```
daxpryce, line 22: since our system deps aren't pinned to specific packages, we're not really achieving the goal of this; still, it doesn't hurt

.github/workflows/perf.yml
```diff
@@ -0,0 +1,26 @@
...
+    name: Run Perf Test from main
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+      - name: Build Perf Container
+        run: |
+          docker build --build-arg GIT_COMMIT_ISH="$GITHUB_SHA" -t perf -f scripts/perf/Dockerfile scripts
```
daxpryce, line 16: We could rely on "HEAD", but there's a subtle inconsistency error if someone manages to push a new push to main sometime between this job *starting* to run and the docker build step actually pulling at this commit.

So we're specifically going with the commit-that-was-at-the-start-of-the-job-schedule.  If GIT_COMMIT_ISH is not set, we default to HEAD.

scripts/perf/Dockerfile
```diff
@@ -0,0 +1,31 @@
+#Copyright(c) Microsoft Corporation.All rights reserved.
+#Licensed under the MIT license.
+
+FROM ubuntu:jammy
```
harsha-simhadri, line 4: should we perf tests for windows too?

harsha-simhadri (commented):
Is the json output on time deltas archived on the CI run logs, or posted somewhere else?

harsha-simhadri (approved):

PR #424, opened 2023-08-14
Change "convert_labels_string_to_int" to use non-negative continuous integer(starts from 0)

yiylin:
 
#### Reference Issues/PRs
when unv_label == "", it does not use label Id "0";
when unv_label != "", it does not use label Id "1";

After my changing, the label ids always start from 0 and continuous.

### DIFF
--- a/include/utils.h
+++ b/include/utils.h
@@ -172,7 +172,7 @@ inline void convert_labels_string_to_int(const std::string &inFileName, const st
             token.erase(std::remove(token.begin(), token.end(), '\r'), token.end());
             if (string_int_map.find(token) == string_int_map.end())
             {
-                uint32_t nextId = (uint32_t)string_int_map.size() + 1;
+                uint32_t nextId = (uint32_t)string_int_map.size();
                 string_int_map[token] = nextId;
             }
             lbls.push_back(string_int_map[token]);


Sanhaoji2 (approved):

.github/workflows/labels.yml
```diff
@@ -77,7 +77,7 @@ jobs:
         run: |
           dist/bin/build_memory_index --data_type uint8 --dist_fn l2 --FilteredLbuild 90 --data_path data/rand_uint8_10D_10K_norm50.0.bin --label_file data/zipf_labels_50_10K.txt --index_path_prefix data/index_l2_zipf_uint8_10D_10K_norm50_wlabel_nouniversal
           dist/bin/build_disk_index --data_type uint8 --dist_fn l2  --FilteredLbuild 90 --data_path data/rand_uint8_10D_10K_norm50.0.bin --label_file data/zipf_labels_50_10K.txt --index_path_prefix data/disk_index_l2_zipf_uint8_10D_10K_norm50_wlabel_nouniversal -R 16 -L 32 -B 0.00003 -M 1
-          dist/bin/search_memory_index --data_type uint8 --dist_fn l2 --filter_label 5 --fail_if_recall_below 70 --index_path_prefix data/index_l2_zipf_uint8_10D_10K_norm50_wlabel_nouniversal --query_file data/rand_uint8_10D_1K_norm50.0.bin --recall_at 10 --result_path temp --gt_file data/l2_zipf_uint8_10D_10K_norm50.0_10D_1K_norm50.0_gt100_wlabel_nouniversal -L  16 32
```
rakri, line 80: What is the reason for this change?

include/utils.h
```diff
@@ -172,7 +172,7 @@ inline void convert_labels_string_to_int(const std::string &inFileName, const st
             token.erase(std::remove(token.begin(), token.end(), '\r'), token.end());
             if (string_int_map.find(token) == string_int_map.end())
             {
-                uint32_t nextId = (uint32_t)string_int_map.size() + 1;
```
rakri, line 175: I see. Hope we don't make any assumptions elsewhere in the code which hardcodes 0 for universal label. Did you check this? Currently, if UNV is found, we have 0, 2, 3, ... and otherwise we have 1, 2, 3, .. for integer range.

.github/workflows/labels.yml
yiylin: I downloaded the data and could run the search_memory_index successfully. So I guess not-enough-memory causes the failure in the CI/CD pipeline

yiylin: reverted my changing

include/utils.h
yiylin: this function always assign "0" to universal label. so when UNV is found, it is the same as before. otherwise, we should not use "0" as special id

rakri: yes, makes sense. i just want to make sure we dnt assume that 0 means universal label in other places of code. will confirm once and approve.

rakri (approved):

PR #444, opened 2023-08-29
Python type hints and returns actually align this time.

### LINKED ISSUE #433: [BUG] diskannpy: batch_search is mistyped
## Expected Behavior

According to its [type annotations](https://github.com/microsoft/DiskANN/blob/4162c2118908cf55e719c0cdfdcd50b9069bef8f/python/src/_static_disk_index.py#L148), `batch_search` is expected to return a named tuple.

## Actual Behavior

It returns a normal tuple.

daxpryce:
 
Fixes #433 .

We were simply returning a bare tuple instead of the NamedTuple we claimed we were going to return. For those who use don't care about the named tuple accessor methods, this had no material difference. But for those who were relying upon access like `query_response.distances` to work, it would fail. 

This fix now ensures we're actually returning a `QueryResponse` or a `QueryBatchResponse`

### DIFF
--- a/python/tests/test_dynamic_memory_index.py
+++ b/python/tests/test_dynamic_memory_index.py
@@ -72,12 +72,15 @@ class TestDynamicMemoryIndex(unittest.TestCase):
                 )
 
                 k = 5
-                diskann_neighbors, diskann_distances = index.batch_search(
+                batch_response = index.batch_search(
                     query_vectors,
                     k_neighbors=k,
                     complexity=5,
                     num_threads=16,
                 )
+                self.assertIsInstance(batch_response, dap.QueryResponseBatch)
+
+                diskann_neighbors, diskann_distances = batch_response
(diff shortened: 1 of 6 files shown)


python/tests/test_dynamic_memory_index.py
```diff
@@ -115,7 +118,9 @@ def test_single(self):
                 index.batch_insert(vectors=index_vectors, vector_ids=generated_tags)
 
                 k = 5
-                ids, dists = index.search(query_vectors[0], k_neighbors=k, complexity=5)
+                response = index.search(query_vectors[0], k_neighbors=k, complexity=5)
+                self.assertIsInstance(response, dap.QueryResponse)
+                ids, dists = response
```
jonmclean, line 123: Horay for testing for a bug!

jonmclean:
Looks straight forward

jonmclean (approved):

PR #460, opened 2023-09-22
read file in one time

Sanhaoji2:


src/pq_flash_index.cpp
```diff
@@ -649,51 +667,88 @@ inline bool PQFlashIndex<T, LabelT>::point_has_label(uint32_t point_id, LabelT l
...
         throw diskann::ANNException(std::string("Failed to open file ") + label_file, -1);
     }
+    infile.seekg(0, std::ios::end);
+    size_t file_size = infile.tellg();
 
-    std::string line, token;
+    std::string buffer(file_size, ' ');
+
+    infile.seekg(0, std::ios::beg);
+    infile.read(&buffer[0], file_size);
```
yiylin, line 681: just wonder if others are ok with this changing. For our usage, this file size is not very big, so it is fine.

yiylin (approved):

src/pq_flash_index.cpp
harsha-simhadri: It does seem excessive to read the entire file into memory before counting filters.  Can it be chunked?

Sanhaoji2: As the memory usage of the label vector after loading is more than 2/3 of label file size in most scenario (uint16 to represent one label, uint32 for the offset of every node), the implement is just like we need double memory usage in the label loading time, I think it's still fine?

harsha-simhadri (approved):

PR #465, opened 2023-10-03
Handle io_setup error properly

### LINKED ISSUE #456: [BUG] crash when loading >64 StaticDiskIndex or when num_threads>64, linux
A crash occurs under any of the following circumstances:
- Multiple `StaticDiskIndex` objects exist in memory and the sum of their `num_threads` settings is greater than 64.
- A single `StaticDiskIndex` is constructed with `num_threads` > 64.
- A `StaticDiskIndex` object is initialized on a machine with more than 64 logical cores and `num_threads` = 0

The crash:

```
io_setup() failed; returned -11, errno=2:No such file or directory
terminate called after throwing an instance of 'std::system_error'
  what():  Operation not permitted
Fatal Python error: Aborted
```

That first line is printed by this code: https://github.com/microsoft/DiskANN/blob/fee17e6a34646c164444101983fd4df76b37fe6e/src/linux_aligned_file_reader.cpp#L152

I think `errno=2:No such file or directory` should be ignored here because `io_setup` isn't setting `errno`. Sometimes `errno=0:Success` is printed instead.

## Expected Behavior

Expect ability to construct an arbitrary number of `StaticDiskIndex`'s and to use all the threads available on the host-system.

## Actual Behavior

When the total number of threads used by static disk indexes exceeds 64, the program aborts.

## Example Code

```python
from tempfile import TemporaryDirectory

import diskannpy
import numpy as np

def load_multiple(index_dir: str, num_threads: int, num_copies: int):
    """initializes num_copies of StaticDiskIndex, each with num_threads threads"""
    [
        diskannpy.StaticDiskIndex(index_dir, num_threads, 0)
        for _ in range(num_copies)
    ]

with TemporaryDirectory() as tmpdir:
    diskannpy.build_disk_index(
        np.random.rand(10, 32).astype(np.float32),
        "l2",
        tmpdir,
        16,  # complexity,
        16,  # graph_degree,
        0.1,  # search_memory_maximum,
        1,  # build_memory_maximum
        1,  # num_threads
    )

    load_multiple(tmpdir, 1, 64)  # works
    load_multiple(tmpdir, 1, 64 + 1)  # aborts

    load_multiple(tmpdir, 64, 1)  # works
    load_multiple(tmpdir, 64 + 1, 1)  # aborts

    load_multiple(tmpdir, 16, 4)  # works
    load_multiple(tmpdir, 16 + 1, 4)  # aborts
```

## Error

```
Setting up thread-specific contexts for nthreads: 1
io_setup() failed; returned -11, errno=2:No such file or directory
terminate called after throwing an instance of 'std::system_error'
  what():  Operation not permitted
Fatal Python error: Aborted

Current thread 0x00007fa78b059000 (most recent call first):
  File "$$REDACTED$$/site-packages/diskannpy/_static_disk_index.py", line 102 in __init__
  File "$$REDACTED$$.py", line 140 in <listcomp>
  File "$$REDACTED$$.py", line 139 in load_multiple
...
```

## Your Environment

* Ubuntu 22.04.2 LTS
* diskannpy 0.6.0
* `cat /proc/sys/fs/aio-max-nr` == 65536

## Additional Details

`io_setup()` is part of a system library for asynchronous io called `libaio` https://man7.org/linux/man-pages/man2/io_setup.2.html . `io_setup` returns -11 to indicate: `EAGAIN` (try again) which is used to indicate the limit set in `/proc/sys/fs/aio-max-nr` has been exceeded.

Note `io_setup` negates the returned errors because libaio is cool like that.

*Hot Tip: you can monitor current `aio-nr` usage with `watch -n 0.1 cat /proc/sys/fs/aio-nr`*

ShawnZhong:
 
#### Reference Issues/PRs

Fixes #456

Quoting from  #456

> io_setup() is part of a system library for asynchronous io called libaio https://man7.org/linux/man-pages/man2/io_setup.2.html . io_setup returns -11 to indicate: EAGAIN (try again) which is used to indicate the limit set in /proc/sys/fs/aio-max-nr has been exceeded.

### DIFF
--- a/src/linux_aligned_file_reader.cpp
+++ b/src/linux_aligned_file_reader.cpp
@@ -147,10 +147,14 @@ void LinuxAlignedFileReader::register_thread()
     if (ret != 0)
     {
         lk.unlock();
-        assert(errno != EAGAIN);
-        assert(errno != ENOMEM);
-        std::cerr << "io_setup() failed; returned " << ret << ", errno=" << errno << ":" << ::strerror(errno)
-                  << std::endl;
+        if (ret == -EAGAIN)
+        {
+            std::cerr << "io_setup() failed with EAGAIN: Consider increasing /proc/sys/fs/aio-max-nr" << std::endl;
+        }
+        else
+        {
+            std::cerr << "io_setup() failed; returned " << ret << ": " << ::strerror(-ret) << std::endl;
+        }
     }
     else
     {


harsha-simhadri (approved):
Thanks for this change. Should we return the ret value?

PR #471, opened 2023-10-09
read from MemoryMappedFile when EXEC_ENV_OLS is defined

hliu18:
This has already been handled mostly except a few places. The PR fixes those missed places.

### DIFF
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -64,7 +64,7 @@ jobs:
         if: success() || failure()
         run: |
           dist/bin/build_disk_index --data_type uint8 --dist_fn l2 --universal_label 0  --FilteredLbuild 90 --data_path data/rand_uint8_10D_10K_norm50.0.bin --label_file data/rand_labels_50_10K.txt --index_path_prefix data/disk_index_l2_rand_uint8_10D_10K_norm50_wlabel -R 32 -L 5 -B 0.00003 -M 1
-          dist/bin/search_disk_index --data_type uint8 --dist_fn l2 --filter_label 10 --fail_if_recall_below 50 --index_path_prefix data/disk_index_l2_rand_uint8_10D_10K_norm50_wlabel --result_path /tmp/res --query_file data/rand_uint8_10D_1K_norm50.0.bin --gt_file data/l2_rand_uint8_10D_10K_norm50.0_10D_1K_norm50.0_gt100_wlabel --recall_at 5 -L 5 12 -W 2 --num_nodes_to_cache 10 -T 16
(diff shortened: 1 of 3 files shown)


src/pq_flash_index.cpp
```diff
@@ -626,7 +624,7 @@ void PQFlashIndex<T, LabelT>::get_label_file_metadata(std::string map_file, uint
 
     diskann::cout << "Labels file metadata: num_points: " << num_pts << ", #total_labels: " << num_total_labels
                   << std::endl;
-    infile.close();
+    infile.seekg(0);
```
ltan1ms, line 627: is this seek to beginning necessary?

src/pq_flash_index.cpp
```diff
@@ -777,15 +769,45 @@ int PQFlashIndex<T, LabelT>::load_from_separate_paths(uint32_t num_threads, cons
 
     this->_num_points = npts_u64;
     this->_n_chunks = nchunks_u64;
+#ifdef EXEC_ENV_OLS
+    if (files.fileExists(labels_file))
```
ltan1ms, line 773: optional: I feel argument name "files" is a somewhat misleading name, it naturally sound like files in file system. Maybe better to use memFiles or memMappedFiles

ltan1ms (approved):

src/pq_flash_index.cpp
hliu18: yes. The original code opens a new stream and read and close. It does not affect later reads from the stream (from the beginning). If we don't reset the cursor, reads after this point will not read correct data. Actually it may not read any data if we have already read to the end of the stream.

src/pq_flash_index.cpp
hliu18: sounds reasonable it's not related to this PR though.

.github/workflows/labels.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         if: success() || failure()
         run: |
           dist/bin/build_disk_index --data_type uint8 --dist_fn l2 --universal_label 0  --FilteredLbuild 90 --data_path data/rand_uint8_10D_10K_norm50.0.bin --label_file data/rand_labels_50_10K.txt --index_path_prefix data/disk_index_l2_rand_uint8_10D_10K_norm50_wlabel -R 32 -L 5 -B 0.00003 -M 1
-          dist/bin/search_disk_index --data_type uint8 --dist_fn l2 --filter_label 10 --fail_if_recall_below 50 --index_path_prefix data/disk_index_l2_rand_uint8_10D_10K_norm50_wlabel --result_path /tmp/res --query_file data/rand_uint8_10D_1K_norm50.0.bin --gt_file data/l2_rand_uint8_10D_10K_norm50.0_10D_1K_norm50.0_gt100_wlabel --recall_at 5 -L 5 12 -W 2 --num_nodes_to_cache 10 -T 16
```
hliu18, line 67: /tmp/res causes failure on Windows when creating a new file under /tmp while /tmp does not exist.

PR #477, opened 2023-10-16
Memory and SSD Filter index Improvement in Load time

NeelamMahapatro:

### DIFF
--- a/src/pq_flash_index.cpp
+++ b/src/pq_flash_index.cpp
@@ -526,10 +526,9 @@ void PQFlashIndex<T, LabelT>::generate_random_labels(std::vector<LabelT> &labels
 }
 
 template <typename T, typename LabelT>
-std::unordered_map<std::string, LabelT> PQFlashIndex<T, LabelT>::load_label_map(const std::string &labels_map_file)
+std::unordered_map<std::string, LabelT> PQFlashIndex<T, LabelT>::load_label_map(std::basic_istream<char> &map_reader)
 {
     std::unordered_map<std::string, LabelT> string_to_int_mp;
-    std::ifstream map_reader(labels_map_file);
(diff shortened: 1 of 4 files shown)


src/pq_flash_index.cpp
```diff
@@ -563,29 +563,44 @@ LabelT PQFlashIndex<T, LabelT>::get_converted_label(const std::string &filter_la
...
+
+    std::string label_str;
+    size_t cur_pos = 0, next_pos = 0;
+
+    while (cur_pos < file_size && cur_pos != std::string::npos)
     {
-        std::istringstream iss(line);
-        while (getline(iss, token, ','))
+        next_pos = file_content.find('\n', cur_pos);
+        if (next_pos == std::string::npos)
```
yiylin, line 580: this looks like to drop the last line if the last line does not end with "\n"? I do not know if this is ok for current existing data. have you tested it?

src/pq_flash_index.cpp
```diff
@@ -608,51 +623,89 @@ inline bool PQFlashIndex<T, LabelT>::point_has_label(uint32_t point_id, LabelT l
...
+    std::string label_str;
+    size_t cur_pos = 0;
+    size_t next_pos = 0;
+
+    while (cur_pos < file_size && cur_pos != std::string::npos)
     {
-        std::istringstream iss(line);
-        std::vector<uint32_t> lbls(0);
+        next_pos = buffer.find('\n', cur_pos);
+        if (next_pos == std::string::npos)
```
yiylin, line 661: this looks like to drop the last line if the last line does not end with "\n"? I do not know if this is ok for current existing data. have you tested it?

yiylin (approved):

Issue #455, opened 2023-09-18
[BUG] Loading a static disk index always triggers a warning, false-positive


bddap:
[This code](https://github.com/microsoft/DiskANN/blob/b8b6caf8c2bfd92be9326fd85946c78dc33eea7f/python/src/_static_disk_index.py#L87) passes a `1` to [this function](https://github.com/microsoft/DiskANN/blob/b8b6caf8c2bfd92be9326fd85946c78dc33eea7f/python/src/_common.py#L212).

The `_ensure_index_metadata()` function sees that `max_vectors` is set to `1` and [emits a warning](https://github.com/microsoft/DiskANN/blob/b8b6caf8c2bfd92be9326fd85946c78dc33eea7f/python/src/_common.py#L229C45-L229C45).

## Expected Behavior

Expecting absence of warning, it's not relevant when using a static disk index.

## Actual Behavior

This warning is printed:

```
.../diskannpy/_common.py:236: UserWarning: The number of vectors in the saved index equals max_vectors parameter. Any insertions will fail.
```

## Example Code

```python
import diskannpy
import numpy as np
import tempfile

with tempfile.TemporaryDirectory() as tmpdir:
    diskannpy.build_disk_index(
        np.array([[0.0] * 32] * 1000, dtype=np.float32),
        "l2",
        tmpdir,
        75,
        60,
        1,
        1,
        1,
    )
    diskannpy.StaticDiskIndex(tmpdir, 1, 0)
```

## Your Environment
* Ubuntu 22.04.2 LTS
* diskannpy 0.6.0

PR #482, opened 2023-11-03
Adding Filtered Index support to Python bindings

daxpryce:
 
Allows for the building of and searching of a filtered memory index, by providing a simple list[list[Hashable]] mapping at build time, and enabling it with a flag on load.

#### Any other comments?
I am not totally confident I know the full set of expectations, behaviors, and bounding edge cases for test in this particular scenario. @harsha-simhadri if you could explain precisely what edge cases I should properly test for other than the basics in this PR, I would be deeply appreciative!


python/src/_builder.py
```diff
@@ -174,8 +174,10 @@ def build_memory_index(
     num_pq_bytes: int = defaults.NUM_PQ_BYTES,
     use_opq: bool = defaults.USE_OPQ,
     vector_dtype: Optional[VectorDType] = None,
-    filter_complexity: int = defaults.FILTER_COMPLEXITY,
     tags: Union[str, VectorIdentifierBatch] = "",
+    filter_labels: Optional[list[list[Hashable]]] = None,
```
daxpryce, line 178: I'm not even confident this is the right signature, since it's just being serialized to a file anyway. Should I change this to `Optional[list[list[str]]]`?

python/src/_common.py
```diff
@@ -211,6 +211,7 @@ def _ensure_index_metadata(
     distance_metric: Optional[DistanceMetric],
     max_vectors: int,
     dimensions: Optional[int],
+    warn_size_exceeded: bool = False,
```
daxpryce, line 214: this fixes a bug whose number escapes me

daxpryce:
Closes #455

python/src/_common.py
daxpryce: #455

python/src/_builder.py
bryantower: It should probably just be str, why do you think it should be Hashable?

bryantower (commented):
It all looks pretty good to me. Did you want to wait on a comment from Harsha? I can approve this once we resolve the Hashable vs str issue.

bryantower (approved):

PR #486, opened 2023-11-08
flex_cblas_sdot and flex_cblas_snrm2

FJafargholi:
Using the new preprocessor flag, moved 2 of MKL methods into wrapper methods.

### DIFF
--- a/apps/utils/compute_groundtruth.cpp
+++ b/apps/utils/compute_groundtruth.cpp
@@ -26,8 +26,10 @@
 #else
 #include <stdlib.h>
 #endif
+
 #include "filter_utils.h"
 #include "utils.h"
+#include "../../src/math_utils.cpp"
 
 // WORKS FOR UPTO 2 BILLION POINTS (as we use INT INSTEAD OF UNSIGNED)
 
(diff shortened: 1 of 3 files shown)


apps/utils/compute_groundtruth.cpp
```diff
@@ -26,8 +26,10 @@
 #else
 #include <stdlib.h>
 #endif
+
 #include "filter_utils.h"
 #include "utils.h"
+#include "../../src/math_utils.cpp"
```
SMovaghati, line 32: can you include #include "math_utils.h"?

src/math_utils.cpp
```diff
@@ -4,13 +4,42 @@
...
 namespace math_utils
 {
 
+#ifdef USE_OPENBLAS
+using Flex_INT = int;
+#else
+using Flex_INT = MKL_INT;
+#endif
+
+float flex_cblas_sdot(const int64_t N, const float *X, const int incX, const float *Y, const int incY) noexcept
```
SMovaghati, line 25: can add these to math.h

SMovaghati (commented):
we should make sure their test projects are running and passing at each step, otherwise it would be hard to figure out which part is causing trouble.
maybe we can find one of their tests that hits all these functions and use that one (maybe test_insert_delete_consolidate)?

apps/utils/compute_groundtruth.cpp
FJafargholi: No, it doesn't know the path.

PR #490, opened 2023-11-09
Label Loading Fix

NeelamMahapatro:

### DIFF
--- a/src/pq_flash_index.cpp
+++ b/src/pq_flash_index.cpp
@@ -810,7 +810,7 @@ int PQFlashIndex<T, LabelT>::load_from_separate_paths(uint32_t num_threads, cons
         }
 #endif
         parse_label_file(infile, num_pts_in_label_file);
-        assert(num_pts_in_label_file == this->_num_points);
+        assert(num_pts_in_label_file == this->num_points);
 
 #ifndef EXEC_ENV_OLS
         infile.close();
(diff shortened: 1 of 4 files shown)


src/pq_flash_index.cpp
```diff
@@ -1766,6 +1766,20 @@ template <typename T, typename LabelT> char *PQFlashIndex<T, LabelT>::getHeaderB
...
+    {
+        if (str[start] == ch)
+        {
+            return start;
+        }
+    }
+
+    return std::string::npos;
+}
+
```
yiylin, line 1782: image

yiylin: so strange to have this error

src/pq_flash_index.cpp
```diff
@@ -810,7 +810,7 @@ int PQFlashIndex<T, LabelT>::load_from_separate_paths(uint32_t num_threads, cons
         }
 #endif
         parse_label_file(infile, num_pts_in_label_file);
-        assert(num_pts_in_label_file == this->_num_points);
+        assert(num_pts_in_label_file == this->num_points);
```
yiylin, line 813: why there are two "num_points"? can we make them to be one?

yiylin (approved):

src/pq_flash_index.cpp
NeelamMahapatro: yes, not sure of this. will check later

src/pq_flash_index.cpp
NeelamMahapatro: there is one num_points in pqflashindex. but since its assert statement, release build didnot fail I think

PR #498, opened 2023-12-11
only move the queue items if the new item index is less than size

SMovaghati:

Before, it used to always move the queue items, but that is not necessary if the item to add is at the end of the queue (and queue still has capacity).

### DIFF
--- a/include/neighbor.h
+++ b/include/neighbor.h
@@ -78,7 +78,7 @@ class NeighborPriorityQueue
             }
         }
 
-        if (lo < _capacity)
+        if (lo < _size)
         {
             std::memmove(&_data[lo + 1], &_data[lo], (_size - lo) * sizeof(Neighbor));
         }


include/neighbor.h
```diff
@@ -78,7 +78,7 @@ class NeighborPriorityQueue
             }
         }
 
-        if (lo < _capacity)
+        if (lo < _size)
```
theantony, line 81: @SMovaghati to confirm my understanding, this change avoids calling `memmove(..., ..., 0)` which is a logical no-op when 1) the new element is appended to the end of the queue and 2) the queue is not full. Is that right?

(Only in this scenario `_lo == _size`, otherwise always `lo < _size`)

theantony (approved):

include/neighbor.h
SMovaghati: That is correct. If we are not adding an item at the end of the queue, then we always need the memmove to open up a space in the middle.

Sanhaoji2 (approved):

mmthomas:
The code  as is currenly correct; there is no rush to merge this change.  However, it does allocate extra memory in the vector to support the current insertion strategy, and it keeps track of its own size and capacity.

We have a larger improvment to this whole class that simplifies the logic, reduces the memory requried, and removes the need for and explicit memmove, allowing std:vector to manage the insertion optimally.

@SMovaghati we should bring in all of our changes to this class instead of this small change.

mmthomas (changes requested):
@SMovaghati I think we should bring our entire changes to this class for this PR; it simplifies the logic, it removes the extra size and capacity state, the need to allocate an extra item, and uses the built-in vector element moving when required.

PR #506, opened 2024-01-19
add16bytes tag type

Sanhaoji2:

### DIFF
--- a/include/natural_number_map.h
+++ b/include/natural_number_map.h
@@ -26,9 +26,6 @@ template <typename Key, typename Value> class natural_number_map
 {
   public:
     static_assert(std::is_trivial<Key>::value, "Key must be a trivial type");
-    // Some of the class member prototypes are done with this assumption to
-    // minimize verbosity since it's the only use case.
-//    static_assert(std::is_trivial<Value>::value, "Value must be a trivial type");
 
     // Represents a reference to a element in the map. Used while iterating
     // over map entries.
--- a/include/natural_number_map.h
+++ b/include/natural_number_map.h
@@ -28,7 +28,7 @@ template <typename Key, typename Value> class natural_number_map
     static_assert(std::is_trivial<Key>::value, "Key must be a trivial type");
     // Some of the class member prototypes are done with this assumption to
     // minimize verbosity since it's the only use case.
-    static_assert(std::is_trivial<Value>::value, "Value must be a trivial type");
+//    static_assert(std::is_trivial<Value>::value, "Value must be a trivial type");
 
     // Represents a reference to a element in the map. Used while iterating
     // over map entries.
--- a/include/utils.h
+++ b/include/utils.h
@@ -27,6 +27,7 @@ typedef int FileHandle;
 #include "windows_customizations.h"
 #include "tsl/robin_set.h"
 #include "types.h"
+#include "tag_uint128.h"
 #include <any>
 
 #ifdef EXEC_ENV_OLS
@@ -1006,6 +1007,18 @@ inline void prefetch_vector_l2(const char *vec, size_t vecsize)
 void block_convert(std::ofstream &writr, std::ifstream &readr, float *read_buf, uint64_t npts, uint64_t ndims);
 
 DISKANN_DLLEXPORT void normalize_data_file(const std::string &inFileName, const std::string &outFileName);
+
+
+inline std::string get_tag_string(std::uint64_t tag)
+{
+    return std::to_string(tag);
+}
+
+inline std::string get_tag_string(tag_uint128 tag)
+{
+    std::string str = std::to_string(tag._data2) + "_" + std::to_string(tag._data1);
+    return str;
+}
 
 }; // namespace diskann
 
--- a/include/utils.h
+++ b/include/utils.h
@@ -1008,7 +1008,6 @@ void block_convert(std::ofstream &writr, std::ifstream &readr, float *read_buf, 
 
 DISKANN_DLLEXPORT void normalize_data_file(const std::string &inFileName, const std::string &outFileName);
 
-
 inline std::string get_tag_string(std::uint64_t tag)
 {
     return std::to_string(tag);
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -12,6 +12,7 @@
 #include "tsl/robin_map.h"
 #include "tsl/robin_set.h"
 #include "windows_customizations.h"
+#include "tag_uint128.h"
 #if defined(DISKANN_RELEASE_UNUSED_TCMALLOC_MEMORY_AT_CHECKPOINTS) && defined(DISKANN_BUILD)
 #include "gperftools/malloc_extension.h"
 #endif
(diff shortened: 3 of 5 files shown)


src/index.cpp
```diff
@@ -2022,14 +2023,10 @@ std::pair<uint32_t, uint32_t> Index<T, TagT, LabelT>::_search_with_filters(const
                                                                            float *distances)
 {
     auto converted_label = this->get_converted_label(raw_label);
-    if (typeid(uint64_t *) == indices.type())
```
rakri, line 2025: i think indices refers to actual location of vectors, not the tags. it is used for a bulk index (not a streaming scenario)  where we implicitly assume 0th vector has index= 0, 1st vector has index 1, etc. In some ads use case, they wanted the index itself to be either uint64 or uint32, hence this type casting. This is different from tags, so please leave this if condition in place.

src/index.cpp
```diff
@@ -2100,7 +2097,7 @@ std::pair<uint32_t, uint32_t> Index<T, TagT, LabelT>::search_with_filters(const
                 TagT tag;
                 if (_location_to_tag.try_get(best_L_nodes[i].id, tag))
                 {
-                    indices[pos] = (IdType)tag;
```
rakri, line 2103: same here, i think the existing search_with_filters code itself has got index and tags confused. so please let us discuss with Gopal to finalize this.

rakri (changes requested):

include/natural_number_map.h
```diff
@@ -26,9 +26,6 @@ template <typename Key, typename Value> class natural_number_map
 {
   public:
     static_assert(std::is_trivial<Key>::value, "Key must be a trivial type");
-    // Some of the class member prototypes are done with this assumption to
-    // minimize verbosity since it's the only use case.
-    static_assert(std::is_trivial<Value>::value, "Value must be a trivial type");
```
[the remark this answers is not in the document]
Sanhaoji2: tag_128 is struct and it's not trivial type, for usage of natural_number_map in index object, it don't need to be trivial type, the type with copyable should be enough.

include/utils.h
```diff
@@ -1007,6 +1008,17 @@ void block_convert(std::ofstream &writr, std::ifstream &readr, float *read_buf,
 
 DISKANN_DLLEXPORT void normalize_data_file(const std::string &inFileName, const std::string &outFileName);
 
+inline std::string get_tag_string(std::uint64_t tag)
+{
+    return std::to_string(tag);
+}
+
+inline std::string get_tag_string(tag_uint128 tag)
+{
```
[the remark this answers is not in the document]
Sanhaoji2: it's used in print log, like "std::cout << get_tag_string(tag)", if it's moved to the member funtion of tag_uint128, we have to check the type of tag before print, as base type like uint64 isn't an object.

src/index.cpp
Sanhaoji2: we use the same search function to support static index and streaming index:
https://github.com/microsoft/DiskANN/blob/main/src/index.cpp#L2103

That means the tag type must support to convert to IdType even we don't go through the streaming part as the template instantiate. And as the tag may not be a basic type(uint32, uint64, etc), we can't instantiate the template like:

Index<T=uint8, TagT=tag_uint128, LabelT=uint32, IdType=uint64> because the following code will compile error (line 2100):
TagT tag;
if (_location_to_tag.try_get(best_L_nodes[i].id, tag))
{
    indices[pos] = (IdType)tag;
}

The code I removed here is trying to convert IdType to uint32 or uint64, that will trigger a template instantiate for all index template define in line 3327 including Index<T=uint8, TagT=tag_uint128, LabelT=uint32>:
https://github.com/microsoft/DiskANN/blob/main/src/index.cpp#L3327

Sanhaoji2: The code is already use force cast to convert Tag to IdType, so I think it should be safe here, and I also think define uint64 tag but want to return uint32 indices doesn't make sense.

The TagT is uint32 by default, and the indices is uint32 or uint64, so nothing is broken to remove this code.

If we still need to keep this code, we have to separate the searching funtion of static index and streaming index, like:
search_with_filters_tags();

src/index.cpp
Sanhaoji2: reverted, but still need to keep:
indices[pos] = best_L_nodes[i].id;

As best_L_nodes[i].id is uint32, we can't cast to tag_uint128

src/index.cpp
```diff
@@ -3349,19 +3349,16 @@ template DISKANN_DLLEXPORT class Index<uint8_t, int64_t, uint16_t>;
 template DISKANN_DLLEXPORT class Index<float, uint64_t, uint16_t>;
 template DISKANN_DLLEXPORT class Index<int8_t, uint64_t, uint16_t>;
 template DISKANN_DLLEXPORT class Index<uint8_t, uint64_t, uint16_t>;
+template DISKANN_DLLEXPORT class Index<float, tag_uint128, uint16_t>;
+template DISKANN_DLLEXPORT class Index<int8_t, tag_uint128, uint16_t>;
+template DISKANN_DLLEXPORT class Index<uint8_t, tag_uint128, uint16_t>;
 
 template DISKANN_DLLEXPORT std::pair<uint32_t, uint32_t> Index<float, uint64_t, uint32_t>::search<uint64_t>(
     const float *query, const size_t K, const uint32_t L, uint64_t *indices, float *distances);
-template DISKANN_DLLEXPORT std::pair<uint32_t, uint32_t> Index<float, uint64_t, uint32_t>::search<uint32_t>(
```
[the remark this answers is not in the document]
Sanhaoji2: reverted.

src/index.cpp
rakri: I think separating is better. Ads team wanted IndexType = uint32 or uint64 for static index. So the IndexType is used for static build, and TagT is used for dynamic index, and there is bug in existing code itself even before your PR. 

To fix this, can you just add "bool use_filters = false, const std::string raw_label="" as default arguments in all variants of search_with_tags (with and without _ and in abstract_index and index) to enable filter support in dynamic indices? If you want I can draft a PR for this and you can rebase. Let me know if you want to do this or me.

rakri: on discussing with Gopal, I think it is best to go ahead with this PR and fix that issue in next PR.

rakri (approved):

src/index.cpp
Sanhaoji2: Change search_with_tags() to support filtered search, and remove the code of streaming in search_with_filters().

include/abstract_index.h
```diff
@@ -120,7 +121,8 @@ class AbstractIndex
     virtual void _set_start_points_at_random(DataType radius, uint32_t random_seed = 0) = 0;
     virtual int _get_vector_by_tag(TagType &tag, DataType &vec) = 0;
     virtual size_t _search_with_tags(const DataType &query, const uint64_t K, const uint32_t L, const TagType &tags,
-                                     float *distances, DataVector &res_vectors) = 0;
+                                     float *distances, DataVector &res_vectors, bool use_filters,
+                                     const std::string filter_label) = 0;
```
rakri, line 125: why there is no default args here?

rakri (approved):

gopalrs (approved):

rakri (approved):

rakri:
Adding 128-byte tag support for dynamic index.

PR #510, opened 2024-02-06
Version bump 0.7.0rc2->0.7.0

daxpryce:
Preparing diskannpy for 0.7.0 release (filter support, static memory indices only)

0.7.0rc2 has been published on PyPI as a pre-release for a while. I'm using it pretty frequently without any troubles, so we should just release it as a final 0.7.0 cut.


jonmclean (approved):
Changes make sense and are approved. AlmaLinux updated their GPG key so we need the new one in order to build/publish correctly.  https://almalinux.org/blog/2023-12-20-almalinux-8-key-update/

PR #518, opened 2024-02-26
add wait() method to AlignedFileReader

hliu18:
Add wait() method for asynchronous IOs where there is no callback support in the underlying implementation.
Caller can issue a number of IOs asynchronously and then block and wait for all the IO results to come back.


ltan1ms (approved):

harsha-simhadri:
Could you please add a description to the PR that provides context?

PR #519, opened 2024-02-29
Fix the wrong behavior when saving tags without calling any lazy_delete.

NingyuanChen:
When there is no lazy_deletes called, _data_compacted will remain true, which will make compact_data no-op. However, save tags has the assumption that compact_data is called so that tags are in the right position.


ltan1ms (approved):
great finding!

PR #520, opened 2024-02-29
Extending simplified PQ functions

michael-popov:
- [-] Does this PR add any new dependencies?
- [ x] Does this PR modify any existing APIs?
- [-] Should this result in any changes to our documentation, either updating existing docs or adding new ones?
 
Extending functions for in-memory PQ calculations:
  - added support for generating and using centroids.
  - removed constrain dim % num_pq_chunks == 0.

For testing I modified (but not added to the PR) generate_pq utility. Code for testing loads pivot and centroids data produced by original functions and uses these data to calculate PQ with a new function. Then it compares results with PQ data produced by the original function.

```
    . . .
    diskann::generate_pq_data_from_pivots<float>(data_path, (uint32_t)num_pq_centers, (uint32_t)num_pq_chunks,
                                             pq_pivots_path, pq_compressed_vectors_path, opq);

    const size_t num_centers = 256;
    std::vector<float> pivot_data;
    std::vector<float> centroids;
    std::vector<float> data;
    std::vector<uint8_t> pq;
    std::vector<uint8_t> control_pq;
    int ret;

    ret = generate_pq_pivots_simplified(train_data, train_size, train_dim, num_pq_chunks,
                                       pivot_data, centroids);
    assert(ret == 0);
    assert(pivot_data.size() == num_centers * train_dim);
    assert(centroids.size() == train_dim);

    load_pivots(pq_pivots_path, pivot_data, centroids);
    load_data(data_path, data);
    load_data(pq_compressed_vectors_path, control_pq);

    ret = generate_pq_data_from_pivots_simplified(&data[0], data.size() / train_dim,
                                                 &pivot_data[0], pivot_data.size(),
                                                 train_dim, num_pq_chunks,
                                                 centroids,
                                                 pq);
    assert(ret == 0);
    assert(pq.size() == control_pq.size());
    for (size_t i = 0; i < pq.size(); i++)
    {
        assert(pq[i] == control_pq[i]);
    }

```

### DIFF
--- a/src/pq.cpp
+++ b/src/pq.cpp
@@ -348,29 +348,27 @@ void pq_dist_lookup(const uint8_t *pq_ids, const size_t n_pts, const size_t pq_n
 // Input is provided in the in-memory buffer train_data.
 // Output is stored in the in-memory buffer pivot_data_vector.
 // Simplification is based on the following assumptions:
-//   dim % num_pq_chunks == 0
 //   num_centers == 256 by default
-//   KMEANS_ITERS_FOR_PQ == 15 by default
-//   make_zero_mean is false by default.
 // These assumptions allow to make the function much simpler and avoid storing
 // array of chunk_offsets and centroids.
 // The compiler pragma for multi-threading support is removed from this implementation
 // for the purpose of integration into systems that strictly control resource allocation.
-int generate_pq_pivots_simplified(const float *train_data, size_t num_train, size_t dim, size_t num_pq_chunks,
-                                  std::vector<float> &pivot_data_vector)
+int generate_pq_pivots_simplified(float *train_data, size_t num_train, size_t dim, size_t num_pq_chunks,
+                                  std::vector<float> &pivot_data_vector, std::vector<float> &centroids,
+                                  const bool make_zero_mean, const uint32_t kmeans_iters_for_pq)
 {
-    if (num_pq_chunks > dim || dim % num_pq_chunks != 0)
+    if (num_pq_chunks == 0 || num_pq_chunks > dim)
     {
         return -1;
     }
 
     const size_t num_centers = 256;
-    const size_t cur_chunk_size = dim / num_pq_chunks;
-    const uint32_t KMEANS_ITERS_FOR_PQ = 15;
+    const size_t chunk_size = dim / num_pq_chunks;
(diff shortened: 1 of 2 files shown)


src/pq.cpp
```diff
@@ -776,24 +828,32 @@ int generate_opq_pivots(const float *passed_train_data, size_t num_train, uint32
...
 // for the purpose of integration into systems that strictly control resource allocation.
-int generate_pq_data_from_pivots_simplified(const float *data, const size_t num, const float *pivot_data,
+int generate_pq_data_from_pivots_simplified(float *data, const size_t num, const float *pivot_data,
                                             const size_t pivots_num, const size_t dim, const size_t num_pq_chunks,
-                                            std::vector<uint8_t> &pq)
+                                            const std::vector<float> &centroids, const std::vector<uint32_t> &offsets,
+                                            std::vector<PQ_DATA_TYPE> &pq)
 {
-    if (num_pq_chunks == 0 || num_pq_chunks > dim || dim % num_pq_chunks != 0)
+    if (num_pq_chunks < 2 || num_pq_chunks > dim)
```
rakri, line 841: why if < 2 we return -1?

michael-popov: Return value -1 means functions failed for some reason.

We check input parameters:
 - num_pq_chunks == 0   =>  there is a division by num_pq_chunks a bit later, which will crash the system
 - num_pq_chunks == 1   =>  originally I had offsets.size() == num_pq_chunks, and then I use offsets[1] as the largest possible size for a chunk. If num_pq_chunks == 1, then access offsets[1] will crash the system. Eventually I accepted the wisdom of the original function and added one more element to offsets so offsets.size() == num_pq_chunks+1 but the check of num_pq_chunks value was left the same.

In general, num_pq_chunks==1 does not make sense anyway, right? In this case the whole vector is squashed into a single number 0..255. Too much information is lost. I can't imagine a practical case when num_pq_chunks==1 is useful.

On the other hand, there are many things that I cannot imagine but they are still valid. :)
I can change the check to if (num_pq_chunks == 0) return -1; if required.

rakri: I'd say lets do what the last option is, we can allow num_pq_chunks =1.

harsha-simhadri (approved):

PR #526, opened 2024-03-07
replace callback with Wait() method

hliu18:
replace callback driven wait logic in getNextCompletedRequest with new interface method Wait() on AlignedFileReader directly. details moved into Wait() implementation and becomes transparent to the caller. The change is needed to accommodate lower IO layer implementation that does not support callback.

change affects only code inside USE_BING_INFRA.


ltan1ms (approved):

hliu18:
@harsha-simhadri can you help merge this PR? thanks.

PR #578, opened 2024-08-14
Fixing the hang when search is run on single thread

gopalrs:
The PR fixes the code freeze that occurs when search is run on a single thread.


gopalrs (commented):
Since it is a minor change and is tested, doing just one review.

PR #591, opened 2024-10-15
Update push-test.yml

daxpryce:
When the upload-artifact action was updated to v4, the behavior of it changes such that you cannot create a unified "artifact" by having it add files to an existing artifact name. Instead, they all require a unique artifact name to be uploaded.

This should fix the current workflow build errors.

### DIFF
--- a/.github/workflows/disk-pq.yml
+++ b/.github/workflows/disk-pq.yml
@@ -111,7 +111,7 @@ jobs:
       - name: upload data and bin
         uses: actions/upload-artifact@v4
         with:
-          name: disk-pq
+          name: disk-pq-${{matrix.os}}
           path: |
             ./dist/**
             ./data/**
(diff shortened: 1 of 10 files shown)


daxpryce:
The artifact woes are more extensive than I thought - I will address them all in a single PR, not just one I made through the github file editor 😆

gopalrs:
Thanks Dax, I didn't realize how complicated this change was :)

daxpryce:
Yeah, this is why I didn't upgrade it before, I think the only problem we'll have is the python wheel part since that'll be across like N different artifacts, but it's not insurmountable - just a lot of places to adjust. I'll try to get this together today :)

daxpryce:
This is still failing on one of the dynamic tests but it's getting surprising exit codes and isn't really in scope for *this* PR. Someone more competent in the windows-diskann-dynamic workflow space should probably check to see why sometimes we get random exit codes of 127 and such though. Seems like we have a bug somewhere?

daxpryce:
[laugh]         Dax Pryce reacted to your message:
________________________________

gopalrs:
I will look into the other test failures, Dax. Thanks for fixing the upload issues.

gopalrs (approved):

Issue #544, opened 2024-04-25
[Question] Why require numpy version stick to 1.25?


KuribohG:
I want to build diskannpy for py3.8, however I found this PR: https://github.com/microsoft/DiskANN/pull/396.
Why diskannpy relies on exactly version 1.25 of numpy? I didn't find further explanation in this repo.
Seems this PR is related to subnormal float computation. Did numpy 1.25 has some special features or bug fixes for subnormal floats?

daxpryce:
@KuribohG sorry I'm so late in getting back to you, I just now saw this issue.

There is *no* reason that you need to stick with numpy 1.25. The problem is that the version that is compiled is compiled against the headers in numpy - and when I published to pypi I could pick precisely one version to do that for.

For further flexibility in this regard, we would need to create a conda recipe to build this. I didn't have time to do this myself, though I'd be eager for a PR in this regard if anyone could make it.

If you want to build against a different version of numpy, you are more than welcome to!  To do this, follow these steps (note they are probably rough as they are from my flawed memory, but it might be enough to get you started)

1: Install all the project's prerequisites for your system. The instructions are in the README.md in the root of the repo. If you're building for ubuntu, there's a helper script in [scripts/dev/install-dev-deps-ubuntu.bash](https://github.com/microsoft/DiskANN/blob/main/scripts/dev/install-dev-deps-ubuntu.bash) that will do it all at once.
2: Edit [pyproject.toml](https://github.com/microsoft/DiskANN/blob/main/pyproject.toml#L6) and change the version to 1.26, 1.24, 1.23 etc. 
3: create a virtual environment `python3.11 -m venv venv` and activate it `source venv/bin/activate`
4: `pip install build`
5: python -m build

This should create a wheel that will work with your chosen version of numpy. 

Note that this is ultimately what a conda recipe would do.

daxpryce:
Also note: you can use whatever python version you'd like, >= 3.9

PR #592, opened 2024-10-17
Memory map load vectors_from_file if requested

daxpryce:
This commit adds some basic memmap support to the `diskannpy.vectors_from_file` utility function. You can now return a memory mapped np.array conformant view (np.memmap object) vs. requiring it be loaded fully into memory (np.ndarray style).

#### Any other comments?
Note that I set the default to be mode="r" not "r+", which is the numpy default. I would prefer we don't mess with mutating the vectors prior to handing them to diskann for build, I have no idea if that way even works. I also wanted to leave it open until we figured out if it did.

### DIFF
--- a/python/tests/test_static_memory_index.py
+++ b/python/tests/test_static_memory_index.py
@@ -306,6 +306,8 @@ class TestFilteredStaticMemoryIndex(unittest.TestCase):
             with_filter_but_label_all, _ = index.search(
                 query_vectors[0], k_neighbors=k*2, complexity=128, filter_label="all"
             )
+            # this has begun failing randomly. gopalrs - this *shouldn't* fail, but for some reason the filtered search
+            # will sometimes return fewer than the expected 10k results here. 
             intersection = np.intersect1d(without_filter, with_filter_but_label_all)
             intersect_count = intersection.shape[0]
             self.assertEqual(intersect_count, k*2)
--- a/.github/workflows/disk-pq.yml
+++ b/.github/workflows/disk-pq.yml
@@ -111,7 +111,7 @@ jobs:
       - name: upload data and bin
         uses: actions/upload-artifact@v4
         with:
-          name: disk-pq
+          name: disk-pq-${{matrix.os}}
           path: |
             ./dist/**
             ./data/**
--- a/.github/workflows/dynamic-labels.yml
+++ b/.github/workflows/dynamic-labels.yml
@@ -96,7 +96,7 @@ jobs:
       - name: upload data and bin
         uses: actions/upload-artifact@v4
         with:
-          name: dynamic
+          name: dynamic-labels-${{matrix.os}}
           path: |
             ./dist/**
             ./data/**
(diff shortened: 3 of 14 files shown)


daxpryce:
(note: I'll want to rebase on main after the mostly-fixed-the-build PR is merged)

gopalrs (commented):
Should we update the dependencies? (Do we even have a requirements.txt?)

gopalrs:
Sorry for the delay - GH notifications go to my gmail account, and I don't look at it often enough

daxpryce:
> Should we update the dependencies? (Do we even have a requirements.txt?)

Dependencies are in pyproject.toml - which is where you want dependencies defined for the library you're publishing. requirements.txt are more for applications that don't get published, and thus, won't be imported.

Anyway, the only strong requirement we have in our deps is numpy - which is actually a big challenge, as we use the numpy headers @ cxx compile time and we need to create compiled wheels for every version of numpy, which in turn can't really be published to pypi as there's no structural way to keep the "compiled for python 3.10 and numpy 1.25" separate from "compiled for python 3.11 and numpy 2.0.2". that's what conda is for, really. See https://github.com/microsoft/DiskANN/issues/544 for more details.

-----

Continuing on this topic but also slightly tangential, I have absolutely seen that pyarrow and polars have native (read: something was compiled) support for numpy without it being a fixed version. I have no idea how they've done it, but it may very well be possible and we can be more loosey goosey with our strict numpy requirements.

------

Which then dovetails into the actual answer to your question: we can absolutely upgrade the requirements, but they're basically all build time requirements, not runtime. Numpy is the only one that is a runtime requirement.  Long answer shortened: "maybe, but not right now, we should look into it a bit deeper first to find a more optimal way of handling the numpy dep that isn't 'change the strict numpy 1.25 requirement into a strict numpy 2.0 requirement' because it's all sorts of crap for a bunch of reasons :)

gopalrs (approved):

daxpryce:
@gopalrs

I'm getting occassional failures in one of the unit tests that seems to be around filtered memory indices. In short, my test compares an index search results <without> filters, vs. an index search with filters (but every element has the same label). The expectation is the two result sets will be identical, but that does not seem to be the case.

I've added a comment to the unit test explaining it with your name in it to showcase what I'm doing, but because this is a: sporadic and b: seems to be a result of some changes in the cxx itself (nothing in the python has changed), I wanted to flag it as a test case that maybe the Cxx tests aren't catching or covering.

python/tests/test_static_memory_index.py
```diff
@@ -306,6 +306,8 @@ def test_exhaustive_validation(self):
             with_filter_but_label_all, _ = index.search(
                 query_vectors[0], k_neighbors=k*2, complexity=128, filter_label="all"
             )
+            # this has begun failing randomly. gopalrs - this *shouldn't* fail, but for some reason the filtered search
```
daxpryce, line 309: @gopalrs this is the comment I'm talking about. it works sometimes, other times it doesn't - the inconsistency is going to make it hard to track down, but it should work 100% of the time (it did before at least!)

PR #604, opened 2024-12-03
filtered streaming fix 

Sanhaoji2:

### DIFF
--- a/include/abstract_index.h
+++ b/include/abstract_index.h
@@ -80,8 +80,8 @@ class AbstractIndex
                                                       float *distances);
 
     // insert points with labels, labels should be present for filtered index
-    template <typename data_type, typename tag_type, typename label_type>
-    int insert_point(const data_type *point, const tag_type tag, const std::vector<label_type> &labels);
+    template <typename data_type, typename tag_type>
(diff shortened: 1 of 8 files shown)


rakri:
So the main logical change in this PR, is that we get rid of frozen point concept, but medoids (global or for any label) can never be deleted? Meaning that we need to skip outputting them at search time, if they are deleted..

rakri:
So now location_to_tag and tag_to_location will not contain these medoids once they are deleted.. Are we making sure that some other new inserted node doesnt take up these locations and get inserted there? In other words, we now lose the invariant delete_set.size() + location_to_tag.size() + empty_slots = _max_points?

Sanhaoji2:
the medoids are not moving to delete set, so it won't be assigned to other nodes again in insert operation. 

For the question "Meaning that we need to skip outputting them at search time, if they are deleted", yes, they may skip outputting   if they got deleted, the logic is as the same as using frozen points:

In previous logic we start with frozen points, while the frozen points are selected in searching, we also need to skip them, there are  no different here.

For the question "lose the invariant delete_set.size() + location_to_tag.size() + empty_slots = _max_points", yes.

PR #644, opened 2025-04-08
thread context for omp parallel to pass thread local allocator

ltan1ms:

### DIFF
--- a/include/thread_context.h
+++ b/include/thread_context.h
@@ -1,3 +1,5 @@
+#pragma once
+
 #ifndef _THREAD_CONTEXT_H_
 #define _THREAD_CONTEXT_H_
 


include/thread_context.h
```diff
@@ -0,0 +1,49 @@
+#ifndef _THREAD_CONTEXT_H_
+#define _THREAD_CONTEXT_H_
```
amrmorsey, line 2: Will using `#pragma once` instead be better?

src/index.cpp
```diff
@@ -1605,7 +1613,11 @@ template <typename T, typename TagT, typename LabelT> void Index<T, TagT, LabelT
     {
         diskann::cout << "Starting final cleanup.." << std::flush;
     }
+#ifdef _ANN_ALLOCATOR
+#pragma omp parallel for schedule(dynamic, 2048) firstprivate(context)
```
amrmorsey, line 1617: Should we also call `OmpParallelContext context;` here?

include/thread_context.h
ltan1ms: Fixed now.

PR #660, opened 2025-07-17
Adding support for external logger in disk ann index operations.

d4mator:
 
DiskANN code output various statements to cout and cerr streams. In some cases this code runs as part of the service that cannot use these streams, such as when run as an Autopilot service. This output is crucial for monitoring, debugging.

The change adds an interface for the logging component and allows to specify a custom implementation of the logger. By default logging is done to a console implementation (similar to the existing cout and cerr).

Mostly converted cout and cerr statements from index.cpp, but not everything. Only the most important parts were converted.

### DIFF
--- a/include/logging.h
+++ b/include/logging.h
@@ -1,0 +1,39 @@
+#pragma once
+
+#include "windows_customizations.h"
+
+namespace diskann
+{
+	namespace logging
+	{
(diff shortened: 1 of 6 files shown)


d4mator:
Wrong target branch selected at first, closing this PR

PR #670, opened 2025-09-12
Support for io_uring

myamada9999:
 
#### Reference Issues/PRs

Nothing.

This patch provides a means to switch from aio to io_uring in diskann.
Generally, aio is older, and in recent Linux distributions, io_uring is often adopted due to its superior performance.

https://www.snia.org/sites/default/files/SDC/2019/presentations/Storage_Performance/Kariuki_John_Verma_Vishal_Improved_Storage_Performance_Using_the_New_Linux_Kernel_I.O_Interface.pdf

By installing the liburing-dev package and building with the -DIOURING=True macro, the parts using libaio will be replaced with liburing.

```
# apt-get install liburing-dev
$ cmake -DIOURING=True -DCMAKE_BUILD_TYPE=Release ..
$ make -j
```

Using a system with 128 AMD cores and 8 PCIe Gen5 SSDs configured in RAID 0, we compared the QPS between aio and io_uring. We observed a performance improvement of 9% to 27% with fewer than 100 threads. However, once the thread count exceeded 100 and approached the physical CPU core count of 128, we found that aio performed better. 

It appears that io_uring is not completely superior to aio, so I implemented a patch to allow switching between aio and io_uring. If you have any other good implementation methods or solutions, please feel free to suggest them.

#### Any other comments?

Nothing.


include/aligned_file_reader.h
```diff
@@ -10,10 +10,18 @@
 
 #ifndef _WINDOWS
 #include <fcntl.h>
+#ifndef IOURING
```
suhasjs, line 13: Does using `-DIOURING` choose between io_uring and io_submit APIs? If yes, can you please document this in the README (or inline in the headers)?
Additionally, this means you should also consider linking only one of the two libraries: `DISKANN_ASYNC_LIB` or `DISKANN_ASYNC_LIB_URING` (see the comment by Copilot) above

src/linux_aligned_file_reader_uring.cpp
```diff
@@ -0,0 +1,218 @@
...
+                              << ", ernno=" << errno << "=" << ::strerror(-ret) << ", try #" << n_tries + 1 << std::endl;
+                    exit(-1);
+                }
+                else
+                {
+					io_uring_cq_advance(ctx, n_ops);
+                    break;
+                }
+            }
+			n_tries++;
```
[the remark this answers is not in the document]
suhasjs: This matches the current code flow in `linux_aligned_file_reader.cpp` that uses the AIO stack, so it's okay if the function just 'crashes' the program if IOs start failing.

src/linux_aligned_file_reader_uring.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include "linux_aligned_file_reader_uring.h"
+
+#include <cassert>
+#include <cstdio>
+#include <iostream>
+#include "tsl/robin_map.h"
+#include "utils.h"
+#define QUEUE_DEPTH 1024
+
+void execute_io(IOContext ctx, int fd, std::vector<AlignedRead> &read_reqs, uint64_t n_retries = 0)
```
suhasjs, line 15: Please wrap `execute_io` in an anonymous namespace as done in `linux_aligned_file_reader.cpp` to prevent symbol conflicts.
```cpp
namespace {
    void execute_io(...) { ... }
} // namespace anonymous

```

include/aligned_file_reader.h
myamada9999: I understand. Thank you.
I’m working on my reply, so please wait a moment.

myamada9999: Thanks for the review, @suhasjs  — I’ve addressed your comments.
Let me know if anything else needs to be changed.

src/linux_aligned_file_reader_uring.cpp
myamada9999: Thanks for the review, @suhasjs  — I’ve addressed your comments.
Let me know if anything else needs to be changed.

src/linux_aligned_file_reader_uring.cpp
```diff
@@ -0,0 +1,219 @@
...
+    {
+        std::cerr << "multiple calls to register_thread from the same thread" << std::endl;
+        return;
+    }
+    IOContext ctx = (struct io_uring *)malloc(sizeof(struct io_uring));
+	int ret = io_uring_queue_init(QUEUE_DEPTH, ctx, 0);
+    if (ret != 0)
+    {
+        lk.unlock();
+        if (ret == -EAGAIN)
```
suhasjs, line 142: @myamada9999  Please `free` the `ctx` memory in case of failure to init `io_uring` queues.

myamada9999: @suhasjs I've implemented freeing the ctx memory in the event that initializing the io_uring queues fails. 
Thank you for pointing this out.

PR #672, opened 2025-09-25
Distance computation: Fix SSE instructions

twuebker:
 
Currently, invalid SSE instructions like _mm128_loadu_ps are used. This causes DiskANN compilation to fail on machines that only support SSE and not AVX. Instead, it should be _mm_loadu_ps.

Change the instructions to the correct ones.

### DIFF
--- a/src/distance.cpp
+++ b/src/distance.cpp
@@ -346,10 +346,10 @@ template <typename T> float DistanceInnerProduct<T>::inner_product(const T *a, c
 #else
 #ifdef __SSE2__
 #define SSE_DOT(addr1, addr2, dest, tmp1, tmp2)                                                                        \
-    tmp1 = _mm128_loadu_ps(addr1);                                                                                     \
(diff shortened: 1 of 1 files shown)


src/distance.cpp
```diff
@@ -346,10 +346,10 @@ template <typename T> float DistanceInnerProduct<T>::inner_product(const T *a, c
 #else
 #ifdef __SSE2__
 #define SSE_DOT(addr1, addr2, dest, tmp1, tmp2)                                                                        \
-    tmp1 = _mm128_loadu_ps(addr1);                                                                                     \
-    tmp2 = _mm128_loadu_ps(addr2);                                                                                     \
-    tmp1 = _mm128_mul_ps(tmp1, tmp2);                                                                                  \
-    dest = _mm128_add_ps(dest, tmp1);
+    tmp1 = _mm_loadu_ps(addr1);                                                                                     \
```
harsha-simhadri: @copilot check that the new instructions do run on SSE and are correct equivalants

PR #699, opened 2026-01-23
Fixing flaky test_compute_relative_contrast_with_random_data

arrayka:
# What
- Use a seeded pseudo-random generator in the test

# Why
- To fix flaky test_compute_relative_contrast_with_random_data() test

Example of the failure: https://github.com/microsoft/DiskANN/actions/runs/21272293270/job/61224767878?pr=697
```
    test utils::relative_contrast::relative_contrast_tests::test_compute_relative_contrast_with_random_data ... FAILED

    failures:

    failures:
        utils::relative_contrast::relative_contrast_tests::test_compute_relative_contrast_with_random_data

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 14 filtered out; finished in 0.50s
    
  stderr ───

    thread 'utils::relative_contrast::relative_contrast_tests::test_compute_relative_contrast_with_random_data' panicked at diskann-tools/src/utils/relative_contrast.rs:192:9:
    Mean relative contrast is out of range: 1.2039199
```

### DIFF
--- a/diskann-tools/src/bin/subsample_bin.rs
+++ b/diskann-tools/src/bin/subsample_bin.rs
@@ -9,7 +9,7 @@ use std::path::PathBuf;
 use anyhow::{ensure, Result};
 use clap::Parser;
 use half::f16;
-use rand::{rngs::StdRng, SeedableRng};
+use rand::rngs::StdRng;
 use rand_distr::{Distribution, StandardUniform};
 
 use diskann::utils::VectorRepr;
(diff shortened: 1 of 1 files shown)


diskann-tools/src/bin/subsample_bin.rs
```diff
@@ -43,7 +43,7 @@ fn create_rng(seed: Option<u64>) -> StdRng {
     match seed {
         Some(seed) => StdRng::seed_from_u64(seed),
         None => {
-            let mut system_rng = rand::rng();
+            let mut system_rng = random::create_rnd();
```
hildebrandmw: It should be possible to return `random::create_rnd()` directly since it is already a `StdRng`. But this pattern is fine since it somewhat decouples the returned RNG from that in `diskann_providers::utils::random`.

arrayka: Fixed.

PR #702, opened 2026-01-23
load validation

jinwei14:


src/pq_flash_index.cpp
```diff
@@ -190,6 +190,16 @@ std::vector<bool> PQFlashIndex<T, LabelT>::read_nodes(const std::vector<uint32_t
         {
             uint32_t *node_nhood = offset_to_node_nhood(node_buf);
             auto num_nbrs = *node_nhood;
+            // Validate num_nbrs is within expected bounds
+            if (num_nbrs > _max_degree)
```
limenghao: This internal check can be removed, it is duplicate as the other one traverse all the num_nbrs and check.

PR #724, opened 2026-02-05
save and load max_record_size and leaf_page_size for bftrees

backurs:
This is a small PR, which builds upon an internal PR where we started implementing save/load functionality for bftrees. Since we recently bumped up bftree version to 0.4.7, which implements getter functions for max_record_size and leaf_page_size, we can save and load these variables when saving and load bftrees. This is necessary to do because we might need to increase max_record_size and leaf_page_size if the dimensionality of the vectors is too large

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -105,9 +105,9 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "bf-tree"
-version = "0.4.7"
+version = "0.4.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8cce2075c711f351f0aa52c05e645cc41f1b3cc0cdba1ad12c5f67c121c1bb7d"
+checksum = "dc59924517d26b63895ed60675e01a8dc27ee1d24d2b62cd6a272e7bd0af19a0"
 dependencies = [
  "cfg-if",
  "io-uring",


diskann-providers/src/model/graph/provider/async_/bf_tree/provider.rs
```diff
@@ -2036,15 +2047,21 @@ where
             .map_err(|e| ANNError::log_index_error(format!("Failed to parse metric: {}", e)))?;
 
         let vector_path = BfTreePaths::vectors_bftree(&saved_params.prefix);
-        let mut vector_config = Config::new(&vector_path, saved_params.bytes_vector);
+        let mut vector_config = Config::new(&vector_path, saved_params.params_vector.bytes);
+        vector_config.cb_max_record_size(saved_params.params_vector.max_record_size);
+        vector_config.leaf_page_size(saved_params.params_vector.leaf_page_size);
```
hildebrandmw: Would it make sense to add a method
```rust
impl BfTreeParams {
    fn apply(&self, &mut Config) {
        // do the thing
    } 
}
```
Otherwise, it's too easy to add another parameter and forget to update it everywhere it's used. Which on that subject, using `serde` opens us up to backwards compatibility issues we have yet to address, but maybe let's not worry about that part yet.

backurs: It's a good idea. I implemented function:

```
impl BfTreeParams {
    /// Apply the saved BfTree parameters to a Config.
    pub fn apply(&self, config: &mut Config) {
        config.cb_max_record_size(self.max_record_size);
        config.leaf_page_size(self.leaf_page_size);
    }
}
```
and calling it in 5 places.
Thank you for the suggestion!

hildebrandmw: Should `bytes` be part of this as well?

backurs: Good idea. I replaced `apply` function with `to_config` function:
```
impl BfTreeParams {
    /// Build a BfTree Config from the saved parameters and a file path.
    pub fn to_config(&self, path: &std::path::Path) -> Config {
        let mut config = Config::new(path, self.bytes);
        config.cb_max_record_size(self.max_record_size);
        config.leaf_page_size(self.leaf_page_size);
        config.storage_backend(bf_tree::StorageBackend::Std);
        config
    }
}
```
and I call it in the 5 places.

hildebrandmw: Ahh, I see. The bytes need to be part of the constructor for the config. That's a little annoying, but I think it's better to keep all the initialization together, so let's go with it.

PR #730, opened 2026-02-06
Experimental Multi-Vector Chamfer Distance with SIMD & BLAS Optimizations

suri-kumkaran:
## Summary

Experimental multi-vector support with fast Chamfer distance for `f32` embeddings and benchmarking infrastructure.

### Core Types

- `MultiVector` - Row-major token embeddings
- `TransposedMultiVector` - Block-transposed layout (16 vectors/block, SIMD-optimized)
- `Chamfer<Approach>` - Generic distance using Inner Product (implements `DistanceFunction`)

### Implementations

- `NaiveApproach` - Scalar baseline
- `SimdApproach` - SIMD via `diskann_vector::InnerProduct`
- `TransposedApproach` - Block-transposed SIMD
- `TransposedWithTilingApproach` - Query tiling (transposes docs, processes query pairs)
- `QueryTransposedWithTilingApproach` - Doc tiling (transposes query, processes doc pairs)
- `SgemmApproach` - BLAS SGEMM + SIMD row-max (via faer library)

## Benchmark Results (100 points, 10 iterations)

**Machine:** Intel Core i7-1365U, AVX2 supported, AVX-512 not supported, 32 GB RAM
**Note:** Times are median over 50 measurements, each measuring 10 consecutive distance computations across 100 points.

### Speedup vs SIMD Baseline (Median, Lower Latency = Better)

| Configuration | SIMD (µs) | transposed_simd | transposed_tiling | query_transposed_tiling | sgemm |
|--------------|-----------|-----------------|-------------------|------------------------|-------|
| dim=128, doc=32, query=8 | 2,237 | 1.34x | **1.75x** | 1.05x | 1.13x |
| dim=128, doc=64, query=16 | 9,224 | 1.42x | 2.07x | **2.35x** | 1.48x |
| dim=128, doc=128, query=32 | 47,882 | 1.32x | 1.86x | **2.64x** | 1.88x |
| dim=256, doc=32, query=8 | 4,654 | 1.26x | **1.69x** | 1.13x | 0.96x |
| dim=256, doc=64, query=16 | 25,809 | 1.56x | 1.94x | **2.40x** | 1.87x |
| dim=256, doc=128, query=32 | 101,093 | 1.41x | 1.71x | **2.67x** | 1.96x |
| dim=256, doc=16, query=32 | 8,239 | 1.22x | 1.77x | **2.02x** | 1.57x |
| dim=384, doc=32, query=8 | 8,412 | 1.41x | **1.65x** | 1.30x | 1.24x |
[...]

PR #731, opened 2026-02-06
RFC: Multi-Vector Distance Functions

### LINKED ISSUE #884: Add multi-vector distance kernels
## Provide references (if applicable)
If your feature request is related to a published algorithm/idea, please provide links to 
any relevant articles or webpages.

### LINKED ISSUE #985: RFC: Multi-Vector Distance Function Architecture
# RFC: Multi-Vector Distance Function Architecture

## Description
This RFC proposes the architectural design for introducing multi-vector support (e.g., ColBERT/MaxSim, Chamfer distance) into the DiskANN ecosystem. This design must balance high-performance, specialized distance evaluations with strict API compatibility and standalone usability.

## Goals
- [ ] Define the core mathematical operations required for multi-vector distances, starting with Chamfer and MaxSim.
- [ ] Maintain strict compatibility with DiskANN's `DistanceFunctionMut` trait.
- [ ] Provide a clean API that enables standalone distance function usage without requiring full index integration.
- [ ] Establish the strategy for supporting various datatypes (`f32` initially, with future `f16` and `u8` support) and subsequent quantizations.
- [ ] Outline proposed memory layouts specifically optimized to achieve a 2x+ speedup over baseline SIMD.

**Action Item:** Solicit feedback from maintainers on the `DistanceFunctionMut` implementation and standalone API boundaries before committing to implementation.

suri-kumkaran:
## Summary
RFC proposing high-performance Chamfer distance for multi-vector (ColBERT-style) representations in DiskANN.

## Changes

| File | Description |
|------|-------------|
| `rfcs/00731-multi-vector-distance-functions.md` | RFC proposing Chamfer distance for ColBERT-style multi-vector search |

## RFC Overview

Proposes a **query-transposed tiling** approach for Chamfer distance computation:

- Transposes queries into a block layout, keeps documents in row-major format (no index migration)
- Processes pairs of document vectors together, amortizing query memory loads (~50% bandwidth reduction)
- Reuses a `MaxSim` scratch buffer across distance calls (zero hot-path allocation)
- Achieves up to **2.67x speedup** over SIMD baseline (avg **1.93x** across 10 configurations)

Builds on existing types from `diskann-quantization` (`Mat`, `MatRef`, `MaxSim`, `Chamfer`) and proposes `MaxSim` implementing `DistanceFunctionMut` from `diskann-vector` for ecosystem integration.

Experimental benchmarks available in [PR #730](https://github.com/microsoft/DiskANN/pull/730).

## Review Focus

This PR is the RFC itself — please review the proposal and leave comments directly on the RFC file:

- Is query-transposed tiling the right default over document-transposed approaches?
- API shape: `MaxSim` implementing `DistanceFunctionMut` directly
- `TransposedMultiVector` type design and `Repr` trait integration
- Trade-offs between proposed and alternative approaches

### DIFF
--- a/rfcs/00731-multi-vector-distance-functions.md
+++ b/rfcs/00731-multi-vector-distance-functions.md
@@ -1,0 +1,298 @@
+# Multi-Vector Distance Functions
+
+| | |
+|---|---|
+| **Authors** | Suryansh Gupta |
+| **Contributors** | Suryansh Gupta, Mark Hildebrand |
+| **Created** | 2026-01-06 |
+| **Updated** | 2026-02-06 |
+
+## Summary
+
+This RFC proposes a high-performance Chamfer distance implementation for multi-vector (ColBERT-style late interaction) representations in DiskANN. The design uses a **query-transposed tiling** approach that transposes queries into a block layout while keeping documents in row-major format, achieving up to **2.67x speedup** over SIMD baseline. The implementation builds on existing types from `diskann-quantization` (`Mat`, `MaxSim`, `Chamfer`) and implements `DistanceFunctionMut` from `diskann-vector` for ecosystem compatibility.
+
+## Motivation
+
+### Background
+
+Traditional vector search represents each document as a single embedding. Multi-vector representations (used in models like ColBERT) encode each document/query as a **bag of embeddings** — typically one per token. This enables:
+
+- **Fine-grained matching**: Token-level similarity captures nuanced semantic relationships
+- **Late interaction**: Document embeddings are pre-computed; only lightweight aggregation at query time
+- **Better recall**: If any query token matches any document token well, the document scores high
+
+### Problem Statement
+
+Chamfer distance for multi-vector search requires O(Q × D × Dim) operations per query-document pair, where:
+
+- Q = number of query tokens
+- D = number of document tokens
+- Dim = embedding dimensionality
+
+For typical configurations (Q=32, D=128, Dim=384), this is ~1.5M floating-point operations per pair. Naive implementations become a bottleneck for large-scale search.
+
+### Goals
+
+1. Implement high-performance Chamfer distance starting with `f32` embeddings, with future support for `f16` and `u8` types
+2. Achieve 2x+ speedup over baseline SIMD through memory layout optimization
+3. Maintain compatibility with DiskANN's `DistanceFunctionMut` trait
+4. Provide a clean API that enables standalone distance function usage without full index integration
+5. Achieve performance within 10–20% of `faer` SGEMM-based Chamfer computation, when both our implementation and `faer` are restricted to AVX2 (no AVX-512 on either side)
+
+## Proposal
+
+### Approach: Query-Transposed Tiling
+
+We propose the **query-transposed tiling** approach as the primary Chamfer distance implementation for DiskANN integration. This approach transposes the query into a block-transposed layout, keeps documents in row-major format, and processes pairs of document vectors together to amortize query memory loads. A pre-allocated scratch buffer tracks per-query max similarities and is reused across distance calls.
+
+This is the recommended default because it preserves the existing document storage format (no index migration), while still achieving significant speedups through SIMD tiling.
+
+### Chamfer Distance Definition
+
+For query multi-vector Q and document multi-vector D:
+
+```
+Chamfer(Q, D) = Σᵢ minⱼ -IP(qᵢ, dⱼ)
+```
+
+Since `InnerProduct::evaluate` in `diskann-vector` returns the negated inner product (`-IP`), the kernel finds the minimum negated IP per query vector (equivalent to finding the maximum similarity), then sums across all query vectors. The result is a distance compatible with DiskANN's min-heap.
+
+### Types
+
+The design builds on the multi-vector matrix types already defined in `diskann_quantization::multi_vector`:
+
+#### Query and Document (from `diskann-quantization`)
+
+```rust
+use diskann_quantization::multi_vector::{Mat, MatRef, Standard, QueryMatRef};
+
+/// Owning row-major matrix: rows = tokens, cols = dimensions
+type MultiVector = Mat<Standard<f32>>;
+
+/// Borrowed view into a multi-vector
+type MultiVectorRef<'a> = MatRef<'a, Standard<f32>>;
+```
+
+`Standard<f32>` provides contiguous row-major storage with `as_slice()` for BLAS compatibility and zero-copy views via `MatRef`. `QueryMatRef` (a newtype over `MatRef`) distinguishes query from document matrices for asymmetric distance functions.
+
+#### TransposedMultiVector (new type)
+
+A block-transposed representation for SIMD-optimized inner product computation. The full transposed type will be added in separate PRs; the below is a simplified depiction.
+
+This will implement a `Repr` trait similar to `Standard<T>` defined in `diskann_quantization::multi_vector`, backed by a block-transposed layout analogous to `BlockTranspose<16>`:
+
+```rust
+/// Block-transposed layout for SIMD-parallel inner products.
+/// Groups 16 vectors and stores dimensions contiguously:
+///
+///   Block 0: [v0_d0..v15_d0], [v0_d1..v15_d1], ...
+///   Block 1: [v16_d0..v31_d0], ...
+///
+/// This enables loading 8 values (f32x8) from 8 different vectors
+/// in a single SIMD load.
+pub struct TransposedMultiVector { ... }
+
+impl TransposedMultiVector {
+    pub fn from_view(view: MatRef<'_, Standard<f32>>) -> Self;
+    pub fn num_vectors(&self) -> usize;
+    pub fn vector_dim(&self) -> usize;
+}
+
+impl From<&MultiVector> for TransposedMultiVector { ... }
+```
+
+#### MaxSim (from `diskann-quantization`)
+
+The existing `MaxSim` type in `diskann_quantization::multi_vector::distance` provides a mutable buffer for per-query-vector maximum similarities. The query-transposed tiling approach uses an analogous scratch pattern internally:
+
+```rust
+/// Per-query-vector maximum similarity scores.
+pub struct MaxSim<'a> {
+    scores: &'a mut [f32],
+}
+```
+
+#### Chamfer (from `diskann-quantization`)
+
+The existing unit type `Chamfer` in `diskann_quantization::multi_vector::distance` identifies the Chamfer distance function:
+
+```rust
+/// Unit type identifying the Chamfer (MaxSim) distance function.
+pub struct Chamfer;
+
+impl PureDistanceFunction for Chamfer { ... }
+```
+
+`Chamfer` implements `PureDistanceFunction` from `diskann-vector`, enabling it to be used as a distance identifier throughout the DiskANN ecosystem without carrying computation state.
+
+### Algorithm
+
+The query-transposed tiling approach:
+
+1. Transpose the query into block-transposed layout (once, before distance calls)
+2. For each pair of document vectors `(d1, d2)`:
+   - Load each query block once, compute inner products against both documents simultaneously
+   - Update per-query min negated IPs in the score array in `MaxSim` struct
+3. Handle remainder if document count is odd
+4. Sum scores to produce the final Chamfer distance
+
+```
+// `scores` corresponds to the scores array inside MaxSim
+// InnerProduct::evaluate returns -IP(q, d), so we track min (= best match)
+scores = [f32::MAX; num_queries]  // reused across calls
+
+for i in (0..num_docs step 2):
+    d1, d2 = doc[i], doc[i+1]
+    for each query block:
+        // Load query block ONCE, use for BOTH documents
+        for each query vector q:
+            scores[q] = min(scores[q], -IP(q, d1))
+            scores[q] = min(scores[q], -IP(q, d2))
+
+if num_docs is odd:
+    for each query vector q:
+        scores[q] = min(scores[q], -IP(q, last_doc))
+
+score = sum(scores[i] for i in 0..num_queries)
+```
+
+**Key optimizations**:
+
+- Query blocks loaded once, reused for 2 document vectors — reduces memory bandwidth by ~50%
+- Scratch buffer stored in `MaxSim` struct and reused across distance calls — zero allocation on hot path
+- 4 SIMD accumulators per document vector to hide FMA latency
+- Uses `SIMDMinMax` from `diskann-wide` for hardware-accelerated SIMD reductions
+
+### Trait Implementation
+
+`MaxSim` implements `DistanceFunctionMut` to populate its per-query-vector max similarities. The caller then sums the scores to produce the Chamfer distance. We propose adding a `chamfer_score()` convenience method on `MaxSim` for this:
+
+```rust
+// DistanceFunctionMut populates the MaxSim scores buffer with per-query max similarities
+impl DistanceFunctionMut<&TransposedMultiVector, MatRef<'_, Standard<f32>>, Result<(), MaxSimError>> for MaxSim<'_> {
+    fn evaluate(&mut self, query: &TransposedMultiVector, doc: MatRef<'_, Standard<f32>>) -> Result<(), MaxSimError>;
+}
+
+impl MaxSim<'_> {
+    /// Sums the per-query min negated IPs to produce the Chamfer distance.
+    /// Call after `evaluate()` has populated the scores buffer.
+    ///
+    /// Returns: Σᵢ scores[i]  (each score is already minⱼ -IP(qᵢ, dⱼ))
+    pub fn chamfer_score(&self) -> f32 {
+        self.scores.iter().sum()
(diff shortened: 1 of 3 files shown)


suri-kumkaran:

rfcs/README.md
```diff
@@ -0,0 +1,39 @@
+# DiskANN RFCs
+
+This directory contains **Requests for Comments (RFCs)** — design documents for substantial changes to DiskANN.
+
+## When to Write an RFC
+
+An RFC is required for changes that affect the public API or architecture of the project, or whenever an important design decision needs review from maintainers and stakeholders. Examples include:
```
hildebrandmw: I don't think we are not at the stage where all changes to the public API need an RFC. Obviously, big changes need more thorough consensus, but this project is still developing so rapidly that, for example, adding a variant to a public enum in an internally supporting crate that should not have external users technically changes the public API but, in my opinion, should not require an RFC. Requiring such will bog us down in bureaucracy.

rfcs/README.md
```diff
@@ -0,0 +1,39 @@
...
+| Status | Meaning |
+|--------|---------|
+| **Draft** | Initial proposal, open for discussion |
+| **InReview** | Published and open for comments from maintainers and stakeholders |
+| **Accepted** | Approved by maintainers and merged |
+| **Rejected** | Declined — the RFC documents why |
+
+## How to Submit an RFC
+
+1. Copy [0000-template.md](0000-template.md) to `NNNN-short-title.md` (use the next available number).
```
hildebrandmw: Perhaps we can use PR numbers instead of total ordering? I'm worried that ensuring all RFCs are sequentially ordered otherwise would be a giant pain.

rfcs/README.md
```diff
@@ -0,0 +1,39 @@
...
+2. Fill in all sections. Remove instructional comments.
+3. Open a pull request with the RFC file. The PR description should summarize the proposal.
+4. Discuss in the PR. Update the RFC based on feedback.
+5. Once accepted, the RFC is merged and the status is updated to **Accepted**.
```
hildebrandmw: Merging an RFC should be considered acceptance. This way, there's no need for additional book-keeping.

rfcs/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Index
+
+| # | Title | Status | Author | Resolved Date |
+|---|-------|--------|--------|---------------|
+| [0000](0000-template.md) | RFC Template | - | - | - |
+| [0001](0001-multi-vector-distance-functions.md) | Multi-Vector Distance Functions | InReview | Suryansh Gupta | - |
```
hildebrandmw: I vote to not have an index like this. Without tooling, I think it's highly unlikely that it will be updated.

rfcs/00731-multi-vector-distance-functions.md
```diff
@@ -0,0 +1,299 @@
...
+
+For typical configurations (Q=32, D=128, Dim=384), this is ~1.5M floating-point operations per pair. Naive implementations become a bottleneck for large-scale search.
+
+### Goals
+
+1. Implement high-performance Chamfer distance starting with `f32` embeddings, with future support for `f16` and `u8` types
+2. Achieve 2x+ speedup over baseline SIMD through memory layout optimization
+3. Maintain compatibility with DiskANN's `DistanceFunctionMut` trait
+4. Provide a clean API that enables standalone distance function usage without full index integration
+5. Achieve performance within 10–20% of `faer` SGEMM-based Chamfer computation, when both our implementation and `faer` are restricted to AVX2 (no AVX-512 on either side)
```
hildebrandmw, line 40: Why not use AVX-512 if available? We have support, and it could likely really benefit dense operations like this.

rfcs/00731-multi-vector-distance-functions.md
```diff
@@ -0,0 +1,299 @@
...
+}
+```
+
+The `TransposedMultiVector::from_view` conversion is O(Q × Dim) — a single pass that rearranges the query into block-transposed layout. This cost is amortized over all document comparisons for that query, making it negligible in practice (typically thousands of comparisons per query).
+
+## Trade-offs
+
+### Alternative Approaches
+
+The experimental crate explored six approaches total. The query-transposed tiling approach was selected as the proposal, but the alternatives remain available and may be better for specific workloads.
```
hildebrandmw, line 232: One thing not covered here is how this behaves under multi-threading. it's possible that single threaded tests may look fine, but if effort is not applied to restrict working set sizes to the size of the L1/L2 as much as possible, then multiple-threads could stomp on one-another's usage of L3 when running multiple computations in parallel.

rfcs/0001-multi-vector-distance-functions.md
```diff
@@ -0,0 +1,299 @@
...
+| dim=128, doc=64, query=16 | 9,224 | 1.42x | 2.07x | **2.35x** | 1.48x |
+| dim=128, doc=128, query=32 | 47,882 | 1.32x | 1.86x | **2.64x** | 1.88x |
+| dim=256, doc=32, query=8 | 4,654 | 1.26x | **1.69x** | 1.13x | 0.96x |
+| dim=256, doc=64, query=16 | 25,809 | 1.56x | 1.94x | **2.40x** | 1.87x |
+| dim=256, doc=128, query=32 | 101,093 | 1.41x | 1.71x | **2.67x** | 1.96x |
+| dim=256, doc=16, query=32 | 8,239 | 1.22x | 1.77x | **2.02x** | 1.57x |
+| dim=384, doc=32, query=8 | 8,412 | 1.41x | **1.65x** | 1.30x | 1.24x |
+| dim=384, doc=64, query=16 | 38,162 | 1.30x | 1.47x | **1.70x** | 1.66x |
+| dim=384, doc=128, query=32 | 171,431 | 1.53x | 1.94x | 2.04x | **2.16x** |
+| **Average** | — | **1.38x** | **1.79x** | **1.93x** | **1.59x** |
```
hildebrandmw: Can we also see situations where there are, say, 900+ documents?

rfcs/0001-multi-vector-distance-functions.md
```diff
@@ -0,0 +1,299 @@
...
+## Benchmark Results
+
+**Machine:** Intel Core i7-1365U, AVX2 supported, AVX-512 not supported, 32 GB RAM
+
+Median over 50 measurements. Each measurement computes 10 consecutive distance evaluations across 100 points.
+
+### Speedup vs SIMD Baseline (Median, Lower Latency = Better)
+
+| Configuration | SIMD (µs) | transposed_simd | transposed_tiling | query_transposed_tiling | sgemm |
+|--------------|-----------|-----------------|-------------------|------------------------|-------|
```
hildebrandmw: Small nit - can this table be adjusted to it's readable without rendering the markdown?

harsha-simhadri:
Let us post the RFCs we already have that have shaped code.

rfcs/00731-multi-vector-distance-functions.md
```diff
@@ -0,0 +1,299 @@
...
+use diskann_quantization::multi_vector::{Mat, MatRef, Standard, QueryMatRef};
+
+/// Owning row-major matrix: rows = tokens, cols = dimensions
+type MultiVector = Mat<Standard<f32>>;
+
+/// Borrowed view into a multi-vector
+type MultiVectorRef<'a> = MatRef<'a, Standard<f32>>;
+```
+
+`Standard<f32>` provides contiguous row-major storage with `as_slice()` for BLAS compatibility and zero-copy views via `MatRef`. `QueryMatRef` (a newtype over `MatRef`) distinguishes query from document matrices for asymmetric distance functions.
```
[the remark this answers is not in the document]
suri-kumkaran: The RFC is based on the experimental PR which adds this API.

rfcs/00731-multi-vector-distance-functions.md
```diff
@@ -0,0 +1,299 @@
...
+use diskann_quantization::multi_vector::{Mat, MatRef, Standard};
+use diskann_quantization::multi_vector::distance::MaxSim;
+use diskann_vector::DistanceFunctionMut;
+
+// 1. Query arrives as row-major multi-vector (e.g., 32 tokens × 384 dims)
+let query: Mat<Standard<f32>> = /* from encoder/upstream layers */;
+
+// 2. Transpose query into SIMD-friendly block layout (once per query)
+let transposed_query = TransposedMultiVector::from_view(query.as_ref());
+
```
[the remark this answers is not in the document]
suri-kumkaran: This is only for depiction, need not to be compiled as it is.

rfcs/00731-multi-vector-distance-functions.md
```diff
@@ -0,0 +1,299 @@
...
+// 3. Create MaxSim buffer (reused across all document comparisons for this query)
+let mut buffer = vec![f32::MAX; query.num_vectors()];
+let mut max_sim = MaxSim::new(&mut buffer).unwrap();
+
+// 4. For each candidate document:
+for doc in candidate_documents {
+    // Populate per-query-vector max similarities
+    max_sim.evaluate(&transposed_query, doc.as_ref());
+
```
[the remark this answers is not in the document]
suri-kumkaran: This is only for depiction, need not to be compiled as it is.

rfcs/00731-multi-vector-distance-functions.md
suri-kumkaran: I think I would like to park this for now. We can pick it up in a follow up with a more thorough design for dynamic dispatching that also takes the machine architecture into account.

rfcs/README.md
suri-kumkaran: I think there are two concerns with using PR numbers:
1. We would need to update file names after opening PRs, which is not a very smooth process.
2. Having a total ordering improves readability and overall cleanliness in my opinion, and we only need to ensure the ordering is correct at the time the PR is merged.

rfcs/README.md
suri-kumkaran: Are we not planning to merge rejected RFCs as well, so that we can document proposals that were discussed and ultimately rejected?

rfcs/README.md
suri-kumkaran: Make sense removed.

rfcs/README.md
suri-kumkaran: Make sense, I have relaxed the wordings.

rfcs/0001-multi-vector-distance-functions.md
suri-kumkaran: Done.

rfcs/README.md
hildebrandmw: I buy the argument on a sequential ordering, though to be fair it's just a number. But, is ensuring the ordering is correct at merge time just as much (if not more) of a process headache than updating the name once with the PR number at the beginning and then never having to worry about it again?

rfcs/README.md
hildebrandmw: We can use an RFC label to tag RFC pull requests. Then rejected RFC's will still be available as closed PRs, the final text will not bloat the repo, *and* the full discussion will be available. The latter is arguable more valuable than just the text.

rfcs/00731-multi-vector-distance-functions.md
hildebrandmw: Deferring is fine, but I don't think there need to be a super thorough design planning/review - we already have the infrastructure for dynamic dispatch and micro-architecture specific code generation.

rfcs/README.md
suri-kumkaran: I think ensuring the ordering is correct at merge time may be more of a headache than updating the PR number at the beginning, but the benefit of maintaining sequential order makes it worthwhile.

Also, since RFCs merge will not happen very frequently, we are unlikely to need to update the sequential ordering second time before merge.

rfcs/00731-multi-vector-distance-functions.md
suri-kumkaran: Got it, so what's your suggestion? Should we include the brief design around dynamic dispatch in this RFC itself?

I was thinking we can work on it in separate PRs and RFCs if required in future.

rfcs/0001-multi-vector-distance-functions.md
suri-kumkaran: I think this is a bit far from a real use case because the Chamfer distance is very costly, but I can still benchmark it if you would like. Let me know if you want me to do that.

rfcs/00731-multi-vector-distance-functions.md
suri-kumkaran: That is a great point. While the distance kernel itself is single-threaded, I agree that minimizing the working set is crucial to prevent cache thrashing when multiple threads are running parallel searches (e.g., in a Rayon thread pool).

I feel Query-Transposed Tiling approach tackles this problem fairly well:

1. Reduced Bandwidth Pressure: By holding a query block in vector registers (or L1) and iterating over multiple document vectors, we maximize arithmetic intensity. This reduces the frequency of fetches from L3/Memory compared to the baseline, which theoretically leaves more bandwidth available for other threads.

2. Small Working Set: The scratch buffer (MaxSim) and the query tiles are designed to stay hot in L1/L2. The only data streaming through L3 is the document vectors themselves, which is unavoidable.

That said, theoretical scaling often differs from practice. I can add a 'Throughput Benchmark' to the RFC (running the kernel on all available cores simultaneously) to verify that we don't see sub-linear scaling due to resource contention. Does that sound like a sufficient verification?

rfcs/0001-multi-vector-distance-functions.md
```diff
@@ -0,0 +1,299 @@
...
+impl PureDistanceFunction for Chamfer { ... }
+```
+
+`Chamfer` implements `PureDistanceFunction` from `diskann-vector`, enabling it to be used as a distance identifier throughout the DiskANN ecosystem without carrying computation state.
+
+### Algorithm
+
+The query-transposed tiling approach:
+
+1. Transpose the query into block-transposed layout (once, before distance calls)
```
arkrishn94: By block transpose, do you mean that you transpose some (contiguous) dimensions for a group of queries? 

I.e. if the matrix Q has queries in the rows, then your first block in the bwlo loop is = Q[i:j, k:l]^T?

rfcs/README.md
hailangx: can we just use the date-title format as we used in the internal repo?

rfcs/README.md
suri-kumkaran: Make sense, done.

rfcs/README.md
suri-kumkaran: I moved this to a separate PR #804

rfcs/00731-multi-vector-distance-functions.md
```diff
@@ -0,0 +1,298 @@
...
+Traditional vector search represents each document as a single embedding. Multi-vector representations (used in models like ColBERT) encode each document/query as a **bag of embeddings** — typically one per token. This enables:
+
+- **Fine-grained matching**: Token-level similarity captures nuanced semantic relationships
+- **Late interaction**: Document embeddings are pre-computed; only lightweight aggregation at query time
+- **Better recall**: If any query token matches any document token well, the document scores high
+
+### Problem Statement
+
+Chamfer distance for multi-vector search requires O(Q × D × Dim) operations per query-document pair, where:
```
arrayka, line 26: Have we checked these papers?
- [A Near-Linear Time Algorithm for the Chamfer Distance](https://arxiv.org/abs/2307.03043)
- [Even Faster Algorithm for the Chamfer Distance](https://drops.dagstuhl.de/storage/00lipics/lipics-vol334-icalp2025/LIPIcs.ICALP.2025.76/LIPIcs.ICALP.2025.76.pdf)

The second paper offers O(nd(log log n)) running time which is less than O(d n^2) of the brute force algorithm.

suri-kumkaran: I've taken a high-level look at both papers. Quick take:

Both target a different regime than our `QueryComputer` — they produce a $(1 \pm \varepsilon)$ probabilistic estimate of the **scalar** Chamfer distance between two large point clouds ($n$ in the thousands to billions).

- **Paper 1** (Bakshi et al., NeurIPS '23) covers $\ell_1$ and $\ell_2$ (via SimHash for $\ell_2$).
- **Paper 2** (Feng & Indyk, ICALP '25) is $\ell_1$-only and explicitly leaves $\ell_2$ as an open problem.
- Both share a known barrier: Paper 1's lower bound rules out sub-quadratic algorithms that **report the actual per-point assignment** — only the scalar estimate is cheap.

Our kernel is the opposite:

- Small $N_q$, $N_d$ (token-level multi-vector)
- **Exact**, deterministic
- Returns **per-query-row MaxSim** (not just a scalar)
- Constant-factor optimised: arch specific (like AVX2) micro-kernel, GotoBLAS tiling, fused max-reduce, block-transposed query

There's no large $n$ in our inner loop, so the $\log n$ / $\log \log n$ savings the papers chase don't apply.

**I don't see them as replacements.** The realistic place we could plug in is approximate Chamfer over very long documents ($N_d$ in the thousands), not corpus-level retrieval — for that, ANN over the doc-vector index is still the right tool. Happy to dig in more.

PR #732, opened 2026-02-06
Remove the `experimental_avx512` feature.

hildebrandmw:
This feature is no longer needed as AVX-512 support has officially landed. Since `diskann-vector` no longer has any features, this removes it from the `--no-default-features` as well.


suhasjs (approved):
LGTM

PR #749, opened 2026-02-10
Add API to load pre-built index in wrapped_async

Sanhaoji2:

### DIFF
--- a/diskann-providers/src/index/wrapped_async.rs
+++ b/diskann-providers/src/index/wrapped_async.rs
@@ -346,7 +346,7 @@ where
                     DiskANNIndex::<DP>::new_with_current_thread_runtime(index_config.config, dp);
                 Ok(index)
             }
-            Err(e) => return Err(e),
+            Err(e) => Err(e),
         }
     }
 }


diskann-providers/src/index/wrapped_async.rs
```diff
@@ -317,3 +326,50 @@ where
...
+        match data_provider {
+            Ok(dp) => {
+                let index =
+                    DiskANNIndex::<DP>::new_with_current_thread_runtime(index_config.config, dp);
+                Ok(index)
+            }
+            Err(e) => Err(e),
+        }
+    }
+}
```
hildebrandmw, line 352: I feel like this would be an endless game of whack-a-mole, trying to introduce sync interfaces for all the async interfaces.

What if we did this instead:

Somewhere high in the stack of `diskann-providers`, we can make a function like [`block_on`](https://github.com/microsoft/DiskANN/blob/87a0757be444af6a567cc399fdb79757451b6ab9/diskann-benchmark/src/utils/tokio.rs#L8-L13) that can run any future. Then for these situations where we have an async function/trait that doesn't **really** need a multi-threaded runtime and is not performance sensitive, `block_on` can be used and we don't have to duplicate our interfaces.

hildebrandmw (changes requested):

PR #759, opened 2026-02-11
[multi-vector] Enable cloning of `Mat` and friends.

hildebrandmw:
Enable:

1. Clone for `Mat`
2. `MatRef::to_owned` (as an inherent method)
3. `MatMut::to_owned` (as an inherent method)

All three of these require the underlying representation to implement `NewCloned` to perform the actual clone. This PR implements `NewCloned` for the `Standard` representation.

This is turning out necessary for use in multi-vector scenarios.

### DIFF
--- a/diskann-quantization/src/multi_vector/mod.rs
+++ b/diskann-quantization/src/multi_vector/mod.rs
@@ -66,6 +66,6 @@ pub(crate) mod matrix;
 
 pub use distance::{Chamfer, MaxSim, MaxSimError, QueryMatRef};
 pub use matrix::{
-    Defaulted, LayoutError, Mat, MatMut, MatRef, Overflow, Repr, ReprMut, ReprOwned, SliceError,
-    Standard,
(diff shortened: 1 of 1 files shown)


diskann-quantization/src/multi_vector/matrix.rs
```diff
@@ -242,6 +242,14 @@ pub unsafe trait NewOwned<T>: ReprOwned {
 #[derive(Debug, Clone, Copy)]
 pub struct Defaulted;
 
+/// Create a new [`Mat`] cloned from a view.
+pub trait NewCloned: ReprOwned {
```
arkrishn94, line 246: Can you export this trait from the module?

It's my bad to forget to do this in the original PR but while you're at it, can you also export the other `New*` types?

hildebrandmw: Good call, done!

PR #769, opened 2026-02-11
Implement pipelined search a la PipeANN

PhilipBAdams:
- Does this PR add any new dependencies?
  - No, but it does require a version bump for the io_uring crate to use SQPOLL. Need to test if this is worth having perf-wise
- Does this PR modify any existing APIs? Is the change to the API backwards compatible?
  - This PR adds a parameter to set the search mode, which defaults to existing behavior so is backwards compatible. We also modify the internal ExpandBeam trait.
- Should this result in any changes to our documentation, either updating existing docs or adding new ones?
  - We should consider adding documentation outlining the tradeoffs between pipelined search and standard beam search, and explanations for the pipelined search parameters

This PR is heavily inspired by PipeANN, as described in the [paper](https://www.usenix.org/conference/osdi25/presentation/guo)/[open-source implementation](https://github.com/thustorage/PipeANN). The goal of PipeANN is to process nodes as their individual IOs complete rather than waiting for a full beam, in order to overlap IO and computation and try to keep at least BW IOs in flight at all times. This slide from the PipeANN presentation compares the approach to the existing beam search. One important note is that the search becomes nondeterministic, because the node expansion pattern becomes dependent on IO completion timing. From the PipeANN slides:
image

In order to accomplish this pipelining, we split the ExpandBeam trait into two phases: `submit_expand` and `expand_available`. The old behavior can be maintained by having `submit_expand` be a no op and implementing the existing expand_beam logic in `expand_available`. But by having two phases, we are able to process individual reads from a beam as they complete rather than processing the whole beam monolithically. We then modify the implementation  of `search_internal` to manage the pipeline, submitting new IOs as old ones are processed. I think this two-phase approach should be generic enough to be useful in other search patterns, such as optimistically prefetching some nodes from disk to trade IOPS for latency.

I did benchmarking on a SIFT1M index built with R=64,L=100. In my testing, the pipelined search achieves on-par or better recall at all L, and has meaningful latency and throughput improvments. See this benchmark plot for reference, which was run on the local nvme of an Azure L48s v3 machine: 
thread_sweep

#### Any other comments?
I initially did a direct port [on this branch](https://github.com/PhilipBAdams/DiskANN/tree/users/philip/pipeann-experiment). I then tried to integrate it more closely with the existing search loop/index abstractions. In my testing, this integrated version has similar performance to the direct port. I'm very open to suggestions about how to better integrate into the existing structure -- I think the ExpandBeam modification is necessary, but ideally more code could be shared between the accessors and the parameter handling could be cleaned up. I'm also not sure if some of the pipelining logic in the accessor could be factored out to make it easier to implement pipelined accessors for other storage backends. Please let me know if you have any suggestions or requests here.

~Another thing to note is we update some of the neighbor queue semantics to allow doing adaptive beam width and relaxed monotonic early-termination. I am not fully confident these are correct, especially since adaptive beam is not giving the expected QPS bump when disk IOPS are saturated. I will do some further review, but please let me know if you prefer these optimizations to be separated into a different PR~ These optimizations have been removed and can be added in a later PR when it's clearer how to implement them cleanly

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -20,6 +20,24 @@ coverage:
         threshold: 0%
         informational: false
 
+flag_management:
+  default_rules: # the rules that will be followed for any flag added, generally
+    carryforward: true
+    statuses:
+      - type: project
+        target: auto
+        threshold: 1%
+      - type: patch
+        target: 90%
+  individual_flags: # exceptions to the default rules above, stated flag by flag
+    - name: miri
+      carryforward: true
+      statuses:
+        - type: project
+          target: 0%
+        - type: patch
+          target: 0%
+
 comment:
   layout: "reach,diff,flags,tree"
   behavior: default
--- a/.gitignore
+++ b/.gitignore
@@ -334,3 +334,6 @@ target/
 
 # ignore VS Code local history
 .history/
+
+# Ignore local settings for claude code AI agent
+.claude/settings.local.json


PhilipBAdams:
> One thing I can't help but think is that it adds considerable complexity to the core search algorithm, even in situations where this pattern is not needed. This makes me think that we need to restructure the search interface to keep it more open ended, allowing extensions like this to exist but not mandating it for simple cases. Otherwise, if we use an approach for every new feature that gets added, the complexity will easily get out of hand.

Yeah, this is something I was struggling with. One thing is I think most of the complexity is due to adjustable beam width and relaxed mono rather than the pipelining -- especially because the details of those heuristics are tied up in the search loop instead of being pluggable. I can factor them out for now and see how minimal I can make the change to the core loop.

In the future if there is a clean representation of all query state (current iteration beamwidth, neighbor queue, IOs in flight, etc) then each heuristic like early-stop or beamwidth adjustment can be a state->state function called each iteration, they will be composable

diskann-disk/src/search/pipelined/pipelined_reader.rs
```diff
@@ -0,0 +1,216 @@
...
+    pub fn wait_completions(&mut self) -> ANNResult<Vec<usize>> {
+        if self.in_flight == 0 {
+            return Ok(Vec::new());
+        }
+        self.ring.submit_and_wait(1)?;
+        self.drain_cqes()
+    }
+
+    /// Drain all available CQEs from the completion queue.
+    fn drain_cqes(&mut self) -> ANNResult<Vec<usize>> {
```
hildebrandmw: Could this take a closure to invoke on each completed slot rather than returning a `Vec`? You may be able to avoid allocations in that context.

You'll have to be careful, though, to correctly handle situations where the closure panics.

diskann-disk/src/search/pipelined/pipelined_reader.rs
```diff
@@ -0,0 +1,216 @@
...
+
+/// A pipelined IO reader that wraps `io_uring` for non-blocking submit/poll.
+///
+/// Unlike `LinuxAlignedFileReader` which uses `submit_and_wait` (blocking),
+/// this reader submits reads and polls completions independently, enabling
+/// IO/compute overlap within a single search query.
+pub struct PipelinedReader {
+    ring: IoUring,
+    /// Pre-allocated sector-aligned read buffers, one per slot.
+    slot_bufs: AlignedBoxWithSlice<u8>,
```
hildebrandmw: This almost surely needs to be in an `UnsafeCell`, and possibly the allocation needs to be behind a raw pointer as well. With IOUring writing to the other side of it, this is a violation of the exclusiveness of mutable references.

If this is behind a raw pointer, then we only materialize slices when we know that the kernel won't be mutating the backing data, which should avoid language level UB.

diskann-disk/src/search/pipelined/pipelined_reader.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// `user_data` in the CQE for later retrieval.
+    /// Submit a read for the given sector offset into the specified buffer slot.
+    ///
+    /// # Safety
+    /// The caller must ensure `slot_id` is not currently in-flight (i.e., it has
+    /// been returned by a previous completion or was never submitted). Violating
+    /// this invariant allows the kernel to DMA into a buffer being read, causing
+    /// data corruption. When using a free-list for slot management (see
+    /// `PipelinedScratch::free_slots`), this invariant is structurally guaranteed.
+    pub unsafe fn submit_read(&mut self, sector_offset: u64, slot_id: usize) -> ANNResult<()> {
```
hildebrandmw: Could we make this safer by having the reader managed free slots itself? These slots can be reclaimed in drained CQEs. My worry with forcing the caller to do this correctly is that (1) they won't and there are no safe-guards against it and (2) even if they do manage it correctly, they will almost surely not do so in a way that doesn't lose a slot ID in the event of a panic.

diskann-disk/src/search/pipelined/pipelined_reader.rs
```diff
@@ -0,0 +1,216 @@
...
+            self.in_flight = self.in_flight.saturating_sub(1);
+        }
+        Ok(completed)
+    }
+
+    /// Returns the read buffer for a completed slot.
+    pub fn get_slot_buf(&self, slot_id: usize) -> &[u8] {
+        let start = slot_id * self.slot_size;
+        &self.slot_bufs[start..start + self.slot_size]
+    }
```
hildebrandmw: I don't think this can be a safe function since there could be an inflight write for this slot.

diskann-disk/src/search/pipelined/pipelined_reader.rs
PhilipBAdams: One other option would be merging pipelined_reader into pipelined_accessor which has the slot management currently. Originally they were separate because I had a monolithic port too and wanted to compare the perf of both with the same reader implementation.

I will think about what's easier/clearer in the morning. Do you think there's likely to be another consumer for this kind of reader? I need to think if other storage backend would work correctly with pipelined_accessor

PR #772, opened 2026-02-12
Bump diskann-quantization to edition 2024.

hildebrandmw:
Bump `diskann-quantization` to Rust edition 2024 - addressing instances of the warn-by-default `unsafe-op-in-unsafe-fn` lint.

The automatically generated `flatbuffers` code is explicitly excluded from the new lint.

Note that edition 2024 changes `rustfmt` - so there are significant changes to formatting as well.

### DIFF
--- a/diskann-quantization/src/minmax/multi/meta.rs
+++ b/diskann-quantization/src/minmax/multi/meta.rs
@@ -156,7 +156,7 @@ where
     unsafe fn drop(self, ptr: NonNull<u8>) {
         let slice_ptr = std::ptr::slice_from_raw_parts_mut(ptr.as_ptr(), self.bytes());
 
-        // Safety: inherited from caller. All implementations that create a new owned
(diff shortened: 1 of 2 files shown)


diskann-quantization/Cargo.toml
```diff
@@ -5,7 +5,7 @@ description.workspace = true
 authors.workspace = true
 documentation.workspace = true
 license.workspace = true
-edition.workspace = true
+edition = "2024"
```
arkrishn94, line 8: For my undestanding - why not set this for the whole workspace?

arkrishn94 (approved):

diskann-quantization/Cargo.toml
hildebrandmw: The issue is that most of the crates still on 2021 require upgrade PRs like this. Instead of one massive PR, I think it makes sense to tackle them one-at-a-time. The end goal is to switch everything to 2024.

PR #773, opened 2026-02-12
(RFC) Refactor search interface with unified SearchDispatch trait

narendatha:
### Summary

This PR refactors the search interface in `diskann` to provide a unified entry point via the `SearchDispatch` trait, improving code organization and API clarity. This also removes (un)popular SearchParams and search specific structs. This PR also removes (un)popular SearchParamsand replaces with search specific structs, so it is clear what params are used and which ones are ignored. 

#### New Architecture

- **`SearchDispatch` trait** - A unified dispatch interface that all search types implement, enabling a single `index.search()` entry point
- **Separate search type files**:
  - `dispatch.rs` - Core trait definition
  - `graph_search.rs` - Standard k-NN graph search (`GraphSearch`, `RecordedGraphSearch`)
  - `range_search.rs` - Range-based search (`RangeSearch`, `RangeSearchOutput`)
  - `multihop_search.rs` - Label-filtered search (`MultihopSearch`)
  - `diverse_search.rs` - Diversity-aware search (`DiverseSearch`, feature-gated)

#### API Changes

| Before | After |
|--------|-------|
| `index.execute_search(..., parameters, ...)` | `index.search(..., search_params, ...)` |
| `index.search_recorded(...)` | `index.debug_search(...)` |
| Multiple search methods | Single `search()` with dispatch |

#### Code Organization

- Moved internal implementations (`range_search_internal`, `multihop_search_internal`) to their respective search type files
- Moved `NotInMutWithLabelCheck` helper to `multihop_search.rs`
- Added `From` impls for `SearchParams` -> `GraphSearch` and `RangeSearchParams` -> `RangeSearch`
- Removed convenience methods (`graph_search`, `multihop_search`, `range_search`) from `DiskANNIndex` to avoid encouraging direct use
- Added local test helper functions where needed for backwards compatibility

#### Other

- Renamed `search_recorded` -> `debug_search` to clearly signal debug-only intent
- Bumped version to 0.46.0

### Usage Example

```rust
use diskann::graph::{GraphSearch, RangeSearch, MultihopSearch};

// Standard k-NN search
let params = GraphSearch::new(10, 100, None)?;
let stats = index.search(&strategy, &context, &query, &params, &mut output).await?;

// Range search
let params = RangeSearch::new(100, 0.5)?;
let result = index.search(&strategy, &context, &query, &params, &mut ()).await?;

// Filtered search
let params = MultihopSearch::new(GraphSearch::new(10, 100, None)?, &filter);
let stats = index.search(&strategy, &context, &query, &params, &mut output).await?;
```

### Why flat_search is not included

`flat_search` remains a separate method because it has unique constraints incompatible with `SearchDispatch`: (1) it requires an additional generic bound `SearchAccessor<'a>: IdIterator<I>` that forces the strategy's accessor to support iteration over all IDs - a capability not all strategies provide, and (2) it takes an extra `vector_filter` closure parameter that other search types don't need. Since `flat_search` is fundamentally a linear scan with filtering rather than a graph traversal, keeping it as a distinct method with its own signature is cleaner than complicating the trait to accommodate it.

### Testing

- All existing tests pass
- `cargo build --workspace` 
- `cargo test -p diskann -p diskann-providers -p diskann-disk`

### DIFF
--- a/diskann/src/graph/index.rs
+++ b/diskann/src/graph/index.rs
@@ -2143,16 +2143,16 @@ where
     /// let result = index.search(params, &strategy, &context, &query, &mut ()).await?;
     /// // result.ids and result.distances contain the matches
     /// ```
-    pub fn search<'a, S, T, O: 'a, OB, P>(
-        &'a self,
+    pub fn search<S, T, O, OB, P>(
+        &self,
         search_params: P,
-        strategy: &'a S,
-        context: &'a DP::Context,
-        query: &'a T,
-        output: &'a mut OB,
-    ) -> impl SendFuture<ANNResult<P::Output>> + 'a
+        strategy: &S,
+        context: &DP::Context,
+        query: &T,
+        output: &mut OB,
+    ) -> impl SendFuture<ANNResult<P::Output>>
     where
-        P: super::search::Search<DP, S, T, O, OB> + 'a,
+        P: super::search::Search<DP, S, T, O, OB>,
         T: ?Sized,
         OB: ?Sized,
     {


diskann/src/graph/search/dispatch.rs
```diff
@@ -0,0 +1,36 @@
...
+
+/// Trait for search parameter types that execute their own search logic.
+///
+/// Each search type (graph search, range search, etc.) implements
+/// this trait to define its complete search behavior. The [`DiskANNIndex::search`]
+/// method delegates to the `dispatch` method.
+///
+/// The `dispatch` method takes `&mut self` to support search types that need to
+/// record state during execution (e.g., [`RecordedGraphSearch`] for path recording).
+pub trait SearchDispatch<DP, S, T: ?Sized, O, OB: ?Sized>
```
hildebrandmw: Should we just call this `Search` and have it live in `mod.rs`?

diskann/src/graph/search/dispatch.rs
```diff
@@ -0,0 +1,36 @@
...
+
+use crate::{ANNResult, graph::index::DiskANNIndex, provider::DataProvider};
+
+/// Trait for search parameter types that execute their own search logic.
+///
+/// Each search type (graph search, range search, etc.) implements
+/// this trait to define its complete search behavior. The [`DiskANNIndex::search`]
+/// method delegates to the `dispatch` method.
+///
+/// The `dispatch` method takes `&mut self` to support search types that need to
```
hildebrandmw: There's no need to spell this out. It can be inferred from the signature.

diskann/src/graph/search/diverse_search.rs
```diff
@@ -0,0 +1,171 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+//! Diversity-aware search (feature-gated).
```
hildebrandmw: No need to mention this is feature gated. If we take an approach similar to [quantization](https://github.com/microsoft/DiskANN/blob/e093583d89c90ab3044168e407c90d1194b67927/diskann-quantization/Cargo.toml#L63-L67), the gated-ness of this module will be automatically documented, and in a way that will not go out-of-date if this is stabilized.

diskann/src/graph/search/diverse_search.rs
```diff
@@ -0,0 +1,171 @@
...
+                .into_ann_result()?;
+
+            Ok(stats.finish(result_count as u32))
+        }
+    }
+}
+
+//=============================================================================
+// Internal Implementation
+//=============================================================================
```
hildebrandmw: Here and elsewhere, can we use the consistent bannerstyle
```rust
/////////////////////////////
// Internal Implementation //
/////////////////////////////
```
I know this is a nit-pick, but it bugs me 😢

diskann/src/graph/search/diverse_search.rs
```diff
@@ -0,0 +1,171 @@
...
+/// * `beam_width` - Optional beam width for parallel exploration
+/// * `diverse_params` - Diversity-specific parameters
+/// * `k_value` - Number of results to return
+pub(crate) fn create_diverse_scratch<DP, P>(
+    index: &DiskANNIndex<DP>,
+    l_value: usize,
+    beam_width: Option<usize>,
+    diverse_params: &DiverseSearchParams<P>,
+    k_value: usize,
+) -> SearchScratch<DP::InternalId, DiverseNeighborQueue<P>>
```
hildebrandmw: Can this be an inherent method on `DiverseSearch`?

diskann/src/graph/search/diverse_search.rs
```diff
@@ -0,0 +1,171 @@
...
+where
+    DP: DataProvider,
+    P: AttributeValueProvider<Id = DP::InternalId>,
+{
+    let attribute_provider = diverse_params.attribute_provider.clone();
+    let diverse_queue = DiverseNeighborQueue::new(
+        l_value,
+        // SAFETY: k_value is guaranteed to be non-zero by SearchParams validation by caller
+        #[allow(clippy::expect_used)]
```
hildebrandmw: When allowing lints - prefer to use `expect` rather than `allow`. The former will result in an error if the condition is not met.

diskann/src/graph/search/diverse_search.rs
```diff
@@ -0,0 +1,171 @@
...
+        NonZeroUsize::new(k_value).expect("k_value must be non-zero"),
```
hildebrandmw: This is not necessarily communicated up the call stack. We can fix this by requiring it in `GraphSearch` probably.

diskann/src/graph/search/graph_search.rs
hildebrandmw, line 1: Suggestion: Maybe we can call this `KNN` since it's the canonical k-nearest-neighbors search?

diskann/src/graph/search/graph_search.rs
```diff
@@ -0,0 +1,279 @@
...
+    provider::{BuildQueryComputer, DataProvider},
+    utils::IntoUsize,
+};
+
+/// Parameters for standard graph-based ANN search.
+///
+/// This is the primary search mode, using the Vamana graph structure for efficient
+/// approximate nearest neighbor traversal.
+#[derive(Debug, Clone, Copy)]
+pub struct GraphSearch {
```
hildebrandmw: This is a big complaint I have with `SearchParams` that (1) needs to get fixed there and (2) should not be repeated here.

There is an invariant that must be upheld between `k` and `l`. While this is checked in the constructor - making these fields public renders that check useless. These fields should be private, `NonZeroUsize`, and the conversion from `SearchParams` (which has the same issue) should be `TryFrom` instead of `From`.

We can then use accessor methods to retrieve the `k` and `l` value without needing to make them public.

diskann/src/graph/search/graph_search.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// Number of results to return (k in k-NN).
+    pub k: usize,
+    /// Search list size - controls accuracy vs speed tradeoff.
+    pub l: usize,
+    /// Optional beam width for parallel graph exploration.
+    pub beam_width: Option<usize>,
+}
+
+impl GraphSearch {
```
hildebrandmw: Can this replace `SearchParams` entirely?

diskann/src/graph/search/graph_search.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// Create new graph search parameters.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if `l < k` or if any value is zero.
+    pub fn new(
+        k: usize,
+        l: usize,
+        beam_width: Option<usize>,
+    ) -> Result<Self, super::super::SearchParamsError> {
```
hildebrandmw: Small nit: I'm all for fully qualifying paths, but `super::super` is a little much. Since `crate::graph` is already used for some imports, why not add this to that import?

diskann/src/graph/search/graph_search.rs
```diff
@@ -0,0 +1,279 @@
...
+        Self {
+            k: params.k_value,
+            l: params.l_value,
+            beam_width: params.beam_width,
+        }
+    }
+}
+
+/// Implement SearchDispatch for SearchParams to provide backwards compatibility.
+/// This treats SearchParams as an alias for GraphSearch.
```
hildebrandmw: I'm all for throwing away `SearchParams` given my above stated issues. In that context, there's no need for comments about backwards compatibility.

diskann/src/graph/search/graph_search.rs
```diff
@@ -0,0 +1,279 @@
...
+        glue::{SearchExt, SearchPostProcess, SearchStrategy},
+        index::{DiskANNIndex, SearchStats},
+        search::record::NoopSearchRecord,
+        search_output_buffer::SearchOutputBuffer,
+    },
+    provider::{BuildQueryComputer, DataProvider},
+    utils::IntoUsize,
+};
+
+/// Parameters for standard graph-based ANN search.
```
hildebrandmw: The search function `fn search` on `DiskANNIndex` has a decent amount of documentation that is not copied over here. Can we include that (and make sure the documentation is accurate and concise)?

diskann/src/graph/search/graph_search.rs
```diff
@@ -0,0 +1,279 @@
...
+
+/// Graph search with traversal path recording.
+///
+/// Records the path taken during search for debugging or analysis.
+#[derive(Debug)]
+pub struct RecordedGraphSearch<'r, SR: ?Sized> {
+    /// Base graph search parameters.
+    pub inner: GraphSearch,
+    /// The recorder to capture search path.
+    pub recorder: &'r mut SR,
```
hildebrandmw: I like how this layers. This is nice!

diskann/src/graph/search/multihop_search.rs
```diff
@@ -0,0 +1,310 @@
...
+        strategy: &'a S,
+        context: &'a DP::Context,
+        query: &'a T,
+        output: &'a mut OB,
+    ) -> impl SendFuture<ANNResult<Self::Output>> {
+        let params = SearchParams {
+            k_value: self.inner.k,
+            l_value: self.inner.l,
+            beam_width: self.inner.beam_width,
+        };
```
hildebrandmw: There's a little more of `SearchParams` being awkward. We're kind of playing hot-potato between it an `GraphSearch`. Strongly recommend we pick one (the latter) and eat the cost of updating everything to use it.

diskann/src/graph/search/range_search.rs
```diff
@@ -0,0 +1,371 @@
...
+    #[allow(clippy::too_many_arguments)]
+    pub fn with_options(
+        max_returned: Option<usize>,
+        starting_l: usize,
+        beam_width: Option<usize>,
+        radius: f32,
+        inner_radius: Option<f32>,
+        initial_slack: f32,
+        range_slack: f32,
+    ) -> Result<Self, super::super::RangeSearchParamsError> {
```
hildebrandmw: My same comments about `SearchParams` can be repeated for this struct.

diskann/src/graph/index.rs
```diff
@@ -2363,48 +2115,75 @@ where
...
-    /// * `output` - A mutable buffer to store the search results. Must be pre-allocated by the caller.
-    /// * `search_record` - A mutable reference to a search record object that will record the path taken during the search.
+    /// - [`search::GraphSearch`]: Standard graph-based ANN search
+    /// - [`search::MultihopSearch`]: Label-filtered search with multi-hop expansion
+    /// - [`search::RangeSearch`]: Range-based search within a distance radius
+    /// - [`search::DiverseSearch`]: Diversity-aware search (feature-gated)
     ///
-    /// # Returns
+    /// For flat (brute-force) search, use [`Self::flat_search`] directly due to its
+    /// unique iterator type constraints.
```
hildebrandmw: I'd recommend avoiding mention of `flat_search`.

diskann/src/graph/index.rs
```diff
@@ -2363,48 +2115,75 @@ where
...
-    ) -> impl SendFuture<ANNResult<SearchStats>>
+    pub fn debug_search<'a, S, T, O, OB, SR>(
+        &'a self,
+        strategy: &'a S,
+        context: &'a DP::Context,
+        query: &'a T,
+        search_params: &'a SearchParams,
+        output: &'a mut OB,
+        search_record: &'a mut SR,
+    ) -> impl SendFuture<ANNResult<SearchStats>> + 'a
```
hildebrandmw: Can we redirect it to use the `RecordedSearch` parameter and get rid of this method?

Cargo.toml
```diff
@@ -46,22 +46,22 @@ undocumented_unsafe_blocks = "warn"
 
 [workspace.dependencies]
 # Base And Numerics
-diskann-wide = { path = "diskann-wide", version = "0.45.0" }
-diskann-vector = { path = "diskann-vector", version = "0.45.0" }
-diskann-linalg = { path = "diskann-linalg", version = "0.45.0" }
-diskann-utils = { path = "diskann-utils", default-features = false, version = "0.45.0" }
-diskann-quantization = { path = "diskann-quantization", default-features = false, version = "0.45.0" }
-diskann-platform = { path = "diskann-platform", version = "0.45.0" }
+diskann-wide = { path = "diskann-wide", version = "0.46.0" }
```
hildebrandmw: Let's save on bumping the versions in PRs until we figure out our release process a bit better.

diskann-providers/src/index/diskann_async.rs
```diff
@@ -215,6 +215,57 @@ pub(crate) mod tests {
...
     fn no_modify(_: &mut diskann::graph::config::Builder) {}
 
+    //////////////////////////
+    // Test helper functions //
+    //////////////////////////
+
+    use diskann::graph::index::SearchStats;
+
+    /// Test helper: performs multihop search using the dispatch API.
+    async fn multihop_search<DP, S, T, O, OB>(
```
hildebrandmw: I'd recommend updating these to use the new API. If it turns out the new API is less ergonomic than the bespoke methods, this would be a good time to find out :smile:

hildebrandmw (commented):
Thanks @narendatha! I think this does a good job of tying together the disparate APIs we've gathered. I'm pretty strongly in favor of getting rid of the legacy `SearchParams` and `RangeSearchParams` to avoid clutter, and updating the structs defined in this PR to not expose all the methods as public and therefore render the constructor invariant checks invalid.

diskann/src/graph/search/diverse_search.rs
narendatha: Good to follow conventions, done.

diskann/src/graph/search/diverse_search.rs
narendatha: fixed

diskann/src/graph/index.rs
```diff
@@ -2363,136 +2115,47 @@ where
...
     ///
-    /// * `strategy` - The search strategy to use for accessing and processing elements.
-    /// * `context` - The context to pass through to providers.
-    /// * `query` - The query vector for which nearest neighbors are sought.
-    /// * `search_params` - Parameters controlling the search behavior, such as search depth (`l_value`) and beam width.
-    /// * `output` - A mutable buffer to store the search results. Must be pre-allocated by the caller.
+    /// - [`search::KnnSearch`]: Standard k-NN graph-based search
+    /// - [`search::MultihopSearch`]: Label-filtered search with multi-hop expansion
+    /// - [`search::RangeSearch`]: Range-based search within a distance radius
+    /// - [`search::DiverseSearch`]: Diversity-aware search (feature-gated)
```
hildebrandmw: Quick naming/export observation, I'd like to hear your thoughts. The links here scope the search parameters to the `search` module, but they are also re-exported through `graph`.

There *is* a world where we structure it like
```
diskann::graph::search::KNN;
diskann::graph::search::MultiHop;
diskann::graph::search::Range;
diskann::graph::search::Diverse;
```
That is, we can drop the `Search` suffixes, use the module for namespacing, and do not reexport through `graph` (the `Search` trait itself should probably be reexported though).

What do you think?

diskann/src/graph/index.rs
```diff
@@ -2363,136 +2115,47 @@ where
...
-        query: &T,
-        search_params: &SearchParams,
-        output: &mut OB,
-    ) -> impl SendFuture<ANNResult<SearchStats>>
+    /// // Range search (note: uses () as output buffer, results in Output type)
+    /// let mut params = RangeSearch::new(100, 0.5)?;
+    /// let result = index.search(&strategy, &context, &query, &mut params, &mut ()).await?;
+    /// // result.ids and result.distances contain the matches
+    /// ```
+    pub fn search<'a, S, T, O: 'a, OB, P>(
```
hildebrandmw: IMO, `search_params` should probably come first now since it's the most important argument at this level for determining what kind of search is about to get run.

diskann/src/graph/index.rs
```diff
@@ -2363,136 +2115,47 @@ where
...
+        &'a self,
+        strategy: &'a S,
+        context: &'a DP::Context,
+        query: &'a T,
+        search_params: &'a mut P,
```
hildebrandmw: Taking `search_params` by mutable reference is a bit unfortunate for the caller. I know why we did this, but it has an unfortunate side-effect of making it harder to work with in cases where mutability is not needed.

To that end, what if we do this:

1. Take `search_params` by value.
2. Implement `Search` for both `KnnSearch` and `&KnnSearch` - making sure that `KnnSearch` by value simply delegates to the `&KnnSearch` implementation.

It does make generic plumbing when mutability is required more tricky (i.e., we'd need HRTB all over the place), but I think that might be preferable to always forcing mutability?

diskann/src/graph/search/mod.rs
```diff
@@ -3,5 +3,90 @@
...
+    ///
+    /// # Returns
+    ///
+    /// Returns `Self::Output` which varies by search type (e.g., [`SearchStats`](super::index::SearchStats)
+    /// for k-NN, [`RangeSearchOutput`] for range search).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if there is a failure accessing elements or computing distances.
+    fn dispatch<'a>(
```
hildebrandmw: You can drop the explicit lifetime `'a` here. Additionally, when you drop the lifetime, you will be able to get rid of the explicit lifetime in `DiskANNIndex::search` as well.

hildebrandmw: Small preference for calling this method `search`, following the convention of traits having a single method named after themselves.

diskann/src/graph/search/knn_search.rs
```diff
@@ -0,0 +1,362 @@
...
+/// - `beam_width`: Optional parallel exploration width
+///
+/// # Example
+///
+/// ```ignore
+/// use diskann::graph::{search::KnnSearch, Search};
+///
+/// let mut params = KnnSearch::new(10, 100, None)?;
+/// let stats = index.search(&strategy, &context, &query, &mut params, &mut output).await?;
+/// ```
```
hildebrandmw: I'd say that 90% of the time, I'm a fan of attaching documentation to the trait implementations and allowing that to be discovered while perusing rustdoc.

However, this feels like a situation where the algorithmic level documentation would be *more* discoverable if attached to the `KnnSearch` struct (and other searchers) themselves **or** as module-level documentation. The trait-impl can then include a cross-reference to the struct level documentation.


Basically - these structs exist almost exclusively to implement `Search`. So struct-level documentation feels appropriate.

Thinking of how this would look for users of the docs, I think it's cleaner. What do you think?

diskann/src/graph/search/knn_search.rs
```diff
@@ -0,0 +1,362 @@
...
+#[derive(Debug, Error)]
+pub enum KnnSearchError {
+    #[error("l_value ({l_value}) cannot be less than k_value ({k_value})")]
+    LLessThanK { l_value: usize, k_value: usize },
+    #[error("beam width cannot be zero")]
+    BeamWidthZero,
+}
+
+impl From<KnnSearchError> for ANNError {
+    fn from(err: KnnSearchError) -> Self {
```
hildebrandmw, line 42: Recommend attaching `#[track_caller]` to all conversions to `ANNError` as it will improve the quality of the error message if backtraces are not collected.

diskann/src/graph/search/knn_search.rs
```diff
@@ -0,0 +1,362 @@
...
+/// This is the primary search mode, using the Vamana graph structure for efficient
+/// approximate nearest neighbor traversal.
+#[derive(Debug, Clone, Copy)]
+pub struct KnnSearch {
+    /// Number of results to return (k in k-NN).
+    k_value: NonZeroUsize,
+    /// Search list size - controls accuracy vs speed tradeoff.
+    l_value: NonZeroUsize,
+    /// Optional beam width for parallel graph exploration.
+    beam_width: Option<usize>,
```
hildebrandmw: Since `beam_width` cannot be zero, we can embed it in an `Option<NonZeroUsize>` to formalize the guarantee.

diskann-benchmark/src/backend/index/search/knn.rs
```diff
@@ -49,8 +49,9 @@ pub(crate) fn run<I>(
                 .search_l
                 .iter()
                 .map(|search_l| {
-                    let search_params =
-                        diskann::graph::SearchParams::new(run.search_n, *search_l, None).unwrap();
+                    let k = NonZeroUsize::new(run.search_n).expect("search_n must be non-zero");
+                    let l = NonZeroUsize::new(*search_l).expect("search_l must be non-zero");
+                    let search_params = diskann::graph::KnnSearch::new(k, l, None).unwrap();
```
hildebrandmw: One small thing I noticed: since `KnnSearch::new` already checks the `k <= l` invariant, we could also check for non-zero there and simplify the caller's life.

This is especially relevant because I'm seeing `expect` being applied to the `NonZeroUsize` constructors where-as before those paths wouldn't necessarily panic.

hildebrandmw (changes requested):
Thanks @narendatha, I'm really liking how this is coming together!

diskann/src/graph/index.rs
narendatha: As always, you amaze me with your keep observation. Here it turned out mut was not needed at all. I added it when trying to handle few scenarios and flat_search, but now that it is not needed, this was a missed left over and you caught it. Fixed this, trivial fix.

diskann-benchmark-core/src/search/graph/knn.rs
```diff
@@ -310,7 +313,7 @@ mod tests {
         let rt = crate::tokio::runtime(2).unwrap();
         let results = search::search(
             knn.clone(),
-            graph::SearchParams::new(nearest_neighbors, 10, None).unwrap(),
+            graph::search::Knn::new(nearest_neighbors.get(), 10, None).unwrap(),
```
hildebrandmw: Small nit: I think we can revert the `NonZeroUsize` change to `nearest_neighnbors` now that `Knn::new()` does the check.

diskann-benchmark-core/src/search/graph/multihop.rs
```diff
@@ -19,11 +19,11 @@ use crate::search::{self, Search, graph::Strategy};
 ///
 /// This is intended to be used in conjunction with [`search::search`] or [`search::search_all`]
 /// and provides some basic additional metrics for the latter. Result aggregation for
-/// [`search::search_all`] is provided by the [`search::graph::knn::Aggregator`] type (same
-/// aggregator as [`search::graph::KNN`]).
+/// [`search::search_all`] is provided by the [`search::graph::search::Knn::Aggregator`] type (same
+/// aggregator as [`search::graph::search::Knn`]).
```
hildebrandmw: I think these changes may have gotten grouped in with a larger find + replace. The original paths are still the intended target of the intra-doc link.

You can use
```
cargo rustdoc --package diskann-benchmark-core --features bigann -- -D rustdoc::broken-intra-doc-links
```
to help find the intra-doc links that may have gotten broken.

diskann-benchmark-core/src/search/graph/multihop.rs
```diff
@@ -152,7 +154,7 @@ mod tests {
 
     #[test]
     fn test_multihop() {
-        let nearest_neighbors = 5;
+        let nearest_neighbors = NonZeroUsize::new(5).unwrap();
```
hildebrandmw: Similar observation that this change could probably be reverted.

diskann-benchmark/src/backend/index/search/knn.rs
```diff
@@ -49,8 +49,9 @@ pub(crate) fn run<I>(
                 .search_l
                 .iter()
                 .map(|search_l| {
-                    let search_params =
-                        diskann::graph::SearchParams::new(run.search_n, *search_l, None).unwrap();
+                    let k = run.search_n;
+                    let l = *search_l;
+                    let search_params = diskann::graph::search::Knn::new(k, l, None).unwrap();
```
hildebrandmw: This can likely be reverted now that they no longer need to be `NonZeroUsize`.

diskann-benchmark/src/backend/index/search/range.rs
```diff
@@ -7,20 +7,20 @@ use std::{num::NonZeroUsize, sync::Arc};
 
 use diskann_benchmark_core::{self as benchmark_core, search as core_search};
 
-use crate::{backend::index::result::RangeSearchResults, inputs::async_::GraphRangeSearch};
+use crate::{backend::index::result::RangeResults, inputs::async_::GraphRange};
```
hildebrandmw: Not necessarily a blocker - but these internal name changes might not be needed. There is potentially still some value in keeping the name "Search" around as the results since they are not grouped into a `search` module.

Double checking that this change was intentional :smile:

diskann-benchmark/src/backend/index/result.rs
```diff
@@ -249,11 +249,11 @@ impl std::fmt::Display for DisplayWrapper<'_, [SearchResults]> {
 }
 
 ////////////////////////
-// RangeSearchResults //
+// RangeResults //
 ////////////////////////
```
hildebrandmw, line 253: Very tiny formatting nit here.

diskann-providers/src/index/diskann_async.rs
```diff
@@ -215,6 +215,8 @@ pub(crate) mod tests {
     // Callbacks for use with `simplified_builder`.
     fn no_modify(_: &mut diskann::graph::config::Builder) {}
 
+    //////////////////////////
+    // Test helper functions //
```
hildebrandmw: Is this a stray change?

diskann-tools/src/utils/search_index_utils.rs
```diff
@@ -27,7 +27,7 @@ pub struct TruthSetWithAssociatedData<Data: GraphDataType> {
     pub index_dimension: usize,
 }
 
-pub struct RangeSearchTruthSet {
+pub struct RangeTruthSet {
```
hildebrandmw: This struct suffered a similar rename that I'm not entirely sure is intentional. Was it intentional, or did it get caught in a larger automated rename?

diskann/src/graph/search/knn_search.rs
```diff
@@ -0,0 +1,342 @@
...
+/// let stats = index.search(params, &strategy, &context, &query, &mut output).await?;
+/// ```
+#[derive(Debug, Clone, Copy)]
+pub struct Knn {
+    /// Number of results to return (k in k-NN).
+    k_value: NonZeroUsize,
+    /// Search list size - controls accuracy vs speed tradeoff.
+    l_value: NonZeroUsize,
+    /// Optional beam width for parallel graph exploration.
+    beam_width: Option<NonZeroUsize>,
```
hildebrandmw: A very minor observation: I've seen several situations where defaulting the `beam_width` to 1 has to be done manually since this is an Option.

One approach we could take is to accept `Option<NonZeroUsize>` as the argument type for `beam_width` (this is more ergonomic than accepting `NonZeroUsize` since `NonZeroUsize::new` automatically returns an `Option`).

Then, the `Knn::new()` could apply the zero default if needed, which would avoid the unnecessary application of default logic in places.

The big question is whether or not a `None` value for beam-width could ever be used to supply a default other than 1. If not, then I vote we make this change while we're already going through an API break.

diskann/src/graph/search/knn_search.rs
```diff
@@ -0,0 +1,342 @@
...
+        k_value: usize,
+        l_value: usize,
+        beam_width: Option<usize>,
+    ) -> Result<Self, KnnSearchError> {
+        if k_value == 0 {
+            return Err(KnnSearchError::KZero);
+        }
+        if l_value == 0 {
+            return Err(KnnSearchError::LZero);
+        }
```
hildebrandmw: We can avoid `unsafe` in the constructor by structurally doing
```rust
let k_value = NonZeroUsize::new(k_value).ok_or(KnnSearchError::KZero)?;
```
Not that the current approach is incorrect, but we might as well fuse the non-zero construction with the check and avoid some `unsafe` maintenance while we're at it.

diskann/src/graph/index.rs
```diff
@@ -2363,136 +2116,47 @@ where
...
+    /// // result.ids and result.distances contain the matches
+    /// ```
+    pub fn search<'a, S, T, O: 'a, OB, P>(
+        &'a self,
+        search_params: P,
+        strategy: &'a S,
+        context: &'a DP::Context,
+        query: &'a T,
+        output: &'a mut OB,
+    ) -> impl SendFuture<ANNResult<P::Output>> + 'a
```
hildebrandmw: Now that the lifetimes have been removed from the `Search` trait - I'm fairly sure we can remove the `'a`` lifetime here.

hildebrandmw (commented):
Thanks Naren - I took one more pass and found a few kind of small things. Please have a look.

One thing I would suggest is highlighting the API breaking change for `SearchParams` and `RangeSearchParams` in the PR description to serve as an upgrade reference when consuming the new version.

diskann-benchmark-core/src/search/graph/multihop.rs
narendatha: Good catch.

diskann-benchmark/src/backend/index/search/range.rs
narendatha: Thanks for your depth again. This is in fact unintentional as part of find/replace. though I thought I carefully checked the changes, I missed it. I found other places such change is not good. So I have reverted those and they have "search" in their names as it was before, while keeping original intended changes alive.

diskann/src/graph/search/knn_search.rs
narendatha: I removed the option.  This simplies code, so, applied this suggestion.

hildebrandmw:
I don't expect regressions as part of this PR since it's shuffling around things high in the stack. My usual benchmark machine is being moved at the moment, but when it's back online I will run some sanity checks.

@narendatha - for future reference, running tests in the manner above does not run any benchmarks, nor does it add anything that isn't already run as part of the CI process. For now, `diskann-benchmark` should be used for A/B testing on non-trivial datasets.

PR #774, opened 2026-02-13
[quantization] Add `as_raw_ptr`.

hildebrandmw:
Add publicly exported `as_raw_ptr` for retrieving the base pointer of `Mat`, `MatRef`, and `MatMut`.

The term "raw" is meant to denote that this is the byte pointer - not necessarily the pointer to the actual element type stored in these containers.

### DIFF
--- a/diskann-quantization/src/multi_vector/matrix.rs
+++ b/diskann-quantization/src/multi_vector/matrix.rs
@@ -1466,6 +1466,10 @@ mod tests {
                     check_mat_ref(mat.reborrow(), repr, ctx);
                     check_mat_mut(mat.reborrow_mut(), repr, ctx);
                     check_rows(mat.rows(), repr, ctx);
+
+                    // Check reborrow preserves pointers.
+                    assert_eq!(mat.as_raw_ptr(), mat.reborrow().as_raw_ptr());
+                    assert_eq!(mat.as_raw_ptr(), mat.reborrow_mut().as_raw_ptr());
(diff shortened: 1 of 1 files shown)


diskann-quantization/src/multi_vector/matrix.rs
```diff
@@ -655,9 +655,9 @@ impl<T: ReprOwned> Mat<T> {
         Self { ptr, repr }
     }
 
-    #[cfg(test)]
-    fn as_ptr(&self) -> NonNull<u8> {
-        self.ptr
+    /// Return the base pointer for the [`Mat`].
+    pub fn as_raw_ptr(&self) -> *const u8 {
+        self.ptr.as_ptr()
     }
```
arkrishn94, line 661: I'm curious - why expose these publicly?

arkrishn94: Isn't this a footgun for unsafe reads?

hildebrandmw: I'm doing some work setting up FFIs. It's exceedingly convenient to actually be able to construct a `Mat` and then decompose it into a pointer and lengths for testing. Raw pointer reads are already unsafe and there is no other way of getting access to the base pointer at the moment. Exposing it via `as_raw_ptr` is infinitely safer than forcing something like `mat.get_row(0).unwrap().as_ptr()`.

PR #786, opened 2026-02-19
Adding GraphParams to be able to save graph parameters of index to SavedParams

backurs:
This PR addresses the following issue:

We want to save alpha, l_build, backedge_ratio and vector_dtype somewhere and the best place to do it (in my opinion) is SavedParams.
For that we need to save GraphParams in BfTreeProviderParameters and in BfTreeProvider. This is what this PR does.

### DIFF
--- a/diskann-providers/src/model/graph/provider/async_/bf_tree/provider.rs
+++ b/diskann-providers/src/model/graph/provider/async_/bf_tree/provider.rs
@@ -1838,19 +1838,26 @@ pub enum VectorDtype {
     I8,
 }
 
-impl VectorDtype {
-    /// Derive the `VectorDtype` from a concrete [`VectorRepr`] type parameter.
-    #[allow(clippy::panic)]
-    pub fn from_type<T: VectorRepr>() -> Self {
-        let name = std::any::type_name::<T>();
-        match name {
-            "f32" => Self::F32,
-            "half::f16" | "f16" => Self::F16,
-            "i8" => Self::I8,
-            "u8" => Self::U8,
-            _ => panic!("unsupported VectorRepr type: {}", name),
-        }
-    }
+/// A trait for mapping concrete vector element types to their [`VectorDtype`]
+/// discriminant at compile time.
+pub trait AsVectorDtype {
+    const DATA_TYPE: VectorDtype;
+}
+
+impl AsVectorDtype for f32 {
+    const DATA_TYPE: VectorDtype = VectorDtype::F32;
+}
+
+impl AsVectorDtype for half::f16 {
+    const DATA_TYPE: VectorDtype = VectorDtype::F16;
+}
+
+impl AsVectorDtype for i8 {
+    const DATA_TYPE: VectorDtype = VectorDtype::I8;
+}
+
+impl AsVectorDtype for u8 {
+    const DATA_TYPE: VectorDtype = VectorDtype::U8;
 }
 
 /// Graph configuration parameters persisted alongside the index.
--- a/diskann-providers/src/model/graph/provider/async_/bf_tree/mod.rs
+++ b/diskann-providers/src/model/graph/provider/async_/bf_tree/mod.rs
@@ -10,8 +10,8 @@ mod vector_provider;
 
 // Accessors
 pub use provider::{
-    BfTreePaths, BfTreeProvider, BfTreeProviderParameters, CreateQuantProvider, FullAccessor,
-    GraphParams, Hidden, Index, QuantAccessor, QuantIndex, StartPoint, VectorDtype,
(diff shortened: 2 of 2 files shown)


diskann-providers/src/model/graph/provider/async_/bf_tree/provider.rs
```diff
@@ -1814,6 +1825,17 @@ pub struct SavedParams {
...
+}
+
+/// Graph configuration parameters persisted alongside the index.
+/// These are needed to reconstruct the `DiskANNIndex` config on load.
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct GraphParams {
+    pub l_build: usize,
+    pub alpha: f32,
+    pub backedge_ratio: f32,
+    pub vector_dtype: String,
```
hildebrandmw: Is this the data type as in `f32`/`f16` etc.? It's probably better for it to be an enum instead of a raw string.

diskann-providers/src/model/graph/provider/async_/bf_tree/provider.rs
```diff
@@ -1814,6 +1825,17 @@ pub struct SavedParams {
     pub params_vector: BfTreeParams,
     pub params_neighbor: BfTreeParams,
     pub quant_params: Option<QuantParams>,
+    pub graph_params: Option<GraphParams>,
+}
+
+/// Graph configuration parameters persisted alongside the index.
+/// These are needed to reconstruct the `DiskANNIndex` config on load.
+#[derive(Serialize, Deserialize, Clone, Debug)]
```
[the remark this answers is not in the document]
backurs: We need it because `BfTreeProviderParameters` has `Debug` derive and it has a field `pub graph_params: Option<GraphParams>,`. Thus, it needs `Debug` derive as well.

diskann-providers/src/model/graph/provider/async_/bf_tree/provider.rs
backurs: Good point. I changed the line to `pub vector_dtype: VectorDtype,` and added code
```
/// The element type of the full-precision vectors stored in the index.
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
#[serde(rename_all = "lowercase")]
pub enum VectorDtype {
    F32,
    F16,
    U8,
    I8,
}

impl VectorDtype {
    /// Derive the `VectorDtype` from a concrete [`VectorRepr`] type parameter.
    #[allow(clippy::panic)]
    pub fn from_type<T: VectorRepr>() -> Self {
        let name = std::any::type_name::<T>();
        match name {
            "f32" => Self::F32,
            "half::f16" | "f16" => Self::F16,
            "i8" => Self::I8,
            "u8" => Self::U8,
            _ => panic!("unsupported VectorRepr type: {}", name),
        }
    }
}
```

diskann-providers/src/model/graph/provider/async_/bf_tree/provider.rs
```diff
@@ -1814,6 +1825,17 @@ pub struct SavedParams {
...
+pub struct GraphParams {
+    pub l_build: usize,
+    pub alpha: f32,
+    pub backedge_ratio: f32,
```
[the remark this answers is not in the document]
backurs: I added comments to explain the values.

diskann-providers/src/model/graph/provider/async_/bf_tree/provider.rs
hildebrandmw: mmmm, this isn't the best approach. First `std::any::type_name` is not a stable representation and can change at any time. But more-so, almost any use of `std::any::type_name` or `std::any::TypeId` in patterns like this is almost always a bad idea. Nothing is preventing passing something like `MinMax` to this type and you only discover that it doesn't work at runtime.

Instead, I'd suggest something like
```rust
trait AsVectorDtype {
    const DATA_TYPE: VectorDtype;
}
```
and use that trait. Misuses will be caught at compile time, it's guaranteed not to break, and is much friendlier on the compiler to optimize.

backurs: Thank you for the suggestion and for describing the alternative! I added the following code:

```
pub trait AsVectorDtype {
    const DATA_TYPE: VectorDtype;
}

impl AsVectorDtype for f32 {
    const DATA_TYPE: VectorDtype = VectorDtype::F32;
}

impl AsVectorDtype for half::f16 {
    const DATA_TYPE: VectorDtype = VectorDtype::F16;
}

impl AsVectorDtype for i8 {
    const DATA_TYPE: VectorDtype = VectorDtype::I8;
}

impl AsVectorDtype for u8 {
    const DATA_TYPE: VectorDtype = VectorDtype::U8;
}
```

Is this what you meant by your comment?

PR #789, opened 2026-02-20
git ignore: Ignore local settings for claude code AI agent

arrayka:
# Why
- To prevent checkins of local preferences for Claude code CLI


metajack:
I don't have a super strong opinion, but in general things like this belong in your ~/.gitignore not in the repo one.

PR #793, opened 2026-02-22
Expand BetaFilter for External SearchStrategy Implementations

hailangx:
- [ y] Does this PR have a descriptive title that could go in our release notes?
- [n ] Does this PR add any new dependencies?
- [y ] Does this PR modify any existing APIs?
- [ y] Is the change to the API backwards compatible?
- [n ] Should this result in any changes to our documentation, either updating existing docs or adding new ones?

The extra O generic is needed because the original 2-parameter form     
  SearchStrategy<Provider, T> defaults O to Provider::InternalId, which is   u32. That means BetaFilter could only produce search results as u32      (internal IDs).                                                         
                                                                          
  In diskann-garnet, the unfiltered path uses FullPrecision, which        
  implements SearchStrategy<Provider, T, O> for any O — so it can output
  GarnetId directly into SearchResults<GarnetId>. But BetaFilter was      
  locked to O = u32, forcing diskann-garnet to use the
  FilteredSearchResults adapter to convert u32 → GarnetId after the       
  search.

  By adding O to BetaFilter's impl:

  impl<Provider, Strategy, T, I, O> SearchStrategy<Provider, T, O> for    
  BetaFilter<Strategy, I>
  where
      O: Send,
      Strategy: SearchStrategy<Provider, T, O>,

  BetaFilter becomes transparent to the output type — it delegates O to   
  whatever the inner strategy uses. If the inner strategy (FullPrecision) 
  can output GarnetId, then BetaFilter wrapping it can too. This
  eliminates the need for FilteredSearchResults entirely.

  The O: Send bound is required because SearchStrategy needs O: Send for  
  the search output buffer to work across async boundaries.

#### Any other comments?

Removed previous solution to reimport the beta's internal:
Why these types need to be public:

  - BetaFilter's blanket impl (SearchStrategy<Provider, T>) hard-codes the output to the provider's internal ID     
  (u32). Consumers with a different external ID type (like GarnetId) can't use BetaFilter directly with their output
   buffer — they're forced to write an adapter (FilteredSearchResults) that accepts u32 and manually converts.      
  - If BetaAccessor/BetaComputer/Unwrap are re-exported, consumers can write:
  impl<T: VectorRepr> SearchStrategy<GarnetProvider<T>, [T], GarnetId>
      for BetaFilter<FullPrecision, u32>
  - naming the associated types directly, and the conversion     
  happens in the post-processor — no wrapper needed.

### DIFF
--- a/diskann-providers/src/model/graph/provider/layers/mod.rs
+++ b/diskann-providers/src/model/graph/provider/layers/mod.rs
@@ -11,4 +11,4 @@
 //! * [`BetaFilter`]
 
 mod betafilter;
-pub use betafilter::{BetaAccessor, BetaComputer, BetaFilter, Unwrap as BetaFilterUnwrap};
+pub use betafilter::BetaFilter;
--- a/diskann-providers/src/model/graph/provider/layers/betafilter.rs
+++ b/diskann-providers/src/model/graph/provider/layers/betafilter.rs
@@ -112,12 +112,13 @@ where
 ///
 /// The [`BetaComputer`] then uses this ID to consult the filter predicate and adjust the
 /// distance accordingly.
-impl<Provider, Strategy, T, I> SearchStrategy<Provider, T> for BetaFilter<Strategy, I>
+impl<Provider, Strategy, T, I, O> SearchStrategy<Provider, T, O> for BetaFilter<Strategy, I>
 where
     T: ?Sized,
     I: VectorId,
+    O: Send,
     Provider: DataProvider<InternalId = I>,
-    Strategy: SearchStrategy<Provider, T>,
+    Strategy: SearchStrategy<Provider, T, O>,
 {
     /// An accessor that returns the ID in addition to the element yielded by the inner
     /// accessor.


hildebrandmw:
Quick question: would it make sense to update `BetaFilter` to be friendlier to what you are trying to do? I also think there is potential with Naren's planned post-process changes that we may be able to shim a custom post-process layer to do the id translation.

hailangx:
> Quick question: would it make sense to update `BetaFilter` to be friendlier to what you are trying to do? I also think there is potential with Naren's planned post-process changes that we may be able to shim a custom post-process layer to do the id translation.
do you mean adding the 3-param impl so BetaFilter is transparent to the  external ID type?

diskann-providers/src/model/graph/provider/layers/mod.rs
```diff
@@ -11,4 +11,4 @@
 //! * [`BetaFilter`]
 
 mod betafilter;
-pub use betafilter::BetaFilter;
+pub use betafilter::{BetaAccessor, BetaComputer, BetaFilter, Unwrap as BetaFilterUnwrap};
```
hildebrandmw: I wouldn't be opposed to just renaming `Unwrap` to `BetaFilterUnwrap` to avoid the rename at the re-export level (`Unwrap` is entirely internal at the moment, so we can rename it).

Another option is to make the `betafilter` module public and simplify the names so these get accessed like
```rust
betafilter::Strategy;
betafilter::Accessor;
betafilter::Computer;
betafilter::Unwrap;
```
If we are going to prefix `BetaFilter` anyways, we *could* drop the prefix entirely and instead use the module name as the prefix.

Up to you (though I would recommend in either case making `Unwrap`'s naming consistent`).

hildebrandmw (approved):

PR #798, opened 2026-02-24
[quantization] 8bit distance kernels and ZipUnzip

arkrishn94:
This PR introduces heterogeneous inner-product kernels for 8-bit bitslices; specifically with 4-bit, 2-bit and 1-bit bitslices. 

The goal is to enable fast kernels for full-precision like queries with quantized vectors (spherical, minmax etc.). In the benchmark, we see the `u8xu4` kernel is ~2x faster than its `f32xu4` counterpart.

For AVX2 capable architectures, the 4-bit and 2-bit kernels are implemented using the [`_mm256_maddubs_epi16`](https://doc.rust-lang.org/beta/core/arch/x86_64/fn._mm256_maddubs_epi16.html) intrinsic acting on blocks of 32 byte-sized dimensions for the `u8xu4` kernel and 64 dimensions for the `u8xu2` kernel. Some care needed to be taken to make sure that for these  specific kernels, the intrinsic doesn't saturate when doing the madds. For the 1-bit kernel, we implement a simple masked horizontal add strategy on blocks of size 32. 

`Scalar` fallback is implemented for `Neon` and for now `V4` architecture gets retargeted to `V3` for these kernels.

Support to compute `u8xu4`, `u8xu2` and `u8xu1` distances with minmax quantized vectors is available mostly out of the box.

## ZipUnzip 

A new trait `ZipUnzip` has been added to diskann-wide to implement vectorized zipping and unzipping logic - the zipping merges two halved vectors into a full vector by interleaving elements from each half vector, and, the unzipping performs the inverse transformation on the full vector. 
- It's currently implemented for `i8x32`, `i16x16`, `i32x8`, `u8x32`, `u32x8` and `f16x16`. 
- It's implemented for `Scalar`, `V3`, `V4` and `Neon` architectures. 

# Benchmark
We ran the benchmark as a flat scan of vectors, making sure to clear the cache at every run and on a count that exceeds the L3 cache size for the machine.
```
Total latency in ms, COUNT=150K, AMD EPYC 7763

Kernel             dim=256    dim=384    dim=896     
─────────────────────────────────────────────────────────────
u8×u4  (new)         8.92       9.93      17.71     
u8×u2  (new)         9.87      13.10      21.53    
u8×u1  (new)         6.09       7.52      14.62     

f32×u4              15.96      20.99      40.67
f32×u2              13.88      18.46      35.64
f32×u1              13.53      17.93      34.65

u8×u8                8.40      10.00      19.72
u4×u4                7.33      11.58      15.49

f32×f32             16.56      23.16      47.76
```


diskann-quantization/Cargo.toml
```diff
@@ -31,6 +31,7 @@ uninlined_format_args = "allow"
 cfg-if = { workspace = true }
 trybuild = "1.0.101"
 rand_distr = { workspace = true }
+criterion = { workspace = true }
```
arkrishn94: need to remove.

diskann-wide/src/arch/x86_64/v3/i16x8_.rs
```diff
@@ -74,6 +74,24 @@ impl SIMDMulAdd for i16x8 {
...
+    ///
+    /// `_mm_srli_epi16` requires SSE2, which is implied by `V3`.
+    #[inline(always)]
+    fn shr_const<const N: i32>(self) -> Self {
+        // SAFETY: `_mm_srli_epi16` requires SSE2, implied by V3.
+        Self::from_underlying(self.arch(), unsafe {
+            _mm_srli_epi16::<N>(self.to_underlying())
+        })
+    }
+}
```
hildebrandmw: I don't think this is needed. LLVM is able to constant propagate through the emulated fallback to emit the desired instruction.

diskann-wide/src/arch/x86_64/algorithms.rs
```diff
@@ -162,3 +165,47 @@ pub(crate) unsafe fn __load_first_u16_of_16_bytes(
         }
     }
 }
+
+/// Unpack sub-byte fields from a `u8x16` register into a `u8x32` by
```
hildebrandmw: A general question: should this belong in `diskann-wide`? This does not use intrinsic shenanigans, just normal functionality. I guess the reinterpret-chain to hit the 16-bit shift is a bit non-obvious. But I feel this is very specific and can probably live just fine in `diskann-quantization`.

diskann-wide/src/arch/x86_64/algorithms.rs
```diff
@@ -162,3 +165,47 @@ pub(crate) unsafe fn __load_first_u16_of_16_bytes(
...
+/// ```
+/// [0,     1,     2,     3,     4,     5,     6,     7]
+/// |----------------------||------------||------------|
+///             G                 Hi             Lo
+///```
+/// This operation will result in a u8x32 vector with 32 (Lo, Hi) byte-pairs.
+/// For `N = 4` this will unpack the half-bytes into a u8x32 byte vector.
+#[inline(always)]
+pub fn unpack_half_bytes<const N: i32>(arch: V3, input: u8x16) -> u8x32 {
+    assert!(N <= 4);
```
hildebrandmw: Something to add to your toolbox: this should be
```rust
const { assert!(N <= 4) };
```
This forces misuse to be a compile-time error rather than a runtime error. Otherwise, code will compile just fine and only fail when you actually try to run this bit of code.

diskann-wide/src/arch/x86_64/algorithms.rs
```diff
@@ -162,3 +165,47 @@ pub(crate) unsafe fn __load_first_u16_of_16_bytes(
...
+    // Step 1: Shift each 16-bit lane right by N. This positions Hi in the low
+    // N bits of each byte. Bit leakage across the byte boundary within a
+    // 16-bit lane is harmless — the final mask cleans it up.
+    //
+    // input:   [B0,       B1,       B2,       ..., B15      ]
+    // shifted: [B0 >> N,  B1 >> N,  B2 >> N,  ..., B15 >> N ]
+    let shifted: u8x16 = input.reinterpret_simd().shr_const::<N>().reinterpret_simd();
+
```
hildebrandmw: The reinterpret targets should be explicit. Currently, this works because there is only one implementation of `ReinterpretSIMD` for `u8x16`. But this can and will change in the future. At which point, this code will stop inferring.

diskann-wide/src/traits.rs
```diff
@@ -722,6 +722,25 @@ impl<T> SIMDUnsigned for T where
...
+/// Given vectors `a = [a0, a1, ..., a(N-1)]` and `b = [b0, b1, ..., b(N-1)]`,
+/// interleaving produces two half-results:
+///
+/// - **lo**: `[a0, b0, a1, b1, ..., a(N/2-1), b(N/2-1)]`
+/// - **hi**: `[a(N/2), b(N/2), ..., a(N-1), b(N-1)]`
+///
+pub trait Interleave: SIMDVector {
+    /// Interleave elements of `self` and `other`, returning the low and high halves.
+    fn interleave(self, other: Self) -> crate::LoHi<Self>;
+}
```
hildebrandmw, line 770: In the Neon world, this is called Zip/Unzip. Like the `SplitJoin` trait, I think it makes sense to support both operation within a single trait. It's also not completely clear to me that `LoHi` is the right return type. I can see an argument for making it so - but we should then probably update the docs for `LoHi` and implement zipping etc. for arrays (like the `array_splitjoin` macro).

diskann-quantization/src/bits/distances.rs
```diff
@@ -2667,4 +3034,329 @@ mod tests {
...
+    // Heterogeneous: USlice<8> × USlice<M> //
+    ///////////////////////////////////////////
+
+    /// Generate a test helper and test suite for heterogeneous 8×M inner products.
+    ///
+    /// The `$M` parameter is the bit width of the RHS vector (2 or 4).
+    /// `$max_val` is `(1 << M) - 1` — the maximum value a single element can hold.
+    /// `$block_size` is the SIMD block size used by the V3 kernel (32 for 4-bit,
+    /// 64 for 2-bit).
+    macro_rules! heterogeneous_ip_tests_8xM {
```
hildebrandmw, line 3350: Is there a way to do this that doesn't involve putting all the business logic in a giant macro? I ask because large macros are tedious: `rustfmt` generally won't run on macro bodies, which makes them annoying to maintain and compile errors get repeated for each instantiation of the macro.

The other macros used in the tests do not implement business logic - they just do high level plumbing.

diskann-quantization/src/bits/distances.rs
```diff
@@ -2667,4 +3034,329 @@ mod tests {
...
+                    let dist_8bit = Uniform::new_inclusive(0i64, 255i64).unwrap();
+                    let dist_mbit = Uniform::new_inclusive(0i64, $max_val as i64).unwrap();
+
+                    for dim in 0..dim_max {
+                        for trial in 0..trials_per_dim {
+                            Case::new(dim, |_| {
+                                (dist_8bit.sample(&mut *rng), dist_mbit.sample(&mut *rng))
+                            })
+                            .check_with(
+                                &format!(
```
hildebrandmw: You can use `diskann_utils::lazy_format!` instead to defer string formatting until an error is actually hit.

diskann-quantization/src/bits/distances.rs
```diff
@@ -1318,33 +1345,373 @@ macro_rules! impl_fallback_ip {
...
+
+            // Main loop: 4× unrolled, processing 128 elements per iteration.
+            while i + 4 <= blocks {
+                // Block 0
+                // SAFETY: `i + 4 <= blocks` guarantees at least 4 full blocks remain.
+                // Each block needs 32 bytes from `px` and 16 bytes from `py`.
+                let vx = unsafe { u8s_32::load_simd(arch, px.add(32 * i)) };
+                // SAFETY: `i + 4 <= blocks` guarantees 16 bytes readable at `py.add(16 * i)`.
+                let vy = unsafe { u8s_16::load_simd(arch, py.add(16 * i)) };
+                s0 = s0.dot_simd(products(vx, unpack_half_bytes::<4>(arch, vy)), ones);
```
hildebrandmw: You probably want to avoid `dot_simd` in this loop - particularly if one of the arguments is `ones`. This instruction has a relatively high latency, which is a bummer.

Instead, for the unrolled portion, you can accumulate in `i16s`. You can do this I think exactly 4 times without overflowing. Then you can have a single `dot_simd` at the end.

diskann-wide/src/arch/x86_64/mod.rs
```diff
@@ -8,7 +8,7 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use super::{Scalar, Target, Target1, Target2, Target3};
 
 // Internal helpers
-mod algorithms;
+pub mod algorithms;
```
hildebrandmw: Can we send this back to being private?

diskann-quantization/src/bits/distances.rs
```diff
@@ -122,42 +125,41 @@ type USlice<'a, const N: usize, Perm = Dense> = BitSlice<'a, N, Unsigned, Perm>;
 /// * [`diskann_wide::arch::x86_64::V4`] -> [`diskann_wide::arch::x86_64::V3`]
 /// * [`diskann_wide::arch::aarch64::Neon`] -> [`diskann_wide::arch::Scalar`]
 macro_rules! retarget {
-    ($arch:path, $op:ty, $N:literal) => {
+    ($arch:path, $op:ty, ($N:literal, $M:literal)) => {
```
hildebrandmw, line 129: If you add the following match first:
```rust
    ($arch:path, $op:ty, $N:literal) => {
        retarget!($arch, $op, ($N, $N));
    };
```
Then the existing single-argument literals continue to work and the macro invocation with the pair (A, B) is only needed when the bit widths are truly different.

diskann-quantization/src/bits/distances.rs
arkrishn94: Done.

PR #800, opened 2026-02-25
Import diskann-garnet and vectorset

metajack:
This imports the `diskann-garnet` provider and a benchmarking utility for Garnet vector set workloads, `vectorset`, to the repo.

`diskann-garnet`: This is a `DataProvider` implementation for the [Garnet cache service](https://github.com/microsoft/garnet). It currently supports u8 and f32 full precision vector set using cosine distance. Quantization and other features will be added soon.

`vectorset`: is a benchmarking tool for Redis vector sets. Since Garnet speaks the Redis protocol, this allows for testing both Garnet and Redis vector workloads.

Both of these were previously developed in their own repos in their early stages.

### DIFF
--- a/diskann-garnet/src/lib.rs
+++ b/diskann-garnet/src/lib.rs
@@ -17,7 +17,10 @@ use std::{
 use diskann::{
     graph::{
         SearchOutputBuffer,
-        config::{self, defaults::{FILTER_BETA, GRAPH_SLACK_FACTOR}},
+        config::{
+            self,
+            defaults::{FILTER_BETA, GRAPH_SLACK_FACTOR},
+        },
         search,
     },
     utils::VectorRepr,


vectorset/src/loader.rs
```diff
@@ -0,0 +1,336 @@
...
+                            "path {} dimension mismatch: expected {} but got {}",
+                            p.display(),
+                            dim,
+                            d
+                        ));
+                    }
+                }
+            } else {
+                dim = headerless_with_dim.unwrap();
+                num_vectors += file.metadata().await?.len() as usize / dim;
```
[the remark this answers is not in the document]
metajack: I will fix this in a followup PR. Trying to keep this one to just a strict import.

vectorset/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "vectorset"
+version = "1.0.0"
+edition = "2024"
```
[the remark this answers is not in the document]
metajack: It certainly is valid.

diskann-garnet/src/lib.rs
```diff
@@ -0,0 +1,714 @@
+use std::{
```
harsha-simhadri, line 6: Some doc comments on the design of the crate could be useful. You might want to cover points such as:
- term design (what kind of terms exist, what is co-located and what is not)
- FFI design
- id mapping overview
- commentary on Redis API vector instructions and its mapping to this library.
- how to set up a diskann+garnet env and run tests/benchmarks (please ignore if you think it is obvious)

metajack: The FFI design docs are there. I will resurrect the term documentation PR on this repo once this lands and will add some new documentation about setting up Garnet with this. I'm trying to keep this PR strictly to importing code, and addressing changes to the code itself in followups.

harsha-simhadri: ok to post these follow ups as issues and proceed. 
What should be the scope of this review, in your opinion?

metajack: All of diskann-garnet has previously been reviewed either by Kevin or Mark. vectorset could use a light review as it has not been reviewed before.

So the main feedback I'm looking for here is whether the tree layout and such are acceptable and if there are any issues people notice I will handle ones that require modifying the code in followups.

metajack: https://github.com/microsoft/DiskANN/issues/802

vectorset/src/loader.rs
```diff
@@ -0,0 +1,336 @@
...
+    for i in 0..num_queries {
+        let start = i * num_neighbors;
+        let end = start + num_neighbors;
+        map.insert(i as u32, id_dists[start..end].to_vec());
+    }
+
+    Ok(Arc::new(map))
+}
+
+#[allow(dead_code)]
```
harsha-simhadri: do we need these subroutines which are already in the repo, and also marked as dead code?

vectorset/src/loader.rs
```diff
@@ -0,0 +1,336 @@
...
+                }
+
+                let elements_left = buf.len() / mem::size_of::<T>();
+                if !buf.is_empty() && !elements_left.is_multiple_of(self.dim) {
+                    return Err(anyhow!("unexpected EOF"));
+                }
+
+                count = self.batch_size - elements_left / self.dim;
+            }
+
```
harsha-simhadri, line 91: I suppose you'd have to fmt, remove dead code and other lints before checking in.

vectorset/src/loader.rs
```diff
@@ -0,0 +1,336 @@
...
+    file: Mutex<(File, Vec<PathBuf>, usize)>,
+    paths: Vec<PathBuf>,
+    num_vectors: usize,
+    dim: usize,
+    batch_size: usize,
+    headerless: bool,
+    type_: PhantomData<T>,
+}
+
+impl<T: bytemuck::Pod> DatasetLoader<T> {
```
harsha-simhadri, line 28: could we use load_bin directly or make this a wrapper around load_bin

vectorset/src/loader.rs
```diff
@@ -0,0 +1,336 @@
...
+        for _ in 0..num_vectors {
+            let mut v = vec![T::zeroed(); dim];
+            file.read_exact(bytemuck::cast_slice_mut(&mut v)).await?;
+            vectors.push(v);
+        }
+        Ok(Arc::new(vectors))
+    }
+}
+
+pub async fn load_groundtruth<P: AsRef<Path>>(
```
harsha-simhadri: Same here -- can we reuse code in repo?

diskann-garnet/src/alloc.rs
```diff
@@ -0,0 +1,25 @@
...
+
+    #[inline]
+    unsafe fn deallocate(&self, ptr: NonNull<[u8]>, layout: std::alloc::Layout) {
+        // Lint: The given `layout` **should** be the same as that passed to `allocate`,
+        // which must have succeeded for the pointer to be valid in the first place.
+        #[allow(clippy::expect_used)]
+        let layout = layout.align_to(8).expect("invalid layout provided");
+        unsafe { GlobalAllocator.deallocate(ptr, layout) }
+    }
+}
```
harsha-simhadri, line 30: could we add unit tests for all these new files?

vectorset/src/loader.rs
metajack: There are marked dead because we removed start points in diskann-garnet after testing them out. But you are right that we have a loader and latin hypercube in the library already. I'll create a followup task to remove this file and use the diskann infra.

metajack: https://github.com/microsoft/DiskANN/issues/803

diskann-garnet/diskann-garnet.nuspec
```diff
@@ -0,0 +1,24 @@
...
+<!-- Version History
+    | 1.0.0 | Sep 2025 | Initial version |
+-->
+<package>
+  <metadata>
+    <id>diskann-garnet</id>
+    <version>1.0.25</version>
+    <readme>docs/README.md</readme>
+    <authors>DiskANN Team</authors>
+    <owners>Encyclopedia@service.microsoft.com</owners>
```
harsha-simhadri: these internal details should be removed from github

diskann-garnet/src/garnet.rs
```diff
@@ -0,0 +1,504 @@
...
+
+#[derive(Copy, Clone)]
+pub struct Callbacks {
+    read_callback: ReadCallback,
+    write_callback: WriteCallback,
+    delete_callback: DeleteCallback,
+    rmw_callback: ReadModifyWriteCallback,
+}
+
+// TODO: read_iid, read_eid, read_wid,
```
harsha-simhadri: Is this still a todo?

diskann-garnet/src/dyn_index.rs
```diff
@@ -0,0 +1,216 @@
...
+    labels::GarnetQueryLabelProvider,
+    provider::{self, GarnetProvider},
+};
+
+/// Wraps `SearchResults` (FFI output buffer) with a provider and context so that
+/// `SearchOutputBuffer<u32>` can be implemented: each internal ID (u32) is converted
+/// to a `GarnetId` via `to_external_id` and written into the underlying buffer.
+/// Start point (internal ID 0) is skipped since it has no external ID mapping.
+///
+/// TODO: Once diskann-providers >= 0.45.0 exports BetaAccessor/BetaComputer/Unwrap,
```
hildebrandmw: Can we track this TODO? I think it's possible to move the ID translation out of the output buffer now.

diskann-garnet/src/dyn_index.rs
```diff
@@ -0,0 +1,216 @@
...
+    fn insert(&self, context: &Context, id: &GarnetId, data: &[u8]) -> ANNResult<()>;
+
+    fn set_attributes(&self, context: &Context, id: &GarnetId, data: &[u8]) -> ANNResult<()>;
+
+    fn search_vector(
+        &self,
+        context: &Context,
+        data: &[u8],
+        params: &search::Knn,
+        filter: Option<(&GarnetQueryLabelProvider, f32)>,
```
hildebrandmw, line 36: Small suggestion for a follow-up: `params` can be an enum instead of `filter` being an `Option`. This makes it impossible 
accidentally ignore.

diskann-garnet/src/dyn_index.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    fn search_element(
+        &self,
+        context: &Context,
+        id: &GarnetId,
+        params: &search::Knn,
+        filter: Option<(&GarnetQueryLabelProvider, f32)>,
+        output: &mut SearchResults<'_>,
+    ) -> ANNResult<SearchStats> {
```
hildebrandmw, line 104: How often is this method called? We could make `FullAccessor::get_element` a synchronous inherent method and delegate the trait `get_element` method to the sync-implementation.

This would avoid the async machinery.

diskann-garnet/src/garnet.rs
```diff
@@ -0,0 +1,504 @@
...
+
+/// Bitmask for extracting the Term bits from a Context.
+/// Must have enough bits to represent all Term variants (max value is 6, needs 3 bits).
+pub const TERM_BITMASK: u64 = (1 << 3) - 1;
+
+#[derive(Debug)]
+pub enum Term {
+    Vector = 0,
+    Neighbors = 1,
+    #[allow(dead_code)]
```
hildebrandmw: Small suggestion: using `#[expect(dead_code)]` will let you know once this variant is actually used to assist in future cleanup.

diskann-garnet/src/garnet.rs
```diff
@@ -0,0 +1,504 @@
...
+    }
+
+    #[must_use]
+    pub fn delete_eid(&self, ctx: Context, id: &GarnetId) -> bool {
+        let id: &[u8] = id;
+        unsafe { (self.delete_callback)(ctx.0, id.as_ptr(), id.len()) }
+    }
+
+    #[must_use]
+    pub fn rmw_iid<'a, F, T>(&self, ctx: Context, id: u32, write_len: usize, mut f: F) -> bool
```
hildebrandmw, line 331: I assume that panics in `f` would be really bad, right? Should we note that `f` should not panic? And maybe add it to the safety requirement?

diskann-garnet/src/labels.rs
```diff
@@ -0,0 +1,230 @@
+//! Garnet label filtering for DiskANN vector search.
+//!
+//! This module provides [`GarnetQueryLabelProvider`], an implementation of DiskANN's
+//! [`QueryLabelProvider`] trait that checks vector IDs against a pre-computed
+//! bitmap received from Garnet's C# `FilterBitmap`.
+//!
+//! ## Design
+//!
+//! Unlike CDB-DiskANN which calls back to C++ via `AreDocumentsIncluded()`,
```
hildebrandmw: Comparison with CDB doesn't mean much in the open-source code. I'd suggest leaving this as just a description of the design.

diskann-garnet/src/labels.rs
```diff
@@ -0,0 +1,230 @@
...
+///
+/// # Safety
+///
+/// The bitmap pointer must remain valid and unmodified for the lifetime of
+/// this struct. In practice, the struct is created and dropped within a
+/// single FFI search call.
+pub struct GarnetQueryLabelProvider {
+    data: *const u8,
+    len: usize,
+}
```
hildebrandmw, line 42: It's a shame that we can't actually have a lifetime to make this requirement explicit. I think it would be relatively easy to allow the `QueryLabelProvider` impl to have a lifetime (and add the `'static` requirement just at the use point).

diskann-garnet/src/labels.rs
```diff
@@ -0,0 +1,230 @@
...
+
+// Safety: the bitmap data is read-only and the caller guarantees the pointer
+// is valid for the duration of the search call.
+unsafe impl Send for GarnetQueryLabelProvider {}
+unsafe impl Sync for GarnetQueryLabelProvider {}
+
+impl std::fmt::Debug for GarnetQueryLabelProvider {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("GarnetQueryLabelProvider")
```
hildebrandmw: Why not `derive(Debug)` and print the pointer value?

diskann-garnet/src/labels.rs
```diff
@@ -0,0 +1,230 @@
...
+                data: bytes.as_ptr(),
+                len: bytes.len(),
+            }
+        }
+    }
+
+    /// Check if the given internal ID has its bit set in the bitmap.
+    /// Empty bitmap (len == 0) matches all IDs (no filter).
+    /// If the bitmap is smaller than the ID being checked, it's considered a partial bitmap and out-of-range IDs are treated as matching (not filtered out)
+    #[inline(always)]
```
hildebrandmw, line 100: Is the reason that out-of-range IDs return `true` so that we can handle situations where new entries are added while the search is running? In that case, would it make sense to return `false` instead to be conservative?

diskann-garnet/src/dyn_index.rs
hailangx: yes, this can be removed since we are now in the diskann repo , I can merge it into this branch or after this merged https://github.com/microsoft/DiskANN/pull/808

diskann-garnet/src/labels.rs
hailangx: resolve in https://github.com/microsoft/DiskANN/pull/808

diskann-garnet/src/labels.rs
hailangx: the bitmap could be lossy as we might not passing the entire bitmap, for that case we won't know if that node should be filter out or not, so we should treat those as candidate and can be filter out in later stage.

diskann-garnet/src/dyn_index.rs
hildebrandmw: Whatever you think is best. If there is a desire to keep this more of a code import rather than fusing changes, that makes sense to me.

diskann-garnet/src/dyn_index.rs
metajack: It's only called when using VSIM with ELEMENT, which is not something I can think of any real use for, but is a thing documented in the protocol. I lump it in with general debugging related functionality like VLINKS and VEMB (which return the adjacency list for a vector and the reconstructed full vector respectively). It does not need to be fast unless we get some data that says otherwise from real users.

diskann-garnet/src/garnet.rs
metajack: Fixed.

diskann-garnet/src/garnet.rs
metajack: Fixed.

diskann-garnet/src/labels.rs
metajack: Fixed.

diskann-garnet/src/garnet.rs
metajack: Fixed.

diskann-garnet/diskann-garnet.nuspec
metajack: Fixed.

diskann-garnet/src/alloc.rs
metajack: I added tests for alloc.rs. Most of the others have unit tests, and the rest of hte unit tests are actually in Garnet itself.

vectorset/src/loader.rs
metajack: Removed the dead code and the support for headerless files.

vectorset/src/loader.rs
metajack: I could use load_bin I suppose, but it's almost as much code and requires I depend on diskann for a single <10 line function.

Also, load_bin and VectorDataIterator don't really amortize allocations and don't keep track of ids along the way. I did an attempt at using them, and it ended up being pretty messy and I lost the ability to read vectors in batches from disk and use the iterator from multiple threads easily.

I will keep this for now, and if we improve the data loading elsewhere we can unify things then.

diskann-garnet/src/dyn_index.rs
metajack: @hailangx Is this resolved now?

diskann-garnet/src/labels.rs
metajack: @hailangx ?

diskann-garnet/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+mod fsm;
+mod garnet;
+mod labels;
+mod provider;
+#[cfg(test)]
+mod test_utils;
+
+/// Default beta factor for BetaFilter filtered search.
+/// Values < 1.0 bias the search toward vectors matching the filter bitmap.
+const DEFAULT_FILTER_BETA: f32 = 0.5;
```
harsha-simhadri: Could this and any other algo specific defaults have been imported from `diskann` crate?

diskann-garnet/src/dyn_index.rs
hailangx: this is resolved

diskann-garnet/docs/data-design.md
```diff
@@ -0,0 +1,177 @@
...
+
+Attributes are ignored during normal searches but will be accessed during a filtered search. The Vector Set API allows users to request the attributes for search results, so even in a normal search these terms may be accessed in order to return the attributes in the results.
+
+### Attributes Index
+
+*Key*: Attribute name + Attribute value as bytes
+
+When vectors are inserted by the Redis Vector Set API, and arbitrary JSON blob of attributes can be attached.
+
+Attributes index should be created for fields with atrribute filtering need, a list of Internal ID are saved under the Attributes Index Key, if the attribute are present in the associated JSON for the Internal ID
```
harsha-simhadri, line 87: do we create index for each unique key seen, or do we expect user to declare the keys for which inverted index must be built

diskann-garnet/src/fsm.rs
```diff
@@ -0,0 +1,504 @@
...
+                .read_single_wid(ctx.term(Term::Metadata), block_key, &mut block)
+            {
+                break;
+            }
+
+            let mut id = block_id * BLOCK_SIZE_IDS as u32 + BLOCK_SIZE_IDS as u32 - 1;
+
+            for &byte in block.iter().rev() {
+                for bidx in (0..8).rev() {
+                    let used = (byte >> (7 - bidx)) & 0x1 == 0x1;
```
harsha-simhadri: could this bit arithmetic and others in this file be a separate function. It would be nice to have one function that interprets the bits for all FSM functions (as opposed to logic spread around the file).

diskann-garnet/src/fsm.rs
```diff
@@ -0,0 +1,504 @@
...
+            }
+        } else if block_id > *max_block + 1 {
+            return Err(FsmError::IdOutOfRange(id));
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
```
harsha-simhadri, line 383: What is the intended thread safety of FSM, and how can we test it

diskann-garnet/src/provider.rs
```diff
@@ -0,0 +1,884 @@
...
+            neighbor_cache,
+            fsm,
+        })
+    }
+
+    pub fn maybe_set_start_point(
+        &self,
+        context: &Context,
+        point: &[T],
+    ) -> Result<(), GarnetProviderError> {
```
harsha-simhadri, line 152: shall we add an issue to support multiple start points?

diskann-garnet/src/labels.rs
hildebrandmw: I think this question still stands?

diskann-garnet/docs/data-design.md
metajack: I think this is a reasonable question, but is not really pertinent to this PR review. I'm not even sure if Haiyang has a plan yet on that topic to write down here.

diskann-garnet/src/provider.rs
metajack: That sounds fine to me, but it should be an issue against the Garnet repo. We aren't in control of the protocol within diskann-garnet. Once the upstream protocol supports this it should be easy enough to add.

diskann-garnet/src/fsm.rs
metajack: Fixed.

diskann-garnet/src/fsm.rs
metajack: I don't know a good way to test it without some test harness that can do random interleavings. The underlying storage is not thread safe as each read and write are atomic on their own, but groups of them are not.

We keep in-memory threadsafe structures to avoid rereading the FSM on the hot path, so those handle checking if there are reusable IDs quickly and safely and also preventing duplicate new ids where two inserts happen concurrently. The first is handled by the thread safe queue, and the second by the atomic which tracks the next id.

I added some explanation to the module level comments.

diskann-garnet/src/labels.rs
hailangx: I just think it's unnecessary to print raw pointer value and len is enough.  but anyway, use derive and print it out is no harm.

diskann-garnet/src/lib.rs
hailangx: move this to diskann defaults

harsha-simhadri (approved):

PR #805, opened 2026-03-02
Add block-transposed multi-vector representation for SIMD-friendly layouts

suri-kumkaran:
## Summary

Introduces `BlockTransposed<T, GROUP, PACK>`, a new `Repr` implementation that stores multi-vectors in a block-transposed layout optimized for SIMD processing. This enables faster multi-vector distance computations by ensuring data is arranged for efficient vectorized access patterns.

## Key changes

- **`BlockTransposed<T, GROUP, PACK>`**: Generic block-transposed matrix representation where groups of `GROUP` rows are stored transposed, with an optional `PACK` factor for column interleaving (e.g. `PACK=2` for `vpmaddwd`-style instructions).
- **Row proxy types**: `BlockTransposedRow` / `BlockTransposedRowMut` with indexed access, iteration, and `Send`/`Sync` support.
- **Factory methods**: `new_block_transposed`, `from_strided`, `from_matrix_view` — generic over any `T: Copy + Default`.
- **Block-level views**: `block()` / `block_mut()` / `remainder_block()` / `remainder_block_mut()` returning `MatrixView` / `MutMatrixView` with dimensions `(padded_ncols/PACK, GROUP*PACK)` for direct SIMD consumption.
- **Compile-time constraints**: `GROUP > 0`, `PACK > 0`, `GROUP % PACK == 0` enforced at monomorphization.
- **Added tests** including parameterized correctness tests across multiple GROUP/PACK combinations, generic element types (`f32`, `i32`, `u8`), bounds-checking panic tests, and Miri-compatible paths.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -621,7 +621,7 @@ dependencies = [
 
 [[package]]
 name = "diskann"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "anyhow",
  "bytemuck",
@@ -645,7 +645,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-benchmark"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "anyhow",
  "bf-tree",
@@ -682,7 +682,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-benchmark-core"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "anyhow",
  "diskann",
@@ -699,7 +699,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-benchmark-runner"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "anyhow",
  "clap",
@@ -713,7 +713,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-benchmark-simd"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "anyhow",
  "diskann-benchmark-runner",
@@ -730,7 +730,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-disk"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "anyhow",
  "bincode",
@@ -781,7 +781,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-label-filter"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "anyhow",
  "bf-tree",
@@ -804,7 +804,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-linalg"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "approx",
  "cfg-if",
@@ -818,7 +818,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-platform"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "io-uring",
  "libc",
@@ -828,7 +828,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-providers"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "anyhow",
  "approx",
@@ -872,7 +872,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-quantization"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "bytemuck",
  "cfg-if",
@@ -891,7 +891,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-tools"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "anyhow",
  "bincode",
@@ -923,7 +923,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-utils"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "bytemuck",
  "cfg-if",
@@ -938,7 +938,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-vector"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "approx",
  "cfg-if",
@@ -952,7 +952,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-wide"
-version = "0.48.0"
+version = "0.49.0"
 dependencies = [
  "cfg-if",
  "half",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ default-members = [
 resolver = "3"
 
 [workspace.package]
-version = "0.48.0"                                                                                      # Obeying semver
+version = "0.49.0"      # Obeying semver
 description = "DiskANN is a fast approximate nearest neighbor search library for high dimensional data"
 authors = ["Microsoft"]
 documentation = "https://github.com/microsoft/DiskANN"
@@ -48,22 +48,22 @@ undocumented_unsafe_blocks = "warn"
 
 [workspace.dependencies]
 # Base And Numerics
-diskann-wide = { path = "diskann-wide", version = "0.48.0" }
-diskann-vector = { path = "diskann-vector", version = "0.48.0" }
-diskann-linalg = { path = "diskann-linalg", version = "0.48.0" }
-diskann-utils = { path = "diskann-utils", default-features = false, version = "0.48.0" }
-diskann-quantization = { path = "diskann-quantization", default-features = false, version = "0.48.0" }
-diskann-platform = { path = "diskann-platform", version = "0.48.0" }
+diskann-wide = { path = "diskann-wide", version = "0.49.0" }
+diskann-vector = { path = "diskann-vector", version = "0.49.0" }
+diskann-linalg = { path = "diskann-linalg", version = "0.49.0" }
+diskann-utils = { path = "diskann-utils", default-features = false, version = "0.49.0" }
+diskann-quantization = { path = "diskann-quantization", default-features = false, version = "0.49.0" }
+diskann-platform = { path = "diskann-platform", version = "0.49.0" }
 # Algorithm
-diskann = { path = "diskann", version = "0.48.0" }
+diskann = { path = "diskann", version = "0.49.0" }
 # Providers
-diskann-providers = { path = "diskann-providers", default-features = false, version = "0.48.0" }
-diskann-disk = { path = "diskann-disk", version = "0.48.0" }
-diskann-label-filter = { path = "diskann-label-filter", version = "0.48.0" }
+diskann-providers = { path = "diskann-providers", default-features = false, version = "0.49.0" }
+diskann-disk = { path = "diskann-disk", version = "0.49.0" }
+diskann-label-filter = { path = "diskann-label-filter", version = "0.49.0" }
 # Infra
-diskann-benchmark-runner = { path = "diskann-benchmark-runner", version = "0.48.0" }
-diskann-benchmark-core = { path = "diskann-benchmark-core", version = "0.48.0" }
-diskann-tools = { path = "diskann-tools", version = "0.48.0" }
+diskann-benchmark-runner = { path = "diskann-benchmark-runner", version = "0.49.0" }
+diskann-benchmark-core = { path = "diskann-benchmark-core", version = "0.49.0" }
+diskann-tools = { path = "diskann-tools", version = "0.49.0" }
 
 # External dependencies (shared versions)
 anyhow = "1.0.98"
--- a/diskann-providers/src/model/graph/provider/async_/bf_tree/neighbor_provider.rs
+++ b/diskann-providers/src/model/graph/provider/async_/bf_tree/neighbor_provider.rs
@@ -52,10 +52,9 @@ impl<I: VectorId> NeighborProvider<I> {
         self.adjacency_list_index.config()
     }
 
-    /// Create a snapshot of the adjacency list index
-    ///
-    pub fn snapshot(&self) -> std::path::PathBuf {
-        self.adjacency_list_index.snapshot()
+    /// Access the underlying BfTree
+    pub(crate) fn bftree(&self) -> &BfTree {
+        &self.adjacency_list_index
     }
 
     /// Return the maximum degree (number of neighbors per vector)
@@ -360,7 +359,7 @@ mod tests {
         neighbor_provider.set_neighbors(2, &[1, 3, 5]).unwrap();
 
         // Call snapshot - should not panic
-        neighbor_provider.snapshot();
+        neighbor_provider.adjacency_list_index.snapshot();
 
         // Verify data is still accessible after snapshot
         let mut result = AdjacencyList::with_capacity(10);
--- a/diskann-providers/src/model/graph/provider/async_/bf_tree/provider.rs
+++ b/diskann-providers/src/model/graph/provider/async_/bf_tree/provider.rs
@@ -1798,11 +1798,17 @@ pub struct BfTreeParams {
 
 impl BfTreeParams {
     /// Build a BfTree Config from the saved parameters and a file path.
-    pub fn to_config(&self, path: &std::path::Path) -> Config {
+    /// When `is_memory` is true, the config uses an in-memory storage backend,
+    /// ensuring the circular buffer is at least as large as the bf-tree default.
+    pub fn to_config(&self, path: &std::path::Path, is_memory: bool) -> Config {
         let mut config = Config::new(path, self.bytes);
         config.cb_max_record_size(self.max_record_size);
         config.leaf_page_size(self.leaf_page_size);
-        config.storage_backend(bf_tree::StorageBackend::Std);
+        if is_memory {
+            config.storage_backend(bf_tree::StorageBackend::Memory);
+        } else {
+            config.storage_backend(bf_tree::StorageBackend::Std);
+        }
         config
     }
 }
@@ -1826,6 +1832,8 @@ pub struct SavedParams {
     pub params_neighbor: BfTreeParams,
     pub quant_params: Option<QuantParams>,
     pub graph_params: Option<GraphParams>,
+    /// Whether the original model was in-memory (`true`) or on-disk (`false`).
+    pub is_memory: bool,
 }
 
 /// The element type of the full-precision vectors stored in the index.
@@ -1939,6 +1947,35 @@ async fn copy_snapshot_if_needed(
     Ok(())
 }
 
+/// Save a BfTree to disk, handling both in-memory and on-disk cases.
+/// For in-memory trees, uses `snapshot_memory_to_disk` to serialize all records.
+/// For on-disk trees, snapshots in place and copies if the target path differs.
+async fn save_bftree(tree: &BfTree, target_path: std::path::PathBuf) -> ANNResult<()> {
+    if tree.config().is_memory_backend() {
+        tree.snapshot_memory_to_disk(&target_path);
+    } else {
+        let snapshot_path = tree.snapshot();
+        copy_snapshot_if_needed(snapshot_path, target_path).await?;
+    }
+    Ok(())
+}
+
+/// Load a BfTree from a snapshot file, restoring it as in-memory or on-disk
+/// depending on `is_memory`. Builds the Config from `params` internally.
+fn load_bftree(
+    params: &BfTreeParams,
+    snapshot_path: std::path::PathBuf,
+    is_memory: bool,
+) -> Result<BfTree, ANNError> {
+    let config = params.to_config(&snapshot_path, is_memory);
+    if is_memory {
+        BfTree::new_from_snapshot_disk_to_memory(snapshot_path, config)
+            .map_err(|e| ANNError::from(super::ConfigError(e)))
+    } else {
+        BfTree::new_from_snapshot(config, None).map_err(|e| ANNError::from(super::ConfigError(e)))
+    }
+}
+
 // SaveWith/LoadWith for BfTreeProvider with TableDeleteProviderAsync
 
 impl<T> SaveWith<String> for BfTreeProvider<T, NoStore, TableDeleteProviderAsync>
@@ -1972,7 +2009,15 @@ where
             },
             quant_params: None, // No quantization parameters
             graph_params: self.graph_params.clone(),
+            is_memory: self.full_vectors.config().is_memory_backend(),
         };
+
+        // All stores must use the same storage backend.
+        debug_assert_eq!(
+            self.full_vectors.config().is_memory_backend(),
+            self.neighbor_provider.config().is_memory_backend(),
+            "Vector and neighbor stores have mismatched storage backends"
+        );
 
         // Save only essential parameters as JSON
         {
@@ -1985,17 +2030,13 @@ where
         }
 
         // Save vectors and neighbors
-        let vectors_snapshot_path = self.full_vectors.snapshot();
-        let neighbors_snapshot_path = self.neighbor_provider.snapshot();
-
-        // Copy snapshot files to the target prefix location if they differ
-        copy_snapshot_if_needed(
-            vectors_snapshot_path,
+        save_bftree(
+            self.full_vectors.bftree(),
             BfTreePaths::vectors_bftree(&saved_params.prefix),
         )
         .await?;
-        copy_snapshot_if_needed(
-            neighbors_snapshot_path,
+        save_bftree(
+            self.neighbor_provider.bftree(),
             BfTreePaths::neighbors_bftree(&saved_params.prefix),
         )
         .await?;
@@ -2037,15 +2078,11 @@ where
         let metric = Metric::from_str(&saved_params.metric)
             .map_err(|e| ANNError::log_index_error(format!("Failed to parse metric: {}", e)))?;
 
-        let vector_config = saved_params
-            .params_vector
-            .to_config(&BfTreePaths::vectors_bftree(&saved_params.prefix));
-        let neighbor_config = saved_params
-            .params_neighbor
-            .to_config(&BfTreePaths::neighbors_bftree(&saved_params.prefix));
-
-        let vector_index =
-            BfTree::new_from_snapshot(vector_config.clone(), None).map_err(super::ConfigError)?;
+        let vector_index = load_bftree(
+            &saved_params.params_vector,
+            BfTreePaths::vectors_bftree(&saved_params.prefix),
+            saved_params.is_memory,
+        )?;
         let full_vectors = VectorProvider::<T>::new_from_bftree(
             saved_params.max_points,
             saved_params.dim,
@@ -2053,8 +2090,11 @@ where
             vector_index,
         );
 
-        let adjacency_list_index =
-            BfTree::new_from_snapshot(neighbor_config.clone(), None).map_err(super::ConfigError)?;
+        let adjacency_list_index = load_bftree(
+            &saved_params.params_neighbor,
+            BfTreePaths::neighbors_bftree(&saved_params.prefix),
+            saved_params.is_memory,
+        )?;
         let neighbor_provider =
             NeighborProvider::<u32>::new_from_bftree(saved_params.max_degree, adjacency_list_index);
 
@@ -2124,7 +2164,20 @@ where
                 },
             }),
             graph_params: self.graph_params.clone(),
+            is_memory: self.full_vectors.config().is_memory_backend(),
         };
+
+        // All stores must use the same storage backend.
+        debug_assert_eq!(
+            self.full_vectors.config().is_memory_backend(),
+            self.neighbor_provider.config().is_memory_backend(),
+            "Vector and neighbor stores have mismatched storage backends"
+        );
+        debug_assert_eq!(
+            self.full_vectors.config().is_memory_backend(),
+            self.quant_vectors.config().is_memory_backend(),
+            "Vector and quant stores have mismatched storage backends"
+        );
 
         // Save only essential parameters as JSON
         {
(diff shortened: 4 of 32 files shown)


diskann-quantization/src/algorithms/kmeans/mod.rs
```diff
@@ -4,7 +4,6 @@
  */
 
 pub(crate) mod common;
-pub use common::BlockTranspose;
 pub(crate) use common::square_norm;
```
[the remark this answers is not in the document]
suri-kumkaran: Not required, the type `BlockTranspose` is only used internally.

diskann-quantization/src/algorithms/kmeans/lloyds.rs
```diff
@@ -373,7 +374,7 @@ fn update_centroids(mut centers: MutMatrixView<'_, f32>, data: StridedView<'_, f
 pub(crate) fn lloyds_inner(
     data: StridedView<'_, f32>,
     square_norms: &[f32],
-    transpose: &BlockTranspose<16>,
+    transpose: &Mat<BlockTransposed<f32, 16>>,
```
hildebrandmw: Generally, we should try to make sure functions accept `MatRef<'_, BlocTransposed>` instead of `&Mat<BlockTransposed>`. The former can be used in more contexts.

The original code didn't do this because there was no generalized reference version of the `BlockTranspose`.

diskann-quantization/src/multi_vector/matrix.rs
```diff
@@ -557,8 +557,8 @@ where
 /// Memory is automatically deallocated when the matrix is dropped.
 #[derive(Debug)]
 pub struct Mat<T: ReprOwned> {
-    ptr: NonNull<u8>,
-    repr: T,
+    pub(crate) ptr: NonNull<u8>,
+    pub(crate) repr: T,
 }
```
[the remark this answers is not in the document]
hildebrandmw: I agree with this.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,1608 @@
...
+//!
+//! ## `PACK = 2` (super-packed)
+//!
+//! With `GROUP = 4` and `PACK = 2`, adjacent column-pairs are interleaved per row
+//! within each group panel:
+//!
+//! ```text
+//! Col-group(0,1):  [r0_c0, r0_c1, r1_c0, r1_c1, r2_c0, r2_c1, r3_c0, r3_c1]
+//! Col-group(2,3):  [r0_c2, r0_c3, r1_c2, r1_c3, r2_c2, r2_c3, r3_c2, r3_c3]
+//! ```
```
hildebrandmw, line 65: I think it might be helpful to include a diagram like the `PACK = 1` case.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,1608 @@
...
+    /// Helper: wrap a `Box<[T]>` into a [`Mat`] without any further checks.
+    ///
+    /// # Safety
+    ///
+    /// `b.len()` must equal `self.storage_len()`.
+    unsafe fn box_to_mat(self, b: Box<[T]>) -> Mat<Self> {
+        debug_assert_eq!(b.len(), self.storage_len(), "safety contract violated");
+
+        // SAFETY: Box guarantees the returned pointer is non-null.
+        let ptr = unsafe { NonNull::new_unchecked(Box::into_raw(b)) }.cast::<u8>();
```
hildebrandmw: The pattern of leaking boxes into non-null can probably be moved to a helper function like that `as_nonnull` functions in `utils.rs`.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,1608 @@
...
+}
+
+// ════════════════════════════════════════════════════════════════════
+// Row proxy types
+// ════════════════════════════════════════════════════════════════════
+
+/// An immutable view of a single logical row in a block-transposed matrix.
+///
+/// Because the elements of a logical row are strided (not contiguous), this struct
+/// acts as a lightweight proxy that supports indexed access and iteration.
```
hildebrandmw: Very small terminology nit: I don't think it's super helpful to refer to these as proxies. They aren't necessarily acting as a substitute for something else ... they *are* the things (in this case, row and column views).

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,1608 @@
...
+#[derive(Debug, Clone, Copy)]
+pub struct BlockTransposedRow<'a, T, const GROUP: usize, const PACK: usize = 1> {
```
hildebrandmw: Small naming nit: This can be called just `Row` and the iter `RowIter`. If we don't export them from the parent `mod.rs`, then that they belong to `block_transposed` is implicit in their fully qualified path.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,1608 @@
...
+    /// Pointer to the element at `(row, col=0)` in the backing allocation.
+    base: *const T,
```
hildebrandmw: It's possible we might be able to reuse `bits::ptr::SlicePtr` for representing a pointer with a lifetime. As an added benefit, `SlicePtr` is already `NonNull`, so can help `Row` use the niche optimization out-of-the-box. Finally, the `SlicePtr` type already has the correct variance behavior built-in, so we don't need to replicate the `&'a T` and `&'a mut T` phantom data fields.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,1608 @@
...
+}
+
+impl<T: Copy, const GROUP: usize, const PACK: usize> std::ops::Index<usize>
+    for BlockTransposedRow<'_, T, GROUP, PACK>
+{
+    type Output = T;
+
+    #[inline]
+    fn index(&self, col: usize) -> &Self::Output {
+        assert!(
```
hildebrandmw: As a small stylistic suggestion - If `get()` returns a reference, then `index` can be expressed as a call to `get` with a panic on the `None` case - saving one extra unsafe call.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,1608 @@
...
+    fn index_mut(&mut self, col: usize) -> &mut Self::Output {
+        assert!(
+            col < self.ncols,
+            "column index {col} out of bounds (ncols = {})",
+            self.ncols
+        );
+        // SAFETY: bounds checked.
+        unsafe { &mut *self.base.add(col_offset::<GROUP, PACK>(col)) }
+    }
+}
```
hildebrandmw, line 457: Should we also implement `std::ops::Index` for `RowMut`?

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,1608 @@
...
+    for BlockTransposed<T, GROUP, PACK>
+{
+    type Error = SliceError;
+
+    fn new_ref(self, data: &[T]) -> Result<MatRef<'_, Self>, Self::Error> {
+        self.check_slice(data)?;
+
+        // `as_nonnull` relies on slices having non-null base pointers; sanity-check
+        // that an empty allocation is only used with a zero-capacity repr.
+        debug_assert!(!data.is_empty() || self.storage_len() == 0);
```
hildebrandmw: I do not fully understand what this check is accomplishing. Would this not already be accomplished through `check_slice`? Note that empty slices are guaranteed to have non-null base pointers.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,1608 @@
...
+use super::matrix::{
+    Defaulted, LayoutError, Mat, MatMut, MatRef, NewMut, NewOwned, NewRef, Overflow, Repr, ReprMut,
+    ReprOwned, SliceError,
+};
+use crate::utils;
+
+/// Round `ncols` up to the next multiple of `PACK`.
+#[inline]
+fn padded_ncols<const PACK: usize>(ncols: usize) -> usize {
+    utils::div_round_up(ncols, PACK) * PACK
```
hildebrandmw: I think there's some simplification to be had with [`checked_next_multiple_of`](https://doc.rust-lang.org/beta/std/primitive.usize.html#method.checked_next_multiple_of) and [`next_multiple_of`](https://doc.rust-lang.org/beta/std/primitive.usize.html#method.next_multiple_of).

hildebrandmw (commented):
Thanks Suryansh! I've taken a look over and left a few stylistic comments, but there are few larger ones that I will group together here:

1. We should be very careful with the unchecked arithmetic, particularly in `compute_capacity`. We honestly probably want two flavors of this function, one that verifies none of the intermediate steps overflow that gets used in the constructor, and one that assumes everything has already been checked (this would be the current `compute_capacity` function). I would also recommend leaning in to `usize::next_multiple_of` instead of doing the operation manually.

2. I'm getting more convinced that we do not want to use `Mat`/`MatRef`/`MatMut` in public APIs. This forces us into an awkward situation where we need to start adding a bunch of inherent methods to these types, which makes method discoverability a little harder and is not a pattern that can be replicated outside of `diskann-quantization`. Instead, I think it would be cleaner to have thin wrappers:
    ```rust
    struct BlockTransposed<T, const GROUP: usize, const PACK: usize> {
        data: Mat<Repr<T, GROUP, PACK>>,
   }

    struct BlockTransposedRef<'a, T, const GROUP: usize, const PACK:    usize> {
        data: MatRef<'a, Repr<T, GROUP, PACK>>,
    }

    // etc.
    ```
    Inherent methods can be added to these at will and the generated docs for these inherent methods will be clear (since they will not be grouped in with all the other inherent methods). This will also let you safely has `as_slice` as an inherent method to avoid the manual unsafe construction in tests.

    Also, I'll admit a small preference for using method delegation instead of stamping out a giant macro, something like:
    ```rust
    impl<T, const GROUP: usize, const PACK: usize> BlockTransposed<T, GROUP, PACK> {
        fn block(&self, block: usize) -> MatrixView<'_, T> {
            self.reborrow().block(block)
        }
    }
    ```
    But I think that is secondary.

3. In the tests - we should really have one that calls `collect` on a `Mat::row_iter_mut()` and writes to a bunch of rows in like a `thread::scoped` or something. Basically, something to make sure that working on multiple rows concurrently is fine (and when using Miri - that will add another level of safety).

diskann-quantization/src/multi_vector/matrix.rs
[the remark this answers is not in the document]
suri-kumkaran: Fixed.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,1608 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
```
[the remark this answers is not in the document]
suri-kumkaran: Fixed.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,1608 @@
...
+
+    // ── Factory methods ──────────────────────────────────────────────
+
+    /// Construct a default-initialized block-transposed matrix from dimensions.
+    #[allow(clippy::expect_used)]
+    pub fn new_block_transposed(nrows: usize, ncols: usize) -> Self {
+        let repr = BlockTransposed::<T, GROUP, PACK>::new(nrows, ncols)
+            .expect("dimensions should not overflow");
+        Mat::new(repr, Defaulted).expect("infallible")
+    }
```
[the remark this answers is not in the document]
suri-kumkaran: Fixed.

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: Done.

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: Yes, sorry for the redundant check, I have now removed it.

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: Done.

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: Fixed.

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: Done.

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: Done.

diskann-quantization/src/algorithms/kmeans/lloyds.rs
suri-kumkaran: Done.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,1608 @@
...
+/// Compute the total number of `T` elements required to store a block-transposed matrix
+/// of `nrows x ncols` with group size `GROUP` and packing factor `PACK`.
+///
+/// Compile-time constraints (`GROUP > 0`, `PACK > 0`, `GROUP % PACK == 0`) are enforced
+/// by [`BlockTransposed::_ASSERTIONS`]; this function does **not** duplicate them.
+#[inline]
+fn compute_capacity<const GROUP: usize, const PACK: usize>(nrows: usize, ncols: usize) -> usize {
+    let num_blocks = utils::div_round_up(nrows, GROUP);
+    GROUP * padded_ncols::<PACK>(ncols) * num_blocks
+}
```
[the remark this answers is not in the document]
suri-kumkaran: Fixed.

diskann-quantization/src/utils.rs
```diff
@@ -37,6 +37,19 @@ pub(crate) fn as_nonnull_mut<T>(slice: &mut [T]) -> NonNull<T> {
...
 
+/// Leak a `Box<[T]>` and return the pointer as a `NonNull<u8>`.
+///
+/// This is the owned-allocation counterpart of [`as_nonnull`]: it consumes the
+/// box, preventing its destructor from running, and hands back a non-null byte
+/// pointer suitable for storing inside a [`Mat`](super::multi_vector::matrix::Mat).
+///
+/// To reclaim the memory later, reconstruct the `Box` via
+/// `Box::from_raw(slice_from_raw_parts_mut(ptr, len))`.
+pub(crate) fn box_into_nonnull<T>(b: Box<[T]>) -> NonNull<u8> {
```
hildebrandmw: I think this should return `NonNull<T>` and leave it to the caller to do the casting - similar to how the `as_nonnull` functions work. This also simplifies the safety documentation.

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: It was already there, right?

hildebrandmw: Don't mind me. I am blind.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,1608 @@
...
+        let mat = Self::new_block_transposed(nrows, ncols);
+
+        // Fill using linear_index. The allocation is default-initialized so padding
+        // positions already hold `T::default()`.
+        let base: *mut T = mat.ptr.as_ptr().cast::<T>();
+        for row in 0..nrows {
+            for col in 0..ncols {
+                let idx = linear_index::<GROUP, PACK>(row, col, ncols);
+                // SAFETY: idx < storage_len by construction, base points to a valid allocation.
+                unsafe { *base.add(idx) = v[(row, col)] };
```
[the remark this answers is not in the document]
suri-kumkaran: Done.

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: Done.

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: Make sense, done.

diskann-quantization/src/utils.rs
```diff
@@ -37,6 +37,19 @@ pub(crate) fn as_nonnull_mut<T>(slice: &mut [T]) -> NonNull<T> {
...
 }
 
+/// Leak a `Box<[T]>` and return the pointer as a `NonNull<u8>`.
+///
+/// This is the owned-allocation counterpart of [`as_nonnull`]: it consumes the
+/// box, preventing its destructor from running, and hands back a non-null byte
+/// pointer suitable for storing inside a [`Mat`](super::multi_vector::matrix::Mat).
+///
+/// To reclaim the memory later, reconstruct the `Box` via
+/// `Box::from_raw(slice_from_raw_parts_mut(ptr, len))`.
```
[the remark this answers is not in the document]
suri-kumkaran: Fixed.

diskann-quantization/src/utils.rs
suri-kumkaran: Done.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,2315 @@
...
+                                // backing allocation which has exactly `storage_len`
+                                // elements, and our loop visits each position once.
+                                unsafe { *dst = src_row[col] };
+                            }
+                            // Advance even for padding — the default is already there.
+                            dst = dst.wrapping_add(1);
+                        }
+                    } else {
+                        // Entire row is padding — skip PACK positions.
+                        dst = dst.wrapping_add(PACK);
```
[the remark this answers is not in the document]
hildebrandmw: Agreed.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,2315 @@
...
+                    for row in start_row..end_row {
+                        // Reconstruct the pointer from the usize address.
+                        let ptr = ptr_addr as *mut u8;
+                        // SAFETY: Each thread operates on distinct rows. Rows in
+                        // the block-transposed layout occupy non-overlapping memory
+                        // regions, so distinct-row `RowMut`s don't alias. The
+                        // pointer is valid for the lifetime of the scope because
+                        // `mat` outlives the `thread::scope` block.
+                        let mut row_view: RowMut<'_, f32, GROUP, PACK> =
+                            unsafe { repr.get_row_mut(NonNull::new_unchecked(ptr), row) };
```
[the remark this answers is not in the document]
hildebrandmw: To second this - this test can be written using [`Mat::rows_mut`](https://github.com/microsoft/DiskANN/blob/1016934462837c7b25003bb830b9ff8fc54a8889/diskann-quantization/src/multi_vector/matrix.rs#L652-L655) to do the partitioning. This is a good idea because it also directly tests the interface that needs to be sound with respect to the `Repr` implementation.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,2315 @@
...
+    #[inline]
+    pub fn padded_ncols(&self) -> usize {
+        self.data.repr().padded_ncols()
+    }
+
+    /// Alias for [`ncols`](Self::ncols) — the number of logical columns.
+    #[inline]
+    pub fn block_size(&self) -> usize {
+        self.data.repr().ncols()
+    }
```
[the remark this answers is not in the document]
hildebrandmw: Agreed. I would tentatively say to get rid of `block_size` and update callers to use `ncols` directly, rather than having two methods that do the same thing.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,2315 @@
...
+
+    /// Borrow as a mutable [`BlockTransposedMut`].
+    pub fn as_view_mut(&mut self) -> BlockTransposedMut<'_, T, GROUP, PACK> {
+        BlockTransposedMut {
+            data: self.data.as_view_mut(),
+        }
+    }
+
+    // ── Delegated read methods ───────────────────────────────────
```
hildebrandmw, line 1012: Something I was mulling over to try to reduce the maintenance cost of the forwarding methods. We can use a macro like:
```rust
macro_rules! delegate_to_ref {
    ($(#[$m:meta])* $vis:vis fn $name:ident(&self $(, $a:ident: $t:ty)*) $(-> $r:ty)?) => {
        #[doc = concat!("See [`BlockTransposedRef::", stringify!($name), "`].")]
        $(#[$m])* #[inline]
        $vis fn $name(&self $(, $a: $t)*) $(-> $r)? { self.as_view().$name($($a),*) }
    };
    // + an `unsafe fn` arm
}
```
Delegation then becomes
```rust
delegate_to_ref(pub fn nrows(&self) -> usize);
```
The documentation that gets generated automatically cross-references to the actual implementation on `BlockTransposedRef` so that will always stay in-sync.

This doesn't necessarily work for the mutable methods or functions like `get_row`, but anything that goes through `as_view()` can be delegated this way, which I think would go a long way towards slimming these down.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,2315 @@
...
+        unsafe { std::slice::from_raw_parts_mut(self.data.as_raw_mut_ptr().cast::<T>(), len) }
+    }
+
+    /// Return a mutable view over a full block.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `block >= self.full_blocks()`.
+    #[allow(clippy::expect_used)]
+    pub fn block_mut(&mut self, block: usize) -> MutMatrixView<'_, T> {
```
hildebrandmw, line 928: In combination with the delegate macro for the easy `as_view` functions, here's what I think we should do with the mutable methods to consolidate the unsafe code into a single place:
```rust
impl<'a> BlockTransposeMut<'a, GROUP, PACK> {
    // The actual implementation - taking `self` by value to we can pilfer
    // out the contents cleanly.
    fn block_mut_inner(self, block: usize) -> MutMatrixView<'a, T> {
        // do the thing!
    }

    pub fn block_mut(&mut self, block: usize) -> MutMatrixView<'_, T> {
        self.as_mut_view().block_mut_inner(block)
    }
}

impl<'a> BlockTranspose<'a, GROUP, PACK> {
    // See [`BlockTransposeMut::block_mut`].
    pub fn block_mut(&mut self, block: usize) -> MutMatrixView<'_, T> {
        // The inner method receiving by value means we don't end up
        // referencing a temporary, so this all works.
        self.as_mut_view().block_mut_inner(block)
    }
}
```

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,2315 @@
...
+
+/// Compute the linear index for the element at logical `(row, col)` in a block-transposed
+/// layout with group size `GROUP`, packing factor `PACK`, and `ncols` logical columns.
+#[inline]
+fn linear_index<const GROUP: usize, const PACK: usize>(
+    row: usize,
+    col: usize,
+    ncols: usize,
+) -> usize {
+    let pncols = padded_ncols::<PACK>(ncols);
```
hildebrandmw, line 136: We should insert
```rust
if pncols == 0 {
   return 0;
}
```
here. The test `test_row_view_empty` is failing Miri because a non-zero linear index is computed and passed to `ptr::add`. See my top level review comment for some ideas about testing.

diskann-quantization/src/multi_vector/block_transposed.rs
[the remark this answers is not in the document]
suri-kumkaran: Make sense, done.

diskann-quantization/src/multi_vector/block_transposed.rs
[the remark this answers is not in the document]
suri-kumkaran: Make sense, done.

diskann-quantization/src/multi_vector/block_transposed.rs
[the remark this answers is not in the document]
suri-kumkaran: Done.

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: Thanks for the cool idea, done.

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: Done.

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: Fixed.

diskann-quantization/src/algorithms/kmeans/lloyds.rs
```diff
@@ -25,7 +26,7 @@ diskann_wide::alias!(u32s = u32x8);
 //
 // Return the residual distance.
 pub fn distances_in_place(
-    dataset: &BlockTranspose<16>,
+    dataset: BlockTransposedRef<'_, f32, 16>,
     data_norms: &[f32],
     centers: MatrixView<'_, f32>,
     center_norms: &[f32],
```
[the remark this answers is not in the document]
suri-kumkaran: Fixed.

diskann-quantization/src/algorithms/kmeans/plusplus.rs
```diff
@@ -79,7 +80,7 @@ pub(crate) trait MicroKernel {
 
 diskann_wide::alias!(f32s = f32x8);
 
-impl MicroKernel for BlockTranspose<16> {
+impl MicroKernel for BlockTransposed<f32, 16> {
     // Process 16-dimensions concurrently, split across two `Wide`s.
     type Intermediate = (f32s, f32s);
     type RollingSum = f64;
```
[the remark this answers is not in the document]
suri-kumkaran: Fixed.

diskann-quantization/src/utils.rs
```diff
@@ -37,6 +37,19 @@ pub(crate) fn as_nonnull_mut<T>(slice: &mut [T]) -> NonNull<T> {
...
+///
+/// This is the owned-allocation counterpart of [`as_nonnull`]: it consumes the
+/// box, preventing its destructor from running, and hands back a non-null
+/// pointer suitable for storing inside a [`Mat`](super::multi_vector::matrix::Mat).
+///
+/// To reclaim the memory later, reconstruct the `Box` via
+/// `Box::from_raw(std::ptr::slice_from_raw_parts_mut(ptr, len))`.
+pub(crate) fn box_into_nonnull<T>(b: Box<[T]>) -> NonNull<T> {
+    // SAFETY: `Box::into_raw` guarantees the returned pointer is non-null.
+    unsafe { NonNull::new_unchecked(Box::into_raw(b).cast::<T>()) }
```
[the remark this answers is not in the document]
suri-kumkaran: Fixed.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,2105 @@
...
+        }
+    }
+
+    // ════════════════════════════════════════════════════════════════
+    // Test runners — each combination gets the full API surface.
+    // ════════════════════════════════════════════════════════════════
+
+    #[test]
+    fn test_api_pack1_group16() {
+        let row_range = if cfg!(miri) { 127..128 } else { 0..128 };
```
hildebrandmw: The Miri ranges bug me, the only hit one arbitrary "nice" point. Can we make them hit layout boundaries instead? Like `0, 1, GROUP - 1, GROUP, GROPU + 1` rows (and a few multiples there-of) and `0, 1, PACK - 1, PACK, PACK + 1` columns?

It's exactly these kinds of transition zones that can accidentally exercise UB and that Miri will protect agains.

suri-kumkaran: Done.

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,2161 @@
...
+            let offset = self.data.repr().block_offset(self.full_blocks());
+            let stride = self.data.repr().block_stride();
+            // SAFETY: The remainder block exists (`remainder() != 0`),
+            // so `offset + stride` is within the backing allocation.
+            let data: &[T] =
+                unsafe { std::slice::from_raw_parts(self.as_ptr().add(offset), stride) };
+            Some(
+                MatrixView::try_from(data, self.padded_ncols() / PACK, GROUP * PACK)
+                    .expect("base data should have been sized correctly"),
+            )
```
arkrishn94, line 821: Minor, but this can just be `Some(self.block(self.full_blocks()))`?

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -0,0 +1,2161 @@
...
+                    if row < nrows {
+                        // SAFETY: row < nrows is checked by the enclosing `if` condition.
+                        let src_row = unsafe { v.get_row_unchecked(row) };
+                        for p in 0..PACK {
+                            let col = col_base + p;
+                            if col < ncols {
+                                // SAFETY: dst advances sequentially through the
+                                // backing allocation which has exactly `storage_len`
+                                // elements, and our loop visits each position once.
+                                unsafe { *dst = src_row[col] };
```
arkrishn94: Not sure if the compiler has a way of skipping bounds checking here but worth doing unchecked indexing here?

arkrishn94 (approved):
Looks good @suri-kumkaran, thanks for fleshing out this one. My comments are both minor.

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: Oh, the `block` method panics for `block => self.full_blocks()` to make sure clients never use `remainder_block` accidentally.

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: Completely make sense, done.

arkrishn94 (approved):
Looks good, thanks for powering through all the iterating.

Issue #737, opened 2026-02-07
Document and extend index serialization format


harsha-simhadri:
## Is your feature request related to a problem? Please describe.
The current format for saving indices -- inmem and disk -- needs to be documented. It also does not allow for storing start point strategies. 

## Describe the solution you'd like
For both in-mem and SSD indices, extension of current format and documentation.

## Steps
- Agree on requirements, target scenarios, exit criteria.
- Agree on proposed solution.
- Save index. Access artifacts.
- Backward compatibility

PR #807, opened 2026-03-02
RFC #2 : Index Serialization and Backward Compatibility

suhasjs:

## Summary
Introduces a unified serialization contract for DiskANN indices, replacing the current per-crate ad hoc approaches (`SaveWith/LoadWith`) with a single context-first, manifest-driven design.

## Key Design Points
- Version primitive in `diskann-utils`: simple (major, minor, patch) struct with `SemVer` compatibility checks.
- `Load / Save` traits (async, context-first): each component declares a static VERSION, and `load()` dispatches to `load_compatible()` or `load_legacy()` based on the persisted version.
- `SerializationContext / DeserializationContext`: orchestrate manifest navigation, per-component versioning, and artifact resolution via scoped/RAII APIs. Callers never deal with file paths directly.
- Recursive `UUID-metadata.json` manifest: single JSON file with per-component versioning and files arrays. Avoids cascading version bumps when only one component evolves.
- `N-1` backward compatibility: guaranteed within the same major version. Dual-path loading (new manifest vs. legacy) with no required one-time migration.
- File-per-artifact: each binary payload gets its own file with deterministic naming (`UUID-lineage-key.data`).
- `SaveValue / LoadValue`: separate value-level traits for primitives, keeping generic containers (`Vec<T>`) uniform across primitive and composite types.
- Optional preflight/summary: lightweight `can_load` and summary passes for runtime sizing before full materialization.

#### Any other comments?
Link to issue -- #737


rfcs/00002-serialization.md
```diff
@@ -0,0 +1,760 @@
...
+1. Primitive values (`u64`, `u32`, `i32`, `i64`, `f32`, `f16`, `String`) do not create component scopes.
+2. Primitive values are encoded as key/value entries within the current component scope through `put_scalar<T>` / `get_scalar<T>`.
+3. Primitive values do not carry their own `version` nodes.
+4. Persisted component/struct types MUST carry versions as defined in Section 1.1.
+5. Generic containers (for example, `Vec<T>`) MUST work uniformly whether `T` is primitive or composite.
+6. For `Vec<T>` under container key `container`, the component name for the i-th element MUST be `container.index-i`.
+7. If `T` is primitive, each element uses the value contract (`SaveValue` / `LoadValue`) under the element key.
+8. If `T` is composite, each element uses the component contract (`Save` / `Load`) with normal version dispatch.
+9. Field/key names SHOULD be supplied by the caller or derive macro expansion (for example, from field identifiers), not inferred at runtime from primitive values.
+
```
hildebrandmw: I have a few concerns I will try to list and I think I have solutions for most of them.

# Value vs Component Split
The split between `Save` and `SaveValue` is going to be problematic. First, it will require code to look something like
```rust
struct Foo {
    a: usize,
    b: usize,
}

struct Bar {
    foo: Foo,
    baz: usize,
}

impl Save for Bar {
    const VERSION = /*...*/;
    fn save(&self, ctx: ComponentScope<'_>) {
        self.foo.save(ctx);
        self.baz.save_value(baz); // Different API
    }
}
```
The difference in calling API means that it will be impossible for a relatively simple macro to generate this code - it will need to have some awareness of types and then guess whether `save` or `save_value` should be used (macros do not have type information, just syntax).

Second, I don't know how we'd implement
```rust
impl<T> Save for Vec<T>
where
    T: ?
```
We cannot have one implementation for `T: Save` and another for `T: SaveValue` as this will run afoul of [coherence](https://rust-lang.github.io/chalk/book/clauses/coherence.html). Rust does not supply sufficient metaprogramming tools to introspect which definition applies and dispatch to that.

My main point is that primitive, sequence, and aggregate types should all use the same API.

2. There are potential issues with mutating the `context` when creating component scopes: (1) what happens if a scope is created but then nothing is ever inserted and (2) if the solution to #1 is to use `Drop` on the component, can we protect against [`std::mem::forget`](https://doc.rust-lang.org/std/mem/fn.forget.html) from ruining our day?

## Potential Solution
This is what I was able to come up with to solve this problem.

### Common Interface
Using saving as an example, we take a three-layered approach. The first layer entry point for saving **any** object is a free function:
```rust
fn save<T>(object: &T, context: Context<'_>) -> Result<Value>
where
    T: Persist,
{
    object.persist(context)
}
```
The free function uses the second level trait `Persist` (feel free to bike-shed - I am terrible at naming):
```rust
pub trait Persist {
    fn persist(&self, context: Context<'_>) -> Result<Value>;
}
```
The `Persist` trait is where the choice between "components" and "values" resides.

Components use the third-level `Save` trait:
```rust
pub trait Save {
    const VERSION: Version;
    fn save(&self, component: Component<'_>) -> Result<Value>;
}
```
Note that this takes a `Component` instead of a `Context`, and has the `VERSION` associated constant. The third layer is glued to the second layer with a blanket implementation:
```rust
impl<T> Persist for T
where
    T: Save,
{
    fn persist(&self, context: Context<'_>) -> Result<Value> {
        // Consume `context` to create a `Component`.
        // supply the `VERSION` upon creation.
        self.save(context.component(T::VERSION))
    }
}
```
Concrete primitive types can implement `Persist` directly:
```rust
impl Persist for usize {
    fn persist(&self, context: Context<'_>) -> Result<Value> {
        context.primitive().set(self)
    }
}

impl Persist for String {
    fn persist(&self, context: Context<'_>) -> Result<Value> {
        context.primitive().set(self)
    }
}
```
Additionally, containers like `Vec` can implement `Persist` directly:
```rust
// Note that we use just a single bound `Persist` for all savable types.
//
// Now whether `T` is a "value" or a "component" - it will just work.
impl<T> Persist for Vec<T>
where
    T: Persist,
{
    fn persist(&self, context: Context<'_>) -> Result<Value> {
        context.sequence().set(self.iter())
    }
}
```

Now, the recursive implementation for components can look something like
```rust
impl Save for Bar {
    fn save(&self, component: Component<'_>) -> Result<Value> {
         // Note that the interface is now the same for all fields.
         // 
         // This makes it *much* easier to automatically generate code
         // from a macro.
         let value = Value::new([
             ("foo", save(&self.foo, component.field("foo"))?),
             ("baz", save(&self.baz, component.field("baz"))?),
         ]);
         Ok(value)
    }
}
```

Importantly, the blanket `impl<T: Save> Persist for T` does not conflict with primitives nor with `Vec`, so they can coexist.

### Dealing with Scopes
In the above design, instead of aggregating the work-in-progress serialized representation on the context - we instead return `Value`s up the stack, where a `Value` represents the serialized state for some sub-object. This completely side-steps potential issues around accumulating the object mutably in the context.

[Here](https://godbolt.org/z/bGcj45x1z) is a compiling sketch generated with the help of AI tooling, so take it with a grain of salt. But the high-level ideas are mostly captured there.

hildebrandmw (commented):
Thanks Suhas! I sketched out a potential way to unify the saving of values and components in a way to that hopefully helps keep the implementation easily auto-generatable.

One orthogonal issue brought up in offline discussion is how to approach a "standardized" representation for indices. This design provides a lot of flexibility, but schemas are largely ad-hoc. Can you expand on what you were thinking regarding JSON Schema (or any other ideas about using a shareable format).

rfcs/00002-serialization.md
```diff
@@ -0,0 +1,760 @@
...
+### Background
+
+DiskANN internals evolve over time (new fields, reordered layouts, changed semantics). Without explicit format versioning and compatibility rules, upgrades can force full index rebuilds, which is costly for large datasets.
+
+The library currently employs divergent approaches:
+
+| Crate | Mechanism | Notes |
+|-------|-----------|-------|
+| `diskann-disk` | `GraphHeader` with `GraphLayoutVersion` | Sector-aligned; validates disk layout. |
```
arrayka: There are two flavors of DiskANN indices:

1.  **Disk-based index**:  
    The graph is stored on disk as a single blob and is accessed from the hot path during search. This enables reading all required data - vectors, adjacency lists, and associated metadata - in a single read operation. It’s designed for efficient random access with minimal I/O overhead.

2.  **Generic index**:  
    Data is typically accessed via a get/set API, backed by either in-memory or persistent storage. Loading happens outside the hot path, and the format is not designed for random access. The load/save functionality here is mainly a convenience feature.

I’m not fully convinced yet that we need to converge the formats for these two flavors, since their intended usage differs. Put differently: if we do want to converge them, why wouldn’t we converge on the existing disk-based format?

rfcs/00002-serialization.md
```diff
@@ -0,0 +1,760 @@
...
+| `diskann` | `SaveWith` / `LoadWith` with provider contexts | Persists graph (`save_graph`/`load_graph`), vector data (`.data` via `.bin`), and quantization artifacts (PQ/SQ), but without a unified recursive metadata contract. |
+
+### Problem Statement
+
+There is no unified serialization contract across DiskANN crates. Each crate uses its own ad hoc mechanism for versioning and persistence, making it impossible to reliably detect index format versions, guarantee backward compatibility, or evolve components independently. This leads to costly full index rebuilds on upgrades and fragile loader code.
+
+### Goals
+
+1. Reliably detect index format versions from root + component metadata
+2. Load indices saved by the immediate SemVer predecessor (major version) using current code. May also support older versions via explicit legacy converters.
```
arrayka: I’m not sure why we need to promise support for the immediate SemVer predecessor (the previous major version) by default. This should be decided on a case‑by‑case basis.

In general, a DiskANN library with major version **X** should only guarantee that it can load data produced by version **X**. Support for loading data produced by older major versions can be added when it makes sense, but it should be an explicit, deliberate decision rather than a blanket promise.

rfcs/00002-serialization.md
```diff
@@ -0,0 +1,760 @@
...
+**Chosen: file-per-artifact (v1).** Each binary artifact is its own file.
+
+- *Pro:* Simple to implement, debug, and inspect.
+- *Pro:* Works well with existing provider abstractions.
+- *Con:* Many small files on disk for complex indices.
+- *Alternative:* Packed archive format — fewer files but adds complexity; deferred to future work.
+
+### Dual-Path Legacy Loading vs. One-Time Migration
+
+**Chosen: dual-path loading with no required migration.** Readers detect the root format and dispatch to either the new context-first path or the legacy path.
```
arrayka: We need to ensure that M365 is capable of detecting the format to support dual-path loading.

rfcs/00002-serialization.md
```diff
@@ -0,0 +1,760 @@
...
+
+- *Pro:* No forced migration step; legacy indices remain loadable indefinitely.
+- *Pro:* Writers always emit the new format, so the ecosystem naturally converges.
```
arrayka: M365 currently owns the logic for extracting existing raw vectors from the graph blob and using them as input for graph updates. We need to make sure this proposal does not break or regress that functionality.

rfcs/00002-serialization.md
```diff
@@ -0,0 +1,760 @@
+# Index Serialization and Backward Compatibility
+
+| | |
+|---|---|
+| **Authors** | Suhas Jayaram Subramanya |
```
arrayka, line 5: @suhasjs , @harsha-simhadri , let’s iterate on this internally a bit, and then loop in the M365 tech leads for review.

arrayka (changes requested):

rfcs/00002-serialization.md
randybird, line 1: Can we brief how to handle the migration from the original index?
And what is the risk if the original non-version index stays?

rfcs/00002-serialization.md
```diff
@@ -0,0 +1,760 @@
...
+
+impl Version {
+    pub const fn new(major: u32, minor: u32, patch: u32) -> Self {
+        Self { major, minor, patch }
+    }
+
+    /// Returns true if `self` is the immediate SemVer predecessor of `other`
+    /// within the same major (e.g., 0.2.5 -> 0.3.0 or 1.4.2 -> 1.4.3).
+    /// Used to validate required N-1 compatibility.
+    pub fn is_prev_of(&self, other: &Version) -> bool {
```
wuw92: Curious where `is_prev_of` is intended to be used. Section 2 dispatch uses` is_compatible_with`, and I didn't spot `is_prev_of` referenced elsewhere in the RFC.

rfcs/00002-serialization.md
```diff
@@ -0,0 +1,760 @@
...
+Versioning rule:
+- Increment the component `VERSION` whenever any of the following changes:
+  - serialized layout changes,
+  - fields are added/removed/renamed/retyped,
+  - field semantics or interpretation changes,
+  - member usage changes that affect reconstruction behavior.
+
+This rule applies independently to each component, so nested components can evolve at different rates.
+
+### 2. Core `Load` and `Save` Contracts (Context-First, Async)
```
wuw92: Disk-based indices don't load the graph blob into memory, they keep a file handle open and do random I/O on the hot path (e.g., reading a node's vector + neighbors by sector offset given a vector_id). 
How does the RFC propose to handle this access pattern?

rfcs/00002-serialization.md
```diff
@@ -0,0 +1,760 @@
...
+    "l_build": 100,
+    "max_occlusion_size": 750,
+    "max_backedges": 32,
+    "max_minibatch_par": 1,
+    "intra_batch_candidates": "all",
+    "saturate_after_prune": true,
+    "files": [
+      {
+        "key": "disk_graph_payload",
+        "name": "d3f90f3c-6e8d-4baf-875b-bf4c4b2c3f9f-graph-disk_graph_payload.data",
```
wuw92: The current proposal ties deterministic file names in the manifest (UUID-lineage-key.data). In M365, the storage layer assigns its own opaque names (such as index file to "stream_id1", pq file to "stream_id2") to artifacts, and the read API accepts these storage-assigned names. This means DiskANN determines the file names, but the caller's storage layer doesn't necessarily preserve them — the caller would need to parse the metadata to discover what file names DiskANN expects, then manage the mapping to its own storage identifiers.

How does the proposal expect callers to handle environments where file naming is not controlled by DiskANN?

rfcs/00002-serialization.md
```diff
@@ -0,0 +1,760 @@
...
+| Crate | Mechanism | Notes |
+|-------|-----------|-------|
+| `diskann-disk` | `GraphHeader` with `GraphLayoutVersion` | Sector-aligned; validates disk layout. |
+| `diskann` | `SaveWith` / `LoadWith` with provider contexts | Persists graph (`save_graph`/`load_graph`), vector data (`.data` via `.bin`), and quantization artifacts (PQ/SQ), but without a unified recursive metadata contract. |
+
+### Problem Statement
+
+There is no unified serialization contract across DiskANN crates. Each crate uses its own ad hoc mechanism for versioning and persistence, making it impossible to reliably detect index format versions, guarantee backward compatibility, or evolve components independently. This leads to costly full index rebuilds on upgrades and fragile loader code.
+
+### Goals
```
arrayka, line 38: I think it would be valuable to describe user scenarios targeted by this RFC.

rfcs/00002-serialization.md
suhasjs: This proposal is not converging the two formats into one. We can keep the disk-optimized format for disk-based indexes (not impacted by this proposal).

rfcs/00002-serialization.md
suhasjs: Current implementation makes deprecation more explicit through version checks (not just major version).

rfcs/00002-serialization.md
suhasjs: We plan to have save/load traits eventually replace the existing serialization/deserialization functionality. Until migration is complete, we can keep both APIs active and deprecate the existing set once done.

rfcs/00002-serialization.md
suhasjs: Does not change the on-disk format

rfcs/00002-serialization.md
suhasjs: A post-processing step can handle the artifact renaming to maintain compatibility with existing pipelines.

PR #808, opened 2026-03-02
refactor the filter code in the garnet provider 

hailangx:
### Label Provider Refactor

* Changed `GarnetQueryLabelProvider` from a raw pointer-based struct to a lifetime-parameterized struct that borrows a slice.  

### Removal of Obsolete Wrapper

* Removed the `FilteredSearchResults` wrapper, which previously handled conversion between internal and external IDs for search results. This is no longer needed with previous change to expose generic external id in the beta search strategy.

### DIFF
--- a/diskann-garnet/src/labels.rs
+++ b/diskann-garnet/src/labels.rs
@@ -22,30 +22,43 @@ use diskann::graph::index::QueryLabelProvider;
 
 /// A zero-copy bitmap-based label provider for Garnet filtered vector search.
 ///
-/// Borrows the bitmap data for lifetime `'a`. No allocation — bit lookups
-/// are performed directly on the caller's memory.
+/// Holds a raw pointer to the bitmap data owned by the FFI caller.
+/// No allocation — bit lookups are performed directly on the caller's memory.
 ///
-/// For FFI usage, [`from_raw`](Self::from_raw) constructs a
-/// `GarnetQueryLabelProvider<'static>` from a raw pointer; the caller must
-/// guarantee the pointer remains valid for the duration of the search call.
-#[derive(Clone)]
-pub struct GarnetQueryLabelProvider<'a> {
-    data: &'a [u8],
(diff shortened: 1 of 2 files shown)


diskann-garnet/src/labels.rs
```diff
@@ -23,43 +22,30 @@ use diskann::graph::index::QueryLabelProvider;
...
-pub struct GarnetQueryLabelProvider {
-    data: *const u8,
-    len: usize,
+/// For FFI usage, [`from_raw`](Self::from_raw) constructs a
+/// `GarnetQueryLabelProvider<'static>` from a raw pointer; the caller must
+/// guarantee the pointer remains valid for the duration of the search call.
+#[derive(Clone)]
+pub struct GarnetQueryLabelProvider<'a> {
+    data: &'a [u8],
 }
```
hildebrandmw, line 36: Apologies, I don't think I was entirely clear on the feedback in #800, I meant that it's reasonable to have a lifetime in the backend use of Strategies, not that `GarnetQueryLabelProvder` needs a lifetime. Without that change, adding a lifetime bound to `GarnetQueryLabelProvider` does not really solve anything, especially since we need it to be `'static` to use it anyways. And this actually feels worse since it forces the materialization of a `&'static [u8]`. At least when there is a raw pointer involved, programmers are primed to think "something is going on".

Without making changes to the core DiskANN to accept non-'static types, I suggest we keep this as-is.

hailangx: yeah, to remove the static we need to remove it from the query computer trait. 

I don't see much difference regarding the raw pointers when initially it we also have the unsafe convert from raw pointers in this version but just have to mark it under static trick.  And when remove the static from the core DiskANN would it be a nature change here to remove the static

hailangx: @copilot  can you help revert this GarnetQueryLabelProvider  to  original version

PR #824, opened 2026-03-11
[diskann-garnet] Allow configurable metric types

tiagonapoli:

Allow Garnet to configure the Metric type to be used by DiskANN

### DIFF
--- a/diskann-garnet/src/ffi_recall_tests.rs
+++ b/diskann-garnet/src/ffi_recall_tests.rs
@@ -6,7 +6,8 @@
 mod tests {
     use std::{collections::HashMap, ffi::c_void, mem, ptr};
 
-    use diskann_vector::distance::Metric;
+    use diskann_vector::PureDistanceFunction;
+    use diskann_vector::distance::{Cosine, Metric, SquaredL2};
 
     use crate::{
         VectorQuantType, VectorValueType, create_index, drop_index, garnet::Context, insert,
@@ -37,31 +38,13 @@ mod tests {
         }
     }
 
-    fn squared_l2(a: &[f32], b: &[f32]) -> f32 {
-        let mut sum = 0.0f32;
-        for i in 0..a.len() {
-            let diff = a[i] - b[i];
-            sum += diff * diff;
-        }
-        sum
-    }
-
-    fn cosine_distance(a: &[f32], b: &[f32]) -> f32 {
-        let mut dot = 0.0f32;
-        let mut a_norm_sq = 0.0f32;
-        let mut b_norm_sq = 0.0f32;
-        for i in 0..a.len() {
-            dot += a[i] * b[i];
-            a_norm_sq += a[i] * a[i];
-            b_norm_sq += b[i] * b[i];
-        }
-        let a_norm = a_norm_sq.sqrt();
-        let b_norm = b_norm_sq.sqrt();
-        if a_norm > 0.0 && b_norm > 0.0 {
-            1.0 - dot / (a_norm * b_norm)
-        } else {
-            1.0
-        }
+    /// Wraps a `PureDistanceFunction` into a plain `fn` pointer so it can be
+    /// passed to helpers that accept `fn(&[f32], &[f32]) -> f32`.
+    fn distance_fn<T>(a: &[f32], b: &[f32]) -> f32
+    where
+        T: for<'a, 'b> PureDistanceFunction<&'a [f32], &'b [f32], f32>,
+    {
+        T::evaluate(a, b)
     }
 
     /// Brute-force k nearest neighbors. Returns the IDs of the k closest
@@ -286,7 +269,15 @@ mod tests {
     /// Helper: create an L2 index, insert grid vectors, query each, return recall.
     fn run_grid_recall(store: &Store, dimensions: u32, grid_size: usize, k: usize) -> f64 {
         let (ids, vectors) = generate_grid_vectors(dimensions as usize, grid_size);
-        run_recall(store, dimensions, Metric::L2, &ids, &vectors, k, squared_l2)
+        run_recall(
+            store,
+            dimensions,
+            Metric::L2,
+            &ids,
+            &vectors,
+            k,
+            distance_fn::<SquaredL2>,
+        )
     }
 
     #[test]
@@ -341,7 +332,15 @@ mod tests {
     /// Helper: create a cosine index, insert circle vectors, query each, return recall.
     fn run_circle_recall(store: &Store, point_count: usize, radius: f32, k: usize) -> f64 {
         let (ids, vectors) = generate_circle_vectors(point_count, radius);
-        run_recall(store, 2, Metric::Cosine, &ids, &vectors, k, cosine_distance)
+        run_recall(
+            store,
+            2,
+            Metric::Cosine,
+            &ids,
+            &vectors,
+            k,
+            distance_fn::<Cosine>,
+        )
     }
 
     /// Circle with 100 points, radius=1.0, cosine distance, k=5


diskann-garnet/src/lib.rs
```diff
@@ -195,13 +198,22 @@ pub unsafe extern "C" fn create_index(
...
     delete_callback: DeleteCallback,
     rmw_callback: ReadModifyWriteCallback,
 ) -> *const c_void {
+    let metric_type = match metric_type {
+        x if x == Metric::Cosine as i32 => Metric::Cosine,
+        x if x == Metric::InnerProduct as i32 => Metric::InnerProduct,
+        x if x == Metric::L2 as i32 => Metric::L2,
+        x if x == Metric::CosineNormalized as i32 => Metric::CosineNormalized,
+        _ => return ptr::null(),
+    };
```
tiagonapoli: > FFI metric_type is accepted as a raw i32 and validated by comparing against Metric::... as i32. Because Metric has no explicit discriminant values, its numeric representation can change if variants are reordered/extended, which would silently break the FFI contract. Consider introducing a Garnet-specific #[repr(i32)] metric enum with explicit values (or otherwise pin the values) and map that to diskann_vector::Metric.

@metajack copilot had this comment on review, whcch I think makes sense - should I create hardcoded constants in diskann-garnet folder? Or change the Metric enum?

metajack: Technically the Metric variants do have an associated constant. I don't think it's useful to define duplicate ones just to get the type to match.

Probably the better thing to do would be to have a `TryFrom<i32> for Metric` impl defined in `diskann-vector/distance/metric.rs`, and that would basically abstract out this match statement.

metajack: Also, if you add a `From<Metric> for i32` you can then just use `.into()` instead of `as i32` everywhere.

tiagonapoli: I had to add these extensions in diskann-vector, the compiler wouldn't allow to add in diskann-garnet since Metric was defined in another crate

diskann-garnet/src/lib.rs
```diff
@@ -195,13 +197,19 @@ pub unsafe extern "C" fn create_index(
...
     l_build: u32,
     max_degree: u32,
     read_callback: ReadCallback,
     write_callback: WriteCallback,
     delete_callback: DeleteCallback,
     rmw_callback: ReadModifyWriteCallback,
 ) -> *const c_void {
+    let metric_type = match Metric::try_from(metric_type) {
+        Ok(m) => m,
+        Err(_) => return ptr::null(),
```
hildebrandmw, line 212: Not blocking for this PR - but a point of curiosity: is there no mechanism (be-it logging or otherwise) for reporting the nature of an error? In this context - the error is simply dropped, which can make debugging challenging.

diskann-garnet/src/ffi_tests.rs
```diff
@@ -478,91 +557,464 @@ mod tests {
...
+        k: usize,
+    ) -> Vec<String> {
+        let mut scored: Vec<(&String, f32)> = ids
+            .iter()
+            .zip(vectors.iter())
+            .map(|(id, vec)| {
+                let dist: f32 = vec
+                    .iter()
+                    .zip(query.iter())
+                    .map(|(a, b)| (a - b) * (a - b))
```
hildebrandmw: Small nit (non-blocking): Use the `squared_l2` function here?

diskann-garnet/src/ffi_tests.rs
```diff
@@ -478,91 +557,464 @@ mod tests {
...
+    /// Count intersection by grouping results by distance (handles ties).
+    /// When multiple vectors are equidistant, the exact IDs may differ
+    /// between brute-force and ANN, but the distance counts must match.
+    fn distance_based_intersection(
+        vectors: &[Vec<f32>],
+        ids: &[String],
+        query: &[f32],
+        expected: &[String],
+        actual: &[String],
+    ) -> usize {
```
hildebrandmw: This is potentially a brittle test: it relies on distances being exact for what's returned from the index versus what's calculated here "manually". This **does** work for the grid problem because the points are all integers and the vectors are small, thus the different floating point association strategies between the ad-hoc L2 implementation and the one provided by `diskann-vector` aren't showing up.

While not necessarily a blocker for the PR, it **is** worth acknowledging this limitation in case folks are tempted to use this infrastructure for other tests (e.g., the cosine test below, where I noticed this check is omitted).

This could be made to be a more reusable test harness by creating a histogram of discrete buckets instead of using the exact floating-point bit-pattern.

diskann-garnet/src/ffi_tests.rs
```diff
@@ -478,91 +557,464 @@ mod tests {
...
+        vectors: &[Vec<f32>],
+        query: &[f32],
+        k: usize,
+    ) -> Vec<String> {
+        let q_norm: f32 = query.iter().map(|x| x * x).sum::<f32>().sqrt();
+        let mut scored: Vec<(&String, f32)> = ids
+            .iter()
+            .zip(vectors.iter())
+            .map(|(id, vec)| {
+                let dot: f32 = vec.iter().zip(query.iter()).map(|(a, b)| a * b).sum();
```
hildebrandmw: Any reason not to use:
```rust
.map(|(id, vec)| {
    (id, diskann_vector::distance::Cosine::evaluate(query, &*vec))
})
```

diskann-garnet/src/ffi_tests.rs
```diff
@@ -478,91 +557,464 @@ mod tests {
...
+            let mut vec = vec![0.0f32; dimensions];
+            let mut pos = i;
+            let mut coords = Vec::with_capacity(dimensions);
+            for d in (0..dimensions).rev() {
+                let coord = pos % grid_size;
+                vec[d] = coord as f32;
+                coords.push(coord);
+                pos /= grid_size;
+            }
+            coords.reverse();
```
hildebrandmw: There is some existing infrastructure [`diskann`](https://github.com/microsoft/DiskANN/blob/965d31a69161227718bd887ac31ccf0c563e3d8f/diskann/src/graph/test/synthetic.rs#L31-L33) for efficiently generating the grid. Not sure if that would be helpful at all here, but it's there.

diskann-garnet/src/ffi_tests.rs
tiagonapoli: I agree, taht was a poor choice from Copilot - I updated to use a histogram approach with some epsilon

diskann-garnet/src/lib.rs
tiagonapoli: For now we don't have anything in Garnet to log these kinds of issues unfortunately - It's a follow up work from VectorSets with StoreV2, to improve observability. Right now, at least for MetricTypes, the options offered to customers will be through a predefined set of flags in VADD. Then we can have unit tests that will ensure that these predefined set do not cause errors creating the index

diskann-garnet/src/ffi_tests.rs
tiagonapoli: Fixed

diskann-garnet/src/ffi_tests.rs
tiagonapoli: I didn't know about it, thanks for sharing, will update

diskann-garnet/src/ffi_tests.rs
tiagonapoli: Ohh, I didn't know about this helper - but given that here we're using small grids and there's just a small helper function to create the grids, do you think it's fine to continue using this one?

hildebrandmw: Yeah - no biggie. Was just offering it!

diskann-garnet/src/lib.rs
hildebrandmw: Like I said - not a blocker, but definitely something we'll will want to think about sooner rather than later.

PR #828, opened 2026-03-14
Overhaul Post-Processing

hildebrandmw:
Decouple post-processing from `SearchStrategy` so that search methods accept an explicit post-processor parameter. This enables callers to supply custom post-processing logic (reranking, filtering, ID translation) without defining new strategy types.

## Search trait changes

The `Search` trait now requires `S: SearchStrategy<DP, T>` at the trait level, and introduces two new generic parameters on the `search` method:

- `O` - the output type written to the buffer (used to be a parameter of `SearchStrategy`)
- `PP` — the post-processor, bounded by `for<'a> SearchPostProcess<S::SearchAccessor<'a>, T, O> + Send + Sync`
- `OB` — the output buffer, bounded by `SearchOutputBuffer<O> + Send + ?Sized`

Previously, `OB` was a trait-level generic and the post-processor was obtained internally from the strategy. Now both are method-level generics, giving callers control over each independently.

The output "id" type `O` has been removed from `SearchStrategy` entirely, which greatly simplifies things.

At the `DiskANNIndex` level, two entry points are provided:

- `search()` — requires `S: DefaultPostProcessor`, creates the strategy's default processor automatically, then delegates to `search_with()`. This is the common path.
- `search_with()` — requires only `S: SearchStrategy`, accepts any compatible `PP`. This is the extension point for custom post-processing.

## DefaultPostProcessor

Strategies opt in to default post-processing by implementing `DefaultPostProcessor`:

```rust
pub trait DefaultPostProcessor<Provider, T, O>: SearchStrategy<Provider, T> {
    type Processor: for<'a> SearchPostProcess<Self::SearchAccessor<'a>, T, O> + Send + Sync;
    fn default_post_processor(&self) -> Self::Processor;
}
```

The `default_post_processor!` macro covers the common case where the processor is `Default`-constructible. `DefaultSearchStrategy` is a convenience alias for `SearchStrategy + DefaultPostProcessor`.

## Start-point filtering

The in-mem code used to rely on a special `Internal` type to customize post-processing for the inplace delete pipeline.
Now that post-processing is customizable, this workaround is no longer needed.

Start-point filtering is composed at the type level using the existing `Pipeline` mechanism:

- `HasDefaultProcessor::Processor = Pipeline<FilterStartPoints, RemoveDeletedIdsAndCopy>` — filters start points during regular search
- `InplaceDeleteStrategy::SearchPostProcessor = RemoveDeletedIdsAndCopy` — no start-point filtering during delete re-pruning

## Range search

To respond to moving the output buffer to a method level geneirc, range search now writes results directly through a `DistanceFiltered` output buffer wrapper that applies radius and inner-radius filtering inline during post-processing. This replaces the previous approach of allocating intermediate `Vec`s and copying results through an `IdDistance` buffer.

`SearchOutputBuffer` is now implemented for `Vec<Neighbor<I>>`, providing an unbounded growable buffer suitable for range search and other cases where the result count is not known in advance.

The `RangeSearchResults` type has been removed now that a `SearchOutputBuffer` is properly used.

## Inplace Delete
Since the post-processor is now a little decoupled from `SearchStrategy`, rustc is unable to fully resolve the existing bounds for the "caching" provider. Try as I might, I could not get anything to work without adding a `DeleteSearchAccessor` associated type to `InplaceDeleteStrategy`. At the end of the day, I think this is a compromise we'll have to live with.

## Migration guide

**If you implement `SearchStrategy`** — minimal changes required. Remove the `O` parfameter (if needed). Implement `DefaultPostProcessor` with your current post-processor if you want your strategy to work with `index.search()`. Use the `default_post_processor!` macro for simple cases:

```rust
impl DefaultPostProcessor<MyProvider, [f32]> for MyStrategy {
    default_post_processor!(Pipeline<FilterStartPoints, RemoveDeletedIdsAndCopy>);
}
```

**If you call `index.search()`** — no API change for strategies that implement `DefaultPostProcessor` (all built-in strategies do).

**If you need custom post-processing** — use `index.search_with()` and pass your post-processor directly:

```rust
let processor = MyCustomPostProcessor::new(/* ... */);
let stats = index.search_with(params, &strategy, processor, &context, &query, &mut output).await?;
```

**If you implement `Search`** — the method signature has changed. `PP` and `OB` are now method-level generics, and `processor` is an explicit parameter:

```rust
fn search<O, PP, OB>(
    self,
    index: &DiskANNIndex<DP>,
    strategy: &S,
    processor: PP,       // new
    context: &DP::Context,
    query: &T,
    output: &mut OB,
) -> impl SendFuture<ANNResult<Self::Output>>
where
    O: Send,
    PP: for<'a> SearchPostProcess<S::SearchAccessor<'a>, T, O> + Send + Sync,
    OB: SearchOutputBuffer<O> + Send + ?Sized;
```

**If you use range search results** — `RangeSearchOutput` no longer carries `ids` and `distances` fields. Results are written to the output buffer passed to `search()`. Use `Vec<Neighbor<u32>>` (or any `SearchOutputBuffer` impl) to collect them.

**If you implement `InplaceDeleteStrategy`**: Add the `DeleteSearchAccessor` associated type, ensuring it is the same as the accessor selected by the `SearchStrategy`. This is redundant information, but needed to help constrain downstream trait bounds.

### DIFF
--- a/diskann-garnet/src/provider.rs
+++ b/diskann-garnet/src/provider.rs
@@ -760,7 +760,7 @@ impl<'a, T: VectorRepr> SearchPostProcess<FullAccessor<'a, T>, [T], GarnetId> fo
     }
 }
 
-impl<T: VectorRepr> SearchStrategy<GarnetProvider<T>, [T], GarnetId> for FullPrecision {
+impl<T: VectorRepr> SearchStrategy<GarnetProvider<T>, [T]> for FullPrecision {
     type SearchAccessor<'a> = FullAccessor<'a, T>;
     type SearchAccessorError = GarnetProviderError;
     type QueryComputer = T::QueryDistance;
@@ -776,20 +776,6 @@ impl<T: VectorRepr> SearchStrategy<GarnetProvider<T>, [T], GarnetId> for FullPre
 
 impl<T: VectorRepr> DefaultPostProcessor<GarnetProvider<T>, [T], GarnetId> for FullPrecision {
     default_post_processor!(glue::Pipeline<glue::FilterStartPoints, CopyExternalIds>);
-}
-
-impl<T: VectorRepr> SearchStrategy<GarnetProvider<T>, [T], u32> for FullPrecision {
-    type SearchAccessor<'a> = FullAccessor<'a, T>;
-    type SearchAccessorError = GarnetProviderError;
-    type QueryComputer = T::QueryDistance;
-
-    fn search_accessor<'a>(
-        &'a self,
-        provider: &'a GarnetProvider<T>,
-        context: &'a <GarnetProvider<T> as DataProvider>::Context,
-    ) -> Result<Self::SearchAccessor<'a>, Self::SearchAccessorError> {
-        Ok(FullAccessor::new(provider, context, true))
-    }
 }
 
 impl<T: VectorRepr> PruneStrategy<GarnetProvider<T>> for FullPrecision {
--- a/diskann/src/graph/search/range_search.rs
+++ b/diskann/src/graph/search/range_search.rs
@@ -157,16 +157,15 @@ impl Range {
     }
 }
 
-impl<DP, S, T, O> Search<DP, S, T, O> for Range
+impl<DP, S, T> Search<DP, S, T> for Range
 where
     DP: DataProvider,
-    S: glue::SearchStrategy<DP, T, O>,
+    S: glue::SearchStrategy<DP, T>,
     T: Sync + ?Sized,
-    O: Send + Default + Clone,
 {
     type Output = SearchStats;
 
-    fn search<PP, OB>(
+    fn search<O, PP, OB>(
         self,
         index: &DiskANNIndex<DP>,
         strategy: &S,
@@ -176,6 +175,7 @@ where
         output: &mut OB,
     ) -> impl SendFuture<ANNResult<Self::Output>>
     where
+        O: Send,
         PP: for<'a> glue::SearchPostProcess<S::SearchAccessor<'a>, T, O> + Send + Sync,
         OB: SearchOutputBuffer<O> + Send + ?Sized,
     {
--- a/diskann-disk/src/search/provider/disk_provider.rs
+++ b/diskann-disk/src/search/provider/disk_provider.rs
@@ -340,15 +340,8 @@ where
     }
 }
 
-impl<'this, Data, ProviderFactory>
-    SearchStrategy<
-        DiskProvider<Data>,
-        [Data::VectorDataType],
-        (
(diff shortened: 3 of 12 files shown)


diskann-garnet/src/provider.rs
```diff
@@ -810,10 +784,6 @@ impl<T: VectorRepr> SearchStrategy<GarnetProvider<T>, [T], u32> for FullPrecisio
     ) -> Result<Self::SearchAccessor<'a>, Self::SearchAccessorError> {
```
hildebrandmw, line 810: @metajack - I removed the "default" post-processing definition from the `FullPrecision` search strategy targeting internal IDs. As far as I could tell, it isn't used now that `InplaceDeleteStrategy` provides its own post-processor. Happy to add it back though you want!

diskann/src/graph/search/range_search.rs
```diff
@@ -241,63 +236,83 @@ where
...
 
-            // Post-process results
-            let mut result_ids: Vec<O> = vec![O::default(); scratch.in_range.len()];
-            let mut result_dists: Vec<f32> = vec![f32::MAX; scratch.in_range.len()];
-
-            let mut output_buffer = search_output_buffer::IdDistance::new(
-                result_ids.as_mut_slice(),
-                result_dists.as_mut_slice(),
-            );
+            // Post-process results directly into the output buffer, filtering by radius.
```
hildebrandmw, line 239: @hailangx - Can you help me double-check the change in logic here? The big difference is that results are written directly into a `SearchOutputBuffer` (this is possible now that the interface has been relaxed into an "append-only" interface - meaning that dynamically sized buffers can be a thing).

One thing I'm a little worried about is that it's still **possible** to call range-search with a fixed size buffer, and I'm wondering if we should invest in generic machinery to make that impossible. One advantage of still allowing a fixed-size buffer is that it provides a natural way of limiting the number of results - but may be non-obvious to callers. What do you think?

hailangx: are we able to have 'SearchOutputBuffer' indicate the buffer type ? if it's a fixed-size buffer we will stop there, and it should have an upper limit for how much this buffer can grow into.

hildebrandmw: The `size_hint` method of `SearchOutputBuffer` returns `None` if the buffer is unbounded. In either case, though, this will work in a sane way. It the buffer is unbounded, then all found results will get added to the buffer. If the buffer *is* bounded with an upper bound `B`, then only the first `B` items (that pass the predicate) will get added before `BufferState::Full` is returned and post-processing stops.

Does that answer your question?

PR #833, opened 2026-03-16
[minmax] Expose minmax heterogeneous distance api

arkrishn94:
I missed exposing the public api for minmax distance functions from #798. 

Bumped version to 0.49.1

### DIFF
--- a/diskann-quantization/src/minmax/vectors.rs
+++ b/diskann-quantization/src/minmax/vectors.rs
@@ -203,16 +203,16 @@ pub type FullQueryMut<'a> = slice::SliceMut<'a, f32, FullQueryMeta>;
 // Compensated Distances //
 ///////////////////////////
 #[inline(always)]
-fn kernel<const N: usize, const M: usize, F>(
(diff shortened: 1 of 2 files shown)


diskann-quantization/src/minmax/vectors.rs
```diff
@@ -386,19 +378,18 @@ where
...
-            DataRef<'a, NBITS>,
-            DataRef<'b, NBITS>,
+            DataRef<'a, N>,
+            DataRef<'b, M>,
             distances::MathematicalResult<f32>,
         >,
 {
     // 1 - <X, Y> / (|X| * |Y|)
-    fn evaluate(x: DataRef<'_, NBITS>, y: DataRef<'_, NBITS>) -> distances::Result<f32> {
+    fn evaluate(x: DataRef<'_, N>, y: DataRef<'_, M>) -> distances::Result<f32> {
```
arrayka: Let's keep it consistent to other parts of this file. Let's keep using NBITS and MBITS

PR #837, opened 2026-03-16
Add baseline tests for insert/multi-insert.

hildebrandmw:
Add baseline tests for `insert` and `multi_insert`, and address two issues in multi-insert:

1. Relax the graph bootstrap trigger to activate even when some intra-batch candidates are used. This handles a case where a large batch is used with a small number of intra-batch candidates, leading to a degenerate graph. Without this change, the candidates for the `intra-batch-candides = 4` case fails to find the correct nearest neighbors for the corner point searches.
2. Sort the aggregated backedges before adding them to the graph. This change makes multi-insert repeatable regardless of the number of threads. The non-determinism comes from the dynamic load balancing mechanism meant to compensate for slow tasks. While the back edges aggregated are the same, the order in which they are collected differs depending on the order of task completion. Sorting them makes the result repeatable. This *may* introduce a "downward" bias in the backedges, so we could consider sorting them by hash rather than by value, but I don't think it's likely to matter much.

## Baseline Tests
The new test file can be found in `diskann/src/graph/test/cases/grid_insert.rs` and uses the baseline caching infrastructure to add the following baselines:

| Grid | Single | Batch (all) | Batch (chunked) |
|------|--------|-------------|-----------------|
| 1D/100 | yes | ibc=none, ibc=all | no |
| 3D/5 | yes | ibc=none, ibc=all, ibc=max(4) | ibc=none, ibc=all |
| 4D/4 | yes | ibc=none, ibc=all, ibc=max(4) | ibc=none, ibc=all 

Where "Batch (chunked)" partitions the vectors to be inserted into multiple mini-batches.


diskann/src/graph/test/cases/mod.rs
```diff
@@ -4,4 +4,11 @@
  */
 
 mod grid;
+mod grid_insert;
 mod inplace_delete;
+
+/// Set to `true` and recompile to include full adjacency list state in participating
+/// baselines.
+///
+/// Useful for debugging regressions, but produces large baseline files.
```
harsha-simhadri, line 13: could you please add a few more comments about what is being dumped here?

diskann/src/graph/test/cases/grid_insert.rs
```diff
@@ -0,0 +1,748 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+//! Baseline regression tests for single and batch insert followed by search.
```
harsha-simhadri, line 6: does this file subsume the `diskann/src/graph/test/cases/grid.rs` file

harsha-simhadri (approved):

diskann/src/graph/test/cases/mod.rs
hildebrandmw: Can you please clarify the ask? The comment already states that the full adjacency list state is included, which is pretty much the only side-effect of this flag at the moment. I'm not sure if adding more information of what that looks like would add signal.

diskann/src/graph/test/cases/grid_insert.rs
hildebrandmw: This test lives along-side. I renamed `grid.rs` to `grid_search.rs` to hopefully more clearly indicate that it's meant to test the just searching on the structured graph, where-as `grid_insert` also runs a full index build.

metajack (approved):

diskann/src/graph/test/cases/mod.rs
hildebrandmw: @harsha-simhadri - Perf numbers for MSTuring-1M have been posted below. What's your opinion on marking this comment as resolved and merging?

hildebrandmw: Resolving per offline communication.

PR #838, opened 2026-03-16
CI maintenance

hildebrandmw:
Perform the following CI chores:

* Add feature tests to `clippy` and include the `disk-index` feature for `diskann-benchmark`.
* Pass `-Dwarnings` explicitly to the clippy jobs. Previously, this wasn't being propagated from `build.rustflags`, meaning that clippy lints weren't blocking check-in.
* Rewrite the `no-default-features` tests to execute serially instead of as multiple jobs. This reduces the CI cache used for this job from ~850 MB for the 9 tested crates to around 160 MB. The job takes a little longer, but still under 1 minute.

Finally, since clippy now correctly respects `-Dwarnings`, there were two lints that needed fixing: one invalid conversion, and one regarding the disallowed overlay method (which doesn't exist when default features are disabled).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,7 +135,7 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
 
-      - name: Clippy all crates (no default features)
+      - name: Clippy - select crates (no default features)
         run: |
           set -euxo pipefail
           crates=(


.github/workflows/ci.yml
```diff
@@ -79,9 +85,40 @@ jobs:
...
+        with:
+          shared-key: clippy
       - name: "clippy --workspace --all-targets"
         run: cargo clippy --locked --workspace --all-targets --no-deps --config "$RUST_CONFIG"
 
+  clippy-features:
+    strategy:
+      matrix:
+        runner:
+          - ubuntu-latest
```
harsha-simhadri, line 96: what about windows-latest instead of ubuntu-latest to cover a different OS as well.

.github/workflows/ci.yml
```diff
@@ -35,8 +40,9 @@ jobs:
     name: basic checks
     runs-on: ubuntu-latest
     needs:
-      - clippy
       - fmt
+      - clippy
```
harsha-simhadri: lets call this `clippy-no-features`?

hildebrandmw: Good idea - renamed to `clippy-default-features`.

.github/workflows/ci.yml
hildebrandmw: Yeah, good point. I updated the features run to run on the combination of windows, ubuntu-x86, and ubuntu-arm. Though now that I think about it - I should include macos as well shouldn't I?

.github/workflows/ci.yml
```diff
@@ -19,6 +19,11 @@ env:
   # This ensures that we do not overwrite the parameters in `.cargo/config.toml` unnecessarily.
   RUST_CONFIG: 'build.rustflags=["-Dwarnings"]'
   RUST_BACKTRACE: 1
+  # The features we want to explicitly test. For example, the `flatbuffers` features
```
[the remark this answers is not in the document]
hildebrandmw: Updated - thanks.

.github/workflows/ci.yml
harsha-simhadri: ok with just windows-latest and ubuntu-aarch64 to save some time.

hildebrandmw: Yeah, good point. I didn't realize the macos runners were taking hours to acquire.

.github/workflows/ci.yml
```diff
@@ -79,9 +84,41 @@ jobs:
           toolchain: ${{ env.rust_stable }}
           components: clippy
       - uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: clippy
```
metajack: Why the same shared key? You can only have one crate in the cache per version of the crate, so if you compile with different features they will blow each other away. Since you are specifically varying features in these two jobs they will thrash the cache.

.github/workflows/ci.yml
```diff
@@ -62,14 +67,14 @@ jobs:
             exit 1
           fi
 
-  clippy:
+  clippy-default-features:
     strategy:
       matrix:
         runner:
           - ubuntu-latest
           - ubuntu-24.04-arm
```
metajack, line 75: Why is windows omitted on this one?

.github/workflows/ci.yml
```diff
@@ -287,15 +325,16 @@ jobs:
           tool: cargo-nextest
 
       - uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: test-workspace
```
metajack: Same here, because features are varied, the cache with thrash.

.github/workflows/ci.yml
hildebrandmw: I'm realizing I don't fully understand how the cache works. There appears to not be any thrashing:

* Run with [default features](https://github.com/microsoft/DiskANN/actions/runs/23171842353/job/67325237219?pr=838)
* Run with [additional features](https://github.com/microsoft/DiskANN/actions/runs/23171842353/job/67325237205?pr=838)

Both hit in the cache and did not write back to the cache. However, it appears that the one with features *did* include a bit of extra compilation. Probably as you described because the cache wasn't perfect. However, it didn't add a *huge* amount of extra time. So I guess the question is: is reducing our cache utilization better, or is having a perfect cache better?

I don't know.

Right now, our cache usage appears to be around 5.23 GB (if Copilot is to be believed, which at this point is dubious). Sharing keys may only get us back like a GB because of all the different platforms.

In any case, I removed the shared keys. If we start to see exploding CI times from cache misses, maybe we'll have to revisit.

I also noticed that I messed up adding `-Dwarnings` originally - apparently when passed via `build.rustflags` it doesn't get propagated to `clippy`. I just pushed a fix. I noticed a warning in an earlier run, so I'll see if it fails now.

hildebrandmw: Boom: https://github.com/microsoft/DiskANN/actions/runs/23216962205/job/67480199468?pr=838

How about that - our CI is actually working now! 😭

.github/workflows/ci.yml
hildebrandmw: Your right - though instead of thrash, it appears that the imperfect cache entries end up rebuilding a bit more. In any case, I removed the shared-key from this PR.

.github/workflows/ci.yml
hildebrandmw: I was thinking of reducing the cache usage, but now I feel like I focused way too much on that. `windows-latest` is not part of this run.

.github/workflows/ci.yml
hildebrandmw: Decision: no mac runners for now.

PR #848, opened 2026-03-18
Quantization Bootstrapping RFC

### LINKED ISSUE #1054: Quantization Bootstrap RFC
Write an RFC describing how to bootstrap quantization in a fresh index where vectors are added incrementally.

metajack:
This document provides a proposal for quantization bootstrapping, allowing DiskANN to grow incrementally from scratch supporting quantization.

[Rendered](https://github.com/microsoft/DiskANN/blob/e906cda82515111650e334fe818fcea4811630bc/rfcs/00000-quantizer-bootstrap.md)


rfcs/00000-quantizer-bootstrap.md
```diff
@@ -0,0 +1,158 @@
...
+
+As the caller designates a strategy to use, we can implement new
+`BootstrappedQuantized` and `BootstrappedHybrid` strategies that layer over
+existing `FullPrecision`, `Quantized`, and `Hybrid` strategies. These new
+bootstrapped strategies will delegate operation to the existing strategies
+depending on the current phase.
+
+*Open question*: How exactly do we do this?
+
+### Quantization Backfill
```
harsha-simhadri: Would backfill require an iterator over the DataProvider?

If so, is it easy to implement this in a threadsafe way and without overheads to other code paths?

metajack: It will not require an `Iterator` implementation, but the DataProvider will have to track the relevant range so that it knows which vectors to backfill, or iterate somehow over all vectors.

harsha-simhadri: "iterate somehow over all vectors." --> how is this different than supporting an iter trait?

metajack: Iter traits require you to bundle up all the iteration state inside the object, and if you want it to be somewhat sane, you'd probably want an `Iterator<Item=InternalId>` where each thing returned is an unquantized vector. This external iteration is possible but not super convenient in our case. It's much easier to scan the FSM and do the quantizatoin that way, instead of having to package up state and reload the FSM on each call to `next()`.

A naive implementation can just track the max internal id and go one by one looking for unquantized vectors. Slightly better would be recording the phase shift vector and just iterating from `0..phase_shift_id`. Those are both fairly easy to implement Iterator for, but if we do it via FSM scanning then we'd want something that iterates internally to avoid reloading the FSM each call to `next()`.

hildebrandmw: I'm dubious on whether an iterator trait is efficiently implementable in the general case and comes with all kinds of draw-backs: do we need to lock down the data store while the iterator is present to prevent mutation? For the FSM case, there's a whole bunch of state that needs to be tracked. How would we potentially parallelize the operation?

Most of these disappear with interior iteration. And while that introduces other problems, the closer you get to the actual provider (i.e., the less you need to abstract away with traits), the easier an implementation will actually be.

rfcs/00860-quantizer-bootstrap.md
```diff
@@ -0,0 +1,198 @@
...
+the background. Once all vectors are quantized, Phase Three begins the normal
+operation of the quantized index.
+
+### Goals
+
+1. Allow quantized indices to start empty and use full precision data only to
+   operate until sufficient vectors are inserted.
+2. Aside from allowing construction of `DiskANNIndex` without providing
+   quantization tables, there should be no user-visible changes to using the index.
+3. Performance should remain as high as possible during the three phases.
```
arrayka, line 66: It would be helpful to make this more concrete. The statement *“performance should remain as high as possible”* is hard to validate as‑is.

If we’re not making any performance commitments during Phase 2, we should call that out explicitly and state that the system operates on a best‑effort basis with no performance guarantees yet.

rfcs/00860-quantizer-bootstrap.md
```diff
@@ -0,0 +1,198 @@
...
+however it likes.
+
+One possibility is to piggy-back on deletion tracking to track quantization
+status of vectors. For example, in diskann-garnet, a free space map is kept that
+tracks deletes. This could be expanded from 1-bit to 2-bits, and the second bit
+used to track whether the vector is quantized. Alternatively, metadata about the
+allocated range can be kept and used to iterate over the unquantized set.
+
+
+## Trade-offs
```
arrayka, line 182: This section explains the problem clearly in plain English, including how different clients handle it. It would be great to merge it into the Problem Statement section at the beginning of the document.

rfcs/00860-quantizer-bootstrap.md
```diff
@@ -0,0 +1,198 @@
+# Quantization Bootstrapping
```
harsha-simhadri, line 1: Give the RFC a non-zero number. Add license header.

rfcs/00860-quantizer-bootstrap.md
```diff
@@ -0,0 +1,198 @@
...
+### Switching Strategies
+
+DiskANN already has the ability to run multiple strategies including hybrid full
+precision and quantized ones. These strategies represent the high level intent,
+but the data provider can choose alternate implementions depending on the data
+available during the current phase.
+
+#### Insertion and Deletion
+
+Insertion and deletion can remain largely the same in the data provider
```
harsha-simhadri, line 96: Do we support insert-delete-insert for the same ID during bootstrap?

rfcs/00860-quantizer-bootstrap.md
```diff
@@ -0,0 +1,198 @@
...
+index or strategy layering, but the complexity this would introduce seems not
+worth the cost.
+
+## Benchmark Results
+
+Since there is no way to build an index currently until quantization tables are
+built, there is no way to benchmark the first two phases. There should be no
+impact during Phase Three to performance.
+
+## Future Work
```
harsha-simhadri, line 216: You had proposed an implementation plan/PR for garnet provider. Does that align with this RFC?

metajack: Yes

rfcs/00860-quantizer-bootstrap.md
metajack: It's up to the data provider, but given how tricky the phase transition are, it's easy to disable ID resuse until phase 3, and that is what Garnet ended up doing.

rfcs/00860-quantizer-bootstrap.md
metajack: Done, although I didn't add the license header since other RFCs don't have this.

rfcs/00860-quantizer-bootstrap.md
metajack: I'm not sure what commitments we can make in a general RFC that could cover many different providers, but I tried to be clearer about the expectation. Namely that Phase 1 and Phase 3 should basically be the same as full precision or quantized only, and that some minimal overhead will happen in phase 2 because of the extra write and the asynchronous backfill.

rfcs/00860-quantizer-bootstrap.md
metajack: I believe I added a short version of this to background.

metajack:
Ok, I've done a quick update to add the number and address the outstanding suggestions. I think this should be ready to go.

rfcs/00860-quantizer-bootstrap.md
```diff
@@ -0,0 +1,213 @@
...
+tables; for spherical, 100 are needed. In order to create an index, these
+vectors must be provided at creation time in order to build the quantization
+tables, at which point each vector is quantized as it is inserted.
+
+This requirement is easy to fulfill when building indexes from existing
+datasets, but when starting from scratch, there is no ability for DiskANN to
+build a quantized index since the quantization tables are a required part of the
+constructor.
+
+Current deployments of DiskANN work around this issue by not allowing index
```
arrayka, line 33: It might be worth explicitly calling out why this is a problem and what we’re trying to improve. Why do we want to move away from the current solutions?

For example, I see later in this doc: "This
requires the Cosmos DB team to maintain all their own infrastructure for the
flat index and the code around transitioning to the graph index."

We could highlight here that the goal is to provide an out-of-the-box solution that anyone can use, without having to maintain their own infrastructure or build custom logic for transitioning to the graph index.

rfcs/00860-quantizer-bootstrap.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Switching Strategies
+
+DiskANN already has the ability to run multiple strategies including hybrid full
+precision and quantized ones. These strategies represent the high level intent,
+but the data provider can choose alternate implementions depending on the data
+available during the current phase.
+
+Instead of passing different strategies to `insert()` and `search()` based on the 
+current phase, instead some ur-strategy will be used that changes behavior based
```
arrayka: minor: ur-strategy? could you please define this term?

rfcs/00860-quantizer-bootstrap.md
```diff
@@ -0,0 +1,213 @@
...
+
+To avoid complexity of hybrid distance calculations, either full precision
+distances will be used (Phase One and Two) or quantized distances will be used
+(Phase Three). If a hybrid strategy is in use, then the hybrid distances will
+not be used until Phase Three.
+
+Since vector data may be in one of two representations, the `Accessor::Element`
+type should be `Poly<u8>` (this should be over-aligned to the correct alignment
+for the full precision element type), and the data provider should interpret
+based on data size. The distance and query computers will also need
```
arrayka: Do I read it correctly that the proposal is to use `SetElement<Poly<u8>>`?
If so, it weakens type safety - nothing stops code from misinterpreting bytes as the wrong representation.
Is it possible to come up with a strong-typed solution, like enum, Either<A,B> or similar:
enum StoredVector {
    FullPrecision(AlignedVec<f32>),
    Quantized(Vec<u8>),
}

rfcs/00860-quantizer-bootstrap.md
```diff
@@ -0,0 +1,213 @@
...
+flat index and the code around transitioning to the graph index.
+
+This proposal mitigates the downsides while still allowing the integrator to
+retain control over key performance details.
+
+This proposal also entirely encapsulates this inside the `DataProvider`
+implementation. Alternatively, one could attempt to solve this with some kind of
+index or strategy layering, but the complexity this would introduce seems not
+worth the cost. This was previously the direction of the caching provider, which
+proved very unwieldy.
```
arrayka, line 202: It’s not immediately clear to me why index layering adds significant complexity. Index composition feels like a natural approach to this problem, unless there are requirements that are difficult to satisfy with two indices.

rfcs/00860-quantizer-bootstrap.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Benchmark Results
+
+Since there is no way to build an index currently until quantization tables are
+built, there is no way to benchmark the first two phases. There should be no
+impact during Phase Three to performance. Specifically, during phase one, the
```
arrayka, line 208: True, assuming we stick with the same graph search algorithm for the first two phases.
Is a flat_index an alternative here—it might be faster at small scale than graph traversal?

rfcs/00860-quantizer-bootstrap.md
metajack: You are welcome to have a stab at it. I tried very hard with the caching data provider and it was hundreds of lines of trait bounds and several problems were never fully solved. The Rust type system just isn't quite expressive enough for the level of complexity we need. If we copied vectors it would be fine, but because we support owned and borrowed vectors in most interfaces, the lack of more sophisticated handling of higher level trait bounds makes some things inexpressible.

After that fiasco, Mark managed to make a slightly simpler version that replaced the pure layered version, but in the end we've decided to throw it away.

In general we run into most of these issues because you want tight control of performance, and layering works against you a bit there.

rfcs/00860-quantizer-bootstrap.md
metajack: It's not a mutually exclusive alternative, they can both be used. Note that by waiting to build an index later, you are deferring all the pruning of the first inserts, which will make phase two much more expensive.

Probably if you *really* care about early index performance you would want both so that you build your quantized index incrementally as vectors are inserted, but maintain a flat index to use until the quantized one reaches phase 3. However, given that these indexes are typically used to hold millions of vectors and the most data hungry quantizer needs ~50-100k vectors to train, the useful window for better performance is pretty small.

For example, in Garnet with the Spherical1Bit quantizer, we need about 1k vectors. We can easily insert 1-2k vectors a second, so on a production system I would expect the quant bootstrap phases to last seconds to minutes. I don't think the tradeoff in complexity is worth it to get slightly better perf in that small time span.

rfcs/00860-quantizer-bootstrap.md
metajack: This might be possible. I'll add it as an alternative in this RFC.

rfcs/00860-quantizer-bootstrap.md
metajack: Fixed. I had misused the term. I think I meant to use 'uber' not 'ur'. Ur means primitive/original. Like the music scores that Bach wrote are referred to as ur-texts.

rfcs/00860-quantizer-bootstrap.md
metajack: I added more words to that effect.

metajack:
Another pass complete.

harsha-simhadri (approved):
Could you please update the "updated date"

metajack:
I will move this RFC to a design document in the diskann-garnet directory. Closing this.

PR #857, opened 2026-03-20
Add DiskANN benchmark pipeline to GitHub Actions

YuanyuanTian-hh:
### Summary

Adds an automated benchmark pipeline to GitHub Actions, enabling regression detection for every PR. The pipeline builds and searches two public ANN datasets, compares performance against a baseline branch, and gates on configurable thresholds for recall, QPS, latency, and I/O metrics.

### What's Added

| File | Description |
|------|-------------|
| `.github/workflows/benchmarks.yml` | Main workflow  2 parallel jobs (Wikipedia-100K + OpenAI-100K), triggered via `workflow_dispatch` with configurable baseline ref |
| `.github/workflows/benchmarks-aa.yml` | Daily A/A stability test (main vs main) at 9 AM UTC  creates a GitHub issue on failure to notify `@microsoft/diskann-admin` |
| `.github/scripts/benchmark_validate.py` | Single script that diffs two benchmark JSONs, validates against thresholds, writes Markdown summary, and posts PR comments on regression (stdlib only, zero pip dependencies) |
| `diskann-benchmark/perf_test_inputs/wikipedia-100K-disk-index.json` | Benchmark config: 768-dim, inner_product, R=59, L=72, 4 threads |
| `diskann-benchmark/perf_test_inputs/openai-100K-disk-index.json` | Benchmark config: 1536-dim, squared_l2, R=59, L=80, SQ_1_2.0, 4 threads |

### How It Works

1. **Checkout** both current branch and baseline (defaults to `main`)
2. **Download** pre-packaged datasets from [diskann-benchmark-data](https://github.com/YuanyuanTian-hh/diskann-benchmark-data/releases/tag/v1) (sourced from [big-ann-benchmarks](https://github.com/harsha-simhadri/big-ann-benchmarks))
3. **Build & search** disk index on both branches using `diskann-benchmark` crate
4. **Validate**  a single Python script (`benchmark_validate.py`) reads both JSON outputs, compares all metrics, writes a Markdown diff table to the step summary, checks thresholds, and posts a PR comment if regression is detected
5. **Upload** JSON artifacts for 30-day retention

### Datasets

| Dataset | Dimensions | Distance | Vectors | Queries | Groundtruth K |
|---------|-----------|----------|---------|---------|---------------|
| Wikipedia-100K | 768 | inner_product | 100K | 5,000 | 100 |
| OpenAI ArXiv-100K | 1,536 | squared_l2 | 100K | 20,000 | 100 |

Datasets are hosted as release assets on a [separate data repo](https://github.com/YuanyuanTian-hh/diskann-benchmark-data) to avoid bloating the code repo. Can be moved to the BAB repo once write access is available.

### Thresholds

Calibrated from 5 GitHub runner runs (10 observations per metric):

| Metric | Contract | Threshold |
|--------|----------|-----------|
| recall | >= 98.0% | 1% deviation |
| QPS | >= 6.5 | 10% deviation |
| mean_ios | <= 2,410 | 1% deviation |
| mean_comps | <= 33,200 | 1% deviation |
| total_build_time | <= 115s | 10% deviation |
| mean_cpus | N/A | 15% deviation (wider  noisy on shared runners) |
| latency, io_time | N/A | 10% deviation |

### Reliability Testing (A/A)

Ran 20 A/A workflow executions (both sides build identical code) to validate pipeline stability on shared GitHub runners:

| Runs | Passed | Pass Rate |
|------|--------|-----------|
| 20 | 19 | **95%** |

- **Recall** is deterministic: Wikipedia 99.87% 0.005%, OpenAI 99.67% 0.003%
- **QPS** stable at ~9.6 across all runs
- The 1 failure was a false positive on `mean_cpus` (13.19% deviation on a noisy runner, threshold was 10%  now widened to 15%)

### Pipeline Runtime

| Dataset | Job Duration | Baseline Step | Target Step |
|---------|-------------|---------------|-------------|
| Wikipedia-100K | ~25 min | ~11 min | ~12.5 min |
| OpenAI-100K | ~78 min | ~38 min | ~39 min |

Both jobs run in parallel  **~78 min wall-clock** per workflow run, gated by OpenAI-100K.

### Before Merge TODO
- [ ] Remove `push` trigger for feature branch (keep only `workflow_dispatch`)


diskann-benchmark/perf_test_inputs/openai-100K-disk-index.json
```diff
@@ -0,0 +1,39 @@
...
+                    "disk-index-source": "Build",
+                    "data_type": "float32",
+                    "data": "OpenAIArXiv/openai_base.bin.crop_nb_100000",
+                    "distance": "squared_l2",
+                    "dim": 1536,
+                    "max_degree": 59,
+                    "l_build": 64,
+                    "num_threads": 4,
+                    "build_ram_limit_gb": 4.0,
+                    "num_pq_chunks": 192,
```
ChenSunriseJiaBao: for OAI, we may use 384 as num_pq_chunks

diskann-benchmark/perf_test_inputs/openai-100K-disk-index.json
```diff
@@ -0,0 +1,39 @@
...
+                    "quantization_type": "FP",
```
ChenSunriseJiaBao: Please use SQ_1_2.0

diskann-benchmark/perf_test_inputs/openai-100K-disk-index.json
```diff
@@ -0,0 +1,39 @@
...
+    ],
+    "jobs": [
+        {
+            "type": "disk-index",
+            "content": {
+                "source": {
+                    "disk-index-source": "Build",
+                    "data_type": "float32",
+                    "data": "OpenAIArXiv/openai_base.bin.crop_nb_100000",
+                    "distance": "squared_l2",
```
ChenSunriseJiaBao, line 13: consinenormalize

diskann-benchmark/perf_test_inputs/openai-100K-disk-index.json
```diff
@@ -0,0 +1,39 @@
...
+                    "dim": 1536,
+                    "max_degree": 59,
+                    "l_build": 64,
```
ChenSunriseJiaBao: this should be 80 if use SQ_1_2.0

diskann-benchmark/perf_test_inputs/openai-100K-disk-index.json
```diff
@@ -0,0 +1,39 @@
...
+                    "save_path": "openai_100k_benchmark_index"
+                },
+                "search_phase": {
+                    "queries": "OpenAIArXiv/openai_query.bin",
+                    "groundtruth": "OpenAIArXiv/openai-100K",
+                    "search_list": [
+                        2000
+                    ],
+                    "beam_width": 4,
+                    "recall_at": 100,
```
ChenSunriseJiaBao, line 30: please meaure recall@1000

YuanyuanTian-hh: The groundtruth file only has K=100 per query. Setting recall_at: 1000 will cause an index out of bounds.

diskann-benchmark/perf_test_inputs/openai-100K-disk-index.json
YuanyuanTian-hh: The OpenAI ArXiv vectors from big-ann-benchmarks have euclidean groundtruth (the openai-100K groundtruth file uses L2). If we switch to cosine_normalized, recall will likely drop to near 0% because the groundtruth won't match.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,343 @@
...
+  workflow_dispatch:
+    inputs:
+      baseline_ref:
+        description: 'A branch, commit SHA, or tag name to compare the current branch with'
+        required: true
+        default: 'main'
+        type: string
+  push:
+    branches:
+      - 'user/tianyuanyuan/add-benchmark-pipeline'
```
arrayka: Why is this specific branch here?

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,343 @@
...
+    paths:
```
arrayka, line 22: Why do we limit the triggering to these paths?

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,343 @@
...
+
+      - name: Run current branch benchmark
+        working-directory: diskann_rust
+        run: |
+          cargo run -p diskann-benchmark --features disk-index --release -- \
+            run --input-file diskann-benchmark/perf_test_inputs/wikipedia-100K-disk-index.json \
+            --output-file target/tmp/wikipedia-100K_benchmark_crate_target.json
+
+      - name: Generate diff stats (baseline vs target)
+        continue-on-error: true
```
arrayka: why continue-on-error: true?

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,343 @@
...
+            diskann_rust/target/tmp/openai-100K_change.md
+            diskann_rust/target/tmp/openai-100K_benchmark_crate_target.json
+            baseline/target/tmp/openai-100K_benchmark_crate_baseline.json
+          retention-days: 30
+
+  # NOTE: IAI micro-benchmarks are temporarily disabled
+  # due to callgrind not running with Rust version `ms-1.86.0`.
+  # Uncomment when ready to enable:
+  #
+  # micro-benchmark-iai:
```
arrayka: Do we need these comments? when are we planning to uncomment this section?

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,343 @@
...
+        run: |
+          python diskann_rust/.github/scripts/compare_disk_index_json_output.py \
+            --baseline baseline/target/tmp/wikipedia-100K_benchmark_crate_baseline.json \
+            --branch diskann_rust/target/tmp/wikipedia-100K_benchmark_crate_target.json \
+            --out diskann_rust/target/tmp/wikipedia-100K_change.csv
+
+      - name: Convert results to Markdown
+        working-directory: diskann_rust
+        run: |
+          python3 -c "
```
arrayka: 1. do we need this python code? can we avoid using it?
2. if we need it, consider moving it to a py script and reuse (at least in two places).

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,343 @@
...
+# Licensed under the MIT license.
+
+# DiskANN Benchmarks Workflow
+#
+# This workflow runs macro benchmarks comparing the current branch against a baseline.
+# It is manually triggered and requires a baseline reference (branch, tag, or commit).
+
+name: Benchmarks
+
+on:
```
arrayka, line 11: We want A/A testing to happen daily and fail if deviation is more than expected threshold - ideally notifying diskann-admins about failure. Is it enabled?

arrayka (changes requested):

.github/workflows/benchmarks.yml
YuanyuanTian-hh: This code allows the new pipeline to be tested on my branch, I will update it to 'main' before I merge it.

.github/workflows/benchmarks.yml
YuanyuanTian-hh: This code allows the new pipeline to be tested on my branch, I will update it to 'main' before I merge it.

.github/workflows/benchmarks.yml
YuanyuanTian-hh: I added that as a safety net during development: if the comparison script fails, the workflow continues so get the artifact uploads and can debug. It should be removed before merging.

.github/workflows/benchmarks.yml
YuanyuanTian-hh: I added in this PR as a separate workflow benchmarks-aa.yml. 
It runs daily, any deviation beyond the calibrated range fails the job
. On failure, a notify-on-failure job automatically creates a GitHub issue and cc diskann-admin.
 It will activate once this PR is merged to main.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,233 @@
...
+            --output-file target/tmp/wikipedia-100K_benchmark_crate_baseline.json
+
+      - name: Run current branch benchmark
+        working-directory: diskann_rust
+        run: |
+          cargo run -p diskann-benchmark --features disk-index --release -- \
+            run --input-file diskann-benchmark/perf_test_inputs/wikipedia-100K-disk-index.json \
+            --output-file target/tmp/wikipedia-100K_benchmark_crate_target.json
+
+      - name: Validate benchmark results
```
ChenSunriseJiaBao, line 126: please remove this step, as mark's comments, it is better to measure this in benchmark internal, for example you can run target branch with below command

cargo run -p diskann-benchmark --features disk-index --release -- \
            run --input-file diskann-benchmark/perf_test_inputs/wikipedia-100K-disk-index.json \
            --output-file target/tmp/wikipedia-100K_benchmark_crate_target.json \
            --baseline target/tmp/wikipedia-100K_benchmark_crate_baseline.json

then benchmark will check whether there is this kind of baseline file, and retrieval the alert config from diskann-benchmark/perf_test_inputs/wikipedia-100K-disk-index.json, and check whether to report regression issue.

ChenSunriseJiaBao: that means alert rule can be configurable for each benchmark, and we don't depend on python script.

ChenSunriseJiaBao (changes requested):

.github/scripts/benchmark_validate.py
```diff
@@ -0,0 +1,425 @@
...
+        if sr.get("search_l") == search_l:
+            metrics["qps"] = sr.get("qps", 0)
+            metrics["recall"] = sr.get("recall", 0)
+            metrics["mean_latency"] = sr.get("mean_latency", 0)
+            metrics["mean_ios"] = sr.get("mean_ios", 0)
+            metrics["mean_comps"] = sr.get("mean_comparisons", 0)
+            metrics["mean_hops"] = sr.get("mean_hops", 0)
+            metrics["mean_io_time"] = sr.get("mean_io_time", 0)
+            metrics["mean_cpus"] = sr.get("mean_cpu_time", 0)
+            metrics["latency_95"] = sr.get("p999_latency", 0)
```
yjiez: should this be p95?

ChenSunriseJiaBao (approved):

.github/scripts/benchmark_validate.py
YuanyuanTian-hh: Good catch. Modified it, thanks

hildebrandmw (changes requested):
Thanks again for working on this! This is an improvement on the earlier iterations, but I have several systemic concerns about merging this as is.

First, the issue of silent success is not addressed.
As a concrete example, the `THRESHOLDS` dictionary contains a `search_disk_index-search_completed` with a subfield containing a contractual value.
However, I cannot find a place where such a field is emitted.
Despite this, the CI run succeeded, which means that the validation routines are currently allowing contract violations to pass without signaling failure.
This field is not the only such field that is omitted from analysis, nor is this the only avenue by which checks can silently pass.
For example:
* `check_contract` succeeding for unrecognized direction values (this should be an error).
* The computation `"deviation": ((tv - bv) / bv * 100) if bv else 0` will succeed if `bv` is zero but `tv` is non-zero.
* Pervasive use of `.get()` with a default value turns missing fields into values that cannot meaningfully produce a deviation.

Second, the regression style workflow requires manual triggering currently, making it unlikely to be actually run in practice.
If it *did* get run on each PR, though, we'd be increasing our CI time from 10-15 minutes to over 70.
Looking at where [time is spent](https://github.com/microsoft/DiskANN/actions/runs/23830529233/job/69462958319?pr=857), we spend ~3 minutes compiling, ~20 seconds in index construction, and then over 30 minutes conducting searches (as far as I can tell).
Is it possible to get a meaningful signal with fewer queries or via another knob to avoid overwhelming CI?

There are tertiary concerns like the AA benchmark probably not needing to cache artifacts, hard coding of the Rust version, etc.

Taken together, I'm concerned about the long term ability of these pipelines to generate a meaningful signal of a regression, especially as the codebase evolves.
These issues are blocking for me, and I don’t think this should be merged as‑is.

.github/workflows/benchmarks.yml
arrayka: Please avoid code duplication between benchmarks-aa and benchmarks.
Also please keep the yml files as simple and small as possible.
It's ok to do these in a separate PR.

.github/workflows/benchmarks.yml
arrayka: it's still there :)

.github/workflows/benchmarks.yml
arrayka: Please fix it.

arrayka (approved):

PR #858, opened 2026-03-23
Implement determinant-diversity search in the new post-processing pipeline

narendatha:
# Add Determinant-Diversity Search Support

## Summary

This PR implements diversity-maximizing search using determinant-diversity post-processing.
Instead of returning only the k nearest neighbors by distance, determinant-diversity search selects k vectors that maximize the volume spanned by query-scaled vectors, producing more diverse and complementary results.

## Motivation

Standard nearest neighbor search often returns highly similar or redundant results.
Determinant-diversity search helps retrieve complementary results that cover different aspects of a query.
This PR uses determinant maximization of the Gram matrix of query-scaled vectors.

## Algorithm

Each candidate vector is scaled by `(similarity_to_query)^power`, then vectors are greedily selected to maximize the determinant of the Gram matrix (equivalently maximizing spanned volume).

Two variants are implemented:
- **eta = 0**: pure greedy orthogonalization (pivoted QR style)
- **eta > 0**: ridge-regularized variant balancing diversity and numerical stability

Both variants run in **O(n * k * d)** where `n` is number of candidates, `k` is number of selected results, and `d` is vector dimension.

## Performance Characteristics

In benchmark comparison runs (normal search vs determinant-diversity search), we observe the expected trade-off:
- lower QPS due to diversity optimization overhead
- recall trade-off depending on search parameters

### `determinant_diversity_power` (default: 2.0)
Controls relevance vs diversity:
- **Higher (e.g., 3.0+)**: stronger relevance weighting, typically less diversity
- **Lower (e.g., 1.0)**: more diversity emphasis
- **0**: pure diversity maximization without relevance weighting

### `determinant_diversity_eta` (default: 0.01)
Ridge regularization for numerical robustness and relevance bias:
- **Higher (e.g., 0.1+)**: more robust, typically more relevance-biased
- **Lower (e.g., 0.001)**: closer to pure determinant maximization
- **0**: pure greedy orthogonalization (no ridge regularization)

### Recommended Settings
| Use Case | determinant_diversity_power | determinant_diversity_eta |
|----------|-----------------------------|----------------------------|
| Balanced (default) | 2.0 | 0.01 |
| High diversity | 1.0 | 0.001 |
| High relevance | 3.0 | 0.1 |

### Benchmark Configuration (disk-index)
```json
{
  "search_phase": {
    "search_list": [100, 200, 400],
    "recall_at": 10,
    "is_determinant_diversity_search": true,
    "determinant_diversity_eta": 0.01,
    "determinant_diversity_power": 2.0
  }
}
```

## Testing

- unit tests for determinant-diversity post-processing and parameter validation
- `cargo fmt --all`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo check --package diskann-benchmark --features disk-index`

## How to try

- build this branch
- set benchmark input paths to your dataset/index
- run `diskann-benchmark` with determinant-diversity parameters


diskann-benchmark-core/Cargo.toml
```diff
@@ -11,6 +11,7 @@ edition = "2024"
 anyhow.workspace = true
 diskann.workspace = true
 diskann-benchmark-runner = { workspace = true }
+diskann-providers.workspace = true
```
hildebrandmw: One quick request - please do not add `diskann-providers` as a dependency of `diskann-benchmark-core`. It will blow up compile times for everyone.

Instead, I'd prefer if the `KNN` search could be augmented to take a post-processor. The ethos of `diskann-benchmark-core` is to provide utilities that are generally useable and since diversity search uses the KNN API - it feels like there's a generalization opportunity.

narendatha: Thanks, done.

diskann-benchmark-core/src/search/graph/knn.rs
```diff
@@ -41,6 +41,18 @@ where
...
+#[derive(Debug)]
+pub struct KNNWithPostProcessor<DP, T, S, P>
+where
+    DP: provider::DataProvider,
+{
+    index: Arc<graph::DiskANNIndex<DP>>,
+    queries: Arc<Matrix<T>>,
+    strategy: Strategy<S>,
+    post_processor: Strategy<P>,
+}
```
hildebrandmw, line 54: I'm trying to think of a way we can get rid of `KNNWithPostProcessor`. Generally, decorating structs like this (`KNNWithPostProcessor`) is not a viable long-term strategy. It is not extendable and does not set a good pattern for other implementations to follow. For example, will we then need a `RangeSearchWithPostProcessor`? `MultiHopSearchWithPostProcessor`?

Maybe something like
```rust
struct Defferred<S>(S);

impl<DP, T, O, S> SearchPostProcess<DP, T, O> for Deferred<S>
where
    S: DefaultPostProcessor<DP, T, O>,
{
     fn post_process(&self, args...) {
          self.0.default_post_processor().post_process(args...)
     }
}

struct KNN<DP, T, S, P = Deferred<S>> {
     ...
}

// Keep original `new`
impl<DP, T, S> KNN<DP, T, S> {
    fn new(args...)
}

// New method with injectable post-process
impl<DP, T, S, P> KNN<DP, T, S, P> {
    fn new_with(args...)
}
```

diskann-benchmark-runner/src/any.rs
```diff
@@ -397,7 +397,7 @@ mod tests {
         let _: Type<f32> = value.convert::<DataType, _>().unwrap();
 
         // An invalid match should return an error.
-        let value = Any::new(0usize, "random-rag");
+        let value = Any::new(0usize, "random-determinant-diversity");
```
hildebrandmw: Nit: Did find and replace get a little aggressive?

diskann-benchmark/example/openai-disk-determinant-diversity-compare.json
```diff
@@ -0,0 +1,52 @@
+{
+  "search_directories": [
+    "C:/data/openai"
```
hildebrandmw: Should this follow the pattern of the other examples and use a relative path? I know it's designed to run on OpenAI, but it's an example. Shouldn't it be able to run on test-data?

diskann-benchmark/src/inputs/async_.rs
```diff
@@ -123,6 +123,9 @@ pub(crate) struct TopkSearchPhase {
     pub(crate) queries: InputFile,
     pub(crate) groundtruth: InputFile,
     pub(crate) reps: NonZeroUsize,
+    pub(crate) determinant_diversity_eta: Option<f64>,
+    pub(crate) determinant_diversity_power: Option<f64>,
+    pub(crate) determinant_diversity_results_k: Option<usize>,
```
hildebrandmw, line 128: This is not the way to add this functionality. The problem is that this bolting on means two things:

1. All of our existing search benchmarks gain additional monomorphizations, bloating compile time for every combination.
2. With all these tacked on as `Options`, it is way too easy for another backend to accidentally silently ignore these parameters and accept inputs with determinant values set but not actually do anything with them.

Features should be added in a way that strategically compiles and forces future implementations to either explicitly opt-in or opt-out. One such example is the `SearchPhase` enum itself. While it *does* kind of string, it at least forces that all variants either succeed or are explicitly opted out.

diskann-benchmark/src/inputs/async_.rs
```diff
@@ -139,6 +142,44 @@ impl CheckDeserialization for TopkSearchPhase {
                 .with_context(|| format!("search run {}", i))?;
         }
 
+        if self.determinant_diversity_eta.is_some() != self.determinant_diversity_power.is_some() {
```
hildebrandmw, line 145: The mutual exclusivity issue can be solved structurally by grouping all of these into a struct, which can then introduce a reusable deserialization check as well.

diskann-benchmark/src/inputs/disk.rs
```diff
@@ -81,6 +81,12 @@ pub(crate) struct DiskSearchPhase {
     pub(crate) search_list: Vec<u32>,
     pub(crate) recall_at: u32,
     pub(crate) is_flat_search: bool,
+    #[serde(default)]
+    pub(crate) is_determinant_diversity_search: bool,
+    #[serde(default)]
+    pub(crate) determinant_diversity_eta: Option<f64>,
+    #[serde(default)]
+    pub(crate) determinant_diversity_power: Option<f64>,
```
hildebrandmw, line 86: Benchmark inputs should not assume default behavior.

diskann-providers/src/model/graph/provider/async_/determinant_diversity_post_process.rs
```diff
@@ -0,0 +1,515 @@
...
+        }
+    }
+}
+
+impl std::error::Error for DeterminantDiversityError {}
+
+#[derive(Debug, Clone, Copy)]
+pub struct DeterminantDiversitySearchParams {
+    pub top_k: usize,
+    pub determinant_diversity_eta: f64,
```
hildebrandmw: Making fields `pub` defeats the purpose of a checked constructor.

diskann-benchmark/src/backend/index/search/knn.rs
```diff
@@ -129,3 +139,30 @@ where
...
+
+impl<DP, T, S, P> Knn<DP::InternalId>
+    for Arc<core_search::graph::knn::KNNWithPostProcessor<DP, T, S, P>>
+where
+    DP: diskann::provider::DataProvider,
+    core_search::graph::knn::KNNWithPostProcessor<DP, T, S, P>: core_search::Search<
+        Id = DP::InternalId,
+        Parameters = diskann::graph::search::Knn,
+        Output = core_search::graph::knn::Metrics,
+    >,
```
hildebrandmw, line 151: Here, we're seeing exactly the problem that `KNNWithPostProcessor` introduces. It requires replication higher in the stack.

diskann-benchmark/src/backend/index/search/knn.rs
```diff
@@ -35,6 +35,32 @@ pub(crate) fn run<I>(
...
+        recall_k: usize,
+        recall_n: usize,
+    ) -> anyhow::Result<Vec<SearchResults>>;
+}
+
+fn run_search<I, F>(
+    runner: &dyn Knn<I>,
+    groundtruth: &dyn benchmark_core::recall::Rows<I>,
+    steps: SearchSteps<'_>,
+    builder: F,
```
hildebrandmw, line 59: Are these changes necessary?

narendatha:
Closing without merging because new plugin model invalidates approach in this PR.

PR #859, opened 2026-03-24
Overhaul WorkingSet/FillSet and its interaction with Multi-Insert

hildebrandmw:
This is a giant PR. I am **very** sorry for that, this is not my favorite thing to do, but I think it's necessary for compile time reduction and long-term multi-vector compatibility. It undoes a nasty knot in how the `AsElement`, `SearchStrategy`, and `InsertStrategy` traits interact in conjunction with multi-insert and the "working set". Unfortunately, the knot cannot be undone in pieces.

An RFC accompanies this PR, you can see the rendered doc [here](https://github.com/microsoft/DiskANN/blob/907cd7166884727096f8401a9892d5663197b29f/rfcs/00859-fillset.md) for a more detailed look at the problem.

# What does this PR enable

* A nearly 4x reduction in compile time and 42% binary size reduction for a production user by effectively allowing non-slice types to be used in `SearchStrategy` and `InsertStrategy`.
* Fewer `get_element` calls for the same production user in some quant-only builds by allowing the working set to be "pre-populated" with the quantized representation of vectors provided to multi-insert.
* Fewer `get_element` calls for multi-insert due to a relaxation in how aggressively the working set is cleared between back-edge insertions. This reuse can be opted out of by providers that do not want stale entries in their working set.
* Allowing users finer control over the nature of the working set, undoing the "passthrough" hack currently employed by the inmem providers and allowing future optimizations, like packing the contents of the working set into a matrix for better memory locality.

# Key Interface Changes

* The query parameter to `SearchStrategy`, `InsertStrategy`, and `SetElement` is no longer `?Sized`. Instead, syntax like `SearchStrategy<DP, [f32]>` will have to be replaced with `SearchStrategy<DP, &[f32]>`. While this loses a little syntactic niceness - the resulting uniformity in the API is much better.
* `SearchStrategy` gets a new `WorkingSet` associated type and `create_working_set` method.
* A new `MultiInsertStrategy` type is added.
* The insert `Guard` is moved from `SetElement` and is instead put directly on `DataProvider`, mainly to make the HRTB syntax for `SetElement` saner. This is an unfortunate API regression, but proper ID handling/error recovery is kind of next on the list of things to figure out, so I think it's okay to keep in `DataProvider` at the moment.

# Removals
* `FillSet` is gone!
* `AsElement` is gone!
* `Accessor::Extended` is gone!
* `diskann_utils::reborrow::{Lower, AsyncLower}` are gone!
* `diskann::utils::VectorIdBoxedSlice` is gone!

# Suggested Reviewing Order and Main Changes

**diskann**
- `graph/workingset/mod.rs`: The main "workingset" abstraction. With three main parts:
  * `Fill`: The replacement for `FillSet`.
  * `View`: A scoped "working set" that prune interacts with.
  * `AsWorkingSet`: A multi-insert only trait for overlaying the inserted batch.

- `graph/workingset/map.rs`: The official "library" default for a workingset implementation.
  The main type is `Map` and its associated types/traits:
  - `Map`: A hash-map with an optional overlay. See the module level docs for its eviction policy.
  - `Overlay`: A dynamically dispatched wrapper for the batch inserted in `multi_insert`. When present in `Map`, we first check the `Overlay` for an ID before falling back to the hashmap.
  - `Projection`/`Project`: The glue between elements in the `Map` and the `Overlay`. These traits are critical for allowing the storage representation in the `Overlay` to be distinct from the storage mechanism for the entries in the `HashMap`. As a mental model, think of the overlay as being literally a `Matrix` whereas entries in the HashMap are boxed types.
  - `Builder`: A fairly simple builder for `Map`. While this makes constructing a `Map` a little awkward, its role is to ensure `AsWorkingSet` can use pretty much the same API as a raw `Map` construction.

  In addition, there is a blanket implementation for `Fill<Map<...>>` for any accessor `A`. This works out of the box for "simple" full-precision cases and is reasonably easy to defeat when not needed. The goal here was to lower the barrier to entry as much as possible.

- `graph/glue.rs`: A `WorkingSet` associated type is added to `PruneStrategy` and a new `MultiInsertStrategy` is introduced.
  `MultiInsertStrategy` is primarily concerned with how to deal with translating the batch into the final working set and abstracting the element type that can be passed to `multi_insert`.

- `graph/index.rs`: Refactored the use of the working set to use the new API. These changes are largely straight-forward. Pay special attention to `inplace_delete_inner` which now needs to fetch all replacement candidates in one `fill` call. This could actually be considered a bug-fix since the code currently calls `get_element` in a relatively hot loop.

Everything else is supporting work. The test provider intentionally uses the blanket `Fill` implementation for `Map` to enable test coverage of that interface.

**diskann-providers**
- `model/graph/provider/async_/distances.rs`: A custom `WorkingSet`/`View` implementation for the hybrid prunes we currently support. This is a simple wrapper around `diskann::graph::workingset::Map` with a small amount of custom behavior. The debug and BF-Tree providers use this implementation.
- `model/graph/provider/async_/*`: Updated the inmem providers to use the new interface. They use an explicit `PassThrough` implementation for their working set, making the direct indexing behavior during prune explicit and removing a `HashMap` allocation entirely.

  Hybrid PQ prune is handled by instead storing a `HashSet` of full-precision IDs.

- `caching provider`: Feel free to largely ignore this. The caching provider is getting complicated enough at this point that I think we should deprecate it as not worth the effort. Directly integrating a cache into an implementation is probably more effective and easier to maintain.

**everything else**
The rest of the changes are largely mechanical for the `SearchStrategy`/`InsertStrategy`/`SetElement` type parameter change and updating `FillSet` implementations.


diskann-providers/src/model/graph/provider/async_/caching/example.rs
```diff
@@ -201,22 +201,49 @@ impl<'a> cache_provider::AsCacheAccessorFor<'a, debug_provider::FullAccessor<'a>
...
-    ) -> Result<Self::Element<'b>, Self::Error> {
-        cache_provider::get_or_insert(self, cache, id).await
+        cache: &'b mut FullAccessorCache<'a>,
+        state: &'b mut FullAccessorState,
+        itr: Itr,
+    ) -> impl diskann_utils::future::SendFuture<
+        Result<Self::View<'b>, cache_provider::CachingError<Self::Error, CacheAccessError>>,
+    >
+    where
+        Itr: Iterator<Item = u32> + Send + Sync,
```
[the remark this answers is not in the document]
hildebrandmw: This is fine. Impls are allowed to have less restrictive clauses. At least it definitely compiles.

diskann/src/internal/chain.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (a_lo, _) = self.a.size_hint();
+        let (b_lo, _) = self.b.size_hint();
+        #[expect(
+            clippy::expect_used,
+            reason = "internally - we should never even get close"
+        )]
+        let len = a_lo.checked_add(b_lo).expect("Chain length overflow");
+        (len, Some(len))
+    }
```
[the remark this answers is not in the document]
hildebrandmw: This is true. I updated the bound to `ExactSizeIterator`.

diskann/src/graph/workingset/map.rs
```diff
@@ -0,0 +1,2091 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT license.
+ */
+
+//! The default implementation of a [workingset](super).
```
metajack: Should this be capitalized?

diskann/src/graph/workingset/map.rs
```diff
@@ -0,0 +1,2091 @@
...
+                        },
+                    },
+                }
+            }
+            Ok(self.view())
+        }
+    }
+}
+
+impl<K, V, P> Map<K, V, P>
```
metajack, line 344: This block could be combined with the second one. Is it separated just for readability reasons?

diskann/src/graph/workingset/map.rs
```diff
@@ -0,0 +1,2091 @@
...
+
+    /// Build a seeded projected test map with the given overlay and capacity.
+    fn seeded_map_projected(
+        overlay: Overlay<u32, TestProjection>,
+        capacity: Capacity,
+    ) -> TestMapProjected {
+        Builder::new(capacity).with_overlay(overlay).build(16)
+    }
+
+    /// The `TestProjection` ensures that we can apply an arbitrary transformation on the
```
metajack, line 1011: Sentence is cut off

diskann/src/graph/workingset/map.rs
```diff
@@ -0,0 +1,2091 @@
...
+            assert!(!overlay.contains_key(&99));
+        }
+    }
+
+    #[test]
+    fn overlay_is_cheap_to_clone() {
+        let (_batch, overlay) = test_overlay();
+        let cloned = overlay.clone();
+        // Both see the same data.
+        assert_eq!(cloned.get(&10).unwrap(), &[1.0, 2.0]);
```
metajack: Isn't this only testing one of them?

diskann/src/graph/index.rs
```diff
@@ -1740,30 +1722,46 @@ where
...
             let computer = accessor.build_distance_computer().into_ann_result()?;
-            accessor
-                .fill_set(&mut working_set, replace_candidates.iter().copied())
-                .await
-                .allow_transient("observed failure during working set population")?;
 
-            accessor
-                .fill_set(&mut working_set, in_neighbors.iter().copied())
+            // This is the total union of elements that we consider for neighbors.
+            // It's possible there is some repition, but implementations of `Fill` should
```
metajack: typo: repition

metajack (approved):

diskann/src/graph/glue.rs
```diff
@@ -251,10 +252,7 @@ impl<T> HybridPredicate<T> for NotInMut<'_, T> where T: Clone + Eq + std::hash::
 /// ## Error Handling
 ///
 /// Transient errors yielded by `distances_unordered` are acknowledged and not escalated.
-pub trait ExpandBeam<T>: BuildQueryComputer<T> + AsNeighbor + Sized
-where
-    T: ?Sized,
-{
+pub trait ExpandBeam<T>: BuildQueryComputer<T> + AsNeighbor + Sized {
```
arkrishn94, line 255: Just for curiosity - why is the `Sized` bound needed here?

diskann/src/graph/workingset/map.rs
```diff
@@ -0,0 +1,2091 @@
...
+        Itr: ExactSizeIterator<Item = A::Id> + Clone + Send + Sync,
+        Self: 'a,
+    {
+        map.fill(self, itr)
+    }
+}
+
+/////////////////
+// Projections //
+/////////////////
```
arkrishn94, line 553: Does it make sense to move the traits and implementations under Projections somewhere outside workingset that is more visible? I think this might be useful more broadly when generalizing trait bounds to referenced types.

diskann/src/graph/index.rs
```diff
@@ -341,7 +329,7 @@ where
             let num_insert_attempts = insert_retry.map_or(1, |v| v.max_retries().get());
 
             // N.B.: Working set needs to be outlived by `accessor`.
```
arkrishn94: Not needed anymore?

arkrishn94 (commented):
Thanks Mark, this is incredibly deep and hard work! I don't have any major feedback, just some small comments.

rfcs/00859-fillset.md
```diff
@@ -0,0 +1,495 @@
+# Untangling the FillSet Knot
+
+
+| | |
+|---|---|
+| **Authors** | Mark Hildebrand |
+| **Contributors** | Coffee |
```
arrayka: + All-mighty AI :D

rfcs/00859-fillset.md
```diff
@@ -0,0 +1,495 @@
...
+| **Created** | YYYY-MM-DD |
```
arrayka: Minor: fill out dates or remove

rfcs/00859-fillset.md
```diff
@@ -0,0 +1,495 @@
+# Untangling the FillSet Knot
```
arrayka, line 1: why does it start with 00859: 00859-fillset.md?

diskann/src/graph/workingset/map.rs
```diff
@@ -0,0 +1,2091 @@
...
+
+                // This operation cannot underflow since `promoted` should be at most `len`.
+                //
+                // However, incorrect implementations of `ExactSizeIterator` could lie about
+                // the length. This will catch such degenerate implementations in debug builds.
+                //
+                // In release builds - this is still fine because the wrapped result is used
+                // as an upper bound, limited by the current size of `map` which is **not**
+                // degenerate.
+                let extra_needed = len - promoted;
```
suri-kumkaran, line 267: nit: Should we also add `debug_assert!(promoted <= len)`

hildebrandmw: No need - a debug build will panic on underflow anyways. I was in the process of relaying my logic when I looked up and saw pretty much exactly what I was typing in the comment hahaha.

diskann/src/graph/workingset/map.rs
hildebrandmw: The one above adds `Send` and `Sync` bounds to the key and value types, which are not strictly needed for the general access methods. In practice, this is probably not a big deal since any reasonable type will be `Send + Sync`, but this felt a little more precise.

diskann/src/graph/glue.rs
hildebrandmw: The reason is a combination of the disk index and the failing of my past self work around some issues. The disk-index required that `Accessor` and `NeighborAccessor` needed to be implemented on the same type, where-as before they could be separate.

To avoid requiring all existing `Accessors` to opt-in to boilerplate to implement `NeighborAccessor[Mut]` (a terrible idea for many reasons), I used the `DelegateNeighbor/AsNeighbor` traits to:

* Implement `NeighborAccessor for &mut T where T: AsNeighbor`.
* Implement `AsNeighbor for T where T: NeighborAccessor`.

These two bounds means that a given `Accessor` could either implement `NeighborAccessor` directly, or implement `Delegateneighbor`. Though either path, a constraint of `AsNeighbor` on the accessor would allow it to be used like a `NeighborAccessor` through the `&mut T` blanket implementation. This also happens to be the source of the `NeighborAccessor` traits receiving by value, because I was originally envisioning a sequence like
```rust
accessor
    .delegate_neighbors()
    .get_neighbors(id).await?; // <--- Doesn't compile - temporary from `delegate_neighbors` is dropped
```

The combination of taking by value and the `&mut T` blanket implementation is what leads to the sized bound.

**Now**: This whole rube-goldberg machinery could be greatly simplified by using a proc-macro for the `NeighborAccessor` delegation and getting rid entirely of the `AsNeighbor/DelegateNeighbor` stuff. I think past me was just afraid of writing a proc macro? I don't know.

In any case, that's where the `Sized` bound comes from. It is bad, I feel bad, and this will probably change in the near future to be less insane.

diskann/src/graph/workingset/map.rs
hildebrandmw: Maybe? The projections here are super specific to the `Map` that I would hesitate to over generalize, having recently reaped the unfortunate rewards of previous attempts at over generlization.

diskann/src/graph/index.rs
hildebrandmw: Right! Woohoo!

diskann/src/graph/index.rs
hildebrandmw: Thanks - fixed!

diskann/src/graph/workingset/map.rs
hildebrandmw: Yup. You are right. Hopefully now the comment is more accurate (I tested for pointer equality).

diskann/src/graph/workingset/map.rs
hildebrandmw: Yes, I accidentally a word or 10. Thanks

diskann/src/graph/workingset/map.rs
hildebrandmw: Sure, I updated the comment.

diskann/src/graph/index.rs
```diff
@@ -674,29 +632,35 @@ where
...
                 match self
                     .search_and_prune(
                         strategy,
                         context,
-                        vector,
-                        position,
+                        batch,
                         work.all(),
+                        position,
+                        working_set,
```
randybird, line 663: I see the risk was mentioned to be minimal due to the new reuse pattern which leads to the stale cache, just want to understand if we have any before/after recall measurement done to validate this?

rfcs/00859-fillset.md
hildebrandmw: Naturally (though this is the PR that's made me rethink my use of the tool as I'm fairly certain it made the process longer).

rfcs/00859-fillset.md
hildebrandmw: It was a debate we had originally - 859 is the PR number. This way, we just need to update the filename once and are guaranteed to never conflict.

diskann/src/graph/index.rs
hildebrandmw: The RFC has recall validation for both the inmem providers and for an internal database integration. Recall is stable and the internal implementation sees a net reduction in `get_vector` calls as this PR enables caching during multi-insert (and said internal implementation has other means of working around any staleness concerns).

rfcs/00859-fillset.md
hildebrandmw: Yup, sorry. Dates are now present.

PR #865, opened 2026-03-26
Cleanup `diskann-benchmark-runner` and friends.

hildebrandmw:
Simplify benchmark running and registration by:

* Moving the `Input` away from a `dyn`-compatible trait to a simple trait with methods that don't receive by `&self`. That is, instead of 
  ```rust
  trait Input {
      fn tag(&self) -> &'static str;
      fn try_deserialize(&self, ...);
      fn example(&self);
  }
  ```
  we get
  ```rust
  trait Input {
      fn tag() -> &'static str;
      fn try_deserialize(...);
      fn example();
  }
  ```
  The original pattern required an awkward proxy type for input registration (see the `Input` type in `diskann-benchmark`). This new pattern allows the actual input structs to be used directly.
* Ditching the whole overly engineered `dispatcher` (which is cool, but inappropriate for this task) and instead introducing a `Benchmark` trait with the signature (see the code for doc-comments)
  ```rust
  pub trait Benchmark {
      type Input: Input + 'static;
      type Output: Serialize;

      fn try_match(input: &Self::Input) -> Result<MatchScore, FailureScore>;

      fn description(
          f: &mut std::fmt::Formatter<'_>,
          input: Option<&Self::Input>,
      ) -> std::fmt::Result;

      fn run(
          input: &Self::Input,
          checkpoint: Checkpoint<'_>,
          output: &mut dyn Output,
      ) -> anyhow::Result<Self::Output>;
  }
  ```
  The functions `try_match` and `description` match the original `DispatchRule` technique, but instead of relying on a `convert` method and a corresponding closure - we just `run` the benchmark directly.

  This almost entirely removes the need for the various `register!` macros as registration is significantly more straightforward. Additionally, the input and output types are concrete!

These changes remove a lot of boilerplate from `diskann-benchmark`. No need to match on the `Any` type anymore as this happens automatically. While updating `diskann-benchmark`, I was able to replace some of the macros for stamping out benchmarks with generics. Scalar and spherically-quantized index builds are still reliant on macros.

Finally, the backend of `diskann-benchmark-runner` (namely `registry::Benchmarks`) is cleaned up to not rely on the complicated `dispatcher` machinery for matching inputs and providing mismatch diagnostics. Instead, the logic is greatly simplified and just implemented inline.

## Why
While reviewing #857, I was pondering what it would take to add A/B comparison logic directly to our benchmark framework. My thinking is that it would be nice to have something like
```rust
trait Compare: Benchmark {
    type Tolerance: Input;
    fn compare(
        tolerances: &Self::Tolerance,
        before: &Self::Output,
        after: &Self::Output,
    ) -> SomeResult;
}
```
Benchmarks that want A/B comparisons could implement `Compare` and use a `registry::Benchmarks::register_compared` method to register a benchmark that participates in A/B comparison. The backend of `diskann-benchmark-runner` could then use this to inspect the benchmarks that support comparison and this becomes an engine for A/B testing with concretely typed outputs!

The problem, of course, is the current way of benchmark registration is a little overly complicated and makes this kind of extension way too messy. So view this PR as serving two purposes. First, we clean up an overly complicated portion of benchmarking and second, we enable evolution of the benchmarking framework.

## Suggested Reviewing Order
I would actually recommend looking at the changes in `diskann-benchmark` before `diskann-benchmark-runner`. The amount of code that got chopped is quite satisfying. The changes here aren't perfect (the use of the `'static` lifetime for benchmarks is a remnant of the `DispatchRule` approach and fully cleaning that up would be a larger change).

Then, inside `diskann-benchmark-runner`, I suggest:
* `input.rs`: New home of the `Input` trait plus the machinery for dynamically dispatching to inputs.
* `benchmark.rs`: The new `benchmark` trait (plus associated dynamic dispatch machinery).
* `registry.rs`: Updates to the registry to reflect the new traits. In particular, `Benchmarks` is updated so the matching logic is much clear (and not hidden behind a dispatcher). I expanded the tests a little and added a few more baselines to exercise the matching logic more.
* `dispatcher/`: This directory has largely been made obsolete. The `DispatchRule` trait is preserved because it still has some utility in matching data types and composes somewhat with the `Benchmark` trait. We could probably get rid of this directory entirely in the future, but for now I'm preserving it for the sake of a less insane diff.

Feel free to skip `diskann-benchmark-simd`.

### DIFF
--- a/diskann-benchmark/src/inputs/mod.rs
+++ b/diskann-benchmark/src/inputs/mod.rs
@@ -24,6 +24,11 @@ pub(crate) trait Example {
     fn example() -> Self;
 }
 
+// NOTE: The input registration and dispatching isn't prefect. It uses a pattern (like
+// the use of `'static` on the benchmark types) as a byproduct of older ways of doing
+// benchmark selection.
+//
+// In the future, these can be migrated to reduce this legacy cruft.
 macro_rules! as_input {
     ($T:ty) => {
         impl diskann_benchmark_runner::Input for $T {


diskann-benchmark-runner/src/app.rs
```diff
@@ -130,23 +130,10 @@ impl App {
                 let run = Jobs::load(input_file, inputs)?;
                 // Check if we have a match for each benchmark.
                 for job in run.jobs().iter() {
-                    if !benchmarks.has_match(job) {
+                    const MAX_METHODS: usize = 3;
+                    if let Err(mismatches) = benchmarks.debug(job, MAX_METHODS) {
```
arkrishn94, line 142: Just for my understanding - what is happening here with `MAX_METHODS`?

diskann-benchmark-runner/src/input.rs
```diff
@@ -0,0 +1,130 @@
...
+        serialized: &serde_json::Value,
+        checker: &mut Checker,
+    ) -> anyhow::Result<Any>;
+    fn example(&self) -> anyhow::Result<serde_json::Value>;
+}
+
+impl<T> DynInput for Wrapper<T>
+where
+    T: Input,
+{
```
arkrishn94, line 116: For my learning - why is the indirection with `Wrapper` needed?

diskann-benchmark-runner/src/app.rs
hildebrandmw: The fallback routines just prints out the 3 nearest misses, as I've found that is usually enough to get close without completely spamming the terminal output. Basically, when an match for an input cannot be found, this dumb algorithm tries to find the nearest misses to help you out. Which is why `Benchmark::description` should be implemented because that's what gets  used for the diagnostic :smile:.

diskann-benchmark-runner/src/input.rs
hildebrandmw: When inputs are registered, we just have the type, not an actual instance. But to make a trait object around something, we need in instance to which the vtable can be attached. The `Wrapper` is what I use for the instance so we can make the trait object.

That's part of the reason why the original indirection existed in the first place with the input proxy types - I wanted a trait object without necessarily requiring an instance of the corresponding `Input`. Why didn't I do it this way before? Not sure ... maybe I've gotten less bad at Rust? Or maybe it was because the OG version of this had super enum for all benchmarks and I got too fixated on that?

diskann-benchmark-runner/src/benchmark.rs
```diff
@@ -0,0 +1,140 @@
...
+                Some(cast) => T::description(f, Some(cast)),
+                None => write!(
+                    f,
+                    "expected tag \"{}\" - instead got \"{}\"",
+                    T::Input::tag(),
+                    input.tag(),
+                ),
+            },
+            None => {
+                writeln!(f, "tag \"{}\"", <T::Input as Input>::tag())?;
```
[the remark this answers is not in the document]
hildebrandmw: I agree that the newline handling here is pretty bad. Will cleanup in future work.

diskann-benchmark-runner/src/utils/datatype.rs
```diff
@@ -60,11 +60,6 @@ impl std::fmt::Display for DataType {
 #[derive(Debug, Default, Clone, Copy)]
 pub struct Type<T>(std::marker::PhantomData<T>);
 
-/// The `Type` meta variable maps to itself.
-impl<T: 'static> Map for Type<T> {
-    type Type<'a> = Self;
-}
-
 pub const MATCH_FAIL: FailureScore = FailureScore(1000);
```
suri-kumkaran, line 63: Maybe out of scope of this PR but as this is a cleanup PR should we consider unifying this constant with the same name constant in `diskann-benchmark-runner/src/any.rs`.

hildebrandmw: Probably - though the stuff in `any.rs` can probably be simplified now that users don't need to write `DispatchRule`s directly for `Any`.

hildebrandmw:
> Should we put a TODO in the code for "The changes here aren't perfect (the use of the `'static` lifetime for benchmarks is a remnant of the `DispatchRule` approach and fully cleaning that up would be a larger change)."?

Seems reasonable. I attached it to the `as_input!` macro since it's the piece the most central bit of code now and is itself an older pattern.

diskann-benchmark-runner/tests/test-mismatch-1/stdout.txt
```diff
@@ -0,0 +1,15 @@
...
+  "error_when_checked": false
+}
+
+Closest matches:
+
+    1. "type-bench-f32": expected "float32" but found "float16"
+    2. "type-bench-i8": expected "int8" but found "float16"
+    3. "exact-type-bench-f32-1000": expected "float32" but found "float16"
+
+could not find find a benchmark for all inputs
```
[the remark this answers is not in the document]
hildebrandmw: Fixed

diskann-benchmark-runner/src/app.rs
```diff
@@ -112,8 +112,8 @@ impl App {
             // List the available benchmarks.
             Commands::Benchmarks {} => {
                 writeln!(output, "Registered Benchmarks:")?;
-                for (name, method) in benchmarks.methods() {
-                    writeln!(output, "    {}: {}", name, method.signatures()[0])?;
+                for (name, description) in benchmarks.names() {
+                    writeln!(output, "    {}: {}", name, description)?;
```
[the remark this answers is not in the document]
hildebrandmw: Fixed

diskann-benchmark-runner/src/test.rs
```diff
@@ -145,121 +128,119 @@ impl CheckDeserialization for DimInput {
...
 {
-    type Error = anyhow::Error;
+    type Input = TypeInput;
+    type Output = &'static str;
 
-    fn try_match(from: &&'a TypeInput) -> Result<MatchScore, FailureScore> {
-        Type::<T>::try_match(&from.data_type)
+    fn try_match(input: &TypeInput) -> Result<MatchScore, FailureScore> {
+        // Try to match based on data type.
+        // Add a small penalty to `ExactTypeBench` can be more specific if it hits.
```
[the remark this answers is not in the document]
hildebrandmw: Fixed

diskann-benchmark-runner/src/registry.rs
```diff
@@ -27,28 +27,25 @@ impl Inputs {
...
+    pub fn get(&self, tag: &str) -> Option<input::Registered<'_>> {
+        self.inputs.get(tag).map(|v| input::Registered(&**v))
     }
 
     /// Register `input` in the registry.
     ///
     /// Returns an error if any other input with the same [`Input::tag()`] has been registered
     /// while leaving the underlying registry unchanged.
-    pub fn register<T>(&mut self, input: T) -> anyhow::Result<()>
+    pub fn register<T>(&mut self) -> anyhow::Result<()>
```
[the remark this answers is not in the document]
hildebrandmw: Fixed

PR #900, opened 2026-04-03
Benchmark Support for A/B Tests

hildebrandmw:
This adds support for native A/B testing in `diskann-benchmark-runner` with an example implementation added to `diskann-benchmark-simd`. The benefits of having this infrastructure at the Rust level are:

* Code actually performing the before and after checks can run on the concrete types used in the Rust code. They do not need to dig through layers of JSON (and the corresponding maintenance headache associated with changing data structures). The compiler checks that at least structurally the logic is sound.
* Shared infrastructure for running comparison tests locally during development and for CI.
* Stricter checks on correctness and wellformedness of data that all users can take advantage of.

### Concept
The idea is to use one `input.json` file to generate two output JSON files `before.json` and `after.json` for different builds or configurations of the library. The `input.json` is then accompanied by a `tolerances.json`, which contains runtime thresholds for values of interest to help accommodate runtime variability. A regression check takes all such files and performs the following steps:
1. Correctly parse `tolerances.json` and `input.json`.
2. Unambiguously match entries in `tolerances.json` to entries in `input.json`. I went with matching semantics rather than requiring a one-to-one correspondence to make it easier to have a single tolerance entry work as a blanket entry for multiple benchmark runs.
3. Take the resolve `tolerance/input` pairs and match them with a regression-checkable benchmark.
4. Once everything has been correctly set up, run each regression check which
   * Deserializes entries in `before/after.json` into the benchmark's output type or error gracefully if this cannot be done due to an incorrect environment.
   * Runs the before/after check.
   * Aggregates and displays errors/failures/successes with an exit code that indicates success or failure.
5. An invocation of the checking functionality only returns a successful exit code if all of the above steps succeed without error and all the run checks pass.

The matching semantics work like this. Each tolerance entry looks like the following:
```json
{
  "input": {
    "type": "type-tag",
    "content": {},
  },
  "tolerance": {
     "type": "tolerance-type-tag",
     "content": "defined-per-tolerance",
  }
}
```
The content field of "input" need not be deserializable to its corresponding value. Instead, we use the raw JSON of the input and match it as a "subset" against the raw JSON of the `input.json` using the following rules:

* Nulls are always equal.
* Primitive types are equal if their values compare equal.
* An array `x` is a subset of an array `y` if `x.len() <= y.len()` and each entry `i` in `x` is a subset of its corresponding entry in `y` (i.e., we match prefixes).
* An object `x` is a subset of an object `y` if each key in `x` is a key in `y` and each value associated with a key in `x` is a subset of the value of the same entry in `y`.
* Heterogeneous matches are not supported. A `bool` cannot be a subset of an `integer`. This breaks the match instantly.

This means that an empty "content" field will match any struct and thus be a blanket implementation for the input with the same type tag. Or, the "content" field can be refined to be more specific as needed.

Since a single tolerance entry may match multiple inputs, we ensure that the matching is unambiguous using the following rules:

* Each tolerance entry must be matched with at least one entry from "inputs.json".
* Every entry in "inputs.json" must have exactly one match from "tolerances.json".

If there is any ambiguity, the app stops with an error.

### CLI Changes
This adds a new `check` subcommand to the runner CLI with the following options:
* `check skeleton`: Print a skeleton tolerance JSON file.
* `check tolerances [NAME]`: List tolerance kinds, or describe one by name. This is similar to `inputs [NAME]`
* `check verify --tolerances <FILE> --input-file <FILE>`: Validate a tolerance file against an input file. This runs up to step 3 of the checklist above and serves as a pre-flight check before any CI jobs are run. Errors with the setup that can be caught early will be and can thus save CI time.
* `check run --tolerances <FILE> --input-file <FILE> --before <FILE> --after <FILE> [--output-file <FILE>]`: Run regression checks.

### Benchmark Registration Changes
Regression checks are opt-in. Benchmarks that wish to opt-in implement `benchmark::Regression` and the singular `check` method. All logic for the before and after comparison lives in the `check` method. Such benchmarks also need to use `registry::Benchmarks::register_regression` to be correctly tracked as regression compatible. No independent registration of the `Tolerance` associated type is needed.

That is it.

Note that check should not print anything out to `stdout` and instead communicate success/failure solely through its return type to avoid spamming the output.

### Example
You can see this in action in `diskann-benchmark-simd`. Run
```
# Generate the "before" file
cargo run --package diskann-benchmark-simd --release -- \
   run \
   --input-file ./diskann-benchmark-simd/examples/simd-scalar.json \
   --output-file before.json

# Generate the "after" file
cargo run --package diskann-benchmark-simd --release -- \
   run \
   --input-file ./diskann-benchmark-simd/examples/simd-scalar.json \
   --output-file after.json

# Run the before/after check
cargo run --package diskann-benchmark-simd --release -- \
  check run \
  --tolerances ./diskann-benchmark-simd/examples/tolerance.json \
  --input-file ./diskann-benchmark-simd/examples/simd-scalar.json \
  --before before.json \
  --after after.json
```
Depending on the noise in your system, you will see something like the following (note that the `ci` profile should be used for more reliable measurements):
```
Check 1 of 4 ("simd-op-f32xf32-scalar") FAILED:
     Distance,   Dim,   Min Before (ns),   Min After (ns),   Change (%),                                      Remark
====================================================================================================================
   squared_l2,   100,             0.000,            0.000,      invalid,   expected "before" to be greater than zero
   squared_l2,   128,             0.000,            0.000,      invalid,   expected "before" to be greater than zero
   squared_l2,   160,             0.000,            0.000,      invalid,   expected "before" to be greater than zero
   squared_l2,   384,            40.000,           40.000,      0.000 %,
   squared_l2,   768,            80.000,           80.000,      0.000 %,
inner_product,   100,             0.000,            0.000,      invalid,   expected "before" to be greater than zero
inner_product,   128,             0.000,            0.000,      invalid,   expected "before" to be greater than zero
inner_product,   160,             0.000,            0.000,      invalid,   expected "before" to be greater than zero
inner_product,   384,            40.000,            0.000,   -100.000 %,
inner_product,   768,           120.000,           40.000,    -66.667 %,
       cosine,   100,           120.000,           40.000,    -66.667 %,
       cosine,   128,            80.000,            0.000,   -100.000 %,
       cosine,   160,            40.000,           40.000,      0.000 %,
       cosine,   384,           120.000,           80.000,    -33.333 %,
       cosine,   768,           280.000,          160.000,    -42.857 %,
```
Where this particular implementation has decided that a regression cannot be meaningfully detected if an execution time got rounded to zero.

### Suggested Reviewing Order
**`diskann-benchmark-runner`**
* `benchmark.rs`: This contains the new `Regression` trait and the internal type-erasure machinery that is used inside the registry to mark a benchmark as regression-capable.
* `registry.rs`: The new `register_regression` API as well as methods for retrieving and interrogating regression capable benchmarks.
* `internal/regression.rs`: This is the meat and potatoes of this PR. This contains all the logic to perform the outlined steps and has a pretty good module-level documentation that outlines the approach taken.
* `app.rs`: The new `Check` subcommand and routing, this is pretty straight-forward save for the changes to tests (see below).
* `utils/num.rs`: I added two new opinionated utilities to help with writing checks:
  - `relative_change`: Compute the relative change between two values, handling corner cases.
  - `NonNegativeFinite`: Useful to serde-compatible assertions that tolerance entries have the obvious properties the name implies.
* `jobs.rs`: The changes here are to mainly make it easier to interact with the JSON patterns used by the input files to have more uniform handling of said files.

Changes to the testing infrastructure:
* The simple test benchmarks/inputs have been factored into their own submodules and given regression checks.
* The UX tests in `app.rs` have been upgraded to handle tolerance checks as well. This involves having more magically escaped patterns for input/output files and supporting multi-line `stdin.txt` files to allow regression tests to set up their respective environment.

Much of the logic in `internal/regression.rs` is tested via the UX tests because (1) setting up a proper environment for this functionality is challenging and (2) the UX tests provide much better visual information on the expected behavior. A lot of new UX tests have been added.

**`diskann-benchmark-simd`**:
The changes here are largely meant to work as an example of adding regression tests. Since `diskann-benchmark-simd` is not a production critical crate, feel free to just skim this or ignore entirely.

## Future Ideas
There are parts of this that are not perfect:
* Error diagnostics when matching fails could be better.
* This doesn't directly support situations where you change runtime values in the input file to generate the before and after comparison. It *will* work if the changes are sufficiently minimal to not influence benchmark matching. I decided not to include in this PR to keep the changes in API small (the internal changes in this PR are already huge).
* This doesn't natively support long-term regression monitoring where we have lots of `before` files and build a statistical picture of trends. That opens a whole can of worms like schema stability, machine stability (if comparing run times) etc. that are way beyond the scope of what's needed to get basic regression tests running.
* Adding a `--verbose` flag to print out all diagnostics rather than using the opinionated triage of errors then failures then successes.

*Disclaimer*: This PR (though not the PR description) was written with the help of AI but has been heavily edited and is something that I wouldn't be annoyed at reviewing outside of it being +3500 lines (I am sorry).

### DIFF
--- a/diskann-benchmark-runner/src/app.rs
+++ b/diskann-benchmark-runner/src/app.rs
@@ -430,11 +430,11 @@ impl App {
 ///
 /// Within the `stdin.txt` command line, there are several special symbols:
 ///
-/// * $INPUT_FILE - Resolves to `input.json` in the same directory as the `stdin.txt` file.
-/// * $OUTPUT_FILE - Resolves to `output.json` in a temporary directory.
-/// * $TOLERANCES_FILE - Resolves to `tolerances.json` in the test directory.
-/// * $REGRESSION_INPUT_FILE - Resolves to `regression_input.json` test directory.
-/// * $CHECK_OUTPUT_FILE - Resolves to `checks.json` in a temporary directory.
+/// * $INPUT - Resolves to `input.json` in the same directory as the `stdin.txt` file.
+/// * $OUTPUT - Resolves to `output.json` in a temporary directory.
+/// * $TOLERANCES - Resolves to `tolerances.json` in the test directory.
+/// * $REGRESSION_INPUT - Resolves to `regression_input.json` in the test directory.
+/// * $CHECK_OUTPUT - Resolves to `checks.json` in a temporary directory.
 ///
 /// As mentioned - an input JSON file can be included and must be named "input.json" to be
 /// discoverable.
--- a/diskann-benchmark-runner/src/jobs.rs
+++ b/diskann-benchmark-runner/src/jobs.rs
@@ -116,8 +116,8 @@ impl Unprocessed {
 
 /// A partially loaded input file.
 ///
-/// To reach this point, we at least the structure of the input JSON to be correct and
-/// parseable. However, we have not yet mapped the raw JSON of any of the registered inputs.
+/// To reach this point, we require at least the structure of the input JSON to be correct
+/// and parseable. However, we have not yet mapped the raw JSON of any of the registered inputs.
 #[derive(Debug, Serialize, Deserialize)]
 pub(crate) struct Partial {
     /// Directories to search for input files.
--- a/diskann-benchmark-runner/tests/regression/check-verify-4/README.md
+++ b/diskann-benchmark-runner/tests/regression/check-verify-4/README.md
@@ -1,2 +1,2 @@
-Here, we test that if a valid "tolerance" is linked with a valid "input" in "tolerance.json"
+Here, we test that if a valid "tolerance" is linked with a valid "input" in "tolerances.json"
 but there is not registered benchmark where these two tags are linked, we get a reasonable error.


diskann-benchmark-runner/src/internal/regression.rs
```diff
@@ -0,0 +1,860 @@
...
+                } = check;
+                Job {
+                    regression,
+                    tolerance,
+                    input,
+                    before,
+                    after,
+                }
+            })
+            .collect();
```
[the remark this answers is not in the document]
hildebrandmw: Sorry, but this check cannot be meaningfully performed at this stage. And indeed, that is the entire purpose of this PR. Only the associated regression benchmark knows how to both deserialize and validate the `before` and `after` inputs. Don't worry, these checks *will* happen when the job is run and the results are aggregated. It just cannot happen at this step.

diskann-benchmark-simd/src/bin.rs
```diff
@@ -74,4 +87,17 @@ mod tests {
...
+        let input_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("examples")
+            .join("simd-scalar.json");
+        let tolerance_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("examples")
+            .join("tolerance.json");
+
+        let stdout = run_check_test(&input_path, &tolerance_path);
+        println!("stdout = {}", stdout);
+    }
```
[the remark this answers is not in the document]
hildebrandmw: That's not necessarily true: we cannot meaningfully run a full regression test in tests because it uses the wrong optimization settings and the meaningful tests would have to run so long to get reliable data that is not appropriate for a unit test. The use of `unwrap` on the inner application call verifies at least that the structure of the files is sane, which is what this test is verifying.

diskann-benchmark-simd/src/lib.rs
```diff
@@ -561,16 +728,22 @@ trait RunBenchmark {
...
     /// The latencies of individual runs.
     latencies: Vec<MicroSeconds>,
     /// Latency percentiles.
     percentiles: percentiles::Percentiles<MicroSeconds>,
 }
 
+impl RunResult {
+    fn computations_per_latency(&self) -> usize {
+        self.run.num_points.get() * self.run.loops_per_measurement.get()
+    }
```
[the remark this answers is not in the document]
hildebrandmw: This is noted - but this is not a production crate so I don't think it's a big deal.

diskann-benchmark-runner/tests/regression/check-verify-4/README.md
```diff
@@ -0,0 +1,2 @@
+Here, we test that if a valid "tolerance" is linked with a valid "input" in "tolerance.json"
+but there is not registered benchmark where these two tags are linked, we get a reasonable error.
```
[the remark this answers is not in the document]
hildebrandmw: I'm not entirely sure what you're getting at here, I think the readme is correct.

diskann-benchmark-runner/src/jobs.rs
```diff
@@ -105,6 +114,10 @@ impl Unprocessed {
     }
 }
 
+/// A partially loaded input file.
+///
+/// To reach this point, we at least the structure of the input JSON to be correct and
+/// parseable. However, we have not yet mapped the raw JSON of any of the registered inputs.
```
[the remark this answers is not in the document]
hildebrandmw: That's fair, I lost a word. Thanks.

diskann-benchmark-runner/src/app.rs
```diff
@@ -228,16 +428,19 @@ impl App {
 ///
 /// Within the `stdin.txt` command line, there are several special symbols:
 ///
-/// * $INPUT - Resolves to `input.json` in the same directory as the `stdin.txt` file.
-/// * $OUTPUT - Resolves to `output.json` in a temporary directory.
+/// * $INPUT_FILE - Resolves to `input.json` in the same directory as the `stdin.txt` file.
+/// * $OUTPUT_FILE - Resolves to `output.json` in a temporary directory.
+/// * $TOLERANCES_FILE - Resolves to `tolerances.json` in the test directory.
+/// * $REGRESSION_INPUT_FILE - Resolves to `regression_input.json` test directory.
+/// * $CHECK_OUTPUT_FILE - Resolves to `checks.json` in a temporary directory.
```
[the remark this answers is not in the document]
hildebrandmw: Updated.

diskann-benchmark-runner/src/internal/regression.rs
```diff
@@ -0,0 +1,860 @@
...
+    // If we exit on the first iteration, the vector stays empty and thus doesn't allocate.
+    let mut stack = Vec::new();
+    loop {
+        match (haystack, needle) {
+            (Value::Null, Value::Null) => {
+                // Null always matches
+            }
+            (Value::Bool(h), Value::Bool(n)) => false_if!(h != n),
+            (Value::Number(h), Value::Number(n)) => false_if!(h != n),
+            (Value::String(h), Value::String(n)) => false_if!(h != n),
```
ChenSunriseJiaBao, line 424: is there any scenario that we don't want to consider up/low cases?

diskann-benchmark-runner/src/internal/regression.rs
```diff
@@ -0,0 +1,860 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+//! This module contains the tolerance parsing, matching, and running logic.
```
ChenSunriseJiaBao, line 6: This design separates benchmark and regression checking.
Do we consider combining them? such as for each benchmark, there is one output file, and maybe we can simple to add --baseline for target build/config, then comparing the baseline and current output.

diskann-benchmark-runner/src/internal/regression.rs
hildebrandmw: That's a good question - I lean towards "no" for the following reasons:

1. The decision to normalize case in strings is type dependent and even system dependent. For example, it's valid to normalize file-path cases on windows because the OS is generally case insensitive, but most Unix filesystems **are** case sensitive. So the decision to normalize case needs to be made by the input type we are matching against, which leads to problem 2.
2. My desire to have fuzzy matching is basically to make the tolerance files easier to write and maintain. One consequence is that the input JSON that accompanies a tolerance JSON cannot properly deserialize into a valid input instance, meaning we can't use something like `PartialEq` to do the comparison. I had thought about allowing inputs to define an acceptance criteria like
    ```rust
    trait Accept {
        fn accept(&self, raw: &serde_json::Value, partial: &serde_json::Value) -> bool {
           is_subset(raw, partial)
        }
    }
    ```
   where we could delegate the entire matching process to the inputs, rather than using this blanket match for everything. The downside is that this adds yet another trait that inputs need to implement, even if it does end up being just a marker trait most of the time due to the provided implementation.

  But this then opens a can of worms of extending the trait in the future to provide better diagnostics and potentially unintuitive matching behavior, and so I just kept it simple.

diskann-benchmark-runner/src/internal/regression.rs
hildebrandmw: I had thought about that approach but went with this approach here for a few reasons. First, adding a `--baseline` mode to the normal benchmark flow introduces two potential problems.
* It requires that either all benchmarks implement regression checks. This is not something I wanted to do: mandatory regression checks mean that they will often be implemented hurriedly to get things to compile - and a bad regression check is worse than no regression check. 
* Or it requires that the regression capability matching is fused with the normal benchmark dispatching, which seemed like an unnecessary complication.

Reporting is also messier with a fused approach. A separate check flow allows all pass/fail results to be aggregated at once. With a fused regression check, I don't know whether regression checks should be displayed inline with the normal results or rolled up at the very end - and if a check fails do we abort early or not? These should probably all be options, but that makes the implementation more complicated.

Finally, the embedded flow is tedious for tuning tolerances. An after check allows multiple tolerances to be tried and the matching scheme to be refined without rerunning the benchmark constantly.
 
Overall, I felt a separate flow with pre-validation gave a cleaner separation of concerns.

diskann-benchmark-simd/examples/tolerance.json
```diff
@@ -0,0 +1,16 @@
...
+  "checks": [
+    {
+      "input": {
+        "type": "simd-op",
+        "content": {}
+      },
+      "tolerance": {
+        "type": "simd-tolerance",
+        "content": {
+          "min_time_regression": 0.05
```
arrayka, line 11: @hildebrandmw , is it an expected usage of the proposed pattern:
1. Create tolerance.json similar to this:
```
{
    "checks": [
        {
            "input": {
                "type": "disk-index",
                "content": {}
            },
            "tolerance": {
                "type": "disk-index-tolerance",
                "content": {
                    "build_time_regression": 0.10,
                    "qps_regression": 0.10,
                    "recall_regression": 0.01,
                    "mean_ios_regression": 0.01,
                    "mean_comps_regression": 0.01,
                    "mean_latency_regression": 0.15,
                    "p95_latency_regression": 0.15
                }
            }
        }
    ]
}
```

2. Add ~300 lines of Rust code to wire it up including implementation of check_lower_is_better() and check_higher_is_better():
https://github.com/microsoft/DiskANN/pull/912/changes#diff-f77d4bbaebfd48c9785d878c56f274b1aa3414d88b3896aed961acdd3904e194R131-R450
```
...
            comparisons.push(check_higher_is_better(
                "qps",
                b_sr.qps as f64,
                a_sr.qps as f64,
                tolerances.qps_regression,
                &mut passed,
            ));
...
```

Is it possible to avoid this step 2? It feels like the higher-the-better knowledge belongs to tolerance.json.

diskann-benchmark-simd/examples/tolerance.json
```diff
@@ -0,0 +1,16 @@
...
+        }
+      }
+    }
+  ]
+}
```
arrayka, line 16: Thinking aloud:

I would love us to make benchmarking **effortless by default**, with more advanced capabilities unlocking incrementally as needed, where common workflows work out of the box with little to no user configuration. 

I'm thinking about these scenarios (the next one builds on top of previous one):

1.  **Single run, easy sharing**  
    Run a benchmark and export results to Excel, Word, or Markdown for quick sharing with the team.

2.  **Zero‑config before/after comparisons**  
    Run benchmarks before and after a change and automatically produce a table with absolute and relative metric diffs, export the table to Excel, Word, or Markdown — no extra configuration, no diff reasoning needed.

3.  **Optional passed/failed signal for each metric**  
    Introduce metric metadata (`higher_better` / `lower_better`, thresholds) to enable green/red highlighting with minimal effort.

4.  **Add CI support to (3.)**  
    Add CI support via something like --fail_on_error, so the benchmark fails when thresholds are violated.

hildebrandmw: Thank you Alex. Some of these are indeed limitations, but some of these are already achievable:
1. This is the biggest weakness of the current approach of dumping things to JSON and I'd really like to improve the story here. I've learned from experience that generating as rich of results as possible (i.e. `output.json`) is very helpful when it comes to data analysis, but not so much data presentation. Would love for more cycles to be put into this for the tool!
2. You can see a more detailed response [here](https://github.com/microsoft/DiskANN/pull/900#discussion_r3045930571), but the TLDR is that the split into a separate executable supports keeping regression-testing opt-in and tolerance iteration a little easier. That said, the `Display` impls on the `PassFail` structs are where rich formatting can happen in this PR. We could definitely work on infrastructure to make these `Display` implementations prettier though to make the output more useful by default.
3. This is already possible with proper `Display` implementations on the returned `PassFail` structs. The benchmark crate could make these a little more accessible though, but the [colored](https://docs.rs/colored/latest/colored/) crate is available pretty easily for this.
4. This *should* already work: the `check` sub executable will exit with an `Error` when at least one inner check fails with an error or doesn't pass. This in turn exits from `main` with a non-zero exit code that CI can use to indicate a failure. Indeed, **any** error encountered this way will result in a non-zero exit code :smile:.

diskann-benchmark-simd/examples/tolerance.json
hildebrandmw: I would argue that the `higher-the-better` knowledge is exactly what shouldn't live in the `tolerance.json`. The writer of the benchmark knows which direction the change should be in and can structurally enforce that in the code. That said, I would recommend using a macro to stamp out much of the boiler plate. I would also say that reusing `NonNegativeFinite` to describe negative bounds is a little confusing.

ChenSunriseJiaBao (approved):

PR #912, opened 2026-04-07
Add benchmark pipeline with Rust-native A/B validation

YuanyuanTian-hh:
### Summary

Adds an automated benchmark regression pipeline to GitHub Actions. This PR builds on PR #900 (benchmark A/B test framework) by implementing the `Regression` trait for disk-index benchmarks and wiring it into CI workflows. The pipeline builds and searches two public 100K ANN datasets, compares before/after performance, and validates against configurable tolerances  all in typed Rust code with no Python dependencies.

### What's Changed

**Benchmark Runner (from PR #900)**
- Rust-native A/B test framework in `diskann-benchmark-runner`
- `Regression` trait, tolerance matching, `check run` CLI subcommand

**Disk-Index Regression Support (this PR)**

| File | Description |
|------|-------------|
| `diskann-benchmark/src/backend/disk_index/benchmarks.rs` | Implements `Regression` trait for `DiskIndex<T>` with typed before/after comparison of 7 metrics |
| `diskann-benchmark/src/backend/disk_index/search.rs` | Added `Deserialize` to `DiskSearchStats`, `DiskSearchResult` |
| `diskann-benchmark/src/backend/disk_index/build.rs` | Added `Deserialize` to `DiskBuildStats`, exposed `build_time_seconds()` |
| `diskann-benchmark/perf_test_inputs/disk-index-tolerances.json` | Tolerance config: 10% build/QPS, 1% recall/IOs/comps, 15% latency |
| `diskann-benchmark/perf_test_inputs/wikipedia-100K-disk-index.json` | Benchmark config: 768-dim, inner_product, search_list=[200] |
| `diskann-benchmark/perf_test_inputs/openai-100K-disk-index.json` | Benchmark config: 1536-dim, squared_l2, SQ_1_2.0, search_list=[200] |

**CI Workflows**

| File | Description |
|------|-------------|
| `.github/workflows/benchmarks.yml` | PR regression workflow  triggers on `pull_request` to main, validates with `cargo ... check run` |
| `.github/workflows/benchmarks-aa.yml` | Daily A/A stability test (main vs main) with GitHub issue on failure |

### How It Works

1. **Checkout** both current branch and baseline (defaults to `main`)
2. **Download** datasets from [big-ann-benchmarks v0.4.0](https://github.com/harsha-simhadri/big-ann-benchmarks/releases/tag/v0.4.0)
3. **Build & search** disk index on both branches
4. **Validate** with `cargo run -p diskann-benchmark --features disk-index --release -- check run --tolerances ... --before ... --after ...`
5. **Upload** JSON artifacts for 30-day retention

### Regression Checks

The `Regression` trait implementation compares 7 metrics between before and after runs:

| Metric | Direction | Tolerance | Rationale |
|--------|-----------|-----------|-----------|
| build_time | Lower is better | 10% | CPU-bound, moderate noise |
| QPS | Higher is better | 10% | CPU-bound, moderate noise |
| recall | Higher is better | 1% | Algorithmic, near-deterministic |
| mean_ios | Lower is better | 1% | Algorithmic, near-deterministic |
| mean_comparisons | Lower is better | 1% | Algorithmic, near-deterministic |
| mean_latency | Lower is better | 15% | Timing, noisy on shared runners |
| p95_latency | Lower is better | 15% | Timing, noisy on shared runners |

All checks use `relative_change()` which properly handles zero-baseline (returns error, not 0%).

### Datasets

| Dataset | Dimensions | Distance | Vectors | Queries | search_list |
|---------|-----------|----------|---------|---------|-------------|
| Wikipedia-100K | 768 | inner_product | 100K | 5,000 | 200 |
| OpenAI ArXiv-100K | 1,536 | squared_l2 | 100K | 20,000 | 200 |

### Improvements over PR #857

This PR addresses all blocking review comments from PR #857:

| Concern | PR #857 (Python) | PR #912 (Rust) |
|---------|-----------------|----------------|
| Phantom thresholds / silent success |  Orphaned categories silently skipped |  Runner errors if tolerances don't match inputs |
| Bad direction values pass silently |  Falls through to else branch |  Type-safe, no string directions |
| Division by zero masked as 0% |  Returns 0 when baseline=0 |  Returns error ("before must be > 0") |
| Missing fields default to 0 |  `.get(field, 0)` |  Rust deserialization fails on missing fields |
| Manual trigger only |  workflow_dispatch only |  pull_request trigger with path filters |
| CI time ~70 min |  search_list=2000 |  search_list=200, ~10 min per job |
| Hardcoded Rust version |  rust_stable: "1.92" |  toolchain: stable, reads rust-toolchain.toml |
| Python dependencies |  benchmark_validate.py |  Zero Python, pure Rust |


.github/workflows/disk-benchmarks.yml
```diff
@@ -0,0 +1,231 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
arrayka, line 1: is it by design?
change 0.105%, threshold 0.1%. outcome: passed

https://github.com/microsoft/DiskANN/actions/runs/24072874830/job/70214545470

```
Run cargo run -p diskann-benchmark --features disk-index --release -- \
    Finished `release` profile [optimized + debuginfo] target(s) in 0.15s
     Running `target/release/diskann-benchmark check run --tolerances diskann-benchmark/perf_test_inputs/disk-index-tolerances.json --input-file diskann-benchmark/perf_test_inputs/wikipedia-100K-disk-index.json --before ../baseline/target/tmp/wikipedia-100K_benchmark_crate_baseline.json --after target/tmp/wikipedia-100K_benchmark_crate_target.json`
Check 1 of 1 ("disk-index-f32") PASSED:
  Search L=200:          Before           After       Change    Tolerance   Remark
  ==========================================================================================
            build_time,         35.436,         34.957,     -1.352%,         0.1%,   
                   qps,         92.122,         92.219,      0.105%,         0.1%,   
                recall,         91.892,         91.895,      0.004%,         0.1%,   
          mean_latency,      43334.354,      43338.147,      0.009%,         0.1%,   
           p95_latency,      61337.000,      61268.000,     -0.112%,         0.1%,   
              mean_ios,        211.039,        210.927,     -0.053%,         0.1%,   
      mean_comparisons,       5281.971,       5277.767,     -0.080%,         0.1%,   
```

arrayka: Ok. It is by design. It's QPS - the higher the better. It means that 0.105% is good.

Then it looks like formatting of thresholds is misleading. It does not match the config - for example, mean_ios_regression is 0.01 but printed as 0.1%:
```
            "tolerance": {
                "type": "disk-index-tolerance",
                "content": {
                    "build_time_regression": 0.10,
                    "qps_regression": 0.10,
                    "recall_regression": 0.01,
                    "mean_ios_regression": 0.01,
                    "mean_comps_regression": 0.01,
                    "mean_latency_regression": 0.15,
                    "p95_latency_regression": 0.15
                }
            }
```

YuanyuanTian-hh: The 0.1% display was from the test commit where I temporarily set all tolerances to 0.001 to verify failure detection. That commit has been reverted. With the production tolerances (0.10, 0.01, 0.15), the display correctly shows 10.0%, 1.0%, 15.0% respectively.
Here is the test run: 

- success [Add benchmark pipeline with Rust-native A/B validation · microsoft/DiskANN@5de8295](https://github.com/microsoft/DiskANN/actions/runs/24070983036)
- failure [Add benchmark pipeline with Rust-native A/B validation · microsoft/DiskANN@6953701](https://github.com/microsoft/DiskANN/actions/runs/24113710995/job/70353452728#step:11:71)

.github/workflows/disk-benchmarks-aa.yml
```diff
@@ -0,0 +1,237 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT license.
+
+# DiskANN Daily A/A Benchmark Stability Test
```
ChenSunriseJiaBao, line 4: I have two concerns for this A/A benchmark:

1. The target and baseline are same main branch, what is the goal?
2. I can see there are lots of duplication code between job aa-wikipedia-100K and job aa-openai-100K, the difference is only dataset and config json file, we need to ensure our code is simple and clear.

diskann-benchmark/src/backend/disk_index/benchmarks.rs
```diff
@@ -124,8 +128,324 @@ where
...
+/// Tolerance thresholds for disk-index regression checks.
+///
+/// Each field specifies the maximum allowed relative increase (for "lower is better" metrics)
+/// or decrease (for "higher is better" metrics) before a regression is flagged.
+///
+/// For example, `recall_regression: 0.01` means recall must not drop by more than 1%.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub(super) struct DiskIndexTolerance {
+    /// Max allowed relative increase in build time (e.g., 0.10 = 10%).
+    build_time_regression: NonNegativeFinite,
```
ChenSunriseJiaBao, line 151: do we need to define the data type as NonNegativeFinite<f64>?

diskann-benchmark/src/backend/disk_index/benchmarks.rs
```diff
@@ -124,8 +128,324 @@ where
...
+/////////////////////////
+
+/// Tolerance thresholds for disk-index regression checks.
+///
+/// Each field specifies the maximum allowed relative increase (for "lower is better" metrics)
+/// or decrease (for "higher is better" metrics) before a regression is flagged.
+///
+/// For example, `recall_regression: 0.01` means recall must not drop by more than 1%.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub(super) struct DiskIndexTolerance {
```
ChenSunriseJiaBao, line 149: do we need to add some default values as below?

fn default_build_time_tolerance() -> NonNegativeFinite<f64> { /* 0.10 */ }
fn default_recall_tolerance() -> NonNegativeFinite<f64> { /* 0.01 */ }
fn default_qps_tolerance() -> NonNegativeFinite<f64> { /* 0.10 */ }

diskann-benchmark/perf_test_inputs/openai-100K-disk-index.json
```diff
@@ -0,0 +1,39 @@
...
+    ],
+    "jobs": [
+        {
+            "type": "disk-index",
+            "content": {
+                "source": {
+                    "disk-index-source": "Build",
+                    "data_type": "float32",
+                    "data": "OpenAIArXiv/openai_base.bin.crop_nb_100000",
+                    "distance": "squared_l2",
```
ChenSunriseJiaBao, line 13: I saw we will use Inner_Product in wikipedia-100k-disk-index. Is there any reason to we set different distance metric?

diskann-benchmark/perf_test_inputs/openai-100K-disk-index.json
```diff
@@ -0,0 +1,39 @@
...
+                    "dim": 1536,
+                    "max_degree": 59,
+                    "l_build": 80,
+                    "num_threads": 4,
```
ChenSunriseJiaBao: for index build, we can increase thread number to 8 to reduce the build time.

diskann-benchmark/perf_test_inputs/wikipedia-100K-disk-index.json
```diff
@@ -0,0 +1,39 @@
...
+                    "data_type": "float32",
+                    "data": "wikipedia_cohere/wikipedia_base.bin.crop_nb_100000",
+                    "distance": "inner_product",
+                    "dim": 768,
+                    "max_degree": 59,
+                    "l_build": 72,
+                    "num_threads": 4,
+                    "build_ram_limit_gb": 4.0,
+                    "num_pq_chunks": 192,
+                    "quantization_type": "FP",
```
ChenSunriseJiaBao: can we always use SQ_1_2.0?

diskann-benchmark/perf_test_inputs/wikipedia-100K-disk-index.json
```diff
@@ -0,0 +1,39 @@
...
+            "type": "disk-index",
+            "content": {
+                "source": {
+                    "disk-index-source": "Build",
+                    "data_type": "float32",
+                    "data": "wikipedia_cohere/wikipedia_base.bin.crop_nb_100000",
+                    "distance": "inner_product",
+                    "dim": 768,
+                    "max_degree": 59,
+                    "l_build": 72,
```
ChenSunriseJiaBao: L_Build should be 80

.github/workflows/disk-benchmarks-aa.yml
YuanyuanTian-hh: 1. **Goal of A/A**: Alex wants A/A testing to happen daily and fail if deviation is more than expected threshold - notifying diskann-admins about failure. The A/A test runs main vs main (identical code) to detect **environment noise**. If the thresholds breach on identical code, it means runner variability is too high and we need to tune thresholds  not that there's a code regression. This is a standard practice for benchmark pipelines on shared CI runners.

2. **Duplication**: Agreed  refactored to use a matrix strategy. Now there's a single \\a-benchmark\\ job with a matrix over datasets. Much cleaner.

diskann-benchmark/src/backend/disk_index/benchmarks.rs
YuanyuanTian-hh: NonNegativeFinite is a newtype wrapper from the benchmark-runner framework (PR #900). It guarantees at deserialization time that tolerance values are non-negative and finite  preventing invalid configs like negative tolerances or Infinity from being silently accepted. It's the idiomatic type for this use case in the framework.

diskann-benchmark/src/backend/disk_index/benchmarks.rs
YuanyuanTian-hh: Intentionally kept without serde defaults. All tolerance fields are required in the JSON config so that when a new metric is added, the config must be explicitly updated rather than silently falling back to a default. The example() method provides reference values (0.10, 0.01) for the check skeleton / check tolerances CLI commands.

diskann-benchmark/perf_test_inputs/openai-100K-disk-index.json
YuanyuanTian-hh: The two datasets use different distance metrics because their groundtruth files were built with different metrics. OpenAI-ArXiv groundtruth uses squared_l2 (euclidean), while Wikipedia groundtruth uses inner_product. This was also discussed in PR #857  switching OpenAI to a different metric (e.g. cosine_normalized) causes recall to drop to near 0%% because it no longer matches the groundtruth.

diskann-benchmark/perf_test_inputs/openai-100K-disk-index.json
YuanyuanTian-hh: Done  updated both OpenAI and Wikipedia configs to num_threads: 8 for the build phase.

diskann-benchmark/perf_test_inputs/wikipedia-100K-disk-index.json
YuanyuanTian-hh: Done  updated Wikipedia config from FP to SQ_1_2.0 so both datasets use the same quantization type consistently.

diskann-benchmark/perf_test_inputs/wikipedia-100K-disk-index.json
YuanyuanTian-hh: Done  updated l_build from 72 to 80. Same value as the OpenAI config (which was already set to 80 since PR #857 per your earlier review).

diskann-benchmark/src/backend/disk_index/benchmarks.rs
```diff
@@ -124,8 +128,324 @@ where
...
+/// Aggregated result of a disk-index regression check.
+#[derive(Debug, Serialize)]
+struct DiskIndexCheckResult {
+    search_l: u32,
+    comparisons: Vec<MetricComparison>,
+}
+
+impl fmt::Display for DiskIndexCheckResult {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        writeln!(
```
hildebrandmw: Using `diskann_benchmark_runner::utils::fmt::Table` will ensure that the resulting display is always vertically aligned correctly.

diskann-benchmark/src/backend/disk_index/benchmarks.rs
```diff
@@ -124,8 +128,324 @@ where
...
+            ));
+        }
+
+        let search_l = before
+            .search
+            .search_results_per_l
+            .first()
+            .map(|s| s.search_l)
+            .unwrap_or(0);
+        let result = DiskIndexCheckResult {
```
hildebrandmw: This seems a little misleading if multiple `search_l` values are used?

diskann-benchmark/src/backend/disk_index/benchmarks.rs
```diff
@@ -124,8 +128,324 @@ where
...
+                c.after,
+                c.change_pct,
+                c.tolerance_pct * 100.0,
+                c.remark
+            )?;
+        }
+        Ok(())
+    }
+}
+
```
hildebrandmw: If you use something like
```rust
#[derive(Clone, Copy)
enum Direction {
    LowerIsBetter,
    HigherIsBetter,
}
```
The functions `check_lower_is_better` and `check_higher_is_better` can be fused into one relatively cleanly.

.github/workflows/benchmarks-aa.yml
```diff
@@ -0,0 +1,163 @@
...
+          ref: main
+          path: diskann_rust
+          lfs: true
+
+      - name: Checkout main (baseline)
+        uses: actions/checkout@v4
+        with:
+          ref: main
+          path: baseline
+          lfs: true
```
hildebrandmw: Why perform two independent clones and builds? Why not build once and run twice to save CI time?

hildebrandmw (commented):
Thanks! This is headed in a good direction. In addition to the individual comments below, I have a couple small requests regarding the CI .yml files.
* Can these be named more accurately as `disk-benchmarks[-aa].yml`? There will presumably be other benchmarks and we might as well make the distinction early.
* There is considerable duplication in the `.yml` structure as well. We should strive to keep our CI system as tidy as we can from the get go. Is it possible to factor out some reusable components (like dataset downloading) so other pipelines can be brought up more easily and if we need to change URLs, there is one place to do so.
* On tidiness, the file paths for finding datasets is pretty verbose and repetitive. Job level environment variables could go a long way towards cutting this down and making it easier to comprehend at a glance.

diskann-benchmark/src/backend/disk_index/benchmarks.rs
YuanyuanTian-hh: Done  switched to `Table::new(header, nrows)` with columns: Metric, Before, After, Change, Tolerance, Remark.

diskann-benchmark/src/backend/disk_index/benchmarks.rs
YuanyuanTian-hh: Done  added `enum Direction { LowerIsBetter, HigherIsBetter }` and fused both functions into a single `check_metric(name, direction, before, after, tolerance, passed)`.

diskann-benchmark/src/backend/disk_index/benchmarks.rs
YuanyuanTian-hh: Good catch  removed the `search_l` field from `DiskIndexCheckResult`. When multiple `search_l` entries exist, metric names are now prefixed with `L{value}:` (e.g. `L50:qps`, `L50:recall`). When there is only one entry, names stay plain (`qps`, `recall`).

.github/workflows/benchmarks-aa.yml
YuanyuanTian-hh: Fixed  the A/A workflow now checks out once and does `cargo build` once, then runs the binary twice (baseline + target). No duplicate clone or compile.

YuanyuanTian-hh:
All addressed:
- Renamed to `disk-benchmarks.yml` / `disk-benchmarks-aa.yml`.
- Factored out shared setup into `.github/actions/setup-disk-benchmark/action.yml` composite action (Rust toolchain, cargo cache, system deps install, dataset download with `BAB_RELEASE_URL` as single source of truth).
- Added `PERF_INPUTS` workflow-level env var to reduce path repetition.

.github/workflows/disk-benchmarks.yml
```diff
@@ -0,0 +1,142 @@
...
+      baseline_ref:
+        description: 'A branch, commit SHA, or tag name to compare the current branch with'
+        required: true
+        default: 'main'
+        type: string
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'diskann/**'
```
arrayka, line 23: I’d remove these path filters—they’re error-prone, and the directory structure is changing frequently. We don’t want to maintain a growing list of paths. Instead, let’s run the benchmarks on every change.

arrayka (approved):

diskann-benchmark/src/backend/disk_index/benchmarks.rs
```diff
@@ -124,8 +129,295 @@ where
...
+            // Prefix metric names with L value when multiple search_l entries exist.
+            let prefix = if before.search.search_results_per_l.len() > 1 {
+                format!("L{}:", b_sr.search_l)
+            } else {
+                String::new()
+            };
+
+            comparisons.push(check_metric(
+                Box::leak(format!("{}qps", prefix).into_boxed_str()),
```
hildebrandmw: ﻿The fix here is to change `MetricComparison::metric` from `&'static str` to `String`, not to leak memory. `Box::leak` is a permanent allocation that never gets freed.

.github/workflows/disk-benchmarks.yml
```diff
@@ -0,0 +1,142 @@
...
+  RUST_BACKTRACE: 1
+  PERF_INPUTS: diskann-benchmark/perf_test_inputs
+
+defaults:
+  run:
+    shell: bash
+
+permissions:
+  contents: read
+  pull-requests: write  # Required for posting PR comments
```
hildebrandmw: This no longer posts comments and does not need `write` permissions. Please remove.

.github/actions/setup-disk-benchmark/action.yml
```diff
@@ -0,0 +1,51 @@
...
+    description: 'Suffix for the Rust dependency cache key'
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Install Rust
+      uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: stable
```
hildebrandmw: You can follow the pattern used in `ci.yml`. We've changed the way we install Rust so it follows `rust-toolchain.toml`. Basically:
```yaml
- name: Install Rust
  shell: bash
  run: rustup show
```

.github/actions/setup-disk-benchmark/action.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    - name: Cache Rust dependencies
+      uses: Swatinem/rust-cache@v2
+      with:
+        workspaces: diskann_rust -> target
+        key: ${{ inputs.rust-cache-key }}
```
hildebrandmw: We can probably remove caching here? The problem with universally caching everything is that it eats into our precious cache limit. Since these benchmarks already take a while to run, it's probably better to save our cache space for the shorter running CI jobs where cache hits are more impactful.

.github/workflows/disk-benchmarks-aa.yml
```diff
@@ -0,0 +1,137 @@
...
+                ``,
+                `The scheduled A/A benchmark run (main vs main) **failed** on ${date}.`,
+                `This indicates environment noise exceeded the configured thresholds.`,
+                ``,
+                `**Run:** ${runUrl}`,
+                ``,
+                `Please review the benchmark artifacts and determine if thresholds need tuning`,
+                `or if there is a runner environment issue.`,
+                ``,
+                `/cc @microsoft/diskann-admin`,
```
hildebrandmw: Admin is the wrong group to target here. This should target the owners directly responsible for maintaining the integrity of the disk index.

YuanyuanTian-hh:
@microsoft-github-policy-service agree

PR #914, opened 2026-04-07
Forbid Debug in diskann-benchmark

arrayka:
## Problem

The benchmark tools (`diskann-benchmark` and `diskann-benchmark-simd`) allow running benchmarks on code compiled in debug mode without any warnings or errors. This leads to misleading performance results since debug builds:
- Lack compiler optimizations
- Include bounds checking and overflow checks
- Have debug assertions enabled
- Are significantly slower than release builds

Users can accidentally run benchmarks in debug mode and draw incorrect performance conclusions.

## Solution

Added a compile-time check using `cfg!(debug_assertions)` that blocks benchmark execution in debug mode with a clear error message. Users can bypass this with the `--allow-debug` flag if intentional.

The `--allow-debug` flag is scoped to the `run` subcommand:

```rust
Commands::Run {
    // ... other fields
    #[arg(long, action)]
    allow_debug: bool,
}
```

## How to review

The check is implemented in `diskann-benchmark-runner/src/app.rs`. All other changes are limited to test code.

### DIFF
--- a/diskann-benchmark-runner/src/app.rs
+++ b/diskann-benchmark-runner/src/app.rs
@@ -75,7 +75,8 @@ use crate::{
 
 /// Check if we're running in debug mode and error if not allowed.
 fn check_debug_mode(allow_debug: bool) -> anyhow::Result<()> {
-    if cfg!(debug_assertions) && !allow_debug {
+    // Unit tests are treated as debug mode to ensure consistent behavior across builds.
+    if cfg!(any(test, debug_assertions)) && !allow_debug {
         anyhow::bail!(
             "Benchmarking in debug mode produces misleading performance results.\n\
              Please compile in release mode or use the --allow-debug flag to bypass this check."
--- a/diskann-benchmark-runner/tests/benchmark/test-debug-mode-error/README.md
+++ b/diskann-benchmark-runner/tests/benchmark/test-debug-mode-error/README.md
@@ -3,6 +3,3 @@
 This test verifies that running benchmarks in debug mode without the `--allow-debug` flag produces an error.
 
 The test intentionally omits the `--allow-debug` flag to ensure the debug mode check fires and blocks execution with an appropriate error message.
-
-- **In debug builds**: The debug mode check fires and blocks execution with an error message (validated against `stdout.txt`)
-- **In release builds**: No debug assertions, so the benchmark runs normally with empty jobs (validated against `stdout_release.txt`)


diskann-benchmark-runner/src/app.rs
```diff
@@ -353,7 +373,13 @@ mod tests {
...
 
         fn run(&self, tempdir: &Path) {
-            let app = self.parse_stdin(tempdir);
+            let mut app = self.parse_stdin(tempdir);
+
+            // For tests, always allow debug mode to prevent test failures when running
+            // in debug configuration.
+            if let Commands::Run { allow_debug, .. } = &mut app.command {
+                *allow_debug = true;
+            }
```
hildebrandmw: Thanks @arrayka. Instead of universally enabling `allow_debug` in UX tests, can we instead selectively add the `--allow-debug` flag to the UX tests that would now fail? Also, can we add a new UX test that neglects to pass `--allow-debug` to ensure this check fires? Since the test suite always runs with debug assertions, I think this should work.

arrayka: Done.
- I've added -allow-debug to stdin.txt of failing UX tests.
- I've added `tests/benchmark/test-debug-mode-error` test
- In release mode, the UX test runner uses the release version of expected output: `stdout_release.txt`‎ (if it exists).

diskann-benchmark-runner/src/app.rs
```diff
@@ -663,7 +656,19 @@ mod tests {
             let stdout: String =
                 ux::normalize(ux::strip_backtrace(buffer.into_inner().try_into().unwrap()));
             let stdout = ux::scrub_path(stdout, tempdir, "$TEMPDIR");
-            let output = self.dir.join(STDOUT);
+
+            let output = if cfg!(debug_assertions) {
```
hildebrandmw: Can we drop this special case here and instead make the logic in `check_debug_mode`:
```rust
    if cfg!(any(test, debug_assertions)) && !allow_debug {
        anyhow::bail!(
            "Benchmarking in debug mode produces misleading performance results.\n\
             Please compile in release mode or use the --allow-debug flag to bypass this check."
        );
    }
    Ok(())
```
The reason is that bifurcating the tests into release-mode expectations and debug-mode expectations makes the UX tests harder to maintain by not regenerating all expected output in one go. Instead, we need two runs.

Now, this doesn't necessarily test the case that we **don't** get a diagnostic when we **are** compiling in release mode. But since that is so visible and the soon-to-be-added regression tests will exercise that by running benchmarks in release mode, I think it's okay.

arrayka: Ah. Good idea. Done.

PR #928, opened 2026-04-08
migrating multi-hop tests from diskann-providers to diskann

JordanMaples:
This is a first step in the effort to migrate unit tests from diskann_async to diskann #927. If you don't think tests that have been brought over should be moved, they probably shouldn't have been. Please point them out and I'll do my best to put them back where I found them. 

  - Remove multihop and filter-related tests from diskann_async.rs
  - Add multihop.rs test module in diskann with equivalent coverage using test_provider::Provider::grid() instead
 of the inmem provider
  - Add baseline JSON files for callback filtering, even filtering, and two-hop-through-non-matching scenarios
  - Make multihop_search module visible to tests (pub(crate) export in search/mod.rs)

 Key differences from the original tests:
  - Tests exercise the graph algorithm directly via Provider::grid() rather than going through the inmem provider
 with PQ quantization — appropriate since these test traversal behavior, not quantization
  - Start point filtering: the test provider doesn't strip start points from results (unlike the inmem
post-processor), so reject_all_returns_zero_results asserts zero non-start-point results rather than zero total
results


diskann-providers/src/test_utils/search_utils.rs
```diff
@@ -3,15 +3,22 @@
  * Licensed under the MIT license.
  */
 
+//! Test assertion helpers for search result verification.
+//!
+//! The canonical `groundtruth` implementation lives in `diskann::graph::test::search_utils`,
+//! but that module is gated behind `cfg(test)` / `feature = "testing"` and is not available
+//! in non-test builds. This module duplicates the functions needed by `diskann-providers`
+//! and `diskann-disk` so they compile unconditionally.
```
harsha-simhadri, line 11: should we address this duplication?

JordanMaples: I tried to have copilot address the duplication, but it kept running into problems. I'll give it another go today and see if Mark has any ideas on keeping a single source of truth for both modules to pull in cleanly

diskann/src/graph/test/cases/multihop.rs
```diff
@@ -0,0 +1,668 @@
...
+    let grid_size = 4;
+    let query = vec![grid_size as f32; 3];
+    let filter = RejectAllFilter::only([]);
+
+    let (_index, stats, ids, _distances) = run_multihop_search(grid_size, &query, 10, 20, &filter);
+
+    // The start point (u32::MAX) is seeded directly into the candidate set and bypasses
+    // both is_match and on_visit. It may appear in results. All non-start-point results
+    // should be zero since on_visit rejects everything.
+    let non_start_results = (0..stats.result_count as usize)
```
hildebrandmw: We probably shouldn't have a "maybe" in here. The start point is either included or it isn't. We've changed the default for the test provided to be excluded, so is this check still needed?

diskann/src/graph/test/cases/multihop.rs
```diff
@@ -0,0 +1,668 @@
...
+#[test]
+fn early_termination() {
+    let grid_size = 5;
+    let num_points = Grid::Three.num_points(grid_size);
+    let query = vec![grid_size as f32; 3];
+
+    let target = (num_points / 2) as u32;
+    let filter = TerminatingFilter::new(target);
+
+    let (_index, stats, _ids, _distances) = run_multihop_search(grid_size, &query, 10, 40, &filter);
```
hildebrandmw: We can include all of `grid_size`, `query`, `target`, `k` and `l` in a checked-in baseline, along with `stats`, `ids`, and `distances` to make this a much better test.

diskann/src/graph/test/cases/multihop.rs
```diff
@@ -0,0 +1,668 @@
...
+    let boosted_in_baseline = baseline_ids
+        .iter()
+        .take(baseline_stats.result_count as usize)
+        .position(|&id| id == boosted_point);
+    let boosted_in_adjusted = adjusted_ids
+        .iter()
+        .take(adjusted_stats.result_count as usize)
+        .position(|&id| id == boosted_point);
+
+    if filter.hits().contains(&boosted_point) {
```
hildebrandmw: This is a little surprising to have in a test. If the filter does not contain the boosted point, we don't check anything. I feel like we should know whether or not the boosted point was expected.

diskann/src/graph/test/cases/multihop.rs
```diff
@@ -0,0 +1,668 @@
...
+    let query = vec![grid_size as f32; 3];
+
+    let max_visits = 5;
+    let filter = TerminateAfterN::new(max_visits);
+
+    // Large L to ensure we'd visit more without termination.
+    let (_index, _stats, _ids, _distances) =
+        run_multihop_search(grid_size, &query, 10, 100, &filter);
+
+    // Allow some slack for beam expansion.
```
hildebrandmw: This comment is surprisingly cryptic. Again, a baseline capturing everything would be much better.

diskann/src/graph/test/cases/multihop.rs
```diff
@@ -0,0 +1,668 @@
...
+        result_ids.contains(&4),
+        "node 4 should be found via two-hop traversal through non-matching node 3, \
+         got results: {:?}",
+        result_ids
+    );
+
+    // All returned results should be even (matching the filter).
+    for &id in &result_ids {
+        if id == u32::MAX {
+            continue; // start point may appear
```
hildebrandmw: Do we still need this check?

hildebrandmw (commented):
Thanks Jordan! This is moving the tests in the right direction, but we need to be careful about just moving the test infrastructure from `diskann_async.rs` as-is.
The utilities in `search_utils.rs` are extremely awkward for actually running tests, are very sensitive, and don't provide much useful information when they fire.
They've been used in `diskann_async.rs` because it was kind of the best thing we had back then.

My hope is that the new `diskann` can take a higher signal approach using baselines and `VerboseEq`.
Not only does this provide a really good way of viewing the expected results as a whole, it's also great for storing additional metrics.
For example, the stats, ids, and distances from multi-hop search can all be checked in as part of the baseline and get protected for free.

My ask is to not migrate the `search_utils.rs` as is - especially if it means including test methods that aren't actually used by `diskann`.
Also, use the baseline capturing mechanism to capture everything about both test setups and results.
We cannot rely solely on the baseline to protect against regression (someone could check-in a broken baseline in the future), but a baseline in combination with some invariant checks (returned items should be filtered/adjusted) will go a long way toward good tests.

JordanMaples:
converting back to draft as it needs some more human refinement

diskann-providers/src/test_utils/search_utils.rs
JordanMaples: Following my discussion with Mark and having copilot re-evaluate the priorities for this work item, I removed the duplicate implementation and left groundtruth alone.

JordanMaples:
@copilot would you mind taking another review over this pr as it has changed significantly from your previous review.

harsha-simhadri (approved):

hildebrandmw (approved):
Thanks Jordan - sorry this fell through the cracks!

PR #929, opened 2026-04-08
two queue filtered search with max effort

hailangx:
Summary

  - Introduces a new two-queue search algorithm for graph-based vector search that decouples exploration from
  filter evaluation, improving recall for low-selectivity filtered queries
  - Adds a callback-based filter (GarnetFilterProvider) to diskann-garnet alongside the existing bitmap filter, enabling
  per-candidate FFI filter evaluation from Garnet/C#

  Motivation

  The existing beta-filtered search works well when filters are moderately selective, but struggles with low-selectivity
  filters where most candidates are rejected. In those cases, the search converges prematurely because pruning is based
  on distance to filtered results that haven't been found yet. The two-queue approach keeps exploration broad until
  enough filtered results are accumulated.

  Design

  Two-Queue Search (two_queue_search.rs)
  - Maintains two separate queues: a min-heap (candidates) for exploration ordered by distance, and a max-heap
  (filtered_results) for filter-passing neighbors
  - All graph neighbors are explored regardless of filter status; only filter-passing nodes are added to filtered_results
  - Convergence: terminates when filtered_results has enough results and the closest unexplored candidate is farther than
   the worst filtered result
  - Supports a max_candidates hop limit as a safety cap and result_size_factor (default 10) to control result queue
  capacity (k * result_size_factor)
  - Reports termination reason via TwoQueueTermination enum (Exhausted, MaxCandidates, Converged, FilterTerminated)


diskann-benchmark/src/backend/index/benchmarks.rs
```diff
@@ -392,6 +392,50 @@ where
             result.append(AggregatedSearchResults::Topk(search_results));
             Ok(result)
         }
+        SearchPhase::TopkTwoQueueFilter(search_phase) => {
```
hildebrandmw, line 395: I'm a little worried about what adding this universally will do for compile times. Ideally, we'd have a more focused way to add extensions like this that don't encroach on our hard-won compile-time reduction efforts. It would require some backend shuffling, but I think there is a world where search phases behave more like plugins rather than enums, so we can target specific monomorphizations instead of forcing this on all instances. 

What've you observed in terms of compile time differences here?

diskann/src/graph/config/defaults.rs
```diff
@@ -39,3 +39,7 @@ pub const INTRA_BATCH_CANDIDATES: IntraBatchCandidates = IntraBatchCandidates::A
 /// Default beta factor for filtered search.
 /// Values < 1.0 bias the search toward vectors matching the filter.
 pub const FILTER_BETA: f32 = 0.5;
+
+/// Default result queue capacity factor for two-queue filtered search.
+/// The result queue capacity is k * this factor.
+pub const RESULT_SIZE_FACTOR: usize = 10;
```
hildebrandmw, line 45: Shouldn't this live in `two_queue`? That's what it's meant to control, right? Things that live here should be focused on graph building confiugration. To that end, I'm not sure what `FILTER_BETA` is doing here either, but at least for this PR, can this constant be moved nearer to what it's meant to influence?

diskann/src/graph/search/scratch.rs
```diff
@@ -67,6 +68,16 @@ where
     /// Only used during range search
     pub in_range: Vec<Neighbor<I>>,
 
+    /// Filtered results for two-queue search.
+    /// Max-heap of filter-passing neighbors (worst/largest distance on top for pruning).
+    /// Only used during two-queue filtered search.
+    pub filtered_results: BinaryHeap<Neighbor<I>>,
```
hildebrandmw, line 74: This adds two `BinaryHeap`s for **all** users of `SearchScratch`, just for use in two-queue algorithm. Can we instead create these two queues manually in the place where they are needed instead of putting it in a central location? Really, `SearchScratch` needs to get pared down.

diskann/src/neighbor/queue.rs
```diff
@@ -313,11 +313,36 @@ impl<I: NeighborPriorityQueueIdType> NeighborPriorityQueue<I> {
         Some(self.get_unchecked(current))
     }
 
+    /// Like [`closest_notvisited`](Self::closest_notvisited), but ignores the `search_param_l`
+    /// bound and considers all entries in the queue. Use this for resizable/unbounded queues
+    /// (e.g. two-queue filtered search) where exploration should not be capped at L.
+    pub fn closest_notvisited_unbounded(&mut self) -> Option<Neighbor<I>> {
```
hildebrandmw, line 319: Can these be removed? It doesn't look like they are used?

diskann/src/graph/search/two_queue_search.rs
```diff
@@ -0,0 +1,320 @@
...
+/// queue for filter-passing nodes. All neighbors are explored regardless of
+/// filter status, but only matching nodes contribute to convergence.
+#[derive(Debug)]
+pub struct TwoQueueSearch<'q, InternalId> {
+    /// Base graph search parameters (k, ef/l_value, beam_width).
+    pub inner: Knn,
+    /// Filter evaluator for determining node matches.
+    pub filter: &'q dyn QueryLabelProvider<InternalId>,
+    /// Maximum number of hops before stopping search.
+    pub max_candidates: usize,
```
hildebrandmw, line 51: Are there invariants that exist among these fields? If so - they shouldn't be public and this struct should have a dedicated constructor.

diskann/src/graph/search/two_queue_search.rs
```diff
@@ -0,0 +1,320 @@
...
+
+/// Parameters for two-queue filtered search.
+///
+/// This search extends standard graph search by maintaining a separate results
+/// queue for filter-passing nodes. All neighbors are explored regardless of
+/// filter status, but only matching nodes contribute to convergence.
+#[derive(Debug)]
+pub struct TwoQueueSearch<'q, InternalId> {
+    /// Base graph search parameters (k, ef/l_value, beam_width).
+    pub inner: Knn,
```
hildebrandmw, line 47: It doesn't look like `Knn::search_l()` is used at all in the two-queue method (outside of a size hint). If that's the case, then this shouldn't wrap `Knn` and instead capture just the parameters that actually affect the algorithm.

diskann/src/graph/search/two_queue_search.rs
```diff
@@ -0,0 +1,320 @@
...
+                .escalate("start point retrieval must succeed")?;
+            let dist = computer.evaluate_similarity(element.reborrow());
+            let neighbor = Neighbor::new(id, dist);
+            scratch.candidates.push(Reverse(neighbor));
+            scratch.cmps += 1;
+
+            // Check filter on start point
+            if let QueryVisitDecision::Accept(n) = filter.on_visit(neighbor) {
+                scratch.filtered_results.push(n);
+            }
```
hildebrandmw, line 221: What about `QueryVisitDecision::Terminate`?

hildebrandmw (commented):
Thanks Haiyang. There was some mention of fusing predicate checks with element retrieval. Is that something that's still planned?

diskann/src/graph/config/defaults.rs
hailangx: :) I think we wanted the defaults to be in a single place. The **FILTER_BETA ** was moved here pervious comments request last time. 

I would prefer we have a central place for defaults, if you think here should be all graph build related defaults, then we might just create another defaults for all search related.

PR #944, opened 2026-04-13
change file save buffer size and use external buffer to avoid neighbo…

SkyInTheSea:
…r buffer allocation for each vertex when saving

### DIFF
--- a/diskann-providers/src/model/graph/provider/async_/simple_neighbor_provider.rs
+++ b/diskann-providers/src/model/graph/provider/async_/simple_neighbor_provider.rs
@@ -288,6 +288,27 @@ impl storage::bin::GetAdjacencyList for SimpleNeighborProviderAsync<u32> {
         Ok(list)
     }
 
+    /// Optimized version that reuses a pre-allocated buffer to avoid per-call allocation.
+    ///
+    /// This directly copies the adjacency list into the provided buffer, avoiding the
+    /// intermediate `AdjacencyList` allocation that `get_adjacency_list` requires.
+    fn get_adjacency_list_into(&self, i: usize, buffer: &mut Vec<u32>) -> ANNResult<()> {
+        #[cfg(test)]
+        self.num_get_calls.increment();
+
+        // Lint: We don't have a good way of recovering from lock poisoning anyways.
+        #[allow(clippy::unwrap_used)]
+        let _guard = self.locks[i].read().unwrap();
+
+        // SAFETY: We are holding the read lock for `i`.
+        let list = unsafe { self.get_slice(i) };
+
+        // Reuse buffer: clear and copy data directly
+        buffer.clear();
+        buffer.extend_from_slice(list);
+        Ok(())
+    }
+
     fn total(&self) -> usize {
         self.locks.len()
     }
@@ -344,6 +365,32 @@ impl storage::bin::GetAdjacencyList for DiskAdaptor<'_> {
         }
 
         Ok(list)
+    }
+
+    /// Optimized version that reuses a pre-allocated buffer to avoid per-call allocation.
+    ///
+    /// This directly reads neighbors into the buffer and performs the start point remapping
+    /// in-place, avoiding the intermediate `AdjacencyList` allocation.
+    fn get_adjacency_list_into(&self, i: usize, buffer: &mut Vec<u32>) -> ANNResult<()> {
+        // Lint: We don't have a good way of recovering from lock poisoning anyways.
+        #[allow(clippy::unwrap_used)]
+        let _guard = self.provider.locks[i].read().unwrap();
+
+        // SAFETY: We are holding the read lock for `i`.
+        let list = unsafe { self.provider.get_slice(i) };
+
+        // Reuse buffer: clear and copy data directly
+        buffer.clear();
+        buffer.extend_from_slice(list);
+
+        // Remap the in-memory start point to the actual start point
+        for id in buffer.iter_mut() {
+            if *id == self.inmem_start_point {
+                *id = self.actual_start_point;
+            }
+        }
+
+        Ok(())
     }
 
     fn total(&self) -> usize {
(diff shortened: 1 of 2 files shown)


diskann-providers/src/storage/bin.rs
```diff
@@ -100,6 +106,23 @@ pub(crate) trait GetAdjacencyList {
...
     fn get_adjacency_list(&self, i: usize) -> ANNResult<Self::Item<'_>>;
 
+    /// Retrieve the data stored at index `i` into a pre-allocated buffer.
+    ///
+    /// This method allows callers to reuse a buffer across multiple calls,
+    /// avoiding per-call memory allocation overhead. The buffer is cleared
+    /// before being populated with the adjacency list data.
+    ///
+    /// Default implementation falls back to `get_adjacency_list` and copies.
+    fn get_adjacency_list_into(&self, i: usize, buffer: &mut Vec<Self::Element>) -> ANNResult<()>
```
hildebrandmw, line 110: You can make this even more efficient by not adding `get_adjacency_list_into` and instead changing the `Item` to something like this:
```rust
struct Guarded<'a, T> {
    guard: RwLockReadGuard<'a, ()>,
    slice: &'a [T],
}

impl<T> Deref for Guarded<'_, T> {
    type Target = [T];
    fn deref(&self) -> &[T] {
        self.slice
    }
}
```
This bypasses the copy into the temporary buffer and lets you copy directly from the underlying provider into the destination.

This doesn't work for the `DiskAdaptor`, but presumably you aren't using that? Another change would be to change `get_adjacency_list` to accept by `&mut self` and that way `DiskAdaptor` can allocate a local buffer itself.

diskann-providers/src/storage/bin.rs
```diff
@@ -183,7 +206,9 @@ where
     let total = data.total();
     let dim = data.dim();
 
-    let mut writer = provider.create_for_write(path)?;
+    let file = provider.create_for_write(path)?;
+    // Use 16 MB buffer for efficient I/O
+    let mut writer = BufWriter::with_capacity(SAVE_BUFFER_SIZE, file);
```
hildebrandmw: This doesn't need to be fixed at this level. The caller should provide a `StorageWriteProvider` and returns a buffered reader. Otherwise, this runs the risk of situations where if a `StorageWriteProvider` already returns `BufWriter`, we have a `BufWriter` in a `BufWriter`.

hildebrandmw (commented):
A couple quick comments on how we can do slightly better without introducing a new method. Longer term, though, this API was always kind of intended to be a stop gap until a more disciplined approach to the inmem providers and the serialization is worked out.

diskann-providers/src/storage/bin.rs
SkyInTheSea: I don't catch your point here. The optimization is to avoid memory allocation per node to flatten its neighbor list to index file, with a global fixed size buffer we can save allocation time, if we don't want to use a new interface to pass the reference of the global buffer address, we may allocate it beforehand and make it referenceable for get_adjacency_list

metajack:
Please do not submit PRs with no description and a blank template. Write up an appropriate description of what this PR does and why.

diskann-providers/src/storage/bin.rs
hildebrandmw: Right. With your API - you are first copying the data into the externally allocated buffer, then copying it out of that external buffer into the final stream. With the change I showed use using `Item<'a> = Guard<'a, u32>`, you can avoid every copying into the allocated buffer and instead use the `Deref` implementation of `Guard` to copy straight from the underlying store. This does not allocate and removes the extra copy step.

SkyInTheSea: I think I know your meaning now, you want to expose the vertex's neighbor array to final bufwriter to copy, It will save one memcpy operation, but the copy to bufwriter in protection of read lock, which bufwriter is blackbox we don't know when the flush will happen(like if possible part of the copy fill the buffer and trigger flush) and write to disk which is a blocking process and may cause readlock hold for long time compared with a single memcpy, and block other operations like update of the node.

memcopy to a small external buffer is very cheap so generally I prefer to copy it which we can ensure hold readlock for short time.

hildebrandmw: I sympathize with the desire to minimize the duration under which a lock is held, but this leads me to a bigger concern: **why** do we need to minimize the duration the lock is held in this API? The only reason I can think of is because doing so would prevent other concurrent writes to the graph, but if writes are happening concurrently to a save using this API, isn't the save going to be incorrect/not a faithful snapshot? If writes are not happening concurrently, then it doesn't matter if we hold onto the lock?

SkyInTheSea: yes, we need consider the write traffic has possibly to update the vertex's neighbors and be blocked by the stream write which fetch read lock.
it's an real issue for your concern, like if your save take 5 minutes(in our prod case for large diskANN index), the copy of the neighbors dump to disk is actually not a completed graph strictly speaking, but an ongoing update one, hard to evaluate the impact.
That means single ANN object can not handle this save during write case but need to resolved by up layer.
Like our ANN service manages multiple indices to avoid such snapshot issue, which host more than one index and the search result is aggregation of them, one read/write index and all others are readonly, write will be applied on read/write index.
when snapshot or save command reached, current read/write index will be sealed as readonly atomically and a new index is created to serve write, and all delete will be applied to not only the new rw index but also a delete list to play on ro index, there's background thread kept merge ro indices and the deletion list.

diskann-providers/src/model/graph/provider/async_/simple_neighbor_provider.rs
```diff
@@ -346,6 +367,32 @@ impl storage::bin::GetAdjacencyList for DiskAdaptor<'_> {
...
+    ///
+    /// This directly reads neighbors into the buffer and performs the start point remapping
+    /// in-place, avoiding the intermediate `AdjacencyList` allocation.
+    fn get_adjacency_list_into(&self, i: usize, buffer: &mut Vec<u32>) -> ANNResult<()> {
+        // Lint: We don't have a good way of recovering from lock poisoning anyways.
+        #[allow(clippy::unwrap_used)]
+        let _guard = self.provider.locks[i].read().unwrap();
+
+        // SAFETY: We are holding the read lock for `i`.
+        let list = unsafe { self.provider.get_slice(i) };
```
harsha-simhadri, line 379: could this be written with fewer clippy overrides on unwrap and unsafe elements?

PR #947, opened 2026-04-14
Remove OPQ and friends

arkrishn94:
This PR removes support for OPQ and all related code in the repo. The existing implementation is a legacy one, and the current complexity doesn't justify the benefit from this feature (if there is any at all). From what we can tell, no consumer is using this path. 

This should resolve Issue #922 

# Changes 

Deleted functions: 
- `pq_construction.rs`: `generate_optimized_pq_pivots`, `opq_quantize_all_chunks`, `copy_chunk_centroids_to_full_table` from 
- `opq_rotation_matrix` field from `FixedChunkPQTable` and associated calls to it.
- `write_rotation_matrix_data`, `read_opq_rotation_matrix`, `get_rotation_matrix_path` from `PQStorage`
- `DistanceComputerConstructionError` was removed. `ANNError::OPQError` was kept just to not mess with variant numbering of errors. 

Some noteworthy API changes (please review):
- `DistanceComputer::new` and `MultiDistanceComputer::new`: now return `Self` directly instead of `Result<Self, DistanceComputerConstructionError>` (construction is infallible). As a result, the quant providers now return `DistanceComputer` direclty instead of a result. This affects `bf_tree`, `multi_pq` and the in-mem PQ provider.

- Apart from that used Copilot to fix docs that mention OPQ and also remove some OPQ specific tests in pq_construction and fixed_chunk_pq_table. 
- Also I whittled down the `rerankingtest_with_membuf_pq_functions` tests in `pq_construction.rs` only testing necessary cases. This should reduce time spent in CI doing this test from 85s to 18s. 

## `diskann_linalg` dependency 

Deleted the benchmark for dead code paths - chunking_closest_centers_benchmarks.rs in diskann-providers/bench. As a result, was able to remove `diskann_linalg` as a dependency for `diskann-providers`. This has reduced compile time for this crate from 200s to 51s.


diskann/src/error/ann_error.rs
```diff
@@ -883,10 +875,6 @@ pub enum DiskANNError {
     // Error happened when we construct PQ pivot or PQ compressed table
     PQError,
 
-    // OPQ construction error
-    // Error happened when we build the optimized PQ index
-    OPQError,
```
hildebrandmw, line 888: Please keep the `OPQError` variant of the enum. Removing this changes the integer association of the variants, which I'm not completely sure won't break some downstream error reporting. I suspect it won't, but let's be a little safe here.

hildebrandmw (commented):
Awesome! Does this let us remove `diskann-linalg` as a dependency of `diskann-providers`? :smile:

arkrishn94:
Thanks @hildebrandmw, the only dependency of `diskann-linalg` in `diskann-providers` are the two benchmark files - chunking_closest_centers_benchmark.rs in benchmarks and benchmarks_iai. Do you know if you need this?

hildebrandmw:
I certainly don't need these!

hildebrandmw (approved):
Woot!

JordanMaples (approved):

diskann/src/error/ann_error.rs
metajack: You can mark it `#[deprecated]` I think to flag that it is not supposed to be used. Later when we are more confident about testing for the error behaviors, we can remove it. I suspect you can just amrk the next variant with the correct integer and that should keep them fixed. Probably we should start doing that whenever we start making the integer representation important for an enum.

suri-kumkaran (approved):

diskann-providers/src/model/pq/pq_construction.rs
```diff
@@ -1362,7 +1355,7 @@ mod pq_test {
             NUM_PQ_CENTROIDS,
             num_pq_chunks,
             crate::model::pq::pq_construction::NUM_KMEANS_REPS_PQ,
-            make_zero_mean,
```
hildebrandmw, line 1365: Can we leave at least one test for this centroid thing? There's some internal code that needs this to still work because existing compressions use it :cry:

hildebrandmw (commented):
Love the simplification of the long-running test. But can we still at least exercise the centering path? As dubious as its utility is, we have some existing compressed data that still wants this to work. Unless you think it's sufficiently tested elsewhere.

diskann-providers/src/model/pq/pq_construction.rs
arkrishn94: The API as well as the functional correctness are tested in other tests here. There is also a test against SIFT that tests the centering path against a fixed set of pivots. 

I think this path in the rerank test only skips the recall check for the sift dataset. But the deterministic PQ pivot generation in the other tests subsumes this.

hildebrandmw: Cool, thanks for double-checking!

hildebrandmw (approved):

Issue #899, opened 2026-04-03
Maintenance of `diskann-providers`


hildebrandmw:
Exit criteria:  eliminate diskann-providers crate.

The `diskann-providers` crate has become a dumping ground for architectural cruft. Containing it in one place is useful, but it blurs what actually belongs here. Here's my opinionated take on how to tackle this.

I think four things need to happen before we can cleanly extract the inmem providers into their own crate (and similarly split out bf-tree), ordered from easy/mechanical to architecturally involved:

- [ ] Cruft Removal: Consolidate and clean up.
- [ ] Migrate unit/integration tests in `diskann_async.rs` to `diskann`.
- [ ] Move reasonable portions of the PQ code to `diskann-quantization`.
- [ ] Finalize our comprehensive saving and loading story.
- [ ] Make a clean `diskann-inmem` crate and others.

Importantly, I don't think much outside the following justifies creating an auxiliary crate — at least until several rounds of cleansing and extraction have been performed. (I reserve the right to be wrong.)

- [ ] Saving/Loading overhaul. This is a great candidate for a new crate.

### Cruft Removal
There's a bunch of utilities that have been either superseded or are straight up dead code (though not caught due to public exports). These should be audited and either deleted, or upgraded to their new counterparts. A few quick examples:
- [x] `utils/kmeans.rs`: KMeans has been replaced for PQ related purposes by the faster implementation in `diskann-quantization`. The last user of this flow is the disk-index partitioning. It's possible that the BLAS based implementation of the `diskann-providers` is faster than the approach taken in `diskann-quantization` for full-dimensional data. To that end, we should either:
  - Audit and replace the use of KMeans in disk partitioning to use `diskann-quantization` and get rid of the `diskann-providers` implementation.
  - Upgrade the implementation in `diskann-quantization` to be on-par for full-dimensional data.
  - Move the implementation to `diskann-disk` where it's actually used.
- [x] `utils/math_util.rs`: See the above `kmeans` use. Also, vector generation with norm can be replaced by the implementations in `diskann-utils`.
- [ ] Rethink file-centric flow? Many of the implementations in `diskann-providers/utils` derive their staying power because they work in-place on raw files. Is this still needed and if so, is there a cleaner way of decoupling the implementation from files?
- [x] `model/graph/traits/graph_data_types.rs`: This type *should* no longer be needed by non-`diskann-disk` code (and if it is, changing it to a raw `T: VectorRepr` should be straight-forward.
[...]

PR #950, opened 2026-04-15
Remove GraphDataType from diskann-providers

### LINKED ISSUE #946: Remove `GraphDataType` from `diskann-providers` and relocate to `diskann-disk`
`model/graph/traits/graph_data_types.rs`: This type *should* no longer be needed by non-`diskann-disk` code (and if it is, changing it to a raw `T: VectorRepr` should be straight-forward).

We should remove this type from `diskann-providers` and update remaining uses within `diskann-providers` to just use a vector type, ID type, and associated data type independently as needed.

The problem with using this large list of associated types to instantiate generic functions that only use a portion is that it can lead to redundant monomorphizations, and makes uses less clear on whether all associated data types are used or not.

wuw92:
## Summary

Addresses the `graph_data_types.rs` cleanup item from #899.

- **Decouple diskann-providers internal code**: Replace `GraphDataType` bounds with direct type params (`T: VectorRepr`, `A`) in `VectorDataIterator`, `MemoryVectorProviderAsync`, `FastMemoryVectorProviderAsync`. Eliminates the `AdHoc<T>` wrapper from all internal usage and inmem provider type aliases.
- **Move trait to diskann-disk**: `GraphDataType` + `AdHoc` now live in `diskann-disk::data_model`. Delete trait definition, test types, and empty `traits/` module from diskann-providers.
- **Replace boilerplate**: Concrete `GraphDataType` impls in diskann-tools replaced with `type Alias = AdHoc<T>`. Benchmark's `GraphData<T>` replaced with `AdHoc<T>` directly.
- **Decouple relative_contrast.rs**: Use `T: VectorRepr` instead of `Data: GraphDataType` since only `VectorDataType` is accessed.

After this PR, `GraphDataType` exists only in `diskann-disk` (where it belongs as a disk-index concept) and `diskann-tools` (which calls disk-index APIs).


wuw92:

hildebrandmw (approved):
Thanks!

suri-kumkaran (approved):

PR #953, opened 2026-04-16
More tests to increase code coverage for diskann

JordanMaples:
This is in service of #927. 

I was discussing the above issue with @hildebrandmw and he pointed out that one of the problems we have with code coverage is that there is a delta in reported coverage between runs where only diskann is specified and runs where diskann and diskann-providers is specified.

I ran the code coverage tool locally to identify some of these deltas and added a couple tests to gain coverage for index.rs.


metajack (approved):

diskann/src/graph/test/cases/consolidate.rs
```diff
@@ -122,3 +129,119 @@ fn flaky_consolidate_returns_failed_retrieval() {
...
+    let adjacency_lists = generate_2d_square_adjacency_list();
+
+    let index = setup_2d_square(create_2d_unit_square(), adjacency_lists, 4);
+    let ctx = test_provider::Context::new();
+    let strategy = Strategy::new();
+
+    // Delete vertex 3, then try to consolidate it.
+    rt.block_on(index.data_provider.delete(&ctx, &3)).unwrap();
+
+    let result = rt
```
arrayka, line 147: I've created a follow-up task to agree on blocking tests vs async tests: #969

arrayka (approved):

diskann/src/graph/test/cases/consolidate.rs
JordanMaples: Resolving so we can merge.

PR #960, opened 2026-04-20
Replace `AlignedBoxWithSlice` with `Vec` in PQScratch and disk fp vector caches

### LINKED ISSUE #954: Replace `AlignedBoxWithSlice` with plain `Vec`/`Matrix` where alignment is not needed
Tracking issue for removing `AlignedBoxWithSlice` usage **where the alignment is not actually needed**, identified during the file-centric flow audit (#899).

## Discussion

Originating proposal and maintainer confirmation in #899:
- Proposal: https://github.com/microsoft/DiskANN/issues/899#issuecomment-4258561508
- Maintainer confirmation (SIMD handles unaligned / arbitrary-length data): https://github.com/microsoft/DiskANN/issues/899#issuecomment-4264194322

## Scope boundary vs. #907

#907 tracks replacing `AlignedBoxWithSlice` with `diskann-quantization::alloc::Poly` for cases that **do** require alignment (e.g. direct I/O sector-aligned buffers).

This issue tracks the complementary cleanup: callers that use `AlignedBoxWithSlice` but whose consumers (distance computation through `diskann-vector` SIMD, or scalar PQ table lookups) don't actually require alignment. These can be replaced with plain `Vec` / `Matrix`.

## Candidates for further investigation (not yet confirmed)

Each of these uses `AlignedBoxWithSlice` but needs a closer look to confirm whether the alignment is actually load-bearing. If not, they would fit this issue's scope.

- `PQScratch` scratch buffers (`rotated_query`, `aligned_query_float`, `aligned_dist_scratch`, etc.) — consumed by `diskann-vector` SIMD or scalar PQ lookup. Needs to confirm whether any consumer relies on alignment.
- In-memory vector caches (`disk_vertex_provider.rs::aligned_vector_buf`, `data_model/cache.rs::vectors`) — fed into distance computation. Same question as above.

wuw92:
## Summary

Replaces `AlignedBoxWithSlice` with plain `Vec` for three disk-side buffers whose consumers don't require alignment.

### 1. `PQScratch.rotated_query` (`diskann-disk/src/search/pq/pq_scratch.rs`)

- Replace `AlignedBoxWithSlice<f32>` with `Vec<f32>`. Consumers (`populate_chunk_distances` via `diskann-vector` SIMD, `TransposedTable::process_into` via `diskann-quantization` scalar PQ table lookups) handle unaligned / arbitrary-length data, so the alignment is not load-bearing.
- `PQScratch::set` now tolerates `query.len() >= dim`, following the inmem `TableL2::populate` convention, instead of requiring strict equality.

### 2. `DiskVertexProvider::aligned_vector_buf` and `Cache::vectors` (`FP_VECTOR_MEM_ALIGN`)

- Replace `AlignedBoxWithSlice<T>` (32-byte `FP_VECTOR_MEM_ALIGN`) with plain `Vec<T>`. The fp distance kernels reached via `get_vector` go through `diskann-vector` SIMD, which handles unaligned data, so the alignment is not load-bearing.
- Neither buffer has the concurrent insert+search pattern that justifies the `FastMemoryVectorProviderAsync`-style 64-byte cache-line alignment: `aligned_vector_buf` is per-scratch-exclusive; `Cache::vectors` is built once via BFS then shared read-only through `Arc<Cache>`.
- Drop the companion `memory_aligned_dimension = dims.next_multiple_of(8)` padding — it was paired with the 32-byte alignment to feed 8-wide AVX tiles and is no longer justified. `DiskVertexProvider` and `Cache` now size slots to raw `dim`; `get_vector` returns a slice of length `dim` instead of a padded length.


JordanMaples (approved):

metajack (approved):

hildebrandmw:
> * `PQScratch::set` now tolerates `query.len() >= dim`, following the inmem `TableL2::populate` convention, instead of requiring strict equality.

@wuw92 - Note that this tolerance of `query.len() >= dim` is a byproduct of the `AlignedBoxcWithSlice` and the silent length extension to a multiple of 8. It's likely with this change that we can strengthen `TableL2::populate` and friends to require the exact length, which would be great!

wuw92:
> @wuw92 - Note that this tolerance of `query.len() >= dim` is a byproduct of the `AlignedBoxcWithSlice` and the silent length extension to a multiple of 8. It's likely with this change that we can strengthen `TableL2::populate` and friends to require the exact length, which would be great!

Hi @hildebrandmw Thanks for flagging — followed up in #1044.

PR #962, opened 2026-04-20
Adding unit tests for paged_search 

JordanMaples:
All code coverage and testing for paged searching are performed in diskann-providers. This adds a set of tests with baselines to diskann. This is also in service of the #927 work item.


diskann/src/graph/test/cases/paged_search.rs
```diff
@@ -0,0 +1,311 @@
...
+                "duplicate id {} found on page {}",
+                n.id,
+                page_idx
+            );
+        }
+    }
+}
+
+/// assert distances are non-decreasing across full result sequence
+fn assert_non_decreasing_distances(pages: &[Vec<Neighbor<u32>>]) {
```
metajack, line 93: This is not something that we guarantee. An individual page will have monotonicly increasing distances, later pages may have closer vectors.

This test should instead check that each page is monotonic in distance, but enforce no constraints across pages.

metajack (changes requested):

diskann/src/graph/test/cases/paged_search.rs
JordanMaples: kk I'll switch that really quick

metajack (approved):

harsha-simhadri (approved):

PR #963, opened 2026-04-21
Remove synthetic/structured data generation from diskann-providers

### LINKED ISSUE #903: move/Remove Synthetic/structured data from diskann-providers
Synthetic/structured data: Audit the users and move synthetic data generation to more appropriate locations (`diskann-tools` already has its own synthetic data creation?).
    Structured data like the grid points already have a replacement in `diskann`.
      We can remove the implementations in `diskann-providers`.

JordanMaples:
This is a follow up PR for #904, which has an outdated baseline and can be closed with this superseding it. 
Essentially the same work was done here just without the provider/async_/caching changes. This closes #903.

When #953 merges, the from_dim definition will need to be updated to support the '2' case. 

Replace generate_structured_data.rs and generate_synthetic_labels_utils.rs with diskann::graph::test::synthetic::Grid and diskann-tools' own label generator respectively.

- Rewrite GenerateGrid trait impls to delegate to Grid::data/data_as
- Replace all adj list generation calls with Grid::neighbors
- Rewire generate_synthetic_labels binary to diskann-tools version
- Delete both redundant utility files and clean up exports


diskann-providers/src/index/diskann_async.rs
```diff
@@ -295,34 +307,19 @@ pub(crate) mod tests {
...
     impl GenerateGrid for u8 {
         fn generate_grid(dim: usize, size: usize) -> Vec<Vec<Self>> {
-            match dim {
-                1 => utils::generate_1d_grid_vectors_u8(size.try_into().unwrap()),
-                3 => utils::generate_3d_grid_vectors_u8(size.try_into().unwrap()),
-                4 => utils::generate_4d_grid_vectors_u8(size.try_into().unwrap()),
-                _ => panic!("{}-dimensions is not support for grid-generation", size),
-            }
+            grid_to_vecs(&grid_from_dim(dim).data_as(size, |v| v as u8))
         }
```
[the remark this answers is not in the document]
JordanMaples: good call. fixing.

harsha-simhadri (approved):

diskann-providers/src/index/diskann_async.rs
```diff
@@ -278,6 +279,17 @@ pub(crate) mod tests {
...
     }
 
+    pub(crate) fn grid_from_dim(dim: usize) -> Grid {
+        Grid::from_dim(dim)
+            .unwrap_or_else(|| panic!("{dim}-dimensions is not supported for grid-generation"))
```
metajack, line 284: This is a slightly strange construction, which I guess you used just to get string substitution? Normally I'd expect: 

`Grid::from_dim(dim).expect("dimension not supported for grid generation")` or use `format!`.

metajack (approved):

diskann-providers/src/index/diskann_async.rs
JordanMaples: yeah, it was just for the sake of getting the string substitution to maintain the messages as they were before. If we don't care, I can switch it over to expect. I don't feel strongly either way on it.

JordanMaples: oh right, i remember the real reason. Clippy suggested to do it this way.

hildebrandmw: One way you can make this more idiomatic is to use an error type for `Grid::from_dim`:
```rust
#[derive(Debug, Clone, Copy, Error)]
#[error("{}-dimensions not supported", self.0)]
struct DimNotSupported(usize);

impl Grid {
    pub fn from_dim(dim: usize) -> Result<Self, DimNotSupported>;
}
```
Then all users can benefit.

PR #965, opened 2026-04-21
Remove AlignedBoxWithSlice wrapper and add alias to Poly<[T], AlignedAllocator>

### LINKED ISSUE #907: Get rid of `AlignedBoxWithSlice`.
Get rid of AlignedBoxWithSlice. This has literally been replaced with diskann-quantization::alloc::Poly as its implementation. Its one non-trivial method split_into_nonoverlapping_mut_slices can simply be replaced with rayon's ParChunksMut.

JordanMaples:
Closes #907 

This change introduces `AlignedSlice<T>` type alias and the `aligned_slice()` constructor in `diskann-quantization::alloc`, replacing the `AlignedBoxWithSlice` wrapper struct from `diskann-providers`. This eliminates an unnecessary abstraction layer and moves aligned allocation to its proper tier.

- Add `AlignedSlice<T> = Poly<[T], AlignedAllocator>` type alias
- Add `aligned_slice<T>(capacity, alignment)` constructor
- Migrate all 9 consumer files in `diskann-disk`
- Replace `split_into_nonoverlapping_mut_slices` with `chunks_mut`
- Replace `as_slice`/`as_mut_slice` with Deref/DerefMut
- Delete aligned_allocator.rs from `diskann-providers`
- Add 5 tests for `aligned_slice`


JordanMaples:
This replaces #908

metajack (approved):

wuw92 (approved):

diskann-disk/benches/benchmarks/aligned_file_reader_bench.rs
```diff
@@ -33,12 +33,11 @@ pub fn benchmark_aligned_file_reader(c: &mut Criterion) {
...
     let num_read = MAX_IO_CONCURRENCY * 100; // The LinuxAlignedFileReader batches reads according to MAX_IO_CONCURRENCY.  Make sure we have many batches to handle.
-    let mut aligned_mem = AlignedBoxWithSlice::<u8>::new(read_length * num_read, 512).unwrap();
+    let mut aligned_mem =
+        aligned_slice::<u8>(read_length * num_read, PowerOfTwo::new(512).unwrap()).unwrap();
 
     // create and add AlignedReads to the vector
-    let mut mem_slices = aligned_mem
-        .split_into_nonoverlapping_mut_slices(0..aligned_mem.len(), read_length)
-        .unwrap();
+    let mut mem_slices: Vec<&mut [u8]> = aligned_mem.chunks_mut(read_length).collect();
```
hildebrandmw, line 40: Note: we probably never need to ever materializes the slices as a `Vec<&mut [u8]>`. In the context of the benchmark, a `chunks_mut` iterator would have worked just fine. In other situations, `par_chunks` directly would work as well.

hildebrandmw (commented):
Thanks for the cleanup @JordanMaples! This is a step in the right direction, but I have some concrete requests for a follow-up. While this PR removes `AlignedBoxWithSlice` to the letter, it doesn't remove it in spirit as the `AlignedSlice` alias and `aligned_slice` constructor still exist. We should strive for construction of concrete types to not require a helper free function.

Instead, I would like call sites to explicitly compose the container and allocator instead of using the thin helper. We can make this more ergonomic by introducing `const` aliases for all valid `PowerOfTwo` values:
```rust
impl PowerOfTwo {
    // Powers
    const P0: Self = unsafe { Self::new_unchecked(NonZeroUsize::new_unchecked(1)) };
    const P1: Self = unsafe { Self::new_unchecked(NonZeroUsize::new_unchecked(2)) };
    // ...
    const P63: Self = unsafe { Self::new_unchecked(NonZeroUsize::new_unchecked(1 << 63)) };

    // Values
    const V1: Self = Self::P0;
    const V2: Self = Self::P1;
    // ...
    const V9223372036854775808: Self = Self::P63;
}
```
Then in situations where the alignment is known and static, we can replace
```rust
// Before
let buf = aligned_slice::<u8>(sector_len, PowerOfTwo::new(512)?)?;

// After
let buf = Poly::broadcast(
    0u8,
    sector_len, 
    AlignedAllocator::new(PowerOfTwo::V512)
)?;
```
We can also add `const` aliases for `AlignedAllocator` for common alignments to simplify even further:
```rust
let buf = Poly::broadcast(0u8, sector_len, AlignedAllocator::A512)?;
```
This way, both the initialization of the allocator **and** the data are visible at the callsite.

My concrete ask is this:
1. Add `PowerOfTwo`/`AlignedAllocator` const aliases.
2. Delete `aligned_slice.rs` and update everything to use `Poly` directly.

PR #976, opened 2026-04-24
[PQ Cleanup] Part 2: Consolidate `calculate_chunk_offsets*` 

arkrishn94:
.Couple of small independent cleanups for PQ. 

##  Move `calculate_chunk_offsets[_auto]` to `ChunkOffsets` as a constructor.

These two functions are pure prefix-sum math over `dimensions` and `num_pq_chunks`. `ChunkOffsetsBase` / `ChunkOffsetsView` in `diskann-quantization::views` is the natural home for it. I've moved the logic into two constructors - `from_dim` and `from_dim_into` for `ChunkOffsets` and `ChunkOffsetsView` respectively to support both allocation patterns.  

All in-repo call sites have been updated. There might be some overlapping edits with #1010. 

## Minor changes
- `QueryComputer` and `DistanceComputer` had six trampoline impls forwarding `&Vec<u8>` and `&&[u8]` arguments to the canonical `&[u8]` impls. `ElementRef` in the accessor now allows us to get rid of these! This might be conflicting with #1008  
- Move `get_chunk_from_training_data` in pq_construction.rs from public API into tests where it is used.


diskann/src/utils/mod.rs
```diff
@@ -3,8 +3,6 @@
  * Licensed under the MIT license.
  */
 
-pub mod object_pool;
-
 pub mod async_tools;
```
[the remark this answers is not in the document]
arkrishn94: Checked out from #975

diskann-providers/src/model/pq/mod.rs
```diff
@@ -11,10 +11,9 @@ pub use fixed_chunk_pq_table::{
...
 pub use pq_construction::{
     MAX_PQ_TRAINING_SET_SIZE, NUM_KMEANS_REPS_PQ, NUM_PQ_CENTROIDS, accum_row_inplace,
-    calculate_chunk_offsets, calculate_chunk_offsets_auto, generate_pq_data_from_pivots,
-    generate_pq_data_from_pivots_from_membuf, generate_pq_data_from_pivots_from_membuf_batch,
-    generate_pq_pivots, generate_pq_pivots_from_membuf, get_chunk_from_training_data,
-    move_train_data_by_centroid,
+    generate_pq_data_from_pivots, generate_pq_data_from_pivots_from_membuf,
+    generate_pq_data_from_pivots_from_membuf_batch, generate_pq_pivots,
+    generate_pq_pivots_from_membuf, get_chunk_from_training_data, move_train_data_by_centroid,
 };
```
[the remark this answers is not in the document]
arkrishn94: Re-exporting from diskann-providers is an anti-pattern. We want to eventually get rid of this crate.

diskann-utils/src/views.rs
```diff
@@ -673,6 +673,23 @@ impl<'a, T> From<MutMatrixView<'a, T>> for &'a [T] {
...
+where
+    T: Copy + std::ops::AddAssign,
+{
+    assert_eq!(x.ncols(), y.len());
+    x.row_iter_mut().for_each(|row| {
+        std::iter::zip(row.iter_mut(), y.iter()).for_each(|(a, b)| {
+            *a += *b;
+        });
+    });
+}
```
hildebrandmw: Would it make sense to have a `map_rows_mut`/some generalized method instead of a bespoke `accum_row_inplace`?

arkrishn94: Yeah good point. On it.

diskann-quantization/src/views.rs
```diff
@@ -205,6 +207,63 @@ impl<'a> From<ChunkOffsetsView<'a>> for &'a [usize] {
...
+    /// Build a chunk-offset plan that partitions `dimensions` into `num_pq_chunks`
+    /// near-equal chunks. The first `dimensions % num_pq_chunks` chunks are one element
+    /// larger than the rest.
+    ///
+    /// Returns an error if the requested partition is not valid (e.g. `dimensions == 0`,
+    /// `num_pq_chunks == 0`, or `num_pq_chunks > dimensions`).
+    pub fn from_dimensions(
+        dimensions: usize,
+        num_pq_chunks: usize,
+    ) -> Result<Self, ChunkOffsetError> {
```
hildebrandmw: In the spirit of bike-shedding, what about something like
```rust
pub fn partition(
    dim: NonZeroUsize,
    len: NonZerousize,
) -> Result<Sel, ChunkOffsetError> {
    
}
```
Which would be internally consistent with the `dim()/len()` methods on `chunk_offsets`.

Also - bonus points for erroring before allocating if at all possible. The final result can still go through `Self::new` to make sure all the invariants are upheld, but if we can tell right away that things are not going to work out, might as well skip some unnecessary work.

I'm not completely set on `NonZeroUsize`, but do consider using bespoke errors here. Right now, if you pass `num_pq_chunks == 0`, the error is "offsets must have a length of at least 2, found 1" which is a little confusing. Perhaps even more confusing is when we pass `num_pq_chunks > dim`, in which case the error is "offsets must be strictly increasing ...".

diskann-quantization/src/views.rs
```diff
@@ -205,6 +207,63 @@ impl<'a> From<ChunkOffsetsView<'a>> for &'a [usize] {
...
+    ///
+    /// See [`ChunkOffsets::from_dimensions`] for the partitioning rule.
+    ///
+    /// Returns an error if `scratch.len() != num_pq_chunks + 1` or if the requested
+    /// partition is not valid (e.g. `dimensions == 0`, `num_pq_chunks == 0`, or
+    /// `num_pq_chunks > dimensions`).
+    pub fn from_dimensions_into(
+        dimensions: usize,
+        num_pq_chunks: usize,
+        scratch: &'a mut [usize],
```
hildebrandmw, line 256: We can derive `num_pq_chunks` from `scratch.len()`, in which case this can be
```rust
pub fn partition_in(
    dim: usize,
    offsets: &mut [usize],
)
```
Which has a nice symmetry with the comment on suggesting `partition` for the constructing case. It would also be nice if we could avoid mutating `offsets` if we know there will be an invalid configuration.

diskann-utils/src/views.rs
```diff
@@ -388,6 +397,32 @@ where
         self.data.as_mut_slice().chunks_exact_mut(ncols)
     }
 
+    /// Apply `op(row[j], &broadcast[j])` to every entry of every row, broadcasting
```
hildebrandmw: Here's the rabbit hole I went down:

First, I felt that this was far too specific targeting a very niche use case and would instead be better served with
```rust
pub fn try_map_row_mut<F, R>(&mut self, f: F) -> R
where
    F: FnMut(&mut [T::Elem]) -> Result<(), R>
```
Then I realized that this could be trivially implemented with `.row_iter_mut().try_for_each(f)`

At which point, this made me question whether or not this is even needed.

My suggestion is to remove it if possible since it's a very specific function that can already be done almost trivially through existing means.

hildebrandmw (commented):
Turns out you updated the code midway through the review :smile:

That said, I think my comments still hold.

diskann-quantization/src/views.rs
arkrishn94: Thanks Mark. Made sure to catch the errors upfront and hopefully raise them correctly. Lmk if that looks good. 

I'm less convinced about the naming, `from_dim` and `from_dim_into` seem reasonable? I'm not sure I'm following the naming of `partition` and `partition_in` :p

diskann-utils/src/views.rs
arkrishn94: Inlined this operation at the two call-sites. Removed this from `MatrixBase` and the associated testing.

diskann-quantization/src/views.rs
arkrishn94: Modified the derivation of `num_chunks` to be using `scratch.len()`.Made sure to throw errors before doing any work if the configuration is invalid.

diskann-quantization/src/views.rs
hildebrandmw: My rationale for `partition` and `partition_in` is that we are partitioning `dim` into `len` chunks. I think that's more descriptive of the operation than `from_dim`.

diskann-quantization/src/views.rs
```diff
@@ -205,6 +207,79 @@ impl<'a> From<ChunkOffsetsView<'a>> for &'a [usize] {
...
 }
 
+impl ChunkOffsets {
+    /// Build a chunk-offset plan that partitions `dim` into `num_chunks`
+    /// near-equal chunks. The first `dim.get() % num_chunks.get()` chunks are
+    /// one element larger than the rest.
+    ///
+    /// Returns an error if the requested partition is not valid (e.g.
+    /// `num_chunks.get() > dim.get()`).
+    pub fn from_dim(dim: NonZeroUsize, num_chunks: NonZeroUsize) -> Result<Self, ChunkOffsetError> {
```
hildebrandmw: Sorry to keep harping on this, but I would prefer a custom error type for these interfaces consisting of just `TooManyChunks`/`LengthMustBeAtleastTwo`. If most of the chunking APIs go here, we might as well provide a tight-list of the possible failure modes, rather than forcing callers to deal with `NonMonotonic` errors, which would be **out** bug, not theirs.

hildebrandmw (commented):
Thanks! Just one more quick comment about the error type for the new constructors then looks fine to me. I also still prefer `partition/partition_in` for the function names since it's partitioning `dim` into the requested number of chunks.

diskann-quantization/src/views.rs
arkrishn94: That's a reasonable point, let me fix.

diskann-quantization/src/views.rs
```diff
@@ -205,6 +223,83 @@ impl<'a> From<ChunkOffsetsView<'a>> for &'a [usize] {
...
+    ///
+    /// Returns an error if the requested partition is not valid (e.g.
+    /// `num_chunks.get() > dim.get()`).
+    pub fn partition(dim: NonZeroUsize, num_chunks: NonZeroUsize) -> Result<Self, PartitionError> {
+        if num_chunks.get() > dim.get() {
+            return Err(PartitionError {
+                num_chunks: num_chunks.get(),
+                dim: dim.get(),
+            });
+        }
```
arrayka, line 239: Do I understand correctly that previously we could handle num_chunks > dim (since `calculate_chunk_offsets()` didn't have this check), and now we will start throwing an error?

diskann-providers/src/model/pq/mod.rs
```diff
@@ -11,10 +11,9 @@ pub use fixed_chunk_pq_table::{
 mod pq_construction;
 pub use pq_construction::{
     MAX_PQ_TRAINING_SET_SIZE, NUM_KMEANS_REPS_PQ, NUM_PQ_CENTROIDS, accum_row_inplace,
```
arrayka, line 13: Could please provide locations of these two call sites in the PR description:
- Inline `accum_row_inplace` from `diskann-providers\src\model\pq` at the two call-sites.

arrayka (approved):

diskann-quantization/src/views.rs
hildebrandmw: Technically yes, but it would have errored later when constructing the `ChunkOffsets` struct for-real. So really, this improves the error diagnostic by reporting earlier and with a more concrete error.

hildebrandmw (approved):

PR #981, opened 2026-04-28
[diskann-vector] Support truly unaligned distances.

hildebrandmw:
An internal user has a case where full-precision vectors (e.g. `f32`) are stored in completely unaligned buffers (e.g. align of 1), requiring a data copy to align the data before the slices can be safely constructed. However, our distance function implementations use `SIMDVector::load_unaligned` under the hood, which are compatible with under-aligned pointers.

This PR exposes a proper API to the `DistanceProvider` trait (via the `Distance` type) for invoking the SIMD implementations with unaligned pointers.

## Suggested Reviewing Order

* `diskann-wide`: The implementations of `SIMDVector::load*` and `SIMDVector::store*` already support underaligned pointers. This PR updates the documentation and restructures the load/store tests to verify this property (we were already using this property in some of the quantized distance kernels). The new load/store tests successfully pass Miri.

* `unaligned.rs` - a new `UnalignedSlice` is added for unaligned slices. This is just a pointer + length pair with some validity requirements but no alignment requirement. Conversions from `&[T]` and `&[T; N]` are added and the trait `AsUnaligned` replaces the use of `AsRef<[T]>` and the internal `ToSlice` traits.
 
  A test-only `Buffer` is used to purposely offset simple types to exercise the unaligned cases.

* `distance/simd.rs`: The `simd_op` kernel is tweaked to accept `AsUnaligned` instead of `AsRef`. Checks have been added to the existing tests to ensure that the under-unaligned versions are both Miri compatible and yield the exact same results as their properly aligned counterparts.

* `distance/implementation.rs`: The architecture hooks and specialization are changed to use `AsUnaligned`. I've investigated the code generation and the checks for `impl FTarget<...> for Specialize<N, F>` are sufficient to trigger constant propagation and the full unrolling of small fixed-sized kernels.

* `distance/distance_provider.rs`: The `Distance` type is changed to pass `UnalignedSlice`s across the function pointer boundary rather than raw slices. We can keep the existing API for slices trivially via `AsUnaligned`.

## Code Generation
Unfortunately, the order in which functions are code-generated seems to have changed with this PR. That said, the fixed-sized specializations I have spot-checked result in identical assembly with this PR as with main, which is to be expected.


diskann-vector/src/distance/simd.rs
```diff
@@ -688,12 +686,12 @@ fn emit_length_error(xlen: usize, ylen: usize) -> ! {
...
-    T: AsRef<[L]>,
-    U: AsRef<[R]>,
+    T: AsUnaligned<Element = L>,
+    U: AsUnaligned<Element = R>,
     S: SIMDSchema<L, R, A>,
 {
-    let x: &[L] = x.as_ref();
-    let y: &[R] = y.as_ref();
+    let x = x.as_unaligned();
+    let y = y.as_unaligned();
```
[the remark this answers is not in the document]
hildebrandmw: This is correct. I believe I have updated `diskann-wide` to completely relax the alignment requirements.

diskann-vector/src/distance/implementations.rs
```diff
@@ -73,35 +51,31 @@ macro_rules! architecture_hook {
...
 pub(crate) struct Specialize<const N: usize, F>(std::marker::PhantomData<F>);
 
-impl<A, T, L, R, const N: usize, F> diskann_wide::arch::FTarget2<A, T, &[L], &[R]>
+impl<A, T, L, R, const N: usize, F>
+    diskann_wide::arch::FTarget2<A, T, UnalignedSlice<'_, L>, UnalignedSlice<'_, R>>
     for Specialize<N, F>
 where
     A: Architecture,
-    F: for<'a, 'b> diskann_wide::arch::Target2<A, T, &'a [L; N], &'b [R; N]> + Default,
+    F: for<'a, 'b> diskann_wide::arch::Target2<A, T, UnalignedSlice<'a, L>, UnalignedSlice<'a, R>>
```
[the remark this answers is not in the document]
hildebrandmw: Fixed

JordanMaples (approved):

diskann-vector/src/distance/implementations.rs
```diff
@@ -73,35 +51,31 @@ macro_rules! architecture_hook {
...
 
-impl<A, T, L, R, const N: usize, F> diskann_wide::arch::FTarget2<A, T, &[L], &[R]>
+impl<A, T, L, R, const N: usize, F>
+    diskann_wide::arch::FTarget2<A, T, UnalignedSlice<'_, L>, UnalignedSlice<'_, R>>
     for Specialize<N, F>
 where
     A: Architecture,
-    F: for<'a, 'b> diskann_wide::arch::Target2<A, T, &'a [L; N], &'b [R; N]> + Default,
+    F: for<'a, 'b> diskann_wide::arch::Target2<A, T, UnalignedSlice<'a, L>, UnalignedSlice<'b, R>>
+        + Default,
```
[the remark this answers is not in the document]
hildebrandmw: This is fixed.

diskann-vector/src/unaligned.rs
```diff
@@ -0,0 +1,202 @@
...
+    ///
+    /// # Safety
+    ///
+    /// The provided memory region must be valid for reading via [`std::ptr::read_unaligned`]
+    /// for the duration of the associated lifetime.
+    ///
+    /// Note that `ptr` need not be aligned to `std::mem::align_of::<T>()`.
+    ///
+    /// Argument `ptr` may be null if `len == 0`.
+    pub const unsafe fn new(ptr: *const T, len: usize) -> Self {
```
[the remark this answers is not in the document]
hildebrandmw: After perusing various discussions, I have come to the conclusion that this is in fact well-defined behavior (and Miri agrees): https://github.com/rust-lang/unsafe-code-guidelines/issues/472

diskann-vector/src/unaligned.rs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+/// View `self` as an [`UnalignedSlice`].
+pub trait AsUnaligned {
+    /// The element type of the slice.
+    type Element;
+
+    /// Return an [`UnalignedSlice`] view of `self`.
+    fn as_unaligned(&self) -> UnalignedSlice<'_, Self::Element>;
+}
```
arkrishn94, line 91: nit: I got a bit confused between `Unaligned` in `distance_provider.rs` and the output of this trait, which is the actual `UnalignedSlice`. Rename to `AsUnalignedSlice`?

diskann-vector/src/distance/distance_provider.rs
```diff
@@ -46,6 +45,16 @@ pub trait DistanceProvider<T>: Sized + 'static {
...
 
+#[derive(Debug)]
+struct Unaligned<T>(std::marker::PhantomData<T>);
+
+impl<T> diskann_wide::lifetime::AddLifetime for Unaligned<T>
+where
+    T: 'static,
+{
+    type Of<'a> = UnalignedSlice<'a, T>;
+}
```
arkrishn94, line 56: This is probably my lack of knowledge of the `AddLifetime` trait but for my learning- why is this implementation needed for `Unaligned<T>`?

diskann-vector/src/distance/implementations.rs
```diff
@@ -7,58 +7,36 @@ use diskann_wide::{arch::Target2, Architecture, ARCH};
...
         impl<A, T, L, R> diskann_wide::arch::Target2<A, T, L, R> for $functor
         where
             A: Architecture,
-            L: ToSlice,
-            R: ToSlice,
-            $impl: simd::SIMDSchema<L::Target, R::Target, A>,
-            Self: PostOp<<$impl as simd::SIMDSchema<L::Target, R::Target, A>>::Return, T>,
+            L: AsUnaligned,
+            R: AsUnaligned,
+            $impl: simd::SIMDSchema<L::Element, R::Element, A>,
```
arkrishn94, line 21: I noticed that all implementations of `AsUnaligned` route through `impl From<B> for UnalignedSlice<T, '_>`. Out of curiosity - why not get rid of this trait and bound using `Into<UnalignedSlice<T, '_>`?

I'm using this as an opportunity to learn : )

diskann-vector/Cargo.toml
```diff
@@ -22,7 +22,8 @@ iai-callgrind.workspace = true
 rand.workspace = true
 criterion.workspace = true
 rand_distr.workspace = true
-half = { workspace = true, features = ["rand_distr", "num-traits"] }
+half = { workspace = true, features = ["rand_distr", "num-traits", "bytemuck"] }
+bytemuck = { workspace = true, features = ["must_cast"] }
```
arkrishn94, line 26: Is `bytemuck` only needed for tests? If so, maybe useful to put behind a flag? Since I know you care a lot about compile times :)

arkrishn94 (approved):
Looks good Mark, mostly had small comments only. The only callout is the question on the indirection through `AsUnaligned`. Would love to understand why this is needed.

diskann-vector/src/distance/distance_provider.rs
hildebrandmw: (Note that this is a private type to make `Dispatched2` work).

The main problem this is solving is outlined [here](https://github.com/microsoft/DiskANN/blob/f458cf613539a57d17f814da3fc8ac8388db1cdd/diskann-wide/src/arch/mod.rs#L251-L314). The gist is that `Dispatched2` is a function-pointer wrapper, but function types and function-pointer types in Rust are blessed with an automatic HRTB-like syntax that is unavailable to normal types. For example, this works: `f: fn(&f32) -> f32` and the type `fn(&f32) -> f32` is `'static` where-as if you try to instantiate a generic `Vec<&f32>`, you need both a lifetime **and** the resulting type is not `'static`.

So the `AddLifetime` is a workaround for this.

diskann-vector/src/distance/implementations.rs
hildebrandmw: The issue is one of potential trait ambiguity. The `AsUnaligned` as the `Element` as a GAT, which can then be fed unambiguously to `SIMDSchema`. If we used a more generic trait like `From`, we'd need to specify the element-type somehow, which won't work because we'd end up with uncovered generics.

You can try this yourself if you take the code and start trying to use `From`, it gets bad fast :smile:

diskann-vector/src/unaligned.rs
hildebrandmw: Maybe. There is a difference, the `Unaligned` in `distance_providers` is a private internal type where-as this is a public trait/public type. I'm not necessarily opposed to renaming this `AsUnalignedSlice`, but it's use is specific enough that I felt a shorter `AsUnaligned` as fine. Happy to make the name longer though if you feel strongly about it.

diskann-vector/Cargo.toml
hildebrandmw: Note that this is just added to the `dev-dependencies`, so it is in fact only being pulled in for tests :wink:.

diskann-vector/src/unaligned.rs
arkrishn94: I don't feel strongly. No need to change, just a nit

diskann-vector/Cargo.toml
arkrishn94: Ahhh of course, I missed that in the diff view.

arkrishn94 (approved):

arkrishn94 (approved):

PR #983, opened 2026-04-28
[flat index] Flat Search Interface

### LINKED ISSUE #999: Create RFC and initial implementation for flat-index search support

arkrishn94:
This PR introduces a trait interface and a light index to support brute-force search for providers that can be used as/are a flat-index. There is an associated RFC that walks through the interface and associated implementation in `diskann` as a new `flat` module.  

Rendered RFC [link](https://github.com/microsoft/DiskANN/blob/u/adkrishnan/flat-index/rfcs/00983-flat-search.md).

## Motivation

The repo has no first-class surface for brute-force search. This PR adds a small trait hierarchy that gives flat search the same provider-agnostic shape that graph search has, so any backend (in-memory, quantized, disk, remote) can plug in once and reuse a shared algorithm.

## Traits (`flat/strategy.rs`)

**`DistancesUnordered<C>`** — the single trait a backend must implement. Fuses iteration and scoring into one method: the implementation drives a full scan, scoring each element with a precomputed query computer `C`, and invokes a callback with `(id, distance)` pairs. Key associated types:
- `ElementRef<'a>` -- the reference shape `C` scores against.
- `Id` -- the id type yielded to the callback (decoupled from `HasId` so visitors can yield any id shape).
- `C : for<'a> PreprocessedDistanceFunction<Self::ElementRef<'a>, f32>` -- the precomputer query computer.

```rust
pub trait DistancesUnordered<C>: Send + Sync
where
    C: for<'a> PreprocessedDistanceFunction<Self::ElementRef<'a>, f32>,
{
    type ElementRef<'a>;
    type Id;
    type Error: ToRanked + Debug + Send + Sync + 'static;

    fn distances_unordered<F>(
        &mut self,
        computer: &C,
        f: F,
    ) -> impl SendFuture<Result<(), Self::Error>>
    where
        F: Send + FnMut(Self::Id, f32);
}
```

**`SearchStrategy<P, T>`** — factory that creates a `DistancesUnordered` visitor from a provider + context, and builds the per-query computer. Mirrors the graph-side strategy pattern. Two fallible methods:
- `create_visitor` — borrows provider + context, returns a `Visitor`
- `build_query_computer` — preprocesses the query `T` into a `QueryComputer`

```rust
pub trait SearchStrategy<P, T>: Send + Sync
where
    P: DataProvider,
{
    type ElementRef<'a>;
    type Id;
    type QueryComputer: for<'a> PreprocessedDistanceFunction<Self::ElementRef<'a>, f32>;
    type QueryComputerError: StandardError;
    type Visitor<'a>: for<'b> DistancesUnordered<
            Self::QueryComputer,
            ElementRef<'b> = Self::ElementRef<'b>,
            Id = Self::Id,
        >;
    type Error: StandardError;

    fn create_visitor<'a>(&'a self, provider: &'a P, context: &'a P::Context)
        -> Result<Self::Visitor<'a>, Self::Error>;

    fn build_query_computer(&self, query: T)
        -> Result<Self::QueryComputer, Self::QueryComputerError>;
}
```

## Index (`flat/index.rs`)

**`FlatIndex<P>`** — thin `'static` wrapper around a `DataProvider`. Currently we have implemented the naive kNN search algorithm for the flat index. `knn_search` asks the strategy for a visitor, builds the query computer, drives `distances_unordered` through a priority queue, and writes results via `SearchPostProcess`.

## Test infrastructure (`flat/test/`)

A self-contained test provider with dimension-validated `Strategy`, transient-error injection, and a `KnnOracleRun` harness that compares `knn_search` results against a brute-force reference with baseline caching for regression detection.

## Future work

- **Post-processing** — `knn_search` currently uses the graph-side `SearchPostProcess` trait to write results into the output buffer. #1067 will introduce a flat-specific post-processing step that decouples flat search from the graph module's output machinery.
- **Vector ids** - The vector id over which `DistancesUnordered` acts is an associated type, instead of the `InternalId` of the provider. This is due to overly restrictive trait bounds on `VectorId` trait. We plan on relaxing this allowing for more generic id types.


diskann/src/flat/index.rs
```diff
@@ -0,0 +1,109 @@
...
+            let computer = strategy.build_query_computer(query).into_ann_result()?;
+
+            let k = k.get();
+            let mut queue = NeighborPriorityQueue::new(k);
+            let mut cmps: u32 = 0;
+
+            callback
+                .distances_unordered(&computer, |id, dist| {
+                    cmps += 1;
+                    queue.insert(Neighbor::new(id, dist));
```
arrayka, line 82: What’s our extension path for dropping candidates whose distance exceeds a given threshold?
Would this logic live inside callback.distance_unordered()? If so, the cmps wouldn’t include the distance computations for candidates that get discarded - it is an issue.

diskann/src/flat/iterator.rs
```diff
@@ -0,0 +1,147 @@
...
+///
+/// The concrete computer is insantiated and supplied externally
+/// by the [`FlatSearchStrategy`](crate::flat::FlatSearchStrategy).
+///
+/// The default implementation delegates to [`OnElementsUnordered::on_elements_unordered`],
+/// calling `computer.evaluate_similarity` on each element.
+pub trait DistancesUnordered: OnElementsUnordered {
+    /// Drive the entire scan, scoring each element with `computer` and invoking `f` with
+    /// the resulting `(id, distance)` pair.
+    fn distances_unordered<C, F>(
```
hildebrandmw: This needs the concrete type of the distance computer, not a generic. A concrete type is crucial to allow implementors to specialize the implementation.

diskann/src/flat/iterator.rs
```diff
@@ -0,0 +1,147 @@
...
+        &mut self,
+        computer: &C,
+        mut f: F,
+    ) -> impl SendFuture<Result<(), Self::Error>>
+    where
+        C: for<'a> PreprocessedDistanceFunction<Self::ElementRef<'a>, f32> + Send + Sync,
+        F: Send + FnMut(Self::Id, f32),
```
arrayka: FnMut is not friendly to parallel iterators. If we ever want to parallelize the distance calculations for flat index, what is our extension path?

diskann/src/lib.rs
```diff
@@ -13,6 +13,7 @@ pub mod utils;
 pub(crate) mod internal;
 
 // Index Implementations
+pub mod flat;
```
arrayka, line 16: What is the rationale behind adding flat index to `diskann` crate if compared with creating `diskann-flat` crate?

arrayka (changes requested):

diskann/src/flat/iterator.rs
arkrishn94: How so? The mutability bound on `F` is restricted to the scope of the search for a single query. This doesn't prevent us from parallelizing each scan across multiple queries. It just restricts you from parallelizing within the search scope for a query. 

This is an identical pattern to the trait constraints for graph search.

diskann/src/lib.rs
arkrishn94: The provider is defined in `diskann` which is a first-class object for both graph and flat index. 

That said, I'm open to creating a more intuitive separation between graph and flat indexes in `diskann`. Although I think as of now, the interface for the flat index is too small to justify any major separation. But I'm open to ideas here.

diskann/src/flat/index.rs
arkrishn94: This is a good point, let me think about this.

diskann/src/flat/iterator.rs
arrayka: Yep, I'm talking about an option to parallelize distance computations between a single query vector and the indexed vectors - trading off CPU for lower search latency.

diskann/src/flat/post_process.rs
```diff
@@ -0,0 +1,72 @@
...
+/// Post-process the survivor candidates produced by a flat search and
+/// write them into an output buffer.
+///
+/// This is the flat counterpart to [`crate::graph::glue::SearchPostProcess`]. Processors
+/// receive `&mut S` so they can consult any iterator-owned lookup state (e.g., an
+/// `Id -> rich-record` table built up during the scan) when assembling outputs.
+///
+/// The `O` type parameter lets callers pick the output element type (raw `(Id, f32)`
+/// pairs, fully hydrated hits etc.).
+pub trait FlatPostProcess<S, T, O = <S as HasId>::Id>
```
hildebrandmw: One big concern I have is that this does not really share much with the existing graph code. Even though the desire is to share code, post-process routines like diversity search will still need to be implemented twice. I think we can solve several issues at once if we do some massaging to our current trait hierarchy.

Let's assume we do the following: First, add a new trait
```rust
/// A general supertrait like `HasId` that we can use to express 
/// relationships
pub trait HasElementRef {
    ElementRef<'a>;
}

/// Restructure the existing `BuildQueryComputer` as a subtrait of
/// `ElementRef`.
pub trait BuildQueryComputer<T>: HasElementRef {
    type QueryComputerError: std::error::Error + Into<ANNError> + Send + Sync + 'static;
    type QueryComputer: for<'a> PreprocessedDistanceFunction<Self::ElementRef<'a>, f32>
        + Send
        + Sync
        + 'static; // Maybe we can finally drop `'static`?

    fn build_query_computer(
        &self,
        from: T,
    ) -> Result<Self::QueryComputer, Self::QueryComputerError>;
}
```

Then, `Accessor` can add a new subtrait of `BuildQueryComputer` for its need of `distances_unordered`, and the flat `Visitor` can do so as well. Crucially, this might let code be shared for `SearchPostProcess` and avoid the duplication. And also keeps `BuildQueryComputer` a bit more centralized.

diskann/src/flat/strategy.rs
```diff
@@ -0,0 +1,77 @@
...
+/// - `Provider`: the [`DataProvider`] that backs the index.
+/// - `T`: the query type. Often `[E]` for vector queries; can be any `?Sized` type.
+pub trait FlatSearchStrategy<P, T>: Send + Sync
+where
+    P: DataProvider,
+    T: ?Sized,
+{
+    /// The iterator type produced by [`Self::create_callback`]. Borrows from `self` and the
+    /// provider.
+    type Callback<'a>: DistancesUnordered
```
hildebrandmw: Maybe `Visitor` instead?

diskann/src/flat/iterator.rs
```diff
@@ -0,0 +1,147 @@
...
+// Default //
+/////////////
+
+/// Bridges a [`FlatIterator`] into an [`OnElementsUnordered`] by looping over
+/// [`FlatIterator::next`] and reborrowing each element into the closure.
+///
+/// This is the default adapter for providers that implement element-at-a-time iteration.
+/// Providers that can do better (prefetching, SIMD batching, bulk I/O) should implement
+/// [`OnElementsUnordered`] directly.
+pub struct DefaultIteratedOperator<I> {
```
hildebrandmw: Nit: maybe just call this "iterated"?

diskann/src/flat/index.rs
```diff
@@ -0,0 +1,109 @@
...
+                })
+                .await
+                .into_ann_result()?;
+
+            let result_count = processor
+                .post_process(&mut callback, query, queue.iter().take(k), output)
+                .await
+                .into_ann_result()? as u32;
+
+            Ok(SearchStats {
```
hildebrandmw: This should probably be a bespoke return type. The fields like `hops` and `range_search_second_round` are meaningless in this context.

diskann/src/flat/index.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// the survivors to the post-processor.
+    ///
+    /// # Arguments
+    /// - `k`: number of nearest neighbors to return.
+    /// - `strategy`: produces the per-query iterator and the query computer. See [`FlatSearchStrategy`]
+    /// - `processor`: post-processes the survivor candidates into the output type.
+    /// - `context`: per-request context threaded through to the provider.
+    /// - `query`: the query.
+    /// - `output`: caller-owned output buffer.
+    pub fn knn_search<S, T, O, OB, PP>(
```
hildebrandmw: We recently went through a whole thing of adding the `Search` trait to the graph index to avoid the proliferation of search methods on the index. We should probably do the same here.

diskann/src/flat/index.rs
```diff
@@ -0,0 +1,109 @@
...
+/// A `'static` thin wrapper around a [`DataProvider`] used for flat search.
+///
+/// The provider is owned by the index. The index is constructed once at process startup and
+/// shared across requests; per-query state lives in the [`crate::flat::OnElementsUnordered`]
+/// implementation that the [`crate::flat::FlatSearchStrategy`] produces.
+#[derive(Debug)]
+pub struct FlatIndex<P: DataProvider> {
+    /// The backing provider.
+    provider: P,
+    _marker: PhantomData<fn() -> P>,
```
hildebrandmw: Flat index already owns a `P` - the `_marker` field isn't doing anything besides being confusing :smile:.

diskann/src/flat/iterator.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// distance-function HRTB bounds.
+    type ElementRef<'a>;
+
+    /// The concrete element returned by [`Self::next`]. Reborrows to [`Self::ElementRef`].
+    type Element<'a>: for<'b> Reborrow<'b, Target = Self::ElementRef<'b>> + Send + Sync
+    where
+        Self: 'a;
+
+    /// The error type yielded by [`Self::next`].
+    type Error: StandardError;
```
hildebrandmw: On error types, maybe consider `ToRanked` instead of the `Into<ANNError>` that `StandardError` implies? That said, the visitor pattern means the implementation can swallow non-critical errors on its own. So maybe not needed, for the visitor case. But on the iterator case, a `ToRanked` might be a good idea.

hildebrandmw (commented):
Thanks Aditya. Left a few general comments with some ideas on how we might improve our code sharing. In general, I'm not a fan of prefixing everything with `Flat`. We already have the `flat` module so `flat::SearchStrategy` reads fine to me as opposed to `flat::FlatSearchStrategy`, which is a little redundant.

diskann/src/flat/iterator.rs
hildebrandmw: That's a good point. Definitely worth considering the parallelism model. You can riffing off rayon's [`Consumer`](https://github.com/rayon-rs/rayon/blob/c4dac9f82ee03f61e855075b668229a9e4d759ad/src/iter/plumbing/mod.rs#L123-L146) [trait](https://github.com/rayon-rs/rayon/blob/main/src/iter/plumbing/README.md) for ideas on how to generically create multiple parallel search contexts and reduce all the results back together.

I could see this as basically providing a way to create a new priority queue to give to another thread to work in parallel and then merging the results at the end. Actually, we might be able to do an unindexed "split" like thing that would put the job of parallelizing back to the `Index` level, which is what presumably knows how many threads can be used.

arrayka: We've made a decision to solve this in a subsequent PR. A work item to track this question #894

diskann/src/flat/index.rs
arrayka: We've made a decision to solve this in a subsequent PR. A work item to track this question #1001

diskann/src/flat/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+ * Licensed under the MIT license.
+ */
+
+//! Sequential ("flat") search infrastructure.
+//!
+//! This module is the streaming counterpart to the random-access [`crate::provider::Accessor`]
+//! family. It is designed for backends whose natural access pattern is a one-pass scan over
+//! their data — for example append-only buffered stores, on-disk shards streamed via I/O,
+//! or any provider where random access is significantly more expensive than sequential.
+//!
```
harsha-simhadri, line 12: This is a nice description comparing traits that enable algorithms based on random access vs sequential scans. Could this be in a higher level directory, either in providers.rs file or in `diskann/src/agents.md`

rfcs/00983-flat-search.md
```diff
@@ -0,0 +1,228 @@
...
+}
+```
+
+`DefaultIteratedOperator<I>` wraps any `FlatIterator` and implements `OnElementsUnordered`
+(and `DistancesUnordered` by inheritance) by looping over `next()` and reborrowing each
+element. 
+
+
+### The glue: `FlatSearchStrategy`
+
```
harsha-simhadri, line 77: Since we are introducing substantial new machinery, can we think about whether we can reuse some of this for IVF/SPANN type of indices that rely on clustering and then scanning entire data in specific clusters to support queries. 

I can imagine that the `OnElementsUnordered` and `DistancesUnordered` could be adapted to the scope of a cluster. 

And SearchStrategies for IVF would need to be added.

Even if we dont have a fully fleshed our proposal for clustering based indices, it would be ideal to document how the abstractions can be reused or adapted in the near future and avoid another set of abstractions for clustering based indices

rfcs/00983-flat-search.md
```diff
@@ -0,0 +1,228 @@
...
+1. Construct the per-query callback via `strategy.create_callback`.
+2. Build the query computer via `strategy.build_query_computer`.
+3. Drive the scan via `callback.distances_unordered(&computer, ...)`, inserting each
+   `(id, distance)` pair into a `NeighborPriorityQueue<Id>` of capacity `k`.
+4. Hand the survivors (in distance order) to `processor.post_process`.
+5. Return search stats.
+
+Other algorithms (filtered, range, diverse) can be added later as additional methods on
+`FlatIndex`.
```
harsha-simhadri: How would we support predicates on flat index?

diskann/src/flat/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+//! The module mirrors the layering used by graph search:
+//!
+//! | Graph (random access)                     | Flat (sequential)                 |
+//! | :------------------------------------     | :-------------------------------- |
+//! | [`crate::provider::DataProvider`]         | [`crate::provider::DataProvider`] |
+//! | [`crate::graph::DiskANNIndex`]            | [`FlatIndex`]                     |
+//! | [`crate::provider::Accessor`]             | [`FlatIterator`]                  |
+//! | [`crate::graph::glue::SearchStrategy`]    | [`FlatSearchStrategy`]            |
+//! | [`crate::graph::glue::SearchPostProcess`] | [`FlatPostProcess`]               |
+//! | [`crate::graph::Search`]                  | [`FlatIndex::knn_search`]         |
```
harsha-simhadri: This table is very useful.
Could this description be upleveled to `diskann/src/agents.md` or readme.

diskann/src/flat/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+//!
+//! # Hot loop
+//!
+//! Algorithms drive the scan via [`FlatIterator::next`] (lending iterator) or override
+//! [`FlatIterator::on_elements_unordered`] when batching/prefetching wins. The default
```
harsha-simhadri: Is this paragraph supposed to be here. Seems a bit out of place.

diskann/src/flat/post_process.rs
```diff
@@ -0,0 +1,72 @@
...
+//! [`FlatPostProcess`] — terminal stage of the flat search pipeline.
+
+use diskann_utils::future::SendFuture;
+
+use crate::{
+    error::StandardError, flat::OnElementsUnordered, graph::SearchOutputBuffer, neighbor::Neighbor,
+    provider::HasId,
+};
+
+/// Post-process the survivor candidates produced by a flat search and
```
harsha-simhadri: Is the intention to support filters with post_process? If so where is the clause?

diskann/src/flat/index.rs
arkrishn94: Ack, makes sense.

diskann/src/flat/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+//! This module is the streaming counterpart to the random-access [`crate::provider::Accessor`]
+//! family. It is designed for backends whose natural access pattern is a one-pass scan over
+//! their data — for example append-only buffered stores, on-disk shards streamed via I/O,
+//! or any provider where random access is significantly more expensive than sequential.
+//!
+//! # Architecture
+//!
+//! The module mirrors the layering used by graph search:
+//!
+//! | Graph (random access)                     | Flat (sequential)                 |  Shared?  |
```
arrayka: Consider adding `Responsibility` column and provide one-sentence description for each layer.

diskann/src/flat/iterator.rs
```diff
@@ -0,0 +1,152 @@
...
+/// calling `computer.evaluate_similarity` on each element.
+///
+/// ## Note
+///
+/// This is the flat analog to [`crate::provider::DistancesUnordered`] which runs over
+/// a random-access [`crate::provider::Accessor`].
+pub trait DistancesUnordered<T>: OnElementsUnordered + BuildQueryComputer<T> {
+    /// Drive the entire scan, scoring each element with `computer` and invoking `f` with
+    /// the resulting `(id, distance)` pair.
+    fn distances_unordered<F>(
```
arrayka: The algorithm only needs `DistancesUnordered`. Making it a supertrait of `OnElementsUnordered` forces all distance-capable providers to also expose raw element streaming, which leaks a lower-level capability and may block implementations that can compute distances without exposing element refs.

Could we decouple the traits and provide a blanket impl of `DistancesUnordered` for `OnElementsUnordered + BuildQueryComputer` instead:

Make `DistancesUnordered` **independent**, and provide a separate adapter/blanket impl *when* `OnElementsUnordered` is available. Conceptually:

- `trait DistancesUnordered<T> { fn distances_unordered(...) ... }`  (no supertrait)
- `impl<T, S> DistancesUnordered<T> for S where S: OnElementsUnordered + BuildQueryComputer<T> { ... }`

That keeps the algorithm-facing surface minimal, preserves encapsulation, and still keeps the convenience default behavior for types that *do* implement `OnElementsUnordered`.

diskann/src/flat/iterator.rs
```diff
@@ -0,0 +1,152 @@
...
+        &mut self,
+        computer: &<Self as BuildQueryComputer<T>>::QueryComputer,
+        mut f: F,
+    ) -> impl SendFuture<Result<(), <Self as OnElementsUnordered>::Error>>
+    where
+        F: Send + FnMut(<Self as HasId>::Id, f32),
+    {
+        self.on_elements_unordered(move |id, element| {
```
arrayka: All functions (including default implementations) should be covered by unit tests (ideally, in the same file).

diskann/src/flat/index.rs
```diff
@@ -0,0 +1,104 @@
...
+        OB: SearchOutputBuffer<O> + Send + ?Sized,
+        PP: for<'a> SearchPostProcess<S::Visitor<'a>, T, O> + Send + Sync,
+    {
+        async move {
+            let mut visitor = strategy
+                .create_visitor(&self.provider, context)
+                .into_ann_result()?;
+
+            let computer =
+                BuildQueryComputer::build_query_computer(&visitor, query).into_ann_result()?;
```
arrayka: Could you please provide an example of why visitor is needed to build a query computer?

diskann/src/graph/test/provider.rs
```diff
@@ -1032,12 +1032,15 @@ impl provider::HasId for Accessor<'_> {
     type Id = u32;
 }
 
+impl provider::HasElementRef for Accessor<'_> {
```
arrayka: As an option, consider extracting this prerequisite change into a separate PR to keep this one smaller and easier to review.

diskann/src/flat/post_process.rs
arkrishn94: Done, makes sense. Have made this refactored and propagated changes. RFC and description should also reflect this. I like it

diskann/src/flat/iterator.rs
arkrishn94: Makes sense, lmk how the new structure looks.

diskann/src/flat/mod.rs
arkrishn94: Agreed, I need to clean up.

diskann/src/flat/iterator.rs
arkrishn94: - I'm a little confused; how would we define `DistancedUnordered` without the `BuildQueryComputer` trait bound?Specifically, the computer signature in the definition of `distances_unordered`. If possible, can you flesh this out a bit?
- On your suggestion, if we implement the default implementation on any `S : OnElementsUnordered + BuildQueryComputer` then we won't allow a consumer to specialize the implementation for their `S` no?

I'm not sure I'm following why we might want to have implementations of `DistancesUnordered` that don't implement `OnElementsUnordered` :)

diskann/src/flat/iterator.rs
arkrishn94: 100%. Unit and integration testing is coming up

diskann/src/flat/index.rs
arkrishn94: Couple reasons - 
1. `visitor` implements `DistancesUnordered` so it is not unreasonable it should know how to construct the correct type to apply the streaming distance computation.
2. Making the visitor implement `DistancesUnordered` has the advantage of being symmetric to the graph case: where the `search_accessor` implements `ExpandBeam` which must implement `BuildQueryComputer<T>`. This symmetry helps a bit with sharing the post-process trait `SearchPostProcess` for both graph and flat search - since trait bounds are identical for both paths.

diskann/src/graph/test/provider.rs
arkrishn94: Yeah, I'm open to separating this change to a pre-cursor PR. 

My only worry is that, since this refactor is specific to enabling the new flat search trait structure I'm proposing here, I don't want to rush the refactor in an earlier PR only to change the flat search trait architecture.

diskann/src/flat/post_process.rs
arkrishn94: Post-processing for flat is now common with `diskann::graph::glue::SearchPostProcess`. So filter will have to live in the object implementing this trait like in the graph case - e.g. inline_beta_filter. Or it can live in the `Visitor<'_>` itself.

rfcs/00983-flat-search.md
arkrishn94: it's a good point - type-1 bitmap filters can be supported by extending the scan trait (`DistancesUnordered`) with a typed predicate applied before computing distances. This can then be wired through at the top-level search API depending on whether we're doing brute-force or filtered search.

diskann/src/flat/test/harness.rs
```diff
@@ -0,0 +1,142 @@
...
+    neighbor::{BackInserter, Neighbor},
+    test::tokio::current_thread_runtime,
+    utils::VectorRepr,
+};
+
+/// Result of running [`FlatIndex::knn_search`] under the harness alongside a
+/// brute-force ground-truth oracle.
+#[derive(Debug, Clone)]
+pub(crate) struct KnnOracleRun {
+    /// Top-`k` `(id, distance)` pairs in canonical `(distance asc, id asc)` order.
```
arrayka: Minor: shouldn't id be in the first position: `(distance asc, id asc)`.

diskann/src/flat/index.rs
```diff
@@ -107,3 +107,98 @@ impl<P: DataProvider> FlatIndex<P> {
...
+        let provider = flat_provider::Provider::grid(grid, size);
+        let len = provider.len();
+        (FlatIndex::new(provider), len)
+    }
+
+    /// `knn_search` returns a `Send` future, and a shared `&FlatIndex` can serve
+    /// many concurrent searches on a multi-threaded runtime, each producing the
+    /// correct top-k independently.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+    async fn knn_search() {
```
arrayka: I would name the test something like this: `flat_index_supports_multi_threading`

diskann/src/flat/test/provider.rs
```diff
@@ -0,0 +1,442 @@
...
+    pub fn len(&self) -> usize {
+        self.items.len()
+    }
+
+    /// `true` if there are no vectors.
+    pub fn is_empty(&self) -> bool {
+        self.items.is_empty()
+    }
+
+    /// Distance metric the provider was constructed with.
```
arrayka: Why does data provider (which responsibility is to provide data vectors) care about the distance metric?
Why is data provider defining the allowed distance metric?

diskann/src/flat/test/provider.rs
```diff
@@ -0,0 +1,442 @@
...
+    pub fn metric(&self) -> Metric {
+        self.metric
+    }
+
+    /// Snapshot of the per-provider counters.
+    pub fn metrics(&self) -> Metrics {
```
arrayka: Minor: `Metrics` is a little bit confusing with `Metric`. Consider renaming `Metrics` to `Counters` to avoid confusion.

arrayka (approved):

diskann/src/flat/iterator.rs
arkrishn94: Ack. Done.

diskann/src/flat/iterator.rs
```diff
@@ -0,0 +1,615 @@
...
+//////////////
+// Iterator //
+//////////////
+
+/// A lending, asynchronous iterator over the elements of a flat index.
+///
+/// Implementations provide element-at-a-time access via [`Self::next`]. Providers that
+/// only implement `FlatIterator` can be wrapped in [`Iterated`] to obtain a
+/// default [`DistancesUnordered`] implementation.
+pub trait FlatIterator: HasId + HasElementRef + Send + Sync {
```
hildebrandmw: I'm still not convinced that this is justified. If someone can write a coherent `next` implementation and all of the state-tracking shenanigans that that will require, why can they not write a handful more lines of code and simply implement all of `DistancesUnordered`?

diskann/src/flat/iterator.rs
```diff
@@ -0,0 +1,615 @@
...
+    provider::{BuildQueryComputer, HasElementRef, HasId},
+};
+
+/// Fused iterate-and-score primitive over the elements of a flat index.
+///
+/// Implementations drive an entire scan over the underlying data, scoring each
+/// element with the supplied [`BuildQueryComputer::QueryComputer`] and invoking
+/// `f` with the resulting `(id, distance)` pair. The super-trait
+/// [`BuildQueryComputer<T>`] supplies the computer type.
+pub trait DistancesUnordered<T>: HasId + BuildQueryComputer<T> + Send + Sync {
```
hildebrandmw: Why is this in `iterator` and not in `strategy`? Also, with the idea outlined in #1067, should we align on like `SearchExt` or something for these core "do the thing" traits? Not super set on names, but `DistancesUnordered` is a bit long.

diskann/src/flat/iterator.rs
```diff
@@ -0,0 +1,615 @@
...
+    /// The error type yielded by [`Self::distances_unordered`].
+    type Error: ToRanked + Debug + Send + Sync + 'static;
```
hildebrandmw: I suppose the error bound on this super trait can be `Into<ANNError>` instead of `ToRanked`. Since it is coarse grained, implementations can decide internally to accept transient errors or now. I'm not sure I see a compelling reason to return a `ToRanked` error here.

diskann/src/flat/iterator.rs
```diff
@@ -0,0 +1,615 @@
...
+//! Sequential ("flat") access primitives.
+//!
+//! This module defines the traits that flat-search algorithms use to walk every element
+//! of a [`DataProvider`](crate::provider::DataProvider) once.
+//!
+//! * [`DistancesUnordered`]: the single trait flat search consumes. It takes a
+//!   pre-built query computer and a callback, applies the callback to every
+//!   `(id, distance)` pair in the provider, and is the only trait an in-memory
+//!   visitor (such as [`crate::flat::test::provider::Visitor`]) needs to implement.
+//!   The super-traits [`HasId`] and [`BuildQueryComputer`] define the id and
```
hildebrandmw: Can we remove this extra fluff that is already easily discoverable via the type system?

diskann/src/flat/test/provider.rs
```diff
@@ -0,0 +1,406 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+#![allow(dead_code)]
```
hildebrandmw: Why are we allowing `dead_code`?

diskann/src/flat/test/provider.rs
```diff
@@ -0,0 +1,406 @@
...
+};
+
+//////////////
+// Provider //
+//////////////
+
+/// In-memory test provider for flat search.
+#[derive(Debug)]
+pub struct Provider {
+    items: Vec<Vec<f32>>,
```
hildebrandmw: Would a `Matrix` be a better fit? Unless the plan is to let this be modifiable in the future. In which case we'd need locks and all that fun stuff.

diskann/src/flat/test/provider.rs
```diff
@@ -0,0 +1,406 @@
...
+
+impl Provider {
+    /// Construct a provider that owns `items`. Every vector must have the same
+    /// (non-zero) length.
+    pub fn new(items: impl IntoIterator<Item = Vec<f32>>) -> Self {
+        let items: Vec<Vec<f32>> = items.into_iter().collect();
+        assert!(
+            !items.is_empty(),
+            "flat::test::Provider needs at least one item"
+        );
```
hildebrandmw: Can we follow the pattern of the test provider in `graph` and return `Error`s instead of panics? I've been treating the test provider in `graph` as an example of how to at least not do things wrong and to use as a general-purpose provider for downstream crates to rely on. I think it's worth making the API tight.

diskann/src/flat/strategy.rs
```diff
@@ -0,0 +1,78 @@
...
+
+//! [`SearchStrategy`] — glue between [`DataProvider`] and per-query
+//! [`DistancesUnordered`] visitors.
+
+use crate::{error::StandardError, flat::DistancesUnordered, provider::DataProvider};
+
+/// Per-call configuration that knows how to construct a per-query
+/// [`DistancesUnordered<T>`] visitor for a provider.
+///
+/// `SearchStrategy` is the flat counterpart to [`crate::graph::glue::SearchStrategy`]
```
hildebrandmw: What happens when `graph::glue::SearchStrategy` describes itself as the graph counterpart to `flat::SearchStrategy` :laugh:?

diskann/src/flat/strategy.rs
```diff
@@ -0,0 +1,78 @@
...
+/// (disambiguated by module path). A strategy instance carries the per-query setup
+/// recipe; the per-query mutable state lives in the visitor it produces, so a single
+/// strategy may be reused across many searches.
+///
+/// The strategy itself is a pure factory; the visitor it produces carries the
```
hildebrandmw: Style: I'm all for docs, but much of this fluffy and redundant.

diskann/src/flat/strategy.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn exercise_create_visitor() {
+        let provider = flat_provider::Provider::grid(Grid::Two, 3);
+        let context = flat_provider::Context::new();
+        let strategy = Strategy::new();
+
+        let v1 = strategy.create_visitor(&provider, &context).unwrap();
+        let v2 = strategy.create_visitor(&provider, &context).unwrap();
+
+        // The two visitors must occupy distinct stack slots — i.e. holding `v1`
+        // does not preclude constructing `v2`.
```
hildebrandmw: What exactly is this testing? How could this ever not be the case?

diskann/src/provider.rs
```diff
@@ -519,11 +537,16 @@ pub trait BuildQueryComputer<T>: Accessor {
...
-    /// Compute the distances for the elements in the iterator `itr` using the
-    /// `computer` and apply the closure `f` to each distance and ID. The default
-    /// implementation uses on_elements_unordered to iterate over the elements
-    /// and compute the distances using `computer` parameter.
+/// A sub-trait of [`Accessor`] and [`BuildQueryComputer`] that exposes the fused
+/// iterate-and-compute primitive `distances_unordered`.
+///
+/// The default implementation uses [`Accessor::on_elements_unordered`] to iterate over the
+/// elements and computes their distances using the provided `computer`.
+pub trait DistancesUnordered<T>: Accessor + BuildQueryComputer<T> {
```
hildebrandmw: That this somewhat general trait shares the same name as the much more specific trait in the flat index is quite confusing. I support traits having the same name if they serve the same purpose in different modules. For example, `graph::SearchExt` and `flat::SearchExt` - these (potentially hypothetical) traits serve the same purpose for different index types.

hildebrandmw (commented):
Thanks Aditya,

Here's my take: The new `HasElementRef` and `DistancesUnordered` (the one in `provider`, not `flat`) add a hefty boilerplate burden to users of the graph index. Especially in light of something like #1067 which pushes in the other direction and while that PR is still experimental, I'm getting more convinced is the right API for the graph index. This puts us in kind of an awkward place. If we get this in, users consuming a new version will have churn for these little misc traits that will then need to be undone by #1067 (or some spiritual successor) which is not great.

If you want to start programming against this API, what if you introduce your own temporary `TemporaryBuildQueryComputer` (I believe this is the biggest reason why this PR needs to add `HasElementRef` and `DistancesUnrodered`) and we hold off on post processing for a short period of time to bottom out on #1067. If the latter doesn't pan out, then we can continue with these changes and accept it. Otherwise, the flat index implementations (which presumably will be less numerous than the current graph implementations) can be brought inline with the new trait organization and we'll cause less overall churn for our users.

I also remain unconvinced that the `Iterator` is any simpler than just implementing the new `DistancesUnordered` trait directly.

diskann/src/flat/iterator.rs
arkrishn94: I don't expect all (or even most) consumers to implement the flat index using the `FlatIterator` API. This is to provide  an easy entry point for a naive implementation of the flat index for providers. Fwiw it also helps a bit with testing providers the flat index search path for providers that can implement an iteration pattern easily.

diskann/src/flat/iterator.rs
arkrishn94: Fair enough, this should probably move to strategy. On the naming, I think `DistancesUnordered` is pretty good at describing what this trait does. But I'm open to reconsidering the name based on what we end with for graph search.

diskann/src/flat/test/provider.rs
arkrishn94: Thanks, good catch

diskann/src/flat/test/provider.rs
arkrishn94: Fair enough

arkrishn94:
> If you want to start programming against this API, what if you introduce your own temporary `TemporaryBuildQueryComputer` (I believe this is the biggest reason why this PR needs to add `HasElementRef` and `DistancesUnrodered`) and we hold off on post processing for a short period of time to bottom out on #1067. If the latter doesn't pan out, then we can continue with these changes and accept it. Otherwise, the flat index implementations (which presumably will be less numerous than the current graph implementations) can be brought inline with the new trait organization and we'll cause less overall churn for our users.
>

Thanks for the feedback and thorough review @hildebrandmw. To make sure I understand, are you suggesting we temporarily create a disjoint trait structure for the flat index (avoiding supporting post-processing) and then once #1067 settles down, we can re-evaluate how much shared surface we can have for these two indexes (both on the building/construction side and the post-processing)?

diskann/src/flat/iterator.rs
hildebrandmw: I'm not convinced that this is any **easier** an entry point than `DistancesUnordered`. You still need all the `BuildQueryComputer` implementations, there are more associated types, the return type is more complicated, users need to implement state tracking, and users have to know to use the `Iterated` proxy type. If one can manage to do all of that, then it's trivial to implement `DistancesUnordered` directly and be done with it.

diskann/src/flat/iterator.rs
hildebrandmw: We probably need to bikeshed for a bit. I'm not happy with `SearchExt` as a trait name. But if we think about the potential future of `DistancesUnordered`, I'm thinking there is going to have to be machinery around parallelism, right? For example splitting into chunks that are processed independently. At which point, the `DistancesUnordered` functionality is one facet of the overall process.

hildebrandmw:
> > If you want to start programming against this API, what if you introduce your own temporary `TemporaryBuildQueryComputer` (I believe this is the biggest reason why this PR needs to add `HasElementRef` and `DistancesUnrodered`) and we hold off on post processing for a short period of time to bottom out on #1067. If the latter doesn't pan out, then we can continue with these changes and accept it. Otherwise, the flat index implementations (which presumably will be less numerous than the current graph implementations) can be brought inline with the new trait organization and we'll cause less overall churn for our users.
> 
> Thanks for the feedback and thorough review @hildebrandmw. To make sure I understand, are you suggesting we temporarily create a disjoint trait structure for the flat index (avoiding supporting post-processing) and then once #1067 settles down, we can re-evaluate how much shared surface we can have for these two indexes (both on the building/construction side and the post-processing)?

Yeah, that's what I'm proposing. Something to minimize code churn, or at least defer it temporarily until we know that we aren't going to introduce these intermediate traits (`HasElementRef` + `provider::DistancesUnordered`) just to remove them again immediately.

diskann/src/flat/strategy.rs
arkrishn94: Removed

diskann/src/flat/iterator.rs
arkrishn94: Agreed. As discussed offline, till we converge on a shared name for these two traits, let us keep as is. We can follow up with a name change.

diskann/src/flat/test/provider.rs
arkrishn94: Done. That makes sense.

diskann/src/flat/index.rs
arkrishn94: Ended up not adding this to the `Search` trait since as is, the trait is defined for a `DiskANNIndex` and graph strategies and post-processing. Once #1067 settles, we can refine these higher level traits to capture both index types?

diskann/src/flat/iterator.rs
arkrishn94: Resolving this since we've gotten rid of `OnElementsUnordered`, `Iterator` and `BuildQueryComputer`

diskann/src/flat/index.rs
arkrishn94: Renamed

diskann/src/flat/test/provider.rs
arkrishn94: Renamed

diskann/src/flat/iterator.rs
arkrishn94: Removed `Iterator` and related objects

diskann/src/flat/mod.rs
arkrishn94: Will pause on solidifying the docs here until the API stabilizes.

diskann/src/flat/iterator.rs
arkrishn94: Keeping `ToRanked` here in case the caller wants to decide whether to raise/suppress the error. Unless you feel strongly this will not be needed.

arrayka (approved):

hildebrandmw (approved):
Thanks for your patience Aditya!

diskann/src/provider.rs
arkrishn94: Keeping as is for now till #1067 lands. Will converge once we agree on a common name for this family of traits.

rfcs/00983-flat-search.md
arkrishn94: This is a good point. I am working on a short addition to this on how IVF type indices might use `DistancesUnordered`

PR #984, opened 2026-04-28
[diskann-disk] Decouple buffer memory alignment from disk block_size

doliawu:
DiskSectorGraph previously allocated sectors_data aligned to block_size (typically 4096), conflating the on-disk stride with the reader's hardware/OS memory-placement requirement. Move the requirement where it belongs: declare it on the reader as
`AlignedFileReader::MEMORY_ALIGNMENT` (default 1; direct-I/O readers override to 512) and have the buffer allocator honor that.

Drop the per-request alignment check from AlignedRead::new — that constraint is the caller's responsibility (DiskSectorGraph's offsets and lengths are already block_size-aligned by construction). Making construction infallible cleans up call sites.

Refresh VertexProvider / VertexProviderFactory docs to drop stale references to ANNWrapper and JetVertexProvider.

### DIFF
--- a/diskann-disk/src/search/provider/disk_provider.rs
+++ b/diskann-disk/src/search/provider/disk_provider.rs
@@ -653,6 +653,7 @@ where
             query,
         })
     }
+
     fn ensure_loaded(&mut self, ids: &[u32]) -> Result<(), ANNError> {
         if ids.is_empty() {
             return Ok(());
--- a/diskann-disk/src/search/traits/vertex_provider.rs
+++ b/diskann-disk/src/search/traits/vertex_provider.rs
@@ -8,7 +8,7 @@ use diskann::ANNResult;
 
 /// `VertexProvider` is a trait that abstracts the access to Vertex data.
 ///
-/// This trait provides an interface to interact with different types of vertex providers structures such as `DiskVertexProvider` and `JetVertexProvider`.
+/// This trait provides an interface to interact with vertex provider structures such as `DiskVertexProvider`.
 ///
 /// # Types
 ///
@@ -47,7 +47,7 @@ pub trait VertexProvider<Data: GraphDataType>: Send + Sync {
         vertex_id: &Data::VectorIdType,
     ) -> ANNResult<&[Data::VectorIdType]>;
 
-    // Gets the associated data for a given vertex id.
+    /// Gets the associated data for a given vertex id.
     ///
     /// The `get_associated_data` function attempts to retrieve the associated data for the
     /// specified `vertex_id`. This function returns an `ANNResult` that wraps a reference to the associated data.
@@ -80,11 +80,11 @@ pub trait VertexProvider<Data: GraphDataType>: Send + Sync {
     /// If it fails, returns an `ANNError`.
     fn load_vertices(&mut self, vertex_ids: &[Data::VectorIdType]) -> ANNResult<()>;
 
-    /// This function to process the loaded node
+    /// Processes a vertex previously loaded by `load_vertices`, materializing its full-precision vector and adjacency list from the raw on-disk record so they can be served by the `get_*` accessors.
     /// # Parameters
     ///
-    /// * `vertex_id`: A Data::VectorIdType value representing the id of the vertex for which to process.
-    /// * `idx`: A usize value representing the index of the vertex in the loaded node list.
+    /// * `vertex_id`: A Data::VectorIdType value representing the id of the vertex to process.
+    /// * `idx`: A usize value representing the index of the vertex in the slice passed to the preceding `load_vertices` call.
     ///
     /// # Returns
     /// * `ANNResult<()>`: If the operation is successful, returns Ok.
@@ -92,12 +92,12 @@ pub trait VertexProvider<Data: GraphDataType>: Send + Sync {
     /// If it fails, returns an `ANNError`.
     fn process_loaded_node(&mut self, vertex_id: &Data::VectorIdType, idx: usize) -> ANNResult<()>;
 
-    // Returns the number of IO operations performed by the vertex provider.
+    /// Returns the number of IO operations performed by the vertex provider.
     fn io_operations(&self) -> u32;
 
-    // Returns the number of vertices loaded by the vertex provider.
+    /// Returns the number of vertices loaded by the vertex provider.
     fn vertices_loaded_count(&self) -> u32;
 
-    // Clears the members of the vertex provider.
+    /// Clears the members of the vertex provider.
     fn clear(&mut self);
 }
--- a/diskann-disk/src/utils/aligned_file_reader/linux_aligned_file_reader.rs
+++ b/diskann-disk/src/utils/aligned_file_reader/linux_aligned_file_reader.rs
@@ -12,7 +12,7 @@ use diskann_platform::ssd_io_context::IOContext;
 use io_uring::IoUring;
 use libc;
 
-use crate::utils::aligned_file_reader::{traits::AlignedFileReader, AlignedRead};
+use crate::utils::aligned_file_reader::{traits::AlignedFileReader, AlignedRead, A512};
 
 pub const MAX_IO_CONCURRENCY: usize = 128;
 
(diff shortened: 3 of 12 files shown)


diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -653,6 +655,7 @@ where
             query,
         })
     }
+
```
wuw92, line 656: nit: unrelated whitespace, worth dropping?

wuw92:
For awareness, this PR overlaps with @hildebrandmw's review direction on #965: https://github.com/microsoft/DiskANN/pull/965#pullrequestreview-4182727911

His two concrete asks there both touch this PR:

1. **Add `PowerOfTwo` / `AlignedAllocator` const aliases** → the four `PowerOfTwo::new_or_panic(...)` callsites added here would collapse to those aliases, and the helper itself becomes unnecessary.
2. **Delete `aligned_slice.rs`, use `Poly` directly** → the two `aligned_slice(...)` callsites in `disk_sector_graph.rs` would migrate accordingly.

wuw92 (approved):

diskann-quantization/src/num.rs
```diff
@@ -75,6 +75,16 @@ impl PowerOfTwo {
         }
     }
 
+    /// Like [`Self::new`] but panics if `value` is not a power of two. Unlike
+    /// `new(v).unwrap()`, this is usable at item scope (e.g. trait `const`
+    /// declarations) because `Result::unwrap` is not const-stable for this `E`.
+    pub const fn new_or_panic(value: usize) -> Self {
+        match Self::new(value) {
+            Ok(v) => v,
+            Err(_) => panic!("value must be a power of two"),
```
wuw92: error: `panic` should not be present in production code

diskann-disk/src/search/traits/vertex_provider.rs
```diff
@@ -6,98 +6,71 @@
...
-    /// # Parameters
-    ///
-    /// * `vertex_ids`: A slice of Data::VectorIdType values representing the ids of vertices for which to retrieve `VertexAndNeighbors`.
-    ///
-    /// # Returns
-    ///
-    /// * `ANNResult<()>`: If the operation is successful, returns Ok.
-    ///
-    /// If it fails, returns an `ANNError`.
+    /// Issues the I/O to load the given batch of vertices into the
```
arrayka: The purpose of the original formatting was to generate a pretty looking documentation: https://docs.rs/diskann-disk/0.50.1/diskann_disk/search/traits/vertex_provider/trait.VertexProvider.html#tymethod.load_vertices

I'm not sure if we should propose changing of this formatting as a part of this PR.

diskann-disk/src/search/provider/disk_provider.rs
doliawu: it's intentional, there should be an empty line here

diskann-disk/src/utils/aligned_file_reader/linux_aligned_file_reader.rs
```diff
@@ -86,8 +87,12 @@ impl LinuxAlignedFileReader {
 }
 
 impl AlignedFileReader for LinuxAlignedFileReader {
+    /// O_DIRECT requires the buffer pointer to be aligned to the device sector
+    /// size in memory (512 bytes on typical Linux block devices).
+    const MEMORY_ALIGNMENT: PowerOfTwo = PowerOfTwo::V512;
+
     // Read the data from the file by sending concurrent io requests in batches.
-    fn read(&mut self, read_requests: &mut [AlignedRead<u8>]) -> ANNResult<()> {
+    fn read(&mut self, read_requests: &mut [ReadRequest<u8>]) -> ANNResult<()> {
```
arrayka: Do I understand correctly that alignment of ReadRequest's offset and buffers is not enforced anymore, and this can lead to UB, if a caller crafts a ReadRequest with wrong alignment?

Let's think if it is possible to enforce relationship between MEMORY_ALIGNMENT and ReadRequest (ideally, at compile time).

diskann-disk/src/utils/aligned_file_reader/traits/aligned_file_reader.rs
```diff
@@ -4,10 +4,22 @@
...
+    /// Memory-placement alignment for the buffer pointer of every
+    /// `ReadRequest<u8>` passed to [`Self::read`]. Always a power of two — a
+    /// hardware/allocator constraint. Default `1` (no requirement).
+    /// Direct-I/O readers (O_DIRECT, `FILE_FLAG_NO_BUFFERING`) override this
+    /// to the device sector size (typically 512 bytes).
+    ///
+    /// Disk offset and read length alignment are *not* declared here — they
+    /// are determined by the caller (e.g. `DiskSectorGraph` uses `block_size`).
+    /// Pairing a reader with an incompatible `block_size` will fail at I/O time.
+    const MEMORY_ALIGNMENT: PowerOfTwo = PowerOfTwo::V1;
```
arrayka: Do we truly believe that MEMORY_ALIGNMENT has to be power of two? Is it possible that some exotic data providers might have exotic memory alignment requirements like 8120 bytes?

arrayka (changes requested):

diskann-disk/src/utils/aligned_file_reader/traits/aligned_file_reader.rs
doliawu: That will be a bigger problem as the aligned allocation takes PowerOfTwo, I don't believe such case exists.

diskann-disk/src/utils/aligned_file_reader/linux_aligned_file_reader.rs
doliawu: Added runtime check, compile time check is possible for buffer as we can use the Poly type for read, but not for offset and length, it's a non-goal for this PR, but I'm open to further discussion.

diskann-disk/src/search/traits/vertex_provider.rs
doliawu: Kept the format and only correct the outdated info.

arrayka (approved):

arrayka (approved):

diskann-disk/src/utils/aligned_file_reader/traits/aligned_file_reader.rs
```diff
@@ -3,11 +3,141 @@
...
     /// Read the data from the file by sending concurrent io requests in batches.
-    fn read(&mut self, read_requests: &mut [AlignedRead<u8>]) -> ANNResult<()>;
+    fn read(&mut self, read_requests: &mut [ReadRequest<u8>]) -> ANNResult<()>;
+
+    /// Validate that every request in `requests` satisfies all three
+    /// alignment constraints (buffer pointer, buffer length, disk offset)
+    /// for `Self::MEMORY_ALIGNMENT`. Direct-I/O readers should call this
+    /// at the entry of their [`Self::read`] impl. For readers with
+    /// `MEMORY_ALIGNMENT = 1` every check passes trivially.
+    fn validate_alignment(requests: &[ReadRequest<u8>]) -> ANNResult<()> {
```
arkrishn94: It feels weird to have a typed request for the reader that needs validation after its construction. It might be more appropriate for this validation to happen at construction. 

Suggestion: does it make sense to make `ReadRequest` an associated type of `AlignedFileReader` so that it can be constructed with the appropriate alignment check?

arkrishn94 (commented):
Thanks for taking this on Dongliang, just have one comment. I know it's a little more work than might be needed but I think it it'll pay off to keep this API clean.

arkrishn94 (approved):

arkrishn94 (approved):
Nice, thanks Dongliang.

diskann-disk/src/utils/aligned_file_reader/traits/aligned_file_reader.rs
doliawu: good suggestion, updated

arrayka (approved):

PR #994, opened 2026-04-28
remove alignedslice and replace with direct poly calls

JordanMaples:
I synced up with @hildebrandmw after he returned from vacation and learned that I misunderstood the brief of #907 with my #965 pr. Rather than removing the struct and replace it with a type-alias for slightly less overhead, the ask was to just replace all calls. So, this follow-up PR does just that and removes AlignedSlice & aligned_slice and replaces all sites with Poly::broadcast calls. 

This also introduces PowerOfTwo consts for 0..64 P and value mappings and AlignedAllocator consts for 4...4096


JordanMaples:
@copilot could you review again? I believe that I've fully addressed your comments

diskann-quantization/src/num.rs
```diff
@@ -60,6 +60,161 @@ pub struct PowerOfTwo(NonZeroUsize);
...
 pub struct NotPowerOfTwo(usize);
 
+/// PowerOfTwo constants for all valid exponents (0..Usize::BITS-1) for 32 bit architectures
+///
+/// # Safety
+///
+/// 1 << N is a power of two for all N in 0..Usize::BITS-1.
+#[allow(clippy::undocumented_unsafe_blocks)]
+impl PowerOfTwo {
+    pub const P0: Self = unsafe { Self::new_unchecked(NonZeroUsize::new(1 << 0).unwrap()) };
```
hildebrandmw: Looking at this again, we can probably actually use a macro to clean this up and offer a tiny bit more safety:
```rust
macro_rules! constants {
    (($pow:ident, $value:ident) => $shift:literal) => {
        pub const $pow: Self = match Self::new(NonZeroUsize::new(1 << $shift).unwrap()) {
            Ok(v) => v,
            Err(_) => panic!("not a power of two"),
        };

        pub const $value: Self = $pow;
    },
    ($(($pow:ident, $value:ident) => $shift:literal),+ $(,)?) => {
        $(constants!(($pow, $value) => $shift);)+
    }
}

impl PowerOfTwo {
    powers! {
        (P0, V1) => 0,
        (P1, V2) => 1,
        (P2, V4) => 2,
        (P3, V8) => 3,
        ...
    }
}
```

hildebrandmw (approved):

diskann-quantization/src/num.rs
JordanMaples: good call, let's do this.

doliawu (approved):

PR #996, opened 2026-04-29
[benchmark/filtered-search prep] Search Plugins

hildebrandmw:
An issue that has been occurring more and more in `diskann-benchmark` is the inextensibility of adding new search kinds.

Currently a `SearchPhase` enum is used to select between topk, range, beta-filtered, and multi-hop filtered search. There are two places where this enum is matched: the giant `run_search_outer` method and the benchmark run for spherical quantization. Folks wishing to try new experimental post-processing or filtered style searches need to do the following:

1. Add a new `SearchPhase` variant.
2. Update the two match sites independently.

Now, here's the problem. Currently **all** static builds go through these sites, so new search flavors result in ~29 monomorphizations (4 full-precision only, `9 x 2` for scalar quantization, `2 x 2` for PQ, and 3 for spherical) of the search algorithm. If we ever want to implement just a subset of the search kinds, it's also difficult to update `Benchmark::try_match` and `Benchmark::description` accordingly. So adding a new search method has a disproportionate effect on compile times and is impossible or difficult to tweak on a per-index basis.

This PR fixes this by introducing a compile-time plugin-system for search kinds. The main addition is the `Plugin` dyn-compatible trait and the `Plugins` collection, defined in `diskann-benchmark/src/backend/index/search/plugins.rs`. The `Plugin` trait matches a particular input `Kind` (in this PR, just `SearchPhase`, but I left it as a generic to keep it more open-ended if needed) and returns `AggregatedSearchResults`. The `Plugins` struct is simply a collection of `Plugin` trait objects. Importantly, registering a new `Plugin` in `Plugins` is what (most likely) triggers the monomorphization of the corresponding method. The Plugins struct also provides convenience methods for listing, matching, and dispatching registered searches. This is used to implement `Benchmark::try_match` and `Benchmark::description` for users of this system.

Additionally, the new `plugins.rs` introduces 4 ZSTs for our current search kinds: `Topk`, `Range`, `BetaFilter` and `MultihopFilter`.

The benchmark types for index searches (`diskann-benchmark/src/backend/index/{benchmarks.rs, scalar.rs, spherical.rs, product.rs}`) are rewritten to wrap `Plugins` and the implementations of `Benchmark` are updated to query the set of registered searches in `Benchmark::try_match` and `Benchmark::description`. Thus, adding or removing a plugin will automatically be reflected.

The updates to `diskann-benchmark-runner` come in two flavors. First, the updates to `fmt.rs` introduce some utilities for making things look nice (`Indent`, `Delimit`, and `Quote`). The changes to `any.rs` and `app.rs` are targeting a more uniform approach to indenting benchmark related descriptions. With these changes, descriptions no longer need to track their indentation level implicitly: they may assume no indentation and the necessary formatting will happen at a higher level.


diskann-benchmark/src/backend/index/benchmarks.rs
```diff
@@ -55,11 +58,39 @@ use crate::{
 ////////////////////////////
 
 pub(super) fn register_benchmarks(benchmarks: &mut diskann_benchmark_runner::registry::Benchmarks) {
+    // Notes on registration:
```
harsha-simhadri, line 61: Could this note be added to diskann-benchmarks/README.md since that is the first entry point for users

diskann-benchmark/src/backend/index/scalar.rs
```diff
@@ -11,20 +11,52 @@ crate::utils::stub_impl!("scalar-quantization", inputs::async_::IndexSQOperation
 pub(super) fn register_benchmarks(benchmarks: &mut Benchmarks) {
     #[cfg(feature = "scalar-quantization")]
     {
+        use crate::backend::index::search::plugins::Topk;
         use half::f16;
 
+        // NOTE: We just register `Topk` for now to reduce compilation cost.
```
harsha-simhadri, line 17: same here, a note in README could be helpful

harsha-simhadri (approved):

harsha-simhadri:
Not to block this PR, but a similar problem arises when we switch providers. perhaps we should discuss this orthogonal to this PR

diskann-benchmark/src/backend/index/search/plugins.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub(crate) fn run(
+        &self,
+        index: Arc<DiskANNIndex<DP>>,
+        kind: &Kind,
+        parameters: &Params,
+    ) -> anyhow::Result<AggregatedSearchResults> {
+        match self.plugins.iter().find(|p| p.is_match(kind)) {
+            Some(plugin) => plugin.run(index, kind, parameters),
+            None => Err(anyhow::anyhow!(
+                "INTERNAL ERROR: Could not find a suitable search plugin",
```
narendatha: Minor nit: Including the kind in the error message would make debugging easier.

diskann-benchmark/src/backend/index/search/plugins.rs
```diff
@@ -0,0 +1,179 @@
...
+impl Range {
+    /// Returns [`SearchPhaseKind::Range`].
+    pub(crate) fn kind() -> SearchPhaseKind {
+        SearchPhaseKind::Range
+    }
+}
+
+/// A search plugin for beta-filtered search.
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct BetaFilter;
```
narendatha: nit: i prefer keeping the struct name same as searchPhaseKind for easier mental mapping.

narendatha (approved):

diskann-benchmark/src/backend/index/spherical.rs
```diff
@@ -264,199 +334,212 @@ mod imp {
...
-                        }
-                        SearchPhase::Range(search_phase) => {
-                            // Handle Range search phase
+                    // Save construction stats before running queries.
+                    checkpoint.checkpoint(&result)?;
+
+                    for layout in input.query_layouts.iter() {
+                        let search = self
+                            .search
+                            .run(index.clone(), &input.search_phase, layout)?;
```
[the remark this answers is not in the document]
hildebrandmw: It's fine. Loading time is outside the timed loop.

diskann-benchmark/src/backend/index/search/plugins.rs
hildebrandmw: Agreed. Unfortunately, with `kind` being generic, there isn't actually a way to get the kind diagnostics from it. I added a `Debug` bound on `Kind` and used that. Since this is supposed to be an internal logic error, I think that's fine.

diskann-benchmark/src/backend/index/search/plugins.rs
hildebrandmw: Fair enough - done.

diskann-benchmark/src/backend/index/benchmarks.rs
hildebrandmw: Added breadcrumbs in the README about how to add plugins as needed.

diskann-benchmark/src/backend/index/scalar.rs
hildebrandmw: Added to the README.

hildebrandmw:
> Not to block this PR, but a similar problem arises when we switch providers. perhaps we should discuss this orthogonal to this PR

Agreed. That is orthogonal to this PR, but it is a problem when trying to bring on BF-Tree or another provider.

diskann-benchmark/src/backend/index/product.rs
```diff
@@ -73,15 +128,61 @@ mod imp {
...
+                            arg.index_operation.source.data_type()
+                        )
+                    )?;
+
+                    if !self
+                        .quant_search
+                        .is_match(&arg.index_operation.search_phase)
+                    {
+                        writeln!(
+                            f,
```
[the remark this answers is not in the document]
hildebrandmw: Updated to only provide a diagnostic on mismatch.

diskann-benchmark/src/backend/index/benchmarks.rs
```diff
@@ -119,16 +192,39 @@ where
...
+                let data_type = match &arg.source {
+                    IndexSource::Load(load) => &load.data_type,
+                    IndexSource::Build(build) => &build.data_type,
+                };
+                writeln!(
+                    f,
+                    "Data/Query Type: {}",
+                    Why::<datatype::DataType, datatype::Type<T>>::new(data_type)
+                )?;
+
```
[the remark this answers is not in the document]
hildebrandmw: Updated to only apply a diagnostic on mismatch.

harsha-simhadri (approved):

PR #997, opened 2026-04-29
Replace sgemm() panics with Result-based error handling

### LINKED ISSUE #982: diskann_linalg::sgemm() should not panic, but should return Result()

arrayka:
## Summary

Refactors `diskann_linalg::sgemm()` to return `Result<(), SgemmError>` instead of panicking on invalid inputs. This change improves error handling and adds overflow protection for matrix dimension calculations.

**Principle:** DiskANN must not panic in production for recoverable errors; panics are reserved for catastrophic failures indicating undefined behavior.

### 1. New Error Type (`SgemmError`)
- Added `SgemmError` enum with two variants:
  - `InvalidMatrixDimensions`: Matrix has incorrect dimensions
  - `DimensionOverflow`: Multiplication would overflow `usize`
- Implemented `Display` and `std::error::Error` traits for proper error reporting

### 2. Updated `sgemm()` Function Signature
**Before:**
```rust
pub fn sgemm(...) { ... }
```

**After:**
```rust
pub fn sgemm(...) -> Result<(), SgemmError> { ... }
```

### 3. Overflow Protection
Replaced unsafe multiplication with `checked_mul()` for all dimension calculations:
- `m * k` validation (matrix a)
- `k * n` validation (matrix b)
- `m * n` validation (matrix c)

This prevents integer overflow vulnerabilities that could lead to undefined behavior.

### 4. Updated Call Sites
Updated all callers of `sgemm()` to handle the `Result`:
- **diskann-quantization**: Propagates errors through `TransformFailed::SgemmError` variant
- **diskann-disk**: Converts errors to `ANNError` with descriptive messages
- **diskann-linalg tests**: Added wrapper function for compatibility

### 5. Comprehensive Test Coverage
Added 6 new unit tests:
- `test_sgemm_invalid_matrix_a_dimensions`: Validates dimension check for matrix a
- `test_sgemm_invalid_matrix_b_dimensions`: Validates dimension check for matrix b
- `test_sgemm_invalid_matrix_c_dimensions`: Validates dimension check for matrix c
- `test_sgemm_m_times_k_overflow`: Tests overflow detection for `m * k`
- `test_sgemm_k_times_n_overflow`: Tests overflow detection for `k * n`
- `test_sgemm_m_times_n_overflow`: Tests overflow detection for `m * n`

## Review Order

  1. Core Changes (5 min)

  📁 diskann-linalg/src/lib.rs
  - Lines 15-50: New `SgemmError` type
  - Lines 155-202: Overflow protection with `checked_mul()`

  2. Tests (3 min)

  📁 diskann-linalg/src/lib.rs (lines 318-461)
  - 6 new tests covering all error variants
  - Verify error messages are tested exactly

  3. Call Sites (2 min)

  📁 diskann-disk/src/utils/math_util.rs (lines 164-204)
  - Three .map_err() conversions to ANNError

### DIFF
--- a/diskann-disk/src/utils/math_util.rs
+++ b/diskann-disk/src/utils/math_util.rs
@@ -172,7 +172,8 @@ pub fn compute_closest_centers_in_block(
         &ones_a,
         None, // Initialize the destination matrix
         dist_matrix,
-    );
+    )
+    .map_err(|e| ANNError::log_index_error(format_args!("{}", e)))?;
 
     diskann_linalg::sgemm(
         Transpose::None,
@@ -185,7 +186,8 @@ pub fn compute_closest_centers_in_block(
         centers_l2sq,
         Some(1.0), // Add to the destination matrix
         dist_matrix,
-    );
+    )
+    .map_err(|e| ANNError::log_index_error(format_args!("{}", e)))?;
 
     diskann_linalg::sgemm(
         Transpose::None,
@@ -198,7 +200,8 @@ pub fn compute_closest_centers_in_block(
         centers,
         Some(1.0), // Add to the destination matrix.
         dist_matrix,
-    );
+    )
+    .map_err(|e| ANNError::log_index_error(format_args!("{}", e)))?;
 
     if k == 1 {
         center_index
--- a/diskann-disk/src/utils/math_util.rs
+++ b/diskann-disk/src/utils/math_util.rs
@@ -12,7 +12,7 @@
 
 use std::{cmp::Ordering, collections::BinaryHeap};
 
-use diskann::{ANNError, ANNResult};
+use diskann::{ANNError, ANNErrorKind, ANNResult};
 use diskann_linalg::{self, Transpose};
 use diskann_providers::{
     forward_threadpool,
@@ -173,7 +173,7 @@ pub fn compute_closest_centers_in_block(
         None, // Initialize the destination matrix
         dist_matrix,
     )
-    .map_err(|e| ANNError::log_index_error(format_args!("{}", e)))?;
+    .map_err(|e| ANNError::new(ANNErrorKind::IndexError, e))?;
 
     diskann_linalg::sgemm(
         Transpose::None,
@@ -187,7 +187,7 @@ pub fn compute_closest_centers_in_block(
         Some(1.0), // Add to the destination matrix
         dist_matrix,
     )
-    .map_err(|e| ANNError::log_index_error(format_args!("{}", e)))?;
+    .map_err(|e| ANNError::new(ANNErrorKind::IndexError, e))?;
 
     diskann_linalg::sgemm(
         Transpose::None,
@@ -201,7 +201,7 @@ pub fn compute_closest_centers_in_block(
         Some(1.0), // Add to the destination matrix.
         dist_matrix,
     )
-    .map_err(|e| ANNError::log_index_error(format_args!("{}", e)))?;
+    .map_err(|e| ANNError::new(ANNErrorKind::IndexError, e))?;
 
     if k == 1 {
         center_index
--- a/diskann-linalg/src/lib.rs
+++ b/diskann-linalg/src/lib.rs
@@ -3,12 +3,79 @@
  * Licensed under the MIT license.
  */
 
+use std::fmt;
+
 pub mod common;
 pub use common::Transpose;
 
 mod faer;
 use faer::{random_distance_preserving_matrix_impl, sgemm_impl, svd_into_impl};
 use rand::Rng;
+
+/// Error type for SGEMM operations.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SgemmError {
+    /// Matrix `a` has incorrect dimensions.
+    InvalidMatrixADimensions {
+        expected_rows: usize,
+        expected_cols: usize,
+        expected_len: usize,
+        actual_len: usize,
+    },
+    /// Matrix `b` has incorrect dimensions.
+    InvalidMatrixBDimensions {
+        expected_rows: usize,
+        expected_cols: usize,
+        expected_len: usize,
+        actual_len: usize,
+    },
+    /// Matrix `c` has incorrect dimensions.
+    InvalidMatrixCDimensions {
+        expected_rows: usize,
+        expected_cols: usize,
+        expected_len: usize,
+        actual_len: usize,
+    },
+}
+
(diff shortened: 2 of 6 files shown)


diskann-linalg/src/lib.rs
```diff
@@ -3,13 +3,63 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SgemmError {
+    /// Matrix has incorrect dimensions.
+    InvalidMatrixDimensions {
+        matrix_name: &'static str,
+        expected_rows: usize,
+        expected_cols: usize,
+        expected_len: usize,
+        actual_len: usize,
+    },
```
hildebrandmw, line 42: This is a large error type (~48 bytes: 32 bytes for the `usize` values and 16 bytes for `&'static str`). This is mainly a concern when it gets rolled into higher level errors, such as `TransformFailed`. We can do a little bit to bring it down in size.

First, `expected_len` can be computed from `expected_rows` and `expected_cols`, so we can save 8 bytes there. Next, an enum can be used for `matrix_name`:
```rust
#[derive(Debug, Clone, Copy)]
pub enum MatrixName {
    A,
    B,
    C,
}

impl std::fmt::Display for MatrixName {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            Self::A => write!(f, "a"),
            Self::B => write!(f, "b"),
            Self::C => write!(f, "c"), 
        }
    }
}
```
This will drop the size of the error down to 32 bytes, which is still a little big, but better.

diskann-linalg/src/lib.rs
```diff
@@ -189,15 +260,178 @@ mod tests {
...
+            n: usize,
+            k: usize,
+            alpha: f32,
+            a: &[f32],
+            b: &[f32],
+            beta: Option<f32>,
+            c: &mut [f32],
+        ) {
+            sgemm(atranspose, btranspose, m, n, k, alpha, a, b, beta, c).unwrap();
+        }
```
hildebrandmw: I wonder if we should update the signature of `TestProblem::check` to be `Result<(), SgemmError>` for uniformity. If you don't want to repeat all the checking logic, a trick I've used in the past is something like this:
```rust
impl SgemmError {
    fn check(m: usize, n: usize, ...) -> Result<(), Self>;
}
```
Then everything preflight with
```rust
fn sgemm(...) -> Result<(), SgemmError> {
    SgemmError::check(args...)?;
    sgemm_impl(...);
    Ok(())
}
```

diskann-quantization/src/algorithms/transforms/utils.rs
```diff
@@ -8,14 +8,17 @@ use thiserror::Error;
...
+#[derive(Debug, Clone, Error, PartialEq)]
 pub enum TransformFailed {
     #[error("incorrect transform input vector - expected length {expected} but got {found}")]
     SourceMismatch { expected: usize, found: usize },
     #[error("incorrect transform output vector - expected length {expected} but got {found}")]
     DestinationMismatch { expected: usize, found: usize },
     #[error(transparent)]
     AllocatorError(#[from] AllocatorError),
+    #[cfg(feature = "linalg")]
+    #[error("SGEMM operation failed: {0}")]
```
hildebrandmw: You don't need to have `{0}` formatted here. The `#[from]` tag of [thiserror](https://docs.rs/thiserror/latest/thiserror/) automatically implements [`Error::source`](https://doc.rust-lang.org/core/error/trait.Error.html#method.source) for the inner error. Using `ANNError` or  quantization's [`format`](https://github.com/microsoft/DiskANN/blob/f458cf613539a57d17f814da3fc8ac8388db1cdd/diskann-quantization/src/error.rs#L24-L51) will automatically walk the source chain.

diskann-disk/src/utils/math_util.rs
```diff
@@ -185,7 +186,8 @@ pub fn compute_closest_centers_in_block(
         centers_l2sq,
         Some(1.0), // Add to the destination matrix
         dist_matrix,
-    );
+    )
+    .map_err(|e| ANNError::log_index_error(format_args!("{}", e)))?;
```
hildebrandmw: For future reference, you may want to prefer `ANNError::new(ANNErrorKind::IndexError, e)` instead of the old `log*` style constructors. The former defers any kind of string formatting until the entire error is formatter higher in the call stack, where-as the latter forces eager string evaluation lower in the callstack. Not that it matters *that* much though.

hildebrandmw (commented):
Thanks Alex - just a few general error handling comments from my end.

diskann-linalg/src/lib.rs
arrayka: Heh, that was my original plan, but I decided that I'm overengineering it.
Fixed.

diskann-quantization/src/algorithms/transforms/utils.rs
arrayka: Done.

diskann-disk/src/utils/math_util.rs
arrayka: Done + created a task to remove the old log-style constructors: #1003

harsha-simhadri (approved):

diskann-linalg/src/lib.rs
arrayka: I've changed signature of `check` to return `Result<(), CheckError>`, where CheckError is a newly introduced enum:
```
#[derive(Debug, Error)]
pub(crate) enum CheckError {
    #[error("mismatch in test problem. got {:?}, expected {:?}", got, expected)]
    ReferenceError { got: Vec<f32>, expected: Vec<f32> },
    #[error(transparent)]
    SgemmError(#[from] crate::SgemmError),
}
```

diskann-linalg/src/lib.rs
hildebrandmw: Yeah. The only reason I bring it up is because `diskann-quantization` uses heavily nested enums for its error handling due to a user's request to keep allocations through a global allocator to a minimum. And in that context, sizes of error types really starts to add up.

hildebrandmw (approved):

diskann-disk/src/utils/math_util.rs
hildebrandmw: Thanks!

diskann-linalg/src/lib.rs
hildebrandmw: Thanks!

PR #998, opened 2026-04-29
Cleaning up test work and changing the get_degree_stats signature.

JordanMaples:
This is a quick cleanup PR for some of the changes that went in the other day. 

In test/cases/index.rs there were a bunch of paged_search tests that on the whole are lower signal than the baseline-related paged_search tests added in a different PR, so those are being removed.

tests for the QueryLabelProvider were moved to the trait defn location for additional documentation purposes on how we intend the trait to be used. 

In the initial test authoring, I added an IntoIterator trait implementation to the test provider in the goals of testing get_degree_stats. After syncing with Mark, we agreed that the better approach was to change the way that get_degree_stats behaves, now it accepts an iterator to the ids that you want the degree stats for rather than forcing the provider to implement IntoIterator. 

There are still some areas that I'm cleaning up. In particular: helpers.rs, while it does help to set up some tests, naming is not the best and there are still some layers of boilerplate that can be eliminated by spending a bit more time handcrafting a solution.

### DIFF
--- a/diskann-bf_tree-provider/test_data
+++ b/diskann-bf_tree-provider/test_data
@@ -1,1 +1,0 @@
-/home/jordan/diskann/test_data
--- a/diskann-bf_tree-provider/test_data
+++ b/diskann-bf_tree-provider/test_data
@@ -1,0 +1,1 @@
+/home/jordan/diskann/test_data
--- a/diskann-providers/src/index/diskann_async.rs
+++ b/diskann-providers/src/index/diskann_async.rs
@@ -3308,7 +3308,7 @@ pub(crate) mod tests {
             .await
             .unwrap();
         let mut accessor_unsat = inmem::FullAccessor::new(index_unsat.provider());
-        let res_unsat = index_sat
+        let res_unsat = index_unsat
             .get_degree_stats(&mut accessor_unsat, index_unsat.provider().iter())
             .await
             .unwrap();
--- a/diskann-providers/src/index/diskann_async.rs
+++ b/diskann-providers/src/index/diskann_async.rs
@@ -3006,7 +3006,10 @@ pub(crate) mod tests {
 
         let neighbor_accessor = &mut index.provider().neighbors();
         // check that we have an unpruned graph
-        let stats = index.get_degree_stats(neighbor_accessor).await.unwrap();
+        let stats = index
+            .get_degree_stats(neighbor_accessor, index.provider().iter())
+            .await
+            .unwrap();
         assert!(stats.max_degree.into_usize() > max_degree);
 
         // prune graph and check that max_degree is respected
@@ -3014,7 +3017,10 @@ pub(crate) mod tests {
             .prune_range(&FullPrecision, ctx, 0..256)
             .await
             .unwrap();
-        let stats = index.get_degree_stats(neighbor_accessor).await.unwrap();
+        let stats = index
+            .get_degree_stats(neighbor_accessor, index.provider().iter())
+            .await
+            .unwrap();
         assert!(stats.max_degree.into_usize() <= max_degree);
     }
 
@@ -3293,14 +3299,17 @@ pub(crate) mod tests {
             .await
             .unwrap();
         let mut accessor_sat = inmem::FullAccessor::new(index_sat.provider());
-        let res_sat = index_sat.get_degree_stats(&mut accessor_sat).await.unwrap();
+        let res_sat = index_sat
+            .get_degree_stats(&mut accessor_sat, index_sat.provider().iter())
+            .await
+            .unwrap();
 
         let index_unsat = create_retry_saturated_index(NonZeroU32::new(1).unwrap(), false)
             .await
             .unwrap();
         let mut accessor_unsat = inmem::FullAccessor::new(index_unsat.provider());
         let res_unsat = index_sat
-            .get_degree_stats(&mut accessor_unsat)
+            .get_degree_stats(&mut accessor_unsat, index_unsat.provider().iter())
             .await
             .unwrap();
         assert!(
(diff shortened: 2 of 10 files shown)


diskann/src/graph/test/cases/index.rs
```diff
@@ -3,56 +3,24 @@
...
-        filter.on_visit(Neighbor::new(1, 1.0)),
-        QueryVisitDecision::Reject
-    ));
-}
+verbose_eq!(DegreeStats {
+    max_degree,
+    avg_degree,
+    min_degree,
+    cnt_less_than_two,
+});
```
hildebrandmw: This should probably live right next to `DegreeStats` under a `#[cfg(test)]` for better locality.

diskann/src/graph/index.rs
```diff
@@ -95,6 +95,28 @@ pub trait QueryLabelProvider<V: VectorId>: std::fmt::Debug + Send + Sync {
     }
 }
 
+#[test]
+fn query_label_provider_on_visit_default() {
```
hildebrandmw: Thanks for moving this here. Can we move this into a `mod tests` at the bottom of the file instead?

JordanMaples: yeah i can do that.

diskann/src/graph/test/cases/index.rs
JordanMaples: I put it in the mod tests section at the bottom of index.rs. let me know if you'd rather I bring it closer to the definition

hildebrandmw: ~~Nah, bottom is fine (and is the usual location).~~

Sorry, wrong context. Yeah, please put it somewhere really close to the definition. You can guard it with
```rust
#[cfg(test)]
verbose_eq!(DegreeStats { .. })
```
to keep it from leaking into the "public domain".

harsha-simhadri (approved):

diskann/src/graph/index.rs
```diff
@@ -2504,7 +2516,7 @@ where
             let mut total_live_points = 0;
 
             let mut neighbors = AdjacencyList::with_capacity(self.max_degree_with_slack());
-            for id in &self.data_provider {
+            for id in itr {
                 total_live_points += 1;
                 accessor.get_neighbors(id, &mut neighbors).await?;
```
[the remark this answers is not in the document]
JordanMaples: will add special case handling for divide by zero

diskann-providers/src/index/diskann_async.rs
```diff
@@ -3290,14 +3296,17 @@ pub(crate) mod tests {
...
 
         let index_unsat = create_retry_saturated_index(NonZeroU32::new(1).unwrap(), false)
             .await
             .unwrap();
         let mut accessor_unsat = inmem::FullAccessor::new(index_unsat.provider());
         let res_unsat = index_sat
-            .get_degree_stats(&mut accessor_unsat)
+            .get_degree_stats(&mut accessor_unsat, index_unsat.provider().iter())
             .await
             .unwrap();
```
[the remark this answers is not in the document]
JordanMaples: I'll update this to call against index_unsat to correct the existing test behavior.

diskann/src/graph/index.rs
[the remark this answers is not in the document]
JordanMaples: if this is a problem for callers, they can materialize the vec when they need to do it

diskann/src/graph/index.rs
```diff
@@ -2491,9 +2499,13 @@ where
...
+        &self,
+        accessor: &mut NA,
+        itr: Itr,
+    ) -> impl SendFuture<ANNResult<DegreeStats>>
     where
-        for<'a> &'a DP: IntoIterator<Item = DP::InternalId, IntoIter: Send>,
+        Itr: IntoIterator<Item = DP::InternalId, IntoIter: Send> + Send,
         NA: AsNeighbor<Id = DP::InternalId>,
     {
         async move {
```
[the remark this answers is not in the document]
JordanMaples: if this is a problem for callers, they can materialize the vec when they need to do it

diskann/src/graph/test/cases/index.rs
```diff
@@ -108,20 +69,25 @@ async fn test_count_unreachable_isolated_nodes() {
...
+        .get_degree_stats(&mut accessor, index.provider().non_start_points_ids())
+        .await
+        .unwrap();
+    let expected = DegreeStats {
+        max_degree: 2,
+        avg_degree: 2.0,
+        min_degree: 2,
+        cnt_less_than_two: 0,
+    };
+    assert!(stats.verbose_eq(&expected).is_ok());
```
hildebrandmw: I think there's an `assert_vector_eq!` floating around here that will actually print out the reason for inequality on a mismatch.

diskann/src/graph/test/cases/index.rs
```diff
@@ -195,199 +161,26 @@ async fn test_get_degree_stats_mixed() {
...
-        .next_search_results(&ctx, &mut state, 2, &mut small_output)
-        .await;
-    assert!(result.is_err());
+    let expected = DegreeStats {
+        max_degree: 2,
+        avg_degree: 1.25,
+        min_degree: 0,
+        cnt_less_than_two: 2,
+    };
+    assert!(stats.verbose_eq(&expected).is_ok());
```
hildebrandmw: See other note about `assert_verbose_eq!`.

hildebrandmw (commented):
Just a small couple requests - otherwise looks good. Thanks!

diskann/src/graph/test/cases/index.rs
JordanMaples: switched to `assert_eq_verbose`

diskann/src/graph/test/cases/index.rs
JordanMaples: see other

diskann-bf_tree-provider/test_data
```diff
@@ -0,0 +1 @@
+/home/jordan/diskann/test_data
```
hildebrandmw: Is this a dev artifact?

diskann-providers/test_data
```diff
@@ -0,0 +1 @@
+/home/jordan/diskann/test_data
```
hildebrandmw: Test artifact?

hildebrandmw (commented):
Couple dev artifacts I think. Otherwise, look good!

hildebrandmw (approved):

PR #1011, opened 2026-05-04
(diversity) Integrate determinant diversity in tokp full precision search and disk search

### LINKED ISSUE #909: Implement determinant diversity search

narendatha:
### Summary

Wires the determinant-diversity post-processing algorithm as an optional `SearchPlugin` for two search paths:

- **Async graph-index** (full-precision topk via `inmem::FullAccessor`)
- **Disk-index** (via `DiskIndexSearcher` post-processor)

Determinant-diversity reranks search results to increase geometric diversity among the top-k, at a configurable quality/speed tradeoff.

## Motivation

Standard nearest neighbor search often returns highly similar/redundant results. For some scenarios like RAG search, it's more valuable to retrieve diverse, complementary documents that cover different aspects of the query. This PR implements a principled approach based on determinant maximization of the Gram matrix of query-scaled vectors.

## Algorithm

The algorithm scales each candidate vector by `(similarity_to_query)^power` and then greedily selects vectors to maximize the determinant of the Gram matrix (equivalent to maximizing the volume of the parallelepiped spanned by the vectors).

Two variants are implemented:
- **eta = 0**: Pure greedy orthogonalization (pivoted QR decomposition)
- **eta > 0**: Ridge-regularized version that balances diversity with numerical stability

Both variants run in **O(n  k  d)** time where n = candidates, k = results to return, d = dimension.

## Performance Characteristics

Tested on OpenAI embeddings dataset (3072 dimensions, 10K sample):

| Metric | Normal Search | Det_Div Search | Notes |
|--------|---------------|------------|-------|
| QPS | ~150 | ~130 | ~13% overhead for diversity optimization |
| Recall@10 | 98-99% | 87-88% | Expected trade-off for diversity |

The recall reduction is expected - Det_Div search intentionally selects diverse vectors rather than the absolute nearest neighbors.

### Diversity Improvement
We measure diversity using log-determinant of the Gram matrix of scaled vectors (higher = vectors span more volume = more diverse).

- RAG search consistently improves diversity (all improvements >= 0)
- Average improvement of +2.9 to +3.3 in log_det corresponds to **10-27x more volume** spanned by selected vectors

### `power` (default: 2.0)
Controls the trade-off between relevance and diversity:
- **Higher values (e.g., 3.0+)**: Strongly prefer relevance - vectors closer to the query get much higher weight, resulting in less diversity but higher recall
- **Lower values (e.g., 1.0)**: Prefer diversity over relevance - similarity weighting is more uniform, allowing more diverse vectors to be selected
- **Value of 0**: Pure diversity maximization with no relevance weighting

### `eta` (default: 0.01)
Ridge regularization parameter that controls numerical robustness and relevance preference:
- **Higher values (e.g., 0.1+)**: More robust numerically, but biases toward selecting high-relevance vectors over diverse ones
- **Lower values (e.g., 0.001)**: Closer to pure determinant maximization, maximizes diversity but may be less stable
- **Value of 0**: Falls back to pure greedy orthogonalization (no ridge regularization)

### Recommended Settings
| Use Case | `power` | `eta` |
|----------|-------------|-----------|
| Balanced (default) | 2.0 | 0.01 |
| High diversity | 1.0 | 0.001 |
| High relevance | 3.0 | 0.1 |

#### Core algorithm (`diskann-providers`)
- Added `determinant_diversity_post_process()` in `diskann-providers/src/model/graph/provider/async_/determinant_diversity_post_process.rs`
- Two selection strategies controlled by `eta`:
  - `eta == 0`: greedy orthogonalization
  - `eta > 0`: eta-weighted residual selection

#### Async graph-index path (`diskann-benchmark`)
- Added `FullPrecisionVectorAccessor` shim trait + impl for `inmem::FullAccessor<f32, ...>` so the post-processor can fetch real candidate vectors
- Implemented `DeterminantDiversity` plugin that satisfies `glue::SearchPostProcess` using `for<'a, 'b> SearchStrategy` HRTB bound (credit: @hildebrandmw)

#### Disk-index path (`diskann-disk`, `diskann-benchmark`)
- Added `DeterminantDiversityAndFilter` processor to `DiskSearchPostProcessor` enum
- Extended `searcher.search()` to accept an optional `SearchPostProcessorKind::DeterminantDiversity { power, eta }`
- Extended `DiskSearchPhase` JSON input to accept `post_processor`
- Result-counting uses `result_count` from search stats to correctly bound post-processed results

#### Input/config
- New `TopkPostProcessor` serde enum in `diskann-benchmark/src/inputs/post_processor.rs` with validation (`power > 0`, `eta >= 0`)
- Two new example configs: `async-determinant-diversity.json`, `disk-index-determinant-diversity.json`

### Benchmark JSON examples

```json
// async graph-index
"post_processor": { "type": "determinant-diversity", "power": 1.0, "eta": 0.0 }

// disk-index
"post_processor": { "type": "determinant-diversity", "power": 1.0, "eta": 0.5 }
```

### Notes

- Determinant-diversity is only wired for **full-precision topk** on the async path; quantized (PQ/scalar) paths are intentionally excluded
- Spherical, range, and filtered-search plugin variants are unchanged; this post-processor is topk-only


diskann-benchmark/src/backend/index/search/plugins.rs
```diff
@@ -145,9 +149,37 @@ where
...
+impl DeterminantDiversity {
+    pub(crate) fn is_match(phase: &SearchPhase) -> bool {
+        phase
+            .as_topk()
+            .ok()
+            .and_then(|topk| topk.post_processor.as_ref())
+            .is_some_and(|pp| matches!(pp, TopkPostProcessor::DeterminantDiversity { .. }))
+    }
+
+    pub(crate) fn get(phase: &SearchPhase) -> anyhow::Result<(&TopkSearchPhase, f32, f32)> {
```
hildebrandmw: Can we return a richer struct rather than to `f32`? Ideally - the parsed code would validate that a determinant diversity post-processor can be constructed from the parameters (using the actual constructor).

diskann-benchmark/src/backend/index/benchmarks.rs
```diff
@@ -447,17 +451,182 @@ impl<S> Strategy<S> {
 // Topk //
 //------//
 
+/// Execute a topk search with a custom per-query search function, collecting timing and recall.
+///
+/// Encapsulates the outer thread/run/L loops, the per-rep timing harness, per-query latency
+/// collection, and [`SearchResults`] construction. The caller supplies only the actual
+/// per-query search as a closure `search_fn(knn_params, query, output) -> Result<()>`.
+fn run_topk_timed(
```
hildebrandmw: The code in `diskann-benchmark-core` abstracting the loop nest exists for several reasons:

1. It keeps the behavior more or less the same for all users.
2. More importantly, it very carefully minimizes monomorphizations to help keep compile times under control.

Introducing bespoke implementations like this can very quickly get us back to a place where benchmarks take forever to compile and the set of metrics gathered by different search kinds diverges.

I understand that the `KNN` runner in `diskann-benchmark-core` predates parametrizable post-process routines, but we should update it with an additional defaulted generic. I understand that's more work, but it improves the core abstraction and will pay dividends when it comes to exploring additional post-processing routines.

diskann-benchmark/src/backend/index/benchmarks.rs
```diff
@@ -447,17 +451,182 @@ impl<S> Strategy<S> {
...
+        let strategy = common::FullPrecision;
+        let context = DefaultContext;
+        let det_div = post_processor::DeterminantDiversity::new(power, eta);
+
+        let queries = Arc::new(datafiles::load_dataset::<f32>(datafiles::BinFile(
+            &topk.queries,
+        ))?);
+        let groundtruth = datafiles::load_groundtruth(datafiles::BinFile(&topk.groundtruth))?;
+
+        let results = run_topk_timed(topk, &queries, &groundtruth, |params, query, output| {
```
hildebrandmw: Case and point as to why we should use `diskann-benchmark-core`: This runs all the searches single-threaded, but the loop nest loops over the number of threads. Not only will this make the benchmark take forever to run, it's also misleading and means that we're measuring overhead related to `tokio` runtime construction.

diskann-benchmark/src/backend/index/benchmarks.rs
```diff
@@ -447,17 +451,182 @@ impl<S> Strategy<S> {
...
     S: for<'a> glue::DefaultSearchStrategy<DP, &'a [DP::Element]> + Clone + AsyncFriendly,
 {
     fn is_match(&self, phase: &SearchPhase) -> bool {
-        Self::kind() == phase.kind()
+        plugins::Topk::is_match(phase)
     }
 
     fn kind(&self) -> &'static str {
-        Self::kind().as_str()
+        SearchPhaseKind::Topk.as_str()
```
hildebrandmw: We should probably have the plugin types implement an `as_str()` method instead to make sure they stay up-to-date properly and completely avoid `SearchPhaseKind` here.

diskann-benchmark/src/inputs/post_processor.rs
```diff
@@ -0,0 +1,29 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "kebab-case")]
+pub(crate) enum TopkPostProcessor {
+    DeterminantDiversity { power: f32, eta: f32 },
+}
+
+impl CheckDeserialization for TopkPostProcessor {
+    fn check_deserialization(&mut self, _checker: &mut Checker) -> Result<(), anyhow::Error> {
+        match self {
+            TopkPostProcessor::DeterminantDiversity { power, eta } => {
```
hildebrandmw: Shouldn't we have one struct that validates the inputs in the lower level crate and always use that for input validation? That way, there's no risk of these ad-hoc validations ever being wrong.

diskann-providers/src/model/graph/provider/async_/determinant_diversity_post_process.rs
```diff
@@ -0,0 +1,359 @@
...
+    determinant_diversity_eta: f32,
+    determinant_diversity_power: f32,
+) -> Vec<(Id, f32)> {
+    if candidates.is_empty() || query.is_empty() {
+        return Vec::new();
+    }
+
+    let candidates: Vec<_> = candidates
+        .into_iter()
+        .filter(|(_, _, vector)| vector.len() == query.len())
```
hildebrandmw: Should this not be a hard pre-requisite of this function instead of silently ignoring vectors?

diskann-providers/src/model/graph/provider/async_/determinant_diversity_post_process.rs
```diff
@@ -0,0 +1,359 @@
...
+    let distance_range = {
+        let mut min_distance = f32::INFINITY;
+        let mut max_distance = f32::NEG_INFINITY;
+
+        for (_, distance, _) in &candidates {
+            min_distance = min_distance.min(*distance);
+            max_distance = max_distance.max(*distance);
+        }
+
+        (min_distance, max_distance)
```
hildebrandmw: Nit: This should probably stay as either two independently names values, or as a helper struct. A `tuple` is not a particularly intuitive representation.

diskann-providers/src/model/graph/provider/async_/determinant_diversity_post_process.rs
```diff
@@ -0,0 +1,359 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+use diskann_vector::{MathematicalValue, PureDistanceFunction, distance::InnerProduct};
```
hildebrandmw: A module level doc explaining what this is and how it works would be really helpful.

diskann-providers/src/model/graph/provider/async_/determinant_diversity_post_process.rs
```diff
@@ -0,0 +1,359 @@
...
+}
+
+#[inline]
+fn dot_product(a: &[f32], b: &[f32]) -> f32 {
+    <InnerProduct as PureDistanceFunction<&[f32], &[f32], MathematicalValue<f32>>>::evaluate(a, b)
+        .into_inner()
+}
+
+#[cfg(test)]
+mod tests {
```
hildebrandmw: Should we have tests that check the algorithmic behavior?

diskann-providers/src/model/graph/provider/async_/determinant_diversity_post_process.rs
```diff
@@ -0,0 +1,359 @@
...
+    selected
+        .iter()
+        .map(|&idx| {
+            let (id, dist, _) = &candidates[idx];
+            (*id, *dist)
+        })
+        .collect()
+}
+
+fn post_process_greedy_orthogonalization_f32<Id: Copy>(
```
hildebrandmw: These two routines are very similar. Is it possible to merge them?

diskann-providers/src/model/graph/provider/async_/determinant_diversity_post_process.rs
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    let inv_sqrt_eta = 1.0 / eta.sqrt();
+    let mut residuals = Vec::with_capacity(n);
+    let mut norms_sq = Vec::with_capacity(n);
+
+    for (_, distance_to_query, v) in &candidates {
+        let scale =
+            distance_to_similarity(*distance_to_query, distance_range).powf(power) * inv_sqrt_eta;
+        let residual: Vec<f32> = v.iter().map(|&x| x * scale).collect();
```
hildebrandmw: Maybe use a `Matrix` to store the residuals? This will provide better locality in the processing loop and cut down on the number of allocations.

diskann-providers/src/model/graph/provider/async_/mod.rs
```diff
@@ -7,7 +7,9 @@ pub mod experimental;
 pub mod common;
 pub use common::{PrefetchCacheLineLevel, StartPoints, VectorGuard};
 
+mod determinant_diversity_post_process;
 pub(crate) mod postprocess;
+pub use determinant_diversity_post_process::determinant_diversity_post_process;
```
hildebrandmw: Since this is shared with the disk index, maybe move it up instead of down in the `async_` abyss? I know `diskann-providers` is kind of a nightmare at the moment 😢

hildebrandmw (commented):
Thanks Naren! My biggest concern is `diskann-benchmark`. I would really prefer if we generalized `diskann-benchmark-core` to handle post-processing, which would reduce bugs and be more broadly applicable.

On the implementation front, there's a bit of work to do regarding being more opinionated on the nature of inputs, establishing invariants of the diversity search with a bespoke struct (which could help with the previous point), reducing code duplication, testing the implementation, and removing allocations with a `Matrix`.

I'll leave reviewing the `diskann-disk` changes to others.

diskann-providers/src/model/graph/provider/async_/determinant_diversity_post_process.rs
narendatha: done, now it allocates residuals with Matrix::new. now it is replaced by contiguous matrix-based storage.

diskann-providers/src/model/graph/provider/async_/mod.rs
narendatha: good point. now the implementation lives at determinant_diversity.rs. async_ only re-exports it.

diskann-providers/src/model/graph/provider/async_/determinant_diversity_post_process.rs
narendatha: yes, added more tests for algo behavior.

diskann-providers/src/model/graph/provider/async_/determinant_diversity_post_process.rs
narendatha: good catch. implemented greedy_orthogonal_select as unified implementation replaces two nearly-identical functions

diskann-providers/src/model/graph/provider/async_/determinant_diversity_post_process.rs
narendatha: added module level doc

diskann-benchmark/src/backend/index/search/plugins.rs
narendatha: done, created a struct to hold these, it makes code easier to read.

diskann-benchmark/src/backend/index/benchmarks.rs
narendatha: yes, this was a blunder when I look at it with fresh eyes. The fix for previous comment to use an additional defaulted generics solves this problem also.

diskann-benchmark/src/inputs/post_processor.rs
narendatha: done

diskann-providers/src/model/graph/provider/async_/determinant_diversity_post_process.rs
narendatha: added an assert to verify this and also changed the test to verify the panic.

diskann-providers/src/model/graph/provider/async_/determinant_diversity_post_process.rs
narendatha: added a struct that makes the code more readable.

diskann-benchmark/src/backend/index/benchmarks.rs
narendatha: yes, adding this code and the code duplication was very bad idea. Sometimes I over-rely on AI. I am being more cautious with AI code now. Updated the KNN runner with adidtional defaulted generic. it does improve core abstraction.

diskann-benchmark/src/backend/index/benchmarks.rs
narendatha: added as_str methods for plugin types.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
```diff
@@ -0,0 +1,430 @@
...
+use diskann_vector::{MathematicalValue, PureDistanceFunction, distance::InnerProduct};
+
+#[derive(Clone, Copy)]
+struct DistanceRange {
+    min: f32,
+    max: f32,
+}
+
+pub fn determinant_diversity_post_process<Id: Copy>(
+    candidates: Vec<(Id, f32, Vec<f32>)>,
```
arkrishn94: Should we force the input to be a fully allocated set of candidate vectors? Given the algorithm is greedy and we're potentially using this in a overfetch setting, it might make sense to allow more allocation patterns here (like a lazily evaluated iterator for e.g.)?

diskann-providers/src/model/graph/provider/determinant_diversity.rs
```diff
@@ -0,0 +1,430 @@
...
+    if candidates.is_empty() || query.is_empty() {
+        return Vec::new();
+    }
+
+    assert!(
+        candidates
+            .iter()
+            .all(|(_, _, vector)| vector.len() == query.len()),
+        "all candidate vectors must have the same dimension as query"
+    );
```
arkrishn94: Does it make sense to throw errors here instead of panicking?

diskann-providers/src/model/graph/provider/determinant_diversity.rs
```diff
@@ -0,0 +1,430 @@
...
+    min: f32,
+    max: f32,
+}
+
+pub fn determinant_diversity_post_process<Id: Copy>(
+    candidates: Vec<(Id, f32, Vec<f32>)>,
+    query: &[f32],
+    k: usize,
+    determinant_diversity_eta: f32,
+    determinant_diversity_power: f32,
```
arkrishn94: Why not have the correct type for these two parameters so we can be sure that `eta >= 0.0` and `power > 0.0`? You can use `Positive<f32>` for `power` but might need to invent some type for `eta`.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
```diff
@@ -0,0 +1,430 @@
...
+/// out previously selected candidates. The `inv_sqrt_eta` parameter controls the
+/// ridge-regularization scaling:
+///
+/// - `inv_sqrt_eta = 1.0`: exact greedy orthogonalization (eta=0 case)
+/// - `inv_sqrt_eta = 1/sqrt(eta)`: ridge-regularized variant for numerical stability
+fn greedy_orthogonal_select<Id: Copy>(
+    candidates: Vec<(Id, f32, Vec<f32>)>,
+    k: usize,
+    power: f32,
+    inv_sqrt_eta: f32,
```
arkrishn94, line 345: nit: same here on making this typed.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
```diff
@@ -0,0 +1,430 @@
...
+//! # Time Complexity
+//!
+//! O(m³) where m is the number of candidates, due to determinant computation.
+//! In practice, m is typically small (search returns hundreds of candidates,
+//! but only top-k ≪ m are selected).
+//!
+//! # References
+//!
+//! The algorithm is based on diversity-promoting ranking methods for nearest neighbor search,
+//! as used in approximate nearest neighbor indices like DiskANN.
```
arkrishn94, line 52: The math here is reasonably sophisticated. Can we please document this carefully so a reader can follow exactly what is happening in `greedy_orthogonal_select` especially?

diskann-providers/src/model/graph/provider/determinant_diversity.rs
```diff
@@ -0,0 +1,430 @@
...
+    for (i, (_, distance_to_query, v)) in candidates.iter().enumerate() {
+        let scale =
+            distance_to_similarity(*distance_to_query, distance_range).powf(power) * inv_sqrt_eta;
+        let row = residuals.row_mut(i);
+        for (r, &x) in row.iter_mut().zip(v.iter()) {
+            *r = x * scale;
+        }
+        let norm_sq = dot_product(residuals.row(i), residuals.row(i));
+        norms_sq.push(norm_sq);
+    }
```
arkrishn94: Is there value in using a more generalized API here? It might help bring in optimizations easily - e.g. with vectorized building blocks.

arkrishn94 (commented):
Thanks Naren, I am glad this feature is starting to land in the repo. Left a few small comments here but my main question is whether `greedy_orthogonal_select` should be written in a way that is more pluggable and generalizable - 
1. Allocation and format of full precision vectors. Isn't `Vec<(Id, Vec<f32>)>` too restrictive?
2. Should we lean into the matrix API more and its optimized implementations in `diskann-quantization`?

diskann-providers/src/model/graph/provider/determinant_diversity.rs
narendatha: Thanks for pointing out. I prefer to Use Result (error return) for expected or recoverable errors, and use panic! for unrecoverable errors or bugs. It is a bug if someone is sending a query that doesn't match vector dim. we can debate this if you have strong preference.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
narendatha: Thanks, used Positive<f32> for power. Leaving eta as-is now. The algo could evolve and will revist if needed.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
narendatha: Good point. This allocation problem and handing types beyond f32, I will handle in follow up PR.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
narendatha: added comments with algo descirption.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
narendatha: since this is internal fn, not `pub fn`, I am skipping adding typing here.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
narendatha: Good point. Will look more into it and address it in future PRs. This design is good enough for experimentations for now.

diskann-providers/src/model/graph/provider/async_/mod.rs
```diff
@@ -8,9 +8,10 @@ pub mod common;
 pub use common::{PrefetchCacheLineLevel, StartPoints, VectorGuard};
 
 pub(crate) mod postprocess;
-
+// Re-export from parent module for backward compatibility.
+// The algorithm is not async-specific and lives in provider::determinant_diversity.
```
hildebrandmw: What backwards compatibility are we trying to maintain? This PR is adding the feature, no?

Please don't reexport items from places that aren't proper children. I know that `diskann-providers` is not structured well at all, but let's not make it any worse.

diskann-benchmark-core/src/search/graph/knn.rs
```diff
@@ -86,7 +142,14 @@ pub struct Metrics {
     pub hops: u32,
 }
 
-impl<DP, T, S> Search for KNN<DP, T, S>
+impl Metrics {
+    /// Construct a new metrics value.
+    pub fn new(comparisons: u32, hops: u32) -> Self {
+        Self { comparisons, hops }
+    }
+}
```
hildebrandmw: This constructor is not needed.

diskann-benchmark-core/src/search/graph/knn.rs
```diff
@@ -23,7 +23,7 @@ use crate::{
 };
 
 /// A built-in helper for benchmarking the K-nearest neighbors method
```
hildebrandmw, line 25: Here is an alternative that doesn't require duplicating the search loop.
```rust
#[derive(Debug, Clone, Copy)]
pub struct Defaulted;

pub trait AsPostProcessor<'a, S, DP, T>
where
    DP: provider::DataProvider,
    S: glue::SearchStrategy<'a, DP, T>,
{
    type Processor: glue::SearchPostProcess<S::SearchAccessor, T, DP::ExternalId> + Send + Sync;

    fn as_post_processor(&'a self, strategy: &'a S) -> Self::Processor;
}

impl<'a, S, DP, T> AsPostProcessor<'a, S, DP, T> for Defaulted
where
    DP: provider::DataProvider,
    S: glue::DefaultPostProcessor<'a, DP, T, DP::ExternalId>,
{
    type Processor = S::Processor;

    fn as_post_processor(&'a self, strategy: &'a S) -> Self::Processor {
        strategy.default_post_processor()
    }
}

#[derive(Debug)]
pub struct Forwarded<PP>(PP);

impl<'a, S, DP, T, PP> AsPostProcessor<'a, S, DP, T> for Forwarded<PP>
where
    DP: provider::DataProvider,
    S: glue::SearchStrategy<'a, DP, T>,
    PP: glue::SearchPostProcess<S::SearchAccessor, T, DP::ExternalId> + Clone + AsyncFriendly,
{
    type Processor = PP;

    fn as_post_processor(&'a self, _strategy: &'a S) -> Self::Processor {
        self.0.clone()
    }
}

// ...
impl<DP, T, S, PP> Search for KNN<DP, T, S, PP>
where
    DP: provider::DataProvider<Context: Default, ExternalId: search::Id>,
    S: for<'a> glue::SearchStrategy<'a, DP, &'a [T]> + Clone + AsyncFriendly,
    PP: for<'a> AsPostProcessor<'a, S, DP, &'a [T]> + AsyncFriendly,
    T: AsyncFriendly + Clone,
{
    // ...
        let strategy = self.strategy.get(index)?;
        let post_processor = self.post_processor.as_post_processor(&strategy);
        let stats = self
            .index
            .search_with(
                knn_search,
                strategy,
                post_processor,
                &context,
                self.queries.row(index),
                buffer,
            )
   //...
}
```
Would prefer an approach like this that is mostly infallible.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -264,12 +265,37 @@ pub struct RerankAndFilter<'a> {
     filter: &'a (dyn Fn(&u32) -> bool + Send + Sync),
 }
 
+#[derive(Clone, Copy)]
+pub struct DeterminantDiversityAndFilter<'a> {
+    filter: &'a (dyn Fn(&u32) -> bool + Send + Sync),
+    power: f32,
+    eta: f32,
```
hildebrandmw: Why not reuse `DeterminantDiversityParams `?

diskann-providers/src/post_processor.rs
```diff
@@ -0,0 +1,137 @@
...
+impl DeterminantDiversityParams {
+    /// Create and validate new Determinant-Diversity parameters.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if validation fails:
+    /// - `power <= 0.0`: invalid relevance weighting
+    /// - `eta < 0.0`: invalid numerical stability parameter
+    pub fn new(power: f32, eta: f32) -> Result<Self, DeterminantDiversityError> {
+        if power <= 0.0 {
```
hildebrandmw: Adding `.is_finite()` checks wouldn't hurt.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
```diff
@@ -0,0 +1,693 @@
...
+//! - **eta**: Numerical stability parameter (must be >= 0.0). Used for ridge regularization:
+//!   - `eta = 0`: Exact determinant computation (can be numerically unstable for some inputs)
+//!   - `eta > 0`: Ridge-regularized computation for improved numerical stability
+//!
+//! # Variants
+//!
+//! The module provides two implementations:
+//!
+//! - `post_process_with_eta_f32()`: Uses ridge regularization for numerical stability
+//! - `post_process_without_eta_f32()`: Computes exact determinants (faster but less stable)
```
hildebrandmw: Nit: these are now stale.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -878,7 +1034,7 @@ where
     /// Perform a raw search on the disk index.
     /// This is a lower-level API that allows more control over the search parameters and output buffers.
     #[allow(clippy::too_many_arguments)]
-    pub(crate) fn search_internal(
+    pub fn search_internal(
```
hildebrandmw: Why did this get promoted to `public`?

diskann-benchmark/src/index/post_processor/determinant_diversity.rs
```diff
@@ -0,0 +1,67 @@
...
+    determinant_diversity_post_process,
+    inmem::{self, GetFullPrecision},
+};
+use diskann_quantization::num::Positive;
+
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct DeterminantDiversity {
+    power: f32,
+    eta: f32,
+}
```
hildebrandmw: Why is this duplicated here instead of using `DeterminantDiversityParams` from in-mem?

Should this implementation line in `diskann-providers` instead?

diskann-providers/src/model/graph/provider/mod.rs
```diff
@@ -6,3 +6,10 @@ pub mod async_;
 
 // Layers for the async index.
 pub mod layers;
+
+/// Determinant-diversity post-processing algorithm.
+///
+/// This module is not async-specific and is re-exported here for clarity.
+/// It provides diversity-promoting reranking for nearest neighbor search results.
+pub mod determinant_diversity;
+pub use determinant_diversity::determinant_diversity_post_process;
```
hildebrandmw: Suggestion - make the `determinant_diversity` module private and reexpor the method as `determinant_diversity`. Suffixing with `_post_process` adds a lot of characters and little value.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
hildebrandmw: This should lean into proper error handling. This is exporting a public method with *no* documentation on invariants.

Additionally, the error handling is a little inconsistent: a query of length 0 is treated as fine even if the candidates have non-zero length.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
```diff
@@ -0,0 +1,693 @@
...
+    max: f32,
+}
+
+pub fn determinant_diversity_post_process<Id: Copy>(
+    candidates: Vec<(Id, f32, Vec<f32>)>,
+    query: &[f32],
+    k: usize,
+    determinant_diversity_eta: f32,
+    determinant_diversity_power: Positive<f32>,
+) -> Vec<(Id, f32)> {
```
hildebrandmw: Suggestion: This interface would be much cleaner if it was
```rust
pub fn determinant_diversity(
    candidates: MutMatrixView<'_, f32>,
    distances: &[f32],
    query: &[f32],
    k: usize,
    params: &DeterminantDiversityParams,
) -> Vec<usize> {
```
where we return the rows of `candidates` that are selected. This does a few things:
1. Makes `determinant_diversity` non-generic. There's no reason it *needs* to be generic other than it's being used to shuffle around IDs.
2. Reduces allocations - the `Vec<usize>` is already materialized within the algorithm and you can mutate `candidates` directly.
3. Sizing validation checks are O(1).
4. Taking `DeterminantDiversityParams` works as a check that the parameters have been validated properly.

diskann-benchmark/src/inputs/graph_index.rs
```diff
@@ -761,6 +780,14 @@ impl IndexOperation {
         self.source.validate(checker)?;
         self.search_phase.validate(checker)?;
 
+        if has_topk_determinant_diversity(&self.search_phase)
+            && *self.source.data_type() != DataType::Float32
+        {
+            anyhow::bail!(
+                "determinant-diversity post-processor requires graph-index full precision float32 input"
+            );
+        }
```
hildebrandmw: This is such a brittle way of doing this check. No one is going to remember to add all these. For this, I'd prefer you just let the plugin matching fail down-stream as that should (hopefully) always be right and doesn't require the input to have as much knowledge about the backend.

diskann-providers/src/post_processor.rs
```diff
@@ -0,0 +1,137 @@
...
+/// Construction fails if:
+/// - `power <= 0.0` (invalid power weighting)
+/// - `eta < 0.0` (negative stability parameter)
+#[derive(Debug, Clone, Copy)]
+pub struct DeterminantDiversityParams {
+    /// Relevance weighting exponent. Must be > 0.0.
+    power: f32,
+    /// Numerical stability parameter. Must be >= 0.0.
+    eta: f32,
+}
```
hildebrandmw: This is a good struct. It should be used in more places, rather than duplicating its contents in the arguments to `determinant_diversity_post_process`, as a struct + enum in `diskann-disk`, and as a raw post-processor in `diskann-benchmark`.

diskann-providers/src/lib.rs
```diff
@@ -14,6 +14,8 @@ pub mod model;
 
 pub mod common;
 
+pub mod post_processor;
```
hildebrandmw: Please remove the top-level module and shuffle the implementation of `DeterminantDiversityParams` to the `determinant_diversity`, which should be wired to use this struct directly rather than the raw parameters.

diskann-providers/src/model/graph/provider/mod.rs
```diff
@@ -6,3 +6,10 @@ pub mod async_;
 
 // Layers for the async index.
 pub mod layers;
+
+/// Determinant-diversity post-processing algorithm.
+///
+/// This module is not async-specific and is re-exported here for clarity.
```
hildebrandmw: This isn't necessarily re-exported ... it's just exported.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -827,13 +966,15 @@ where
...
     pub fn search(
         &self,
         query: &[Data::VectorDataType],
         return_list_size: u32,
         search_list_size: u32,
         beam_width: Option<usize>,
         vector_filter: Option<VectorFilter<Data>>,
+        post_processor: Option<SearchPostProcessorKind>,
         is_flat_search: bool,
     ) -> ANNResult<SearchResult<Data::AssociatedDataType>> {
```
hildebrandmw, line 989: There are not substantive tests for the new functionality in `diskann-disk` and this API is starting to get bloated with many optionals. My stake in `diskann-disk` is relatively low. Please ensure the maintainers of this code are comfortable with these changes.

diskann-benchmark/src/inputs/post_processor.rs
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+use diskann_benchmark_runner::Checker;
+use diskann_providers::post_processor::DeterminantDiversityParams;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "kebab-case")]
+pub(crate) enum TopkPostProcessor {
+    DeterminantDiversity { power: f32, eta: f32 },
```
hildebrandmw, line 15: This would be cleaner as
```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(tag = "type", rename_all = "kebab-case")]
enum RawTopkPostProcessor {
    DeterminantDiversity { power: f32, eta: f32 },
}

#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(try_from = "RawTopkPostProcessor`, into = "RawTopkPostProcessor")]
pub(crate) enum TopkPostProcessor {
    DeterminantDiversity(DeterminantDiversityParams)
}

// Impl necessary conversions
```
This has the benefit of carrying around a fully verified `DeterminantDiversityParams` struct and is pretty easily extensible for future variants.

diskann-benchmark/src/inputs/disk.rs
```diff
@@ -373,6 +381,14 @@ impl DiskSearchPhase {
             Some(lim) => write_field!(f, "Search IO Limit", format!("{lim}"))?,
             None => write_field!(f, "Search IO Limit", "none (defaults to `usize::MAX`)")?,
         }
+        match &self.post_processor {
+            Some(TopkPostProcessor::DeterminantDiversity { power, eta }) => {
```
hildebrandmw: Should this not live as a `Display` impl on `TopkPostProcessor` instead?

diskann-benchmark/src/backend/index/benchmarks.rs
```diff
@@ -73,6 +74,7 @@ pub(super) fn register_benchmarks(registry: &mut Registry) -> anyhow::Result<()>
     registry.register(
         "graph-index-full-precision-f32",
         FullPrecision::<f32>::new()
+            .search(plugins::DeterminantDiversity)
```
hildebrandmw: Why is determinant diversity first? The most common search kinds should be ordered first.

diskann-benchmark/src/index/search/plugins.rs
```diff
@@ -145,9 +150,49 @@ where
...
+    }
+
+    pub(crate) const fn as_str() -> &'static str {
+        "topk"
+    }
+}
+
+/// A search plugin for determinant-diversity top-k post-processing.
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct DeterminantDiversity;
```
hildebrandmw, line 161: After reviewing this - matching on a sub-enum of `TopK` feels like it's going to be super brittle. Since the plugin implementation already duplicates all the logic that a new `SearchPhase` enum would need anyways, let's must make determinant diversity a new `SearchPhase` enum and keep the one-to-one mapping.

hildebrandmw (commented):
Thanks Naren, this has been a journey. The PR still has a lot of little structural issues that need to be addressed before merging. I've done my best to enumerate the issues and potential fixes in the inline comments. Please read through all the comments first as some of them are inter-related.

Note that this review focused on architecture and integration. I'd like to take one more look at the algorithm when the structural issues are resolved.

Feel free to reach out with any questions!

diskann-benchmark/src/backend/index/benchmarks.rs
narendatha: Makes sense, moved it as last item.

diskann-benchmark/src/inputs/disk.rs
narendatha: done.

diskann-providers/src/model/graph/provider/mod.rs
narendatha: Fixed comment.

diskann-providers/src/model/graph/provider/async_/mod.rs
narendatha: Good points. done.

diskann-disk/src/search/provider/disk_provider.rs
narendatha: Looks like experimental change when I tried previous approaches. Revered this.

diskann-benchmark/src/inputs/post_processor.rs
narendatha: Good idea, makes it cleaner. done.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
narendatha: Missed to fix the documentation when changing code. fixed now.

diskann-providers/src/post_processor.rs
narendatha: added checks for infinity and fixed comments.

diskann-disk/src/search/provider/disk_provider.rs
narendatha: Now using `DeterminantDiversityParams` instead of power and eta directly.

diskann-benchmark-core/src/search/graph/knn.rs
narendatha: removed this.

diskann-providers/src/lib.rs
narendatha: done as part of another comment.

diskann-benchmark/src/index/search/plugins.rs
narendatha: Makes sense. Removed the matching of sub-enum. I made it a new `SearchPhase` enum.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
narendatha: This is excellent suggestion and I admire you looking deep into this code. It makes it cleaner and also improves QPS for larger datasets. done.

diskann-providers/src/model/graph/provider/mod.rs
narendatha: done.

diskann-benchmark/src/index/post_processor/determinant_diversity.rs
narendatha: done, it now uses `DeterminantDiversityParams` from `diskann-providers`

diskann-benchmark-core/src/search/graph/knn.rs
narendatha: This approach is brilliant. It removes two search loops and makes it into one for easier maintenance. done.

diskann-benchmark/src/inputs/graph_index.rs
narendatha: This was causing me discomfort also. I agree with you and removed this, now the error will happen down-stream, when is_match fails, will produce error "Unsupported search phase: "topk-determinant-diversity" - expected one of "topk", "range", "topk-beta-filter", and "topk-multihop-filter". This is confusing but I feel it is acceptable for now. We might start supporting other types soon and then this becomes moot.

diskann-providers/src/post_processor.rs
narendatha: yes, this is now used in all the places I could search and make sure of.

diskann-disk/src/search/provider/disk_provider.rs
narendatha: I added more tests in diskann-disk. You are right about the API bloat. It feels right to handle that in separate PR. I will let the reviewers on disk_provider know about this.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
```diff
@@ -0,0 +1,712 @@
...
+//!
+//! # References
+//!
+//! The algorithm is based on diversity-promoting ranking methods for nearest neighbor search,
+//! as used in approximate nearest neighbor indices like DiskANN.
+
+use diskann_utils::views::MutMatrixView;
+use diskann_vector::{MathematicalValue, PureDistanceFunction, distance::InnerProduct};
+
+use crate::post_processor::DeterminantDiversityParams;
```
hildebrandmw: Small: move the definition of `DeterminantDiversityParams` into this file so it's all self contained?

hildebrandmw (approved):
I think it's there! Thanks Naren!

diskann-benchmark/src/index/post_processor/determinant_diversity.rs
suri-kumkaran: +1, Should we move this whole implementation to `diskann-providers`?

diskann-benchmark-core/src/search/graph/knn.rs
```diff
@@ -68,10 +78,98 @@ where
...
+    type Processor = S::Processor;
+
+    fn as_post_processor(&'a self, strategy: &'a S) -> Self::Processor {
+        strategy.default_post_processor()
+    }
+}
+
+/// Wraps an explicit post-processor for use with [`KNN::with_postprocessor`].
+#[derive(Debug, Clone, Copy)]
+pub struct Forwarded<PP>(PP);
```
suri-kumkaran, line 158: For my understanding - Why do we need `Forwarded<PP>(PP)`?

diskann-providers/src/model/graph/provider/determinant_diversity.rs
narendatha: Done, thanks for pointing out.

diskann-benchmark/src/index/post_processor/determinant_diversity.rs
```diff
@@ -0,0 +1,72 @@
...
+        I: Iterator<Item = Neighbor<A::Id>> + Send,
+        B: SearchOutputBuffer<A::Id> + Send + ?Sized,
+    {
+        let candidates: Vec<Neighbor<A::Id>> = candidates.collect();
+        let store: &inmem::FullPrecisionStore<f32> = accessor.as_full_precision();
+        let mut vectors = Matrix::new(0.0f32, candidates.len(), query.len());
+        let mut ids = Vec::with_capacity(candidates.len());
+        let mut distances = Vec::with_capacity(candidates.len());
+
+        for (i, candidate) in candidates.iter().enumerate() {
```
suri-kumkaran: nit: Since we need to collect here to know the candidate count under the current `Iterator` contract, could this use `candidates.into_iter()` instead of `candidates.iter()`? We do not need `candidates` after this loop, so consuming it releases the temporary `Vec<Neighbor<_>>` before reranking.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
suri-kumkaran, line 1: Should we consider moving this to `diskann-quantization`, WDYT @hildebrandmw, @arkrishn94?

diskann-providers/src/model/graph/provider/determinant_diversity.rs
suri-kumkaran: I also feel we should have some error handelling and/or documentation.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -330,6 +363,125 @@ where
...
+
+        let reranked = determinant_diversity(
+            candidate_vectors.as_mut_view(),
+            &candidate_distances,
+            &query_f32,
+            usize::MAX,
+            &self.params,
+        );
+
+        Ok(output.extend(reranked.into_iter().filter_map(|idx| {
```
suri-kumkaran: nit: This lookup should always succeed because the data was inserted for every candidate above. With aligned vectors, this can become a plain map using `candidate_ids[idx]`, `associated_data[idx]`, and `candidate_distances[idx]`.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -330,6 +363,125 @@ where
...
+            let id = ordered_ids[idx];
+            let distance = candidate_distances[idx];
+            associated_data
+                .get(&id)
+                .copied()
+                .map(|data| ((id, data), distance))
+        })))
```
suri-kumkaran, line 440: Since determinant diversity returns candidate row indexes, could we keep IDs, distances, and associated data in aligned vectors? That would remove `ordered_ids`, replace the `HashMap` with a `Vec`, and avoid the final hash lookup.

diskann-disk/src/search/provider/disk_provider.rs
suri-kumkaran: This makes the API harder to reason about, since several parameters intersect or depend on one another, and it allows invalid combinations to be represented.

I am tagging @dyhyfu and @wuw92 to consider `Diversity` as part of the API cleanup in #1173.

Even if we defer this to a follow up PR, I think the input should be `SearchPostProcessorKind` directly rather than an `Option`. It is unclear what the three possible states represent: `None`, plus the two variants of `SearchPostProcessorKind`.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
hildebrandmw: Let's not make `quantization` a dumping ground for non-quantization related things. `multi-vector` is already pushing it.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
```diff
@@ -0,0 +1,840 @@
...
+///
+/// - The numerator flips the order so that the *closest* candidate gets the
+///   highest similarity (~1) and the *farthest* gets ~0.
+/// - The denominator is clamped to `f32::EPSILON` so a degenerate range
+///   (d_max == d_min) produces a finite, equal score for all candidates
+///   instead of a divide-by-zero.
+/// - The trailing `+ EPSILON` ensures the result is strictly positive, so
+///   that `similarity(d).powf(power)` never produces a hard zero scale (which
+///   would erase a candidate from the QR pivoting and bias selection).
+fn distance_to_similarity(distance: f32, distance_range: DistanceRange) -> f32 {
```
arkrishn94, line 467: nit: LLVM probably already does, but maybe this deserves an `[inline(always)]`?

arkrishn94 (approved):
It's looking good Naren, there's just the one comment on the error handling.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
arkrishn94: It's bespoke enough for now that it can live here :)

suri-kumkaran: We can keep it here for now, but for longer term we should find a good place for things which are not `quantization` and not `providers` and still going to be used as a common algorithm/feature.

narendatha: agree with the comments. we will find right place to move.

diskann-benchmark/src/index/post_processor/determinant_diversity.rs
narendatha: I agree — it's a valid nit, and I've applied,  it's a sound micro-optimization

diskann-disk/src/search/provider/disk_provider.rs
narendatha: Good catch, it's a good simplification. done.

diskann-disk/src/search/provider/disk_provider.rs
narendatha: Done, removed `ordered_ids` also.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
narendatha: made it inline, I guess LLVM might do it, but no need to take chances.

diskann-disk/src/search/provider/disk_provider.rs
narendatha: Agreed with the `Option`. Removed that `Option` now. On the API itself,it's a larger redesign of the whole search surface. It belongs in the dedicated cleanup, not bolted onto this feature PR, so will defer that part.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
narendatha: Done, added error handling and now errors are reutnred.

diskann-benchmark-core/src/search/graph/knn.rs
narendatha: Forwarded<PP> lets the KNN post-processor slot accept two different things through one AsPostProcessor trait, so the search loop isn't duplicated:

Defaulted: use the strategy's default post-processor.
Forwarded(pp): use the caller's explicit post-processor.
It has to be a newtype rather than using PP directly because of coherence: an impl AsPostProcessor for Defaulted plus a blanket impl<PP> AsPostProcessor for PP would overlap (a future PP could be Defaulted), which the compiler rejects. Wrapping in a local Forwarded<PP> makes the two impls provably disjoint. As a bonus, the marker is zero-sized while Forwarded carries the actual processor, and the right impl is picked at compile time with no runtime branch.

diskann-benchmark/src/index/post_processor/determinant_diversity.rs
suri-kumkaran: Looks like we may have missed this comment.

narendatha: Moved it to diskann-providers now.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
```diff
@@ -0,0 +1,932 @@
...
+            } => write!(
+                f,
+                "determinant-diversity received {} distances for {} candidate rows",
+                distances, candidates
+            ),
+        }
+    }
+}
+
+impl std::error::Error for DeterminantDiversityError {}
```
suri-kumkaran: I think we can use `use thiserror::Error;` to simplify this implementation, [example](https://github.com/microsoft/DiskANN/blob/3aa44ace775d2d32e7f9d37d5a99b5f4f26d8e48/diskann-providers/src/model/graph/provider/async_/inmem/scalar.rs#L859-L888)

narendatha: Good point, this does simply impliemnetation, thanks and done.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
```diff
@@ -127,20 +127,26 @@ impl fmt::Display for DeterminantDiversityParams {
 
 /// Error produced when constructing [`DeterminantDiversityParams`] or running
 /// [`determinant_diversity`].
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, thiserror::Error)]
 pub enum DeterminantDiversityError {
     /// Power parameter <= 0.0
```
suri-kumkaran: nit: We can clean up these comments since the error descriptions are already sufficiently descriptive, making these comments redundant.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -330,6 +365,121 @@ where
...
+        }
+
+        let reranked = determinant_diversity(
+            candidate_vectors.as_mut_view(),
+            &candidate_distances,
+            &query_f32,
+            usize::MAX,
+            &self.params,
+        )
+        .map_err(|e| ANNError::new(diskann::ANNErrorKind::DimensionMismatchError, e))?;
```
suri-kumkaran: Could we add `impl From<DeterminantError> for ANNError` (ideally with variant-specific `ANNErrorKind` mapping) and use `?` here instead of `map_err(...)`?

Right now this forces every determinant-diversity failure into `DimensionMismatchError`, which may hide other classes of failures (e.g., invalid params or numerical issues). A centralized conversion would make this callsite cleaner and preserve better error semantics across disk/async paths.

Something like:
```rust
impl From<DeterminantError> for ANNError {
    fn from(err: DeterminantError) -> Self {
        use diskann::ANNErrorKind;
        match err {
            DeterminantError::DimensionMismatch { .. } => {
                ANNError::new(ANNErrorKind::DimensionMismatchError, err)
            }

            DeterminantError::InvalidK { .. }
            | DeterminantError::KExceedsCandidates { .. }
            | DeterminantError::EmptyCandidates
            | DeterminantError::InvalidParameter { .. } => {
                ANNError::new(ANNErrorKind::InvalidArgument, err)
            }

            DeterminantError::NumericalInstability { .. }
            | DeterminantError::SingularMatrix { .. }
            | DeterminantError::Internal { .. } => {
                ANNError::new(ANNErrorKind::LogError, err)
            }
        }
    }
}
```

suri-kumkaran (approved):
Thanks Naren for all the hard work.

diskann-disk/src/search/provider/disk_provider.rs
narendatha: Good call — done. Added `impl From<DeterminantDiversityError> for ANNError` (with `#[track_caller]`) next to the enum and switched both the disk and async callsites to `?`.

Note our enum only has 4 variants, so the mapping is: `InvalidPower`/`InvalidEta` → `IndexConfigError` (we don't have an `InvalidArgument` kind), and `QueryDimensionMismatch`/`DistanceCountMismatch` → `DimensionMismatchError`. This fixes the `DistanceCountMismatch` case that was previously mislabeled and centralizes the conversion for future variants.

diskann-providers/src/model/graph/provider/determinant_diversity.rs
narendatha: good point, done.

suri-kumkaran (approved):

PR #974, opened 2026-04-23
Add A/A benchmark documentation

YuanyuanTian-hh:
## Summary

Add diskann-benchmark/AA_BENCHMARK.md documenting how the daily A/A benchmark stability test is conducted and scheduled.

## Content Covered

- **Purpose**: Detecting environment noise on CI runners (not code regressions)
- **Schedule**: Daily at 9 AM UTC via cron, plus manual \workflow_dispatch\
- **Datasets**: \wikipedia-100K\ and \openai-100K\
- **Execution flow**: Step-by-step description of the benchmark pipeline
- **Tolerance thresholds**: Build time, QPS, recall, mean I/Os, mean comparisons, mean/P95 latency
- **Failure notification**: Auto-created GitHub issue tagging \@microsoft/diskann-disk-maintainers\
- **Comparison**: How A/A differs from the A/B benchmark workflow

## Motivation

This makes it easier for new contributors and maintainers to understand how the A/A stability test works without having to read the YAML workflow directly.

PR #1013, opened 2026-05-04
[v0.51.0] Bump version to 0.51.0

arkrishn94:
Bump version to 0.51.0 due to propagate changes to downstream consumers

## Breaking API changes (AI Generated)

- **`ObjectPool` moved** (#975): now lives in `diskann-utils`. Update imports from `diskann::...::ObjectPool` → `diskann_utils::ObjectPool`.

- **`AlignedSlice` removed** (#994): the `AlignedSlice` abstraction in `diskann-vector` is gone. Code that converted between vector representations through `AlignedSlice` should now use the `Poly` / `CastFromSlice` polymorphic interfaces directly (see `diskann-vector::conversion` and `diskann-quantization::alloc::poly`). Storage that previously held `AlignedSlice` values should hold `Poly<T, A>` instead.

- **`AsThreadPool` generic removed** (#967): functions that previously took `pool: impl AsThreadPool` now take `pool: &RayonThreadPool`. Pass a borrow of an existing pool; remove the generic parameter from your call sites.

- **`sgemm()` returns `Result`** (#997): in `diskann-linalg`, the new signature is:

  ```rust
  pub fn sgemm(
      atranspose: Transpose, btranspose: Transpose,
      m: usize, n: usize, k: usize,
      alpha: f32, a: &[f32], b: &[f32],
      beta: Option<f32>, c: &mut [f32],
  ) -> Result<(), SgemmError>
  ```

  `SgemmError` has variants `InvalidMatrixDimensions { matrix_name, expected_rows, expected_cols, actual_len }` and `DimensionOverflow { matrix_name, rows, cols }`. Replace previous panic-on-bad-input assumptions with explicit handling.

- **Benchmarks are stateful** (#995): the `Benchmark` impls in `diskann-benchmark` are no longer stateless unit structs. Each benchmark type now has a `::new()` constructor (often holding `PhantomData<T>` or plugin state), and registration uses an instance:

  ```rust
  // before
  benchmarks.register("name", MyBench);
  // after
  benchmarks.register("name", MyBench::<T>::new());
  ```
  If you wrote a custom benchmark, give it a `new()` and register an instance. Combined with #996, search-side benchmarks now compose `Plugins<Provider, Phase, Strategy>` and expose builder methods like `.search(plugin)` to register search plugins on the instance.

- **`diskann-benchmark`: `async` → `graph-index`** (#1009): the benchmark category previously named `async` was renamed to `graph-index`. JSON config `type` values and example file names changed accordingly:
  - `async-build` → `graph-index-build`
  - `async-dynamic-run` → `graph-index-dynamic-run`
  - and the same prefix swap for `*-pq`, `*-sq`, `*-spherical-quantization`, etc.

  Update any benchmark config files, scripts, or CI that reference the old `async-*` names.

- **`diskann-disk` buffer alignment decoupled from `block_size`** (#984): code that assumed I/O buffer alignment equals the disk block size should now configure alignment explicitly.

## Non-breaking

- New cache-aware block-transposed Chamfer/MaxSim distance for f32/f16 (#863).
- A/A benchmark documentation (#974); CI publish workflow improvements (#755, #1017); openssl bump (#973); `compute_closest_centers` allocation reduction (#980).
-  **`DistanceComputer` `'static` bound relaxed** (#1007) and **redundant `DistanceFunction` impls removed** (#1008)

**Full Changelog**: https://github.com/microsoft/DiskANN/compare/v0.50.1...v0.51.0


.github/workflows/publish.yml
```diff
@@ -102,13 +102,20 @@ jobs:
             exit 1
           fi
 
-      - name: Verify all crates use workspace version
+      - name: Verify all publishable crates use workspace version
```
harsha-simhadri, line 105: Thanks for fixing this oversight? Shall we separate this into another PR?

arkrishn94: Makes sense, opened #1017

harsha-simhadri (approved):

hildebrandmw (approved):
Thanks!

PR #1014, opened 2026-05-05
Update instructions for agents

harsha-simhadri:


agents.md
```diff
@@ -1,17 +1,46 @@
...
-**Last Updated**: 2026-02-11 (based on v0.45.0, Rust 1.92)
+**Last Updated**: 2026-05-04 (based on v0.50.1, Rust 1.92)
 
 This guide helps coding agents understand how to work efficiently with the DiskANN repository.
 
+> **Relationship with `.github/copilot-instructions.md`**: This file (`agents.md`) is the operational manual — project structure, commands, workflows, and detailed conventions. The [`.github/copilot-instructions.md`](.github/copilot-instructions.md) file contains concise review-time rules that are automatically injected into Copilot code review context. Keep both in sync when conventions change.
+
+---
+
+## Quick Reference
```
harsha-simhadri: we can probably remove since agents would not how to work with Rust workspace

metajack (approved):

.github/copilot-instructions.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Dependency and Build Hygiene
+- Check for changes that introduce new dependencies or increase build times. If a dependency is added, it must be justified in the PR description.
+
+
+## Error Handling
+
+- Do not introduce `panic!` paths for recoverable errors — propagate with `Result` instead.
+- Keep error types small. Avoid large enums/structs that blow up the stack; look for ways to reduce field sizes (e.g., compute derivable fields, use enums instead of `&'static str`).
+- Prefer `ANNError::new(ANNErrorKind::…, e)` over the old `log_*`-style constructors, which force eager string formatting and double-log errors.
```
hildebrandmw, line 15: Nit: There is not `double-log` from these APIs (the name is a misnomer). The biggest issue is eager string formatting.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,64 @@
...
+- When using `thiserror`, rely on `#[from]` for automatic `Error::source` chaining — do not format the inner error in the `#[error("…")]` display string.
+- Include relevant context values (e.g., the kind, key, or dimension) in error messages for debuggability.
+
+## Documentation
+
+- Doc comments and README examples must match actual API signatures and serialized shapes. 
+- Stale examples that fail to compile or deserialize are treated as bugs.
+- Do not leave dead references to APIs that no longer exist.
+- When changing a function signature or removing a parameter, update all doc comments that mention the old signature.
```
hildebrandmw, line 24: More suggestions on documentation:
* Doc and inline comments should describe what the code does. Do not describe behavior by contrasting it with other code, except when referencing documented external behavior.
* Avoid comments that simply restate what is already clear from function signatures or where clauses.
* Do not list functions or types in module docs that `rustdoc` already documents.
* Module-level docs should describe the purpose and structure of the module, not its contents.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Testing
+- Do not drop existing unit tests without a strong reason. 
+- Keep test helpers close to the code they exercise, typically in a `mod tests` at the bottom of the file or in an adjacent test module, guarded with `#[cfg(test)]`.
+- Do not add tests for derived traits (`Clone`, `Debug`, `PartialEq`) or enums unless they have explicit behavior beyond the derive.
+- Test edge cases like empty inputs (e.g., empty iterators) to lock in defined behavior and prevent divide-by-zero or NaN results.
+
+## Rayon and Parallelism
+
+- Never use the global Rayon thread pool. Always execute parallel work within the provided `RayonThreadPool` or `RayonThreadPoolRef`.
```
hildebrandmw, line 44: This guidance is more for `diskann-providers`/`diskann-disk` than lower level crates. Lower level crates like `diskann-quantization` should use the dynamically scoped thread pool, but advertise this using the `Parallelism` enum.

hildebrandmw (commented):
A few small nits, but otherwise looks good.

AGENTS.md
```diff
@@ -0,0 +1,182 @@
+# DiskANN Repository - Agent Onboarding Guide
+
+**Last Updated**: 2026-05-04 (based on v0.50.1, Rust 1.92)
```
magdalendobson, line 3: It looks like a lot of previous commentary on testing is removed and not reproduced anywhere else. Is this purposeful? It seems like useful information that shouldn't be lost.

Same comment on code coverage. Should we add something on both topics to the "Always" list?

AGENTS.md
```diff
@@ -0,0 +1,182 @@
...
+- `diskann-quantization/` - Vector quantization 
+
+**Tier 3: Algorithm & Providers**
+- `diskann/` - Core indexing logic
+- `diskann-providers/` - Hodge-podge of stuff, will be dismantled
+- `diskann-disk/` - Disk-based provider for the index with io_uring support
+- `diskann-label-filter/` - Inverted index for filtered search
+- `diskann-garnet/` - Garnet (Redis-compatible) Provider and FFI endpoints for vector sets
+
+**Tier 4: Infrastructure & Tools**
```
magdalendobson, line 25: "Infrastructure" is a bit of a vague name here, it kind of seems like every tier contains something that could be called infrastructure. Maybe "Benchmarks and Tools" would be better?

AGENTS.md
```diff
@@ -0,0 +1,182 @@
...
+- Add a `// SAFETY:` comment above every `unsafe` block with specific, verifiable preconditions.
+
+---
+
+
+
+
+## Error Handling
+
+There are three regimes of error handling and the strategy to use depends on the regime.
```
magdalendobson, line 80: It sounds like these categories may conform neatly to the tiers discussed above, although maybe there are exceptions. If so it could be good to state that explicitly.

AGENTS.md
```diff
@@ -0,0 +1,182 @@
...
+
+```rust
+// ✅ Good — deferred formatting, precise kind, track_caller
+#[track_caller]
+fn process_vectors(data: &[f32]) -> Result<(), ANNError> {
+    validate(data).map_err(|e| ANNError::new(ANNErrorKind::IndexError, e))?;
+    Ok(())
+}
+
+// ❌ Bad — eager string formatting, double-logs on creation
```
magdalendobson, line 131: See comment from Mark about double-logging

.github/copilot-instructions.md
```diff
@@ -0,0 +1,64 @@
+When performing a code review, check that:
```
magdalendobson, line 1: Can copilot instructions point to the existing AGENTS.md files where applicable? It looks like a lot of instructions are manually duplicated here, so we risk them getting out of sync.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,64 @@
...
+
+ The workspace obeys SemVer. Removing or changing public API signatures (functions, types, re-exports)  requires a version bump, suggested  migration impact in the PR description.
+
+## Dependency and Build Hygiene
+- Check for changes that introduce new dependencies or increase build times. If a dependency is added, it must be justified in the PR description.
+
+
+## Error Handling
+
+- Do not introduce `panic!` paths for recoverable errors — propagate with `Result` instead.
```
magdalendobson, line 13: This probably deserves a mention/move to the top-level AGENTS.md

.github/copilot-instructions.md
arrayka: +1.
Do we really need `.github/copilot-instructions.md`? if so, how is it semantically different from `/AGENTS.md`? Why can't we keep all agent-related instructions in `/AGENTS.md`?

AGENTS.md
```diff
@@ -0,0 +1,182 @@
+# DiskANN Repository - Agent Onboarding Guide
```
arrayka, line 1: I would explicitly follow next structure in this document:
1. How code is organized (Crate Organization)
2. How to contribute (aka How we write code) - and move all stuff from copilot-instructions to this section.
3. How to validate code before submitting a PR + briefly mention that CI will perform extensive validation in `.github/workflows/ci.yml`.

hildebrandmw (approved):

PR #1020, opened 2026-05-05
bf_tree migration away from diskann-providers

### LINKED ISSUE #971: Extract bftree provider into its own crate with appropriate tests.

JordanMaples:
# Migrate bf_tree provider to standalone `diskann-bftree` crate

## Motivation

The long-term goal is to remove `diskann-providers` entirely. The bf_tree provider was one of several components keeping that crate alive — it depended on `diskann-providers` for generic delete infrastructure (`DeletionCheck`/`AsDeletionCheck`/`RemoveDeletedIdsAndCopy`) and PQ quantization types. This PR extracts bf_tree into a standalone `diskann-bftree` crate, cuts those dependencies, simplifies the generics, and switches from PQ to spherical quantization — all to reduce the surface area of `diskann-providers` and move closer to removing it.

### Crate Extraction & Rename

Moved the bf_tree provider from `diskann-providers/src/model/graph/provider/async_/bf_tree/` into a standalone `diskann-bftree` crate (flat module structure: `lib.rs`, `provider.rs`, `vectors.rs`, `quant.rs`, `neighbors.rs`). Updated workspace `Cargo.toml` and CI configuration.

### PQ → Spherical Quantization

- **`QuantVectorProvider`**: Stores `Poly<dyn Quantizer>` (spherical) instead of `Arc<FixedChunkPQTable>`.
- **`QuantQueryComputer`**: Newtype wrapper adapting spherical's `Opaque`-based API to `PreprocessedDistanceFunction<&[u8], f32>`.
- **Serialization**: Uses `Quantizer::serialize` / `iface::try_deserialize` (flatbuffers format).
- **Distance layout**: Uses `QueryLayout::FullPrecision` for better recall with low-bit quantizers.
- Removed criterion/iai benchmarks (`benches/` directory).

### Remove `D` (Delete Provider) Parameter

- `BfTreeProvider<T, Q, D>` → `BfTreeProvider<T, Q>`.
- Hard deletes via bf_tree's native `delete(key)` API replace bitmap-based soft deletes.
- `delete()` removes from all three trees (neighbors, full_vectors, quant_vectors).
- Removed `delete_bitmap_serde.rs`, `NoDeletes`/`TableBasedDeletes` from constructors.

### Simplify to Two Strategies

Removed the `Hybrid` strategy entirely — only `FullPrecision` and `Quantized` remain. `Quantized` search uses `Rerank` post-processor to re-rank candidates using full-precision vectors.

### Reduce `diskann-providers` Coupling

- `FullPrecision` and `Quantized` marker structs moved to `diskann::graph::strategy` (canonical home), re-exported from `diskann-providers`.
- `NoStore` defined locally in `diskann-bftree`.
- `TestCallCount` duplicated locally (conditional compile: real counter in test, no-op in release).
- `AsKey` trait added locally to replace scattered `bytes_of` calls.

### `ToRanked` Error Handling

Replaced blanket `ANNError` with ranked error types for vector access:

- **`VectorError`** (`Deleted` | `NotFound`): transient errors from bf_tree reads.
- **`VectorUnavailable`**: implements `TransientError<ANNError>` — acknowledge (skip) or escalate.
- **`AccessError`** = `RankedError<VectorUnavailable, ANNError>`: type alias used as `GetError`.
- `on_elements_unordered`: skips transient errors, propagates real errors.
- `status_by_internal_id`: transient → `ElementStatus::Deleted`, real → propagate.
- `get_delete_element`: escalates (delete target must exist).
- `Rerank::post_process`: acknowledges transient (skips candidate), propagates real errors.

### Other Cleanups

- `new_empty` removed — `new` requires start points upfront.
- Uses `Map` working set for prune (avoids full refetch).
- `SearchAccessorError` → `Infallible` (search accessor cannot fail).
- Broken intra-doc links fixed.
- Test quantizer helper deduplicated (`create_test_quantizer` in `quant.rs`).

## ⚠️ Breaking Changes

- **Serialization format**: Persisted indexes using the PQ format will not load. Full reindex required.
- **Training API**: Callers must provide `Poly<dyn Quantizer>` instead of `FixedChunkPQTable`.
- **Constructor signatures**: `D` parameter removed, `new_empty` removed, `new` requires start points.
- **Delete semantics**: Deletes are immediate and permanent (hard delete from bf_tree).
- **Error types**: `GetError` is now `AccessError` (ranked) instead of `ANNError`.

## Net Impact

- **-36 net lines** (17 files changed, +1607 / -1643)
- `diskann-bftree` is a self-contained crate with two generic parameters (`T`, `Q`)
- Two indexing strategies: `FullPrecision` and `Quantized` (with `Rerank`)
- No dependency on `DeletionCheck` / `AsDeletionCheck` / `RemoveDeletedIdsAndCopy`
- 23 unit tests + 2 doc tests covering both strategies (insert, search, delete, save/load)

## Open Items

- **Delete ordering**: Discuss whether current delete operation order is sufficient for concurrent access (see review thread).
- **NeighborAccessor ToRanked**: Follow-up to add ranked errors to the core `NeighborAccessor` trait.
- **QueryComputer Debug**: Deferred — needs `Debug` impls in `diskann-quantization` first.


diskann-bf_tree-provider/src/provider/delete_bitmap_serde.rs
```diff
@@ -0,0 +1,176 @@
...
+ * Licensed under the MIT license.
+ */
+
+//! Serialization/deserialization for [`TableDeleteProviderAsync`] delete bitmaps.
+//!
+//! The on-disk format is a sequence of little-endian `u32` words, each holding 32 deletion
+//! bits. Bit `b` of word `w` corresponds to vector ID `w * 32 + b`. The number of words
+//! is `ceil(max_size / 32)`.
+
+use diskann_providers::model::graph::provider::async_::TableDeleteProviderAsync;
```
hildebrandmw: Weren't we getting rid of this?

diskann-bf_tree-provider/src/provider/quant_vector_provider.rs
```diff
@@ -105,30 +92,33 @@ impl QuantVectorProvider {
...
-            self.metric,
-            &T::as_f32(query).into_ann_result()?,
-            Some(self.vec_pool.clone()),
-        )
+        let query_f32 = T::as_f32(query).into_ann_result()?;
+        let inner = self
+            .quantizer
+            .fused_query_computer(
+                &query_f32,
+                QueryLayout::FullPrecision,
```
hildebrandmw, line 93: One minor detail - for index construction, you might want to leave this `SameAsData` so the distances are comparable to the distances computed during pruning.

diskann-bf_tree-provider/src/provider/delete_bitmap_serde.rs
JordanMaples: this was from the earlier "migration only" phase of the changes. This'll be removed when I attack the D param removal.

JordanMaples:
@copilot would you mind giving this a review

Cargo.toml
```diff
@@ -23,7 +23,7 @@ members = [
     "diskann-benchmark-simd",
     "diskann-benchmark",
     "diskann-tools",
-    "vectorset",
+    "vectorset", "diskann-bf_tree",
```
hildebrandmw: `diskann-bf_tree` is quite a name ... maybe `diskann-bftree`? At least something that doesn't mix hyphens and underscores :laugh:

Also - the formatting here is a little off.

diskann-providers/src/model/graph/provider/async_/mod.rs
```diff
@@ -29,9 +27,7 @@ pub use fast_memory_vector_provider::FastMemoryVectorProviderAsync;
 pub mod fast_memory_quant_vector_provider;
 pub use fast_memory_quant_vector_provider::FastMemoryQuantVectorProviderAsync;
 
+pub(crate) mod postprocess;
```
hildebrandmw: Nit: shuffling diff?

diskann-bf_tree/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+bf-tree.workspace = true
+bytemuck = { workspace = true, features = ["must_cast"] }
+diskann.workspace = true
+diskann-providers.workspace = true
+diskann-quantization = { workspace = true, features = ["flatbuffers"] }
+diskann-utils.workspace = true
+diskann-vector.workspace = true
+half = { workspace = true, features = ["bytemuck", "num-traits"] }
+futures-util.workspace = true
+polonius-the-crab = "0.5.0"
```
hildebrandmw: Don't need `polonius-the-crab` now that the caching provider is gone.

diskann-bf_tree/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+ * Licensed under the MIT license.
+ */
+
+//! BfTree-based data provider for DiskANN async indexes.
+//!
+//! This crate provides a [`BfTree`](bf_tree::BfTree)-backed implementation of the DiskANN
+//! [`DataProvider`](diskann::provider::DataProvider) trait, enabling indexes that can
+//! transparently spill to disk for datasets larger than available memory.
+
+pub mod provider;
```
hildebrandmw: If the only contents here are a directory, we can move away from the madness of the current inmem and flatten the hierarchy even further. What about something like this:
```
diskann-bftree/
  src/
    lib.rs
    provider.rs
    vectors.rs
    neighbors.rs
    quant.rs
```
And the following renames:
* `BfTreeProvider` -> `Provider: Users would import as `diskann_bftree::Provider`, which IMO is completely unambiguous.
* `NeighborProvider` -> `Neighbors`.
* `VectorProvider` -> `Vectors`.
* `QuantVectorProvider` -> `QuantVectors`.

diskann-bf_tree/src/provider/vector_provider.rs
```diff
@@ -10,13 +10,13 @@ use std::marker::PhantomData;
...
 use diskann::{
-    ANNError, ANNErrorKind, ANNResult,
     utils::{ErrorToVectorId, TryIntoVectorId, VectorId, VectorRepr},
+    ANNError, ANNErrorKind, ANNResult,
 };
 use thiserror::Error;
 
-use super::super::common::TestCallCount;
 use super::ConfigError;
+use diskann_providers::model::graph::provider::async_::common::TestCallCount;
```
hildebrandmw: Maybe either duplicate `TestCallCount` or drop entirely? Alternatively, we can move this into the `Context` used by the bf-tree provider and always gather these metrics. It's generally pretty useful to have these things.

diskann-bf_tree/src/provider/vector_provider.rs
```diff
@@ -191,6 +191,11 @@ impl<T: VectorRepr, I: VectorId> VectorProvider<T, I> {
         self.get_vector_into(i, &mut vector)?;
         Ok(vector)
     }
+
+    pub(crate) fn delete_vector(&self, i: usize) {
+        let key = bytes_of::<usize>(&i);
```
hildebrandmw: One cleanup to consider: instead of spamming `bytes_of` everywhere, we can do something like
```
trait AsKey {
    fn as_key(&self) -> &[u8]
}

impl AsKey for usize {
    fn as_key(&self) -> &[u8] {
        bytemuck::bytes_of(self)
    }
}
```
Basically some way of centralizing the implementation. Also, should we consider either using `u32` or `u64` instead of `usize`?

diskann-bf_tree/src/provider/quant_vector_provider.rs
```diff
@@ -9,63 +9,55 @@ use std::sync::Arc;
...
+        DistanceComputer, Opaque, OpaqueMut, Quantizer, QueryComputer, QueryLayout,
     },
-    utils::BridgeErr,
 };
+use diskann_vector::{distance::Metric, PreprocessedDistanceFunction};
+
+use super::ConfigError;
+use diskann_providers::model::graph::provider::async_::common::TestCallCount;
+
+pub struct QuantQueryComputer(pub QueryComputer<GlobalAllocator>);
```
hildebrandmw: Perhaps keep the inner computer non-public? We should also add a `Debug` bound here - bug me to fix `diskann-quantization` to add debug implementations to the inner `QueryComputer`.

diskann-bf_tree/src/provider/quant_vector_provider.rs
```diff
@@ -9,63 +9,55 @@ use std::sync::Arc;
...
+    }
+}
 
 pub struct QuantVectorProvider {
     quant_vector_index: BfTree,
     max_vectors: usize,
     num_start_points: usize,
-    pub pq_chunk_table: Arc<FixedChunkPQTable>,
-    metric: Metric,
+    pub quantizer: Arc<Poly<dyn Quantizer>>,
```
hildebrandmw: Looking at this again, I don't think we need this to be an `Arc`. We only needed it for PQ because we hadn't worked out the `Distance/QueryDistance` APIs yet.

Also, the quantizer should probably not be public. Not sure why it was originally.

diskann-bf_tree/src/provider/quant_vector_provider.rs
```diff
@@ -9,63 +9,55 @@ use std::sync::Arc;
...
-    vec_pool: Arc<ObjectPool<Vec<f32>>>,
 }
 
-type DistanceComputer = pq::distance::DistanceComputer<Arc<FixedChunkPQTable>>;
-type QueryComputer = pq::distance::QueryComputer<Arc<FixedChunkPQTable>>;
-
 impl QuantVectorProvider {
     pub fn new_with_config(
-        dist_metric: Metric,
+        _dist_metric: Metric,
```
hildebrandmw: We own the constructor ... we can remove arguments that are no longer used. We have not API compatibility to maintain.

diskann-bf_tree/src/provider/quant_vector_provider.rs
```diff
@@ -9,63 +9,55 @@ use std::sync::Arc;
...
+        self.0
+            .evaluate_similarity(Opaque::new(x))
+            .expect("spherical query distance failed")
+    }
+}
 
 pub struct QuantVectorProvider {
     quant_vector_index: BfTree,
     max_vectors: usize,
     num_start_points: usize,
```
hildebrandmw: Do we even need `max_vectors/num_start_points`? BFTree is a key-value store, so I don't think it's needed at all at this level.

diskann-bf_tree/src/provider/quant_vector_provider.rs
```diff
@@ -105,33 +94,39 @@ impl QuantVectorProvider {
...
-            self.metric,
-            &T::as_f32(query).into_ann_result()?,
-            Some(self.vec_pool.clone()),
-        )
+        let query_f32 = T::as_f32(query).into_ann_result()?;
+        let inner = self
+            .quantizer
+            .fused_query_computer(
+                &query_f32,
+                QueryLayout::SameAsData,
```
hildebrandmw: Universally using `SameAsData` will lead to poor recall for low-bit quantizers. This should default to `FullPrecision`, maybe with a hook to set as `SameAsData` during insert.

diskann-bf_tree/src/provider/quant_vector_provider.rs
```diff
@@ -261,70 +263,103 @@ impl QuantVectorProvider {
...
+            SupportedMetric::SquaredL2,
+            PreScale::None,
+            &mut rng,
+            GlobalAllocator,
+        )
+        .unwrap();
+
+        let imp = iface::Impl::<1>::new(quantizer).unwrap();
+        let poly = Poly::new(imp, GlobalAllocator).unwrap();
+        poly!(iface::Quantizer, poly)
```
hildebrandmw: Small thing: this can be simplified to
```rust
let imp = iface::Impl::<1>::new(quantizer).unwrap();
poly!(iface::Quantizer, imp, GlobalAllocator).unwrap();
```

diskann-bf_tree/src/provider/bf_tree_provider.rs
```diff
@@ -32,27 +36,21 @@ use diskann::{
...
+use diskann_vector::{distance::Metric, DistanceFunction};
 
-use crate::storage::{LoadWith, PQStorage, SaveWith};
-
-use crate::storage::{StorageReadProvider, StorageWriteProvider};
+use super::{
+    neighbor_provider::NeighborProvider, quant_vector_provider::QuantVectorProvider,
+    vector_provider::VectorProvider,
+};
+use diskann_providers::model::graph::provider::async_::{
```
hildebrandmw: I'd highly recommend distances this from the details of the inmem index as much as possible. This will help tremendously when we finally fix the inmem provider.

diskann-bf_tree/src/provider/bf_tree_provider.rs
```diff
@@ -70,7 +68,7 @@ use crate::storage::{StorageReadProvider, StorageWriteProvider};
 /// * `Q`: The full type of the quant vector store. This is not constrained by a trait and
 ///   rather relies on implementation for several concrete types, including:
 ///
-///   - [`BfTreeQuantVectorProviderAsync`]: A Bf-Tree based PQ-based quantized vector store.
+///   - [`BfTreeQuantVectorProviderAsync`]: A Bf-Tree based spherical quantized vector store.
```
hildebrandmw: Run
```
cargo rustdoc --package diskann-bf_tree -- -D rustdoc::broken-intra-doc-links
```
to help find outdated intra-doc links.

diskann-bf_tree/src/provider/bf_tree_provider.rs
```diff
@@ -88,22 +86,18 @@ use crate::storage::{StorageReadProvider, StorageWriteProvider};
...
-///   - Insertion: Quantized vectors are used during the search phase. During the pruning
-///     phase, a hybrid of quantized and full-precision vectors are used.
+/// * [`Quantized`]: Performs all operations (search, pruning, insert) entirely in the
+///   quantized space using spherical distance functions. Post-processing copies candidate
+///   IDs forward without reranking. Fastest option — full-precision vectors are not
+///   touched at query time.
 ///
-///     The ratio of full-precision and quantized vectors is controlled by the
-///     `max_fp_vecs_per_prune` parameter, which adjusts the implementation of [`Fill`].
+/// * [`Hybrid`]: Searches and prunes in the quantized space (same as [`Quantized`]), but
```
hildebrandmw: Why still keep hybrid? `Quantized`, we certainly need. But `Hybrid` is targeting the hybrid PQ pruning, which we do not need.

diskann-bf_tree/src/provider/bf_tree_provider.rs
```diff
@@ -310,10 +253,10 @@ pub struct BfTreeProviderParameters {
     pub graph_params: Option<GraphParams>,
 }
 
-pub type Index<T, D = NoDeletes> = Arc<DiskANNIndex<BfTreeProvider<T, NoStore, D>>>;
-pub type QuantIndex<T, Q, D = NoDeletes> = Arc<DiskANNIndex<BfTreeProvider<T, Q, D>>>;
+pub type Index<T> = Arc<DiskANNIndex<BfTreeProvider<T, NoStore>>>;
```
hildebrandmw: Don't include these type aliases. These should probably live in user code.

diskann-bf_tree/src/provider/bf_tree_provider.rs
```diff
@@ -326,14 +269,9 @@ where
     /// * `delete_precursor`: A precursor type for the delete layer.
     /// * `neighbor_precursor`: A precursor type for the neighbor layer.
     ///   or the neighbor layer
-    pub fn new_empty<TQ, TD>(
-        params: BfTreeProviderParameters,
-        quant_precursor: TQ,
-        delete_precursor: TD,
-    ) -> ANNResult<Self>
+    pub fn new_empty<TQ>(params: BfTreeProviderParameters, quant_precursor: TQ) -> ANNResult<Self>
```
hildebrandmw: I'm not convinced that `new_empty` is a good idea, given the question "what do we do about start points". Is it possible to drop this method in favor of just `new`?

diskann-bftree/src/provider.rs
```diff
@@ -493,9 +417,65 @@ where
...
+        _context: &Self::Context,
+        gid: &Self::ExternalId,
+    ) -> impl std::future::Future<Output = Result<(), Self::Error>> + Send {
+        let id = *gid;
+
+        if let Err(e) = self.neighbor_provider.delete_vector(id) {
+            return std::future::ready(Err(e));
+        }
+        if let Err(e) = self.neighbor_provider.set_neighbors(id, &[]) {
+            return std::future::ready(Err(e));
```
hildebrandmw, line 410: Would this second condition be sufficient? Otherwise, there's a transient situation where concurrent accesses could see a deleted vector, which IIRC we don't handle gracefully yet.

As an aside, we should change the `NeighborAccessor` traits to have `ToRanked` errors instead of `ANNError`.

diskann-bf_tree/src/provider/bf_tree_provider.rs
hildebrandmw: Changing the semantics of deleted vector to being deleted in the underlying BFTree means we can no longer use the `Panics` error. Instead, we need a true `ToRanked` error with a transient when accessing deleted vectors. And this needs to get propagated all the way down to the inherent access methods on the various providers to avoid overheads associated with `ANNError`.

diskann-bf_tree/src/provider/bf_tree_provider.rs
```diff
@@ -1182,452 +1103,518 @@ where
...
-/// This type implements the following traits:
-///
-/// * [`Accessor`] for the [`BfTreeProvider`].
-/// * [`BuildDistanceComputer`] for computing distances among [`distances::pq::Hybrid`]
-///   element types.
-/// * [`Fill`] for populating a mixture of full-precision and quant vectors.
-///
-pub struct HybridAccessor<'a, T, D>
+// Pass-through fill — returns `&Self` which directly accesses the underlying provider.
+impl<T> workingset::Fill<PassThrough> for QuantAccessor<'_, T>
```
hildebrandmw: We really really don't want to use `PassThrough` here. While the prune algorithm caches the result internally to avoid refetching for every pairwise distance, `Map` will allow vectors to be reused across multiple prunes.

diskann-bf_tree/src/provider/bf_tree_provider.rs
```diff
@@ -1182,452 +1103,518 @@ where
...
 where
     T: VectorRepr,
-    D: AsyncFriendly,
+    Q: AsyncFriendly,
 {
-    type DistanceComputerError = ANNError;
-    type DistanceComputer = distances::pq::HybridComputer<T>;
+    type QueryComputer = T::QueryDistance;
+    type SearchAccessor<'a> = FullAccessor<'a, T, Q>;
+    type SearchAccessorError = Panics;
```
hildebrandmw: I think this can be `Infallible` instead?

diskann-bf_tree/src/provider/bf_tree_provider.rs
```diff
@@ -1182,452 +1103,518 @@ where
...
-        // Store the reranked results.
         std::future::ready(Ok(output.extend(reranked)))
     }
 }
 
-/// Perform a search entirely in the quantized space.
-impl<T, D> SearchStrategy<BfTreeProvider<T, QuantVectorProvider, D>, &[T]> for Hybrid
+/// Hybrid strategy: search in quantized space, rerank with full-precision distances,
+/// prune in quantized space.
+impl<T> SearchStrategy<BfTreeProvider<T, QuantVectorProvider>, &[T]> for Hybrid
```
hildebrandmw: We can completely ditch `Hybrid` as a strategy, just using `FullPrecision` and `Quantized`. Quantized search should still rerank, but that's generally expected.

hildebrandmw (changes requested):
Thanks Jordan, we can do quite a bit of cleanup on the crate structure and item names which I think would really start to tighten up the crate.

One question I had for you is what the expected end state is: are we targeting a full clean up or an incremental, non-published state while we continue working? I ask because the bf-tree provider inherited a bit of cruft from earlier iterations of the code, and I'm wondering how aggressiver we want to be in this PR in particular, considering that we still need integration tests and benchmarks.

Cargo.toml
JordanMaples: oh strange, i missed that. Also, yeah this is like the third iteration on a name. I'll just make it diskann-bftree.

diskann-providers/src/model/graph/provider/async_/mod.rs
JordanMaples: I'll move that back.

diskann-bf_tree/src/lib.rs
JordanMaples: sure, i can do that.

diskann-bf_tree/src/provider/bf_tree_provider.rs
JordanMaples: kk I'll remove. I was just trying to keep as much of the same surface area as possible during the transition.

diskann-bf_tree/src/provider/quant_vector_provider.rs
JordanMaples: kk, removing.

diskann-bf_tree/src/provider/quant_vector_provider.rs
JordanMaples: I'll look into how it's used and get back to you

diskann-bf_tree/src/provider/bf_tree_provider.rs
JordanMaples: consider them gone

diskann-bf_tree/src/provider/bf_tree_provider.rs
JordanMaples: Alright, I'll do that instead of silently proceeding

diskann-bf_tree/src/provider/bf_tree_provider.rs
JordanMaples: I'll switch to Map

diskann-bftree/src/provider.rs
```diff
@@ -493,9 +417,65 @@ where
     }
 }
 
+impl<T, Q> Delete for BfTreeProvider<T, Q>
+where
+    T: VectorRepr,
+    Q: AsyncFriendly + QuantDelete,
+{
+    fn release(
```
harsha-simhadri, line 393: Who is the intended caller of this `release` function?

harsha-simhadri (commented):
Thanks for the PR, Jordan.

Could you wire it through diskann-benchmarks and run it in streaming mode to replicate the recall results in the IP-diskann paper? That would be good E2E validation of the provider. Please all latency and total runtime to analyze performance.

diskann-bftree/src/provider.rs
hildebrandmw: It was originally intended as a mechanism for completely a soft-delete (when all incoming edges where removed). However, we have generally moved away from soft-deletes in the usual sense, so this method can probably be removed.

diskann-bf_tree/src/provider/vector_provider.rs
JordanMaples: done

diskann-bf_tree/src/provider/vector_provider.rs
JordanMaples: duping for now, while the context would probably be the better solution, it's a lot of plumbing.

diskann-bf_tree/src/provider/quant_vector_provider.rs
JordanMaples: Removed it for quant. I'll check the vector provider tomorrow.

diskann-bf_tree/src/provider/quant_vector_provider.rs
JordanMaples: switched it

diskann-bf_tree/src/provider/quant_vector_provider.rs
JordanMaples: done

diskann-bf_tree/src/provider/bf_tree_provider.rs
JordanMaples: Only things left from inmem are the save/load and unwraperr stuff. I'll look into that tomorrow.

diskann-bf_tree/src/provider/bf_tree_provider.rs
JordanMaples: fixed.

diskann-bf_tree/Cargo.toml
JordanMaples: gone

diskann-bf_tree/src/provider/quant_vector_provider.rs
JordanMaples: I'll take care of that tomorrow

diskann-bf_tree/src/provider/quant_vector_provider.rs
JordanMaples: removed the arc and made pub(crate)

diskann-bf_tree/src/provider/bf_tree_provider.rs
JordanMaples: done

diskann-bf_tree/src/provider/bf_tree_provider.rs
JordanMaples: done

JordanMaples:
> Thanks for the PR, Jordan.
> 
> Could you wire it through diskann-benchmarks and run it in streaming mode to replicate the recall results in the IP-diskann paper? That would be good E2E validation of the provider. Please all latency and total runtime to analyze performance.

I'll take care of that after I finish addressing Mark's feedback.

diskann-bftree/src/provider.rs
JordanMaples: I'll remove the first conditional then and call that good.

diskann-bftree/src/provider.rs
JordanMaples: can we resolve this and address removing release in a different pr?

diskann-bftree/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+[package]
+name = "diskann-bftree"
+version.workspace = true
+description.workspace = true
+authors.workspace = true
+documentation.workspace = true
+license.workspace = true
+edition.workspace = true
```
harsha-simhadri, line 10: add `publish.workspace = false`

JordanMaples: done

diskann-bftree/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    pub fn get(&self) -> usize {
+        self.count.load(std::sync::atomic::Ordering::Relaxed)
+    }
+}
+
+#[cfg(not(test))]
+#[allow(dead_code)]
+pub(crate) struct TestCallCount {}
```
hildebrandmw, line 148: Suggestion: Not necessary in this PR, but we can do something like [this](https://github.com/microsoft/DiskANN/blob/d7bf689a99fe9849419481b0a0d2bd4a4f4ba450/diskann/src/internal/counter.rs#L48-L56) to always collect call counts with minimal overhead. Just putting it on your radar.

diskann-bftree/src/provider.rs
```diff
@@ -1197,341 +1069,139 @@ where
...
-        self.provider.neighbors()
+impl<'a> From<Opaque<'a>> for OwnedOpaque {
+    fn from(value: Opaque<'a>) -> Self {
+        OwnedOpaque(value.to_vec())
     }
 }
 
-impl<T, D> Accessor for HybridAccessor<'_, T, D>
+// Pass-through view — reads quantized vectors directly from the provider.
+impl<T> workingset::View<u32> for &QuantAccessor<'_, T>
```
hildebrandmw: We can remove this now that we're using `workingset::Map`.

diskann-bftree/src/provider.rs
```diff
@@ -310,82 +228,67 @@ pub struct BfTreeProviderParameters {
...
-    ///   or the neighbor layer
-    pub fn new_empty<TQ, TD>(
-        params: BfTreeProviderParameters,
-        quant_precursor: TQ,
-        delete_precursor: TD,
-    ) -> ANNResult<Self>
+    ///
+    /// # Type Constraints
+    /// * `Self: StartPoint<T>` - The provider must implement the `StartPoint` trait.
+    pub fn new_empty<TQ>(params: BfTreeProviderParameters, quant_precursor: TQ) -> ANNResult<Self>
```
hildebrandmw: I know there was some debate about whether or not to expose this. My vote is to keep it private/unexported for now. It's less of a potential API break to expose it once we have the whole start point situations ironed out then to walk it back.

hildebrandmw (commented):
One more small round, then looks good to me. Thanks!

diskann-bftree/src/lib.rs
JordanMaples: kk, I'll switch this over in the follow-up that enables benchmarking.

diskann-bftree/src/provider.rs
JordanMaples: alright, i'll remove the pub

hildebrandmw (approved):

diskann-bftree/src/provider.rs
```diff
@@ -2285,28 +1943,320 @@ where
...
+        assert!(!neighbor_ids.contains(&4u32));
+    }
 
     #[tokio::test]
     async fn test_data_provider_and_delete_interface() {
         let ctx = &DefaultContext;
-        let provider = BfTreeProvider::new_empty(
+        let num_start_points = 2;
+        let dim = 5;
+        let start_points = Matrix::new(Init(|| 0.0f32), num_start_points, dim);
```
harsha-simhadri: Does this produce two starting points that are all zero-vectors?

diskann-bftree/src/provider.rs
```diff
@@ -2285,28 +1943,320 @@ where
...
+    }
+
+    #[tokio::test]
+    async fn test_quantized_index_multi_insert_search() {
+        let index = create_quant_index();
+        let ctx = &DefaultContext;
+
+        let mut counter = 0.0f32;
+        let data = Matrix::new(
+            Init(move || {
```
harsha-simhadri: What is the rationale/intuition behind testing this kind of vectors? Does it correspond to any realistic embedding pattern or to a structured grid (such as what we are using in test providers)?

diskann-bftree/src/provider.rs
JordanMaples: Yes. This is creating two zero-vector start points. Start point doesn't matter for this test. We're testing delete functionality here. We don't need any sort of accurate search, just that the deletes happened.

diskann-bftree/src/provider.rs
JordanMaples: This is a smoke test that the multi_insert interface is working and you can perform a search on it. There's no real rationale as to why the vectors were created as they were. As of yet, I don't have any intuition on what a realistic embedding pattern looks like. I didn't sample any sort of real data shapes for this because ultimately it doesn't matter for the sake of a smoke test.

 I'll switch the vectors to be uniformly initialized, query on 3 as I'm doing here, and assert that the elements returned are ids 1,2,3,4,5.

diskann-bftree/src/provider.rs
harsha-simhadri: in general, its not a good idea to pick two overlapping start points. If it does not matter, could you please pick two spearate start points or remove one ? approving PR, feel free to address this and merge. thanks

diskann-bftree/src/provider.rs
harsha-simhadri: understand. I am ok to merge, am trying to understand why these data patterns diverge from test providers.

harsha-simhadri (approved):

diskann-bftree/src/provider.rs
JordanMaples: Alright, I made them non-overlapping.

harsha-simhadri (approved):

PR #1025, opened 2026-05-06
Add v4 distance kernels

m3hm3t:

### DIFF
--- a/diskann-quantization/src/bits/distances.rs
+++ b/diskann-quantization/src/bits/distances.rs
@@ -564,6 +564,154 @@ impl Target2<diskann_wide::arch::x86_64::V3, MathematicalResult<u32>, USlice<'_,
 ///
 /// # Implementation Notes
 ///
+/// This implementation is optimized for x86 with the AVX-512 vector extension.
+/// It scales the V3 approach to 512-bit registers: we load data as `u32x16`, shift and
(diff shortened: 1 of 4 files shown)


diskann-quantization/src/bits/distances.rs
```diff
@@ -558,6 +558,154 @@ impl Target2<diskann_wide::arch::x86_64::V3, MathematicalResult<u32>, USlice<'_,
...
+
+                // SAFETY: The same logic applies to `y` because:
+                // 1. It has the same type as `x`.
+                // 2. We've verified that it has the same length as `x`.
+                let vy = unsafe { u32s::load_simd(arch, py_u32.add(i)) };
+
+                let wx: i16s = (vx & mask).reinterpret_simd();
+                let wy: i16s = (vy & mask).reinterpret_simd();
+                let d = wx - wy;
+                s0 = s0.dot_simd(d, d);
```
hildebrandmw, line 624: ~~Thanks for taking the initiative! You can do much better by targeting [this method](https://github.com/microsoft/DiskANN/blob/c50fb2bfd1b0b04a2390f71fe1912deea32a53a5/diskann-wide/src/arch/x86_64/v4/i32x16_.rs#L113-L120). Unlike the dot-product from 16-bit integers, it does the dot product of 64 8-bit numbers **and** the accumulation with a `i32x16` in a single instruction :smile:~~

Ignore me. It's too late in the day. Forgot this was L2.

PR #1027, opened 2026-05-06
Multi-vector MaxSim benchmark with BYOTE factory

### LINKED ISSUE #993: Add benchmarking suite for multi-vector operations
# Add benchmarking suite for multi-vector operations

## Description
Establish a rigorous benchmarking framework to validate the performance and scalability of the new multi-vector architecture. This suite will measure the standalone API's throughput, latency, allowing us to compare our specialized microkernels and memory layouts against established baselines as they are introduced.

## Tasks
- [ ] Set up the core benchmarking harness for the standalone multi-vector distance API.
- [ ] Integrate naive or external baselines for comparative performance tracking.
- [ ] Benchmark the core distance functions across supported datatypes (e.g., `f32`, `f16`) and quantizations.
- [ ] Establish and document the methodology for tracking future hardware-specific microkernel optimizations.

suri-kumkaran:
Adds the multi-vector MaxSim distance benchmark to `diskann-benchmark`
behind a `multi-vector` Cargo feature (off by default). Public library
API stays minimal: internals (`Kernel<A>`, `tiled_reduce`, layouts,
`TileBudget`) remain private; the only external surface is the generic
`build_max_sim<T>` factory (gated by the sealed `MaxSimElement` trait)
plus the `MaxSimKernel<T>` / `Erase<T>` traits.

The factory follows the BYOTE ("Bring your own type erasure") pattern
from RFC #1068.

`MaxSim::new` is now infallible (`-> Self`, no longer `-> Result`) and
`MaxSimError::BufferLengthIsZero` is removed. With the kernel boundary
enforcing the only meaningful precondition (`scores.len() ==
nrows()`), an empty `scores` buffer is a defined no-op — the standalone
"can't be empty" rule was defensive code that caught nothing the size
check doesn't already catch.

## JSON shape

```json
{
  "type": "multi-vector-op",
  "content": {
    "element_type": "float32",
    "isa": "x86-64-v3",
    "runs": [ { "num_query_vectors": 8, "num_doc_vectors": 32, "dim": 128, "loops_per_measurement": 200, "num_measurements": 50 } ]
  }
}
```

## Suggested review order

1. **`distance/isa.rs`** — `MaxSimIsa`, `NotSupported`, `is_available()`.
2. **`distance/kernel.rs`** — `MaxSimKernel<T>` (`Result`-returning),
   `Erase<T>`, `BoxErase`.
3. **`distance/factory.rs`** — sealed `MaxSimElement` + `build_max_sim`;
   `Prepared` (SIMD) and `ReferenceKernel`; parity, size-mismatch,
   zero-doc, and zero-query tests.
4. **`distance/max_sim.rs`** — `MaxSim::new` infallible;
   `BufferLengthIsZero` removed.
5. **`multi_vector/{mod,distance/mod}.rs`** — public re-exports; verify
   `QueryComputer` is gone.
6. **`benchmark/inputs/multi_vector.rs`** — `MultiVectorOp`, `BenchIsa`
   shadow enum, `From<BenchIsa> for MaxSimIsa`.
7. **`benchmark/backend/multi_vector/{mod,driver,kernels}.rs`** —
   `cfg_if!` gate; `Kernel<T>` carrier registered per `MaxSimElement`;
   `try_match` rejects unavailable ISAs at job selection.
8. **Glue + tests** — `backend/mod.rs`, `inputs/mod.rs`, `Cargo.toml`
   feature, `main.rs` integration tests, JSON fixtures.


diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+    fn check_deserialization(&mut self, _checker: &mut Checker) -> Result<(), anyhow::Error> {
+        Ok(())
+    }
+}
+
+macro_rules! write_field {
+    ($f:ident, $field:tt, $($expr:tt)*) => {
+        writeln!($f, "{:>18}: {}", $field, $($expr)*)
+    }
+}
```
hildebrandmw: Since #996, manual alignment shouldn't be needed any more.

diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+
+////////////////////////////
+// Benchmark Registration //
+////////////////////////////
+
+fn register_benchmarks_impl(dispatcher: &mut diskann_benchmark_runner::registry::Benchmarks) {
+    macro_rules! register {
+        ($impl:ident, $t:ty, $tag:literal) => {
+            dispatcher.register_regression($tag, Kernel::<$impl, $t>::new());
+        };
```
hildebrandmw: Nit: The macro is offering only a moderate benefit in terseness. Might be better off without it?

diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+    }
+
+    // Optimized (architecture-dispatched QueryComputer).
+    register!(Optimized, f32, "multi-vector-op-f32-optimized");
```
hildebrandmw: We should probably provide a way of overriding the runtime architecture to test both AVX512 and AVX2 kernels on the same machine. Otherwise, we have no way to compare such implementations against each other ...

diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+                            write!(f, "matched {}", impl_)
+                        } else {
+                            write!(f, "expected {}, got {}", Implementation::$variant, impl_)
+                        }
+                    }
+                }
+            }
+        }
+    };
+}
```
hildebrandmw: This might be cleaner to get rid of `DispatchRule` and use an entirely bespoke, smaller trait. Like:
```
impl TryFrom<Implementation> for Optimized {
    type Error = FailureScore;
    fn try_from(i: Implementation) -> Result<Self, Self::Error> {
        match i {
            Implementation::Optimized => Ok(Self),
            _ => Err(FailureScore(1)),
        }
    }
}
```
and write matching via `TryFrom`.

diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+
+///////////////////
+// Data fixtures //
+///////////////////
+
+const RNG_SEED: u64 = 0x12345;
+
+struct Data<T> {
+    query_data: Box<[T]>,
```
hildebrandmw: We spend so much time writing matrix abstractions. Why not use them? If the constructors are insufficient, that's a sign we have work to do!

diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+                run_loops(run, || {
+                    computer.max_sim(doc, &mut scores);
+                    std::hint::black_box(&mut scores);
+                })
+            }
+        };
+        results.push(result);
+    }
+    Ok(results)
+}
```
hildebrandmw: These two `run` loops are nearly identical. Perhaps a tweak to factor out the common parts? Maybe consider keeping the loop nest the same, but taking a trait object (to reduce compile times) for the distance implementation?

diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+                input: &MultiVectorOp,
+            ) -> Result<Vec<RunResult>, anyhow::Error> {
+                run_reference::<$t>(input)
+            }
+        }
+    };
+}
+
+impl_kernel_for!(f32);
+impl_kernel_for!(f16);
```
hildebrandmw: Since the abstractions these macros are stamping were also introduced in this PR, consider whether or not there's a simpler way to express things via generics that doesn't require macros.

diskann-benchmark-multi-vector/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+| Tag                              | Element | Implementation       |
+| -------------------------------- | ------- | -------------------- |
+| `multi-vector-op-f32-optimized`  | `f32`   | `QueryComputer`      |
+| `multi-vector-op-f16-optimized`  | `f16`   | `QueryComputer`      |
+| `multi-vector-op-f32-reference`  | `f32`   | `Chamfer` / `MaxSim` |
+| `multi-vector-op-f16-reference`  | `f16`   | `Chamfer` / `MaxSim` |
+
+The **optimized** path constructs a `QueryComputer` once per shape (which
+internally selects the best available SIMD kernel for the host) and calls
```
hildebrandmw: If we're using this as a development tool ... it's probably a good idea to allow an override for the architecture.

diskann-benchmark-multi-vector/README.md
```diff
@@ -0,0 +1,136 @@
...
+### Run benchmarks
+
+`run` executes every job in the input file and writes per-measurement
+latencies plus percentiles to the output file:
+
+```bash
+bench run --input-file examples/multi-vector.json --output-file before.json
+```
+
+### Regression check workflow
```
hildebrandmw: The section on regression checks seems to be just a copy+paste of how `diskann-benchmark-runner` works. Does this need to be spelled out in the README like this?

diskann-benchmark-multi-vector/README.md
```diff
@@ -0,0 +1,136 @@
...
+- `examples/test.json` — minimal smoke configuration consumed by the
+  integration tests.
+- `examples/tolerance.json` — default regression thresholds.
+
+## Registered kernels
+
+The crate registers four kernels — one per `(element_type, implementation)`
+pair:
+
+| Tag                              | Element | Implementation       |
```
hildebrandmw: While it's fine to have a list like this, the chances of this getting updated as kernels are added is nearly zero. Maybe instead rely on the `inputs` behavior of the  runner APP and ensure that the returned descriptions are sufficient?

hildebrandmw (commented):
Thanks Suryansh, this will be useful for performance optimization. One big question I had is whether or not this needs to be its own crate, or whether it can be folded into `diskann-benchmark` (which will need some work to put the normal indexing code behind a feature gate to keep compile times low) or as a binary in `diskann-quantization` directly.

The pattern established by `diskann-benchmark-simd` is a little unfortunate so now is the time to think about our approach before we have to deal with a crate proliferation. We also might be able to bury all benchmark related binary crates in a subdirectory to keep the top level organized.

In either case, I don't think we should publish this to `crates.io` - and we should probably stop publishing `diskann-benchmark-simd` as well.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ members = [
     "diskann-benchmark-runner",
     "diskann-benchmark-core",
     "diskann-benchmark-simd",
+    "diskann-benchmark-multi-vector",
```
suhasjs: Do we want to make this a high-level crate? I think it would be good if we had this added as a 'benchmark' under `diskann-benchmark`..

diskann-benchmark-multi-vector/README.md
suhasjs: +1

Cargo.toml
arrayka: How many top-level crates do we actually want to have here? If we end up with 10–20 diskann-benchmark-* crates, is that going to be a problem?

As an alternative, could we keep benchmarks inside each crate: e.g., add a `bench` subfolder and use [[bench]] with `harness = false`, pulling in something like diskann-benchmark-runner as a dev-dependency? That feels more idiomatic in Rust.

diskann-benchmark-multi-vector/README.md
```diff
@@ -0,0 +1,136 @@
...
+internally selects the best available SIMD kernel for the host) and calls
+`chamfer` / `max_sim` inside the timed loop. The **reference** path drives
+the `Chamfer` / `MaxSim` fallback used by the `multi_vector` unit tests —
+useful both as a numerical ground truth and as a baseline to measure SIMD
+speedups against.
+
+## Time normalization
+
+Per-measurement latency is normalized to **nanoseconds per inner-product
+call**, abbreviated `ns/IP`:
```
arkrishn94: Maybe I'm overthinking this but some CPU cycles are spent on reduction here too for `Chamfer`. 

Why not just report the total time here? Users can profile themselves if they want to.

diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+            min_time_regression: EXAMPLE,
+        })?)
+    }
+}
+
+/// Per-run comparison result showing before/after percentile differences.
+#[derive(Debug, Serialize)]
+struct Comparison {
+    run: Run,
+    tolerance: MultiVectorTolerance,
```
arkrishn94: nit: `Tolerance` here should do? The path should allow us to distinguish?

diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+    before_min: f64,
```
arkrishn94: nit: `Positive`?

diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+    T: 'static,
+{
+    type Input = MultiVectorOp;
+    type Output = Vec<RunResult>;
+
+    fn try_match(&self, from: &MultiVectorOp) -> Result<MatchScore, FailureScore> {
+        let mut failscore: Option<u32> = None;
+        if datatype::Type::<T>::try_match(&from.element_type).is_err() {
+            *failscore.get_or_insert(0) += 10;
+        }
```
arkrishn94: I forget - how did we come up a failure score of 10 for this? Should this be something large?

diskann-quantization/src/multi_vector/matrix.rs
```diff
@@ -514,6 +526,22 @@ where
...
 }
 
+// SAFETY: The implementation uses guarantees from `Box` to ensure that the pointer
+// initialized by it is non-null and properly aligned to the underlying type.
+unsafe impl<T, F> NewOwned<Init<F>> for Standard<T>
+where
+    T: Copy,
+    F: FnMut() -> T,
+{
+    type Error = crate::error::Infallible;
```
arkrishn94: Does it make sense to fold the default initialization into this in the spirit of minimizing the number of constructor paths we expose to users?

diskann-benchmark/src/inputs/multi_vector.rs
```diff
@@ -0,0 +1,190 @@
...
+// Enum types //
+////////////////
+
+/// The two distance operations exposed by `QueryComputer`.
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum Operation {
+    Chamfer,
+    MaxSim,
+}
```
arkrishn94: I'm wondering if we should just expose a single operation here? I'm not sure folks will use `MaxSim` much and it might simplify some of your code here?

arkrishn94 (commented):
Thanks Suryansh, I left some relatively minor comments. Can we run this benchmark and post some example numbers in the PR description?

diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+    fn new() -> Self {
+        Self {
+            _type: std::marker::PhantomData,
+        }
+    }
+}
+
+#[derive(Debug, Error)]
+#[error("this kernel handles a different implementation than {0}")]
+pub(crate) struct ImplementationMismatch(Implementation);
```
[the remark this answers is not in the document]
suri-kumkaran: Outdated

diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+                    row.insert(format!("{:.3} %", change * 100.0), 6);
+                    if change > c.tolerance.min_time_regression.get() {
+                        row.insert("FAIL", 7);
+                    }
+                }
+                Err(err) => {
+                    row.insert("invalid", 6);
+                    row.insert(err, 7);
+                }
+            }
```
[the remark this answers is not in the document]
suri-kumkaran: Outdated

diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+                };
+
+                match relative_change(before_min, after_min) {
+                    Ok(change) => {
+                        if change > tolerance.min_time_regression.get() {
+                            passed = false;
+                        }
+                    }
+                    Err(_) => passed = false,
+                };
```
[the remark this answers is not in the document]
suri-kumkaran: Outdated

diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+
+        self.iter().enumerate().for_each(|(row, r)| {
+            let mut row = table.row(row);
+
+            let min_latency = r
+                .latencies
+                .iter()
+                .min()
+                .copied()
+                .unwrap_or(MicroSeconds::new(u64::MAX));
```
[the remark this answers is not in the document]
suri-kumkaran: Outdated

diskann-benchmark-multi-vector/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+        latencies.push(start.elapsed().into());
+    }
+
+    let percentiles = percentiles::compute_percentiles(&mut latencies).unwrap();
+    RunResult {
+        run: run.clone(),
+        latencies,
+        percentiles,
+    }
+}
```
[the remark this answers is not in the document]
suri-kumkaran: Outdated

diskann-benchmark-multi-vector/src/bin.rs
```diff
@@ -0,0 +1,96 @@
...
+    fn check_verify() {
+        let input_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("examples")
+            .join("test.json");
+        let tolerance_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("examples")
+            .join("tolerance.json");
+
+        let stdout = run_check_test(&input_path, &tolerance_path);
+        println!("stdout = {}", stdout);
```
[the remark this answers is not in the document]
suri-kumkaran: Outdated

diskann-benchmark/src/inputs/multi_vector.rs
suri-kumkaran: Yes I choose to only benchmark `MaxSim` now as it's a more lower level API and Chamfer just uses `MaxSim` as it is.

diskann-quantization/src/multi_vector/matrix.rs
suri-kumkaran: Oh I added `from_fn` constructor to `Mat<Standard<T>>` rather than seperate `Init` struct.

diskann-quantization/src/multi_vector/distance/factory.rs
```diff
@@ -0,0 +1,360 @@
...
+            isa,
+            reason: "aarch64 target only",
+        }),
+        MaxSimIsa::Reference => Ok(erase.erase(ReferenceKernel::<f32>::new(query))),
+    }
+}
+
+/// Build a multi-vector MaxSim kernel for `half::f16` queries. Same contract
+/// as [`build_max_sim_f32`].
+pub fn build_max_sim_f16<E: Erase<half::f16>>(
```
hildebrandmw: Baking the element type into the function name makes this annoying to call in a generic context (anyone trying to do that will have to introduce a shim trait to do this dispatch). Perhaps introducing this as a trait instead?

This is already showing itself with the near duplicated functionality in `diskann-benchmark`, forcing the use of a macro.

diskann-quantization/src/multi_vector/distance/query_computer/mod.rs
```diff
@@ -1,290 +0,0 @@
...
-fn build_prepared<T: Copy + Default, A, const GROUP: usize>(
-    arch: A,
-    query: MatRef<'_, Standard<T>>,
-) -> Prepared<A, BlockTransposed<T, GROUP>> {
-    let prepared = BlockTransposed::<T, GROUP>::from_matrix_view(query.as_matrix_view());
-    Prepared { arch, prepared }
-}
-
-#[cfg(test)]
-mod tests {
```
hildebrandmw, line 111: We should keep a version of these tests to ensure the factory is doing something sane.

diskann-benchmark/src/backend/multi_vector/driver.rs
```diff
@@ -0,0 +1,318 @@
...
+    }
+}
+
+//////////////////////
+// Timing harness   //
+//////////////////////
+
+fn run_loops<F>(run: &Run, mut body: F) -> RunResult
+where
+    F: FnMut(),
```
hildebrandmw: Would `&mut dyn FnMut()` work here?

diskann-benchmark/src/backend/multi_vector/driver.rs
```diff
@@ -0,0 +1,318 @@
...
+        std::hint::black_box(&mut scores);
+    })
+}
+
+//////////////////////
+// Result types     //
+//////////////////////
+
+#[derive(Debug, Clone, Copy)]
+pub(super) struct DisplayWrapper<'a, T: ?Sized>(pub(super) &'a T);
```
hildebrandmw: I think there's already a `DisplayWrapper` lurking somewhere in `crate::utils` you can use.

diskann-benchmark/src/backend/multi_vector/README.md
```diff
@@ -0,0 +1,43 @@
+# Multi-vector benchmark — kernel-author workflow
```
hildebrandmw: This feels like a really weird place for this. Maybe include it as module level documentation in `multi_vector/mod.rs`? As one advantage, links will be checked (but we need to turn on doclink checks in CI ...)

hildebrandmw (commented):
I took a quick pass.

diskann-benchmark-multi-vector/src/lib.rs
suri-kumkaran: Yes

diskann-benchmark-multi-vector/src/lib.rs
suri-kumkaran: done

diskann-benchmark-multi-vector/src/lib.rs
suri-kumkaran: Done.

diskann-benchmark-multi-vector/src/lib.rs
suri-kumkaran: Fixed, sorry for the miss.

diskann-benchmark-multi-vector/src/lib.rs
suri-kumkaran: Stale.

diskann-benchmark-multi-vector/src/lib.rs
suri-kumkaran: Fixed.

diskann-benchmark-multi-vector/README.md
suri-kumkaran: Fixed.

diskann-benchmark-multi-vector/README.md
suri-kumkaran: Removed.

Cargo.toml
suri-kumkaran: Merged to `diskann-quantization`.

diskann-benchmark-multi-vector/README.md
suri-kumkaran: Removed the `Chamfer` benchmark. The normalized timing is now more useful given the many different multi-vector configs, and it even allows us to compare the performance of dense vs. sparse use cases.

diskann-benchmark-multi-vector/src/lib.rs
suri-kumkaran: stale

diskann-benchmark-multi-vector/src/lib.rs
suri-kumkaran: stale

diskann-benchmark-multi-vector/README.md
suri-kumkaran: removed

diskann-benchmark-multi-vector/src/lib.rs
suri-kumkaran: `FailureScore` is just an ordinal used to rank "nearest misses" in the registry — no cutoff compares it against anything, so the magnitude doesn't matter, only the ordering (see `diskann-benchmark-runner/src/benchmark.rs:72-82`).

I picked `10` so Q/D-type mismatches rank strictly worse than the arch penalty of `2`/`3` above it. That's the only internal relationship that matters.

Elsewhere in the workspace these tend to be larger (`DATA_TYPE_MISMATCH = 1000` in `utils/mod.rs`, `MATCH_FAIL = 10_000` in the runner). Happy to bump it for consistency, but behaviorally equivalent.

diskann-benchmark/src/backend/multi_vector/README.md
suri-kumkaran: Make sense, moved.

diskann-benchmark/src/backend/multi_vector/driver.rs
suri-kumkaran: Thanks for pointing out.

diskann-benchmark/src/backend/multi_vector/driver.rs
suri-kumkaran: Yes, and it might help a bit in compile times.

diskann-quantization/src/multi_vector/distance/query_computer/mod.rs
suri-kumkaran: Sorry added it back.

diskann-quantization/src/multi_vector/distance/factory.rs
suri-kumkaran: Oh yes, it was screaming for a trait.

diskann-quantization/src/multi_vector/matrix.rs
```diff
@@ -712,6 +712,13 @@ impl<T: NewCloned> Clone for Mat<T> {
 }
 
 impl<T: Copy> Mat<Standard<T>> {
+    /// Construct a [`Mat`] by calling `f` once per element in row-major order.
+    pub fn from_fn<F: FnMut() -> T>(repr: Standard<T>, mut f: F) -> Self {
+        let b: Box<[T]> = (0..repr.num_elements()).map(|_| f()).collect();
+        // SAFETY: `b` has length `repr.num_elements()` by construction.
+        unsafe { repr.box_to_mat(b) }
+    }
```
hildebrandmw, line 720: This definitely needs tests.

diskann-quantization/src/multi_vector/distance/kernel.rs
```diff
@@ -0,0 +1,53 @@
...
+
+//! Object-safe kernel boundary trait plus BYOTE visitor trait.
+
+use crate::multi_vector::{MatRef, Standard};
+
+/// Object-safe interface for computing per-query MaxSim scores.
+///
+/// # Contract
+///
+/// - `scores.len() == self.nrows()` (caller's precondition).
```
hildebrandmw: We should document what happens if this doesn't hold. Preferably return an error. Right now, the SIMD kernels panic - the reference kernels silently don't do anything and leave whatever the initial values of `scores` happens to be. Behavior should be consistent.

diskann-quantization/src/multi_vector/distance/kernel.rs
```diff
@@ -0,0 +1,53 @@
...
+/// - The implementation must populate **all** `nrows()` entries of `scores`.
+///   Callers that derive quantities from the full score vector (e.g. sums)
+///   would silently corrupt their result if any trailing entry were left
+///   unwritten.
```
hildebrandmw: Is this needed? Hopefully it's obvious that if the sizing contract is upheld from the caller, the implementation really should populate the whole thing.

diskann-quantization/src/multi_vector/distance/factory.rs
```diff
@@ -0,0 +1,545 @@
...
+    }
+
+    fn compute_max_sim(&self, doc: MatRef<'_, Standard<T>>, scores: &mut [f32]) {
+        if scores.is_empty() {
+            return;
+        }
+        let query: QueryMatRef<'_, Standard<T>> = self.query.as_view().into();
+        let Ok(mut max_sim) = MaxSim::new(scores) else {
+            return;
+        };
```
hildebrandmw: Why not move this higher and have it pull double duty for the empty scores check?

diskann-quantization/src/multi_vector/distance/factory.rs
```diff
@@ -0,0 +1,545 @@
...
+        let _ = max_sim.evaluate(query, doc);
```
hildebrandmw: This silently doesn't do anything if there is a size mismatch between `query` and `doc`, which is worse than panicking.

diskann-quantization/src/multi_vector/distance/factory.rs
```diff
@@ -0,0 +1,545 @@
...
+            .finish()
+    }
+}
+
+impl<T: Copy> ReferenceKernel<T> {
+    fn new(query: MatRef<'_, Standard<T>>) -> Self {
+        let repr = *query.repr();
+        let src = query.as_slice();
+        let mut idx = 0usize;
+        let owned = Mat::<Standard<T>>::from_fn(repr, || {
```
hildebrandmw: `MatRef::to_owned` exists.

diskann-benchmark/src/backend/multi_vector/driver.rs
```diff
@@ -0,0 +1,303 @@
...
+//////////////////////
+
+/// Object-safe distance executor. The library factory's `Erase` visitor
+/// already produces a `Box<dyn MaxSimKernel<T>>`, but the driver wants its
+/// own narrow trait so the kernel + its assertions are tucked inside one
+/// vtable boundary. Simpler than threading `Box<dyn MaxSimKernel<T>>`
+/// generically through the timing harness.
+pub(super) trait Distance<T: Copy> {
+    fn max_sim(&self, doc: MatRef<'_, Standard<T>>, scores: &mut [f32]);
+}
```
hildebrandmw: Two comments to make here:

1. This is the first user of the BYOTE pattern introduced in this PR, and it immediately doesn't use the BYOTE pattern and instead double dispatches :wink:. Would probably be a good stress test of how ergonomic it is in practice (and set a good example).
2. What additional assertions are needed here that `MaxSimKernel` is not doing. **Why** are additional assertions needed, and if benchmarking is already having to work around limitations in the underlying implementation introduced in this PR, should this not be fixed at the source?

diskann-benchmark/src/backend/multi_vector/driver.rs
```diff
@@ -0,0 +1,303 @@
...
+
+impl<T: Copy> Data<T>
+where
+    StandardUniform: Distribution<T>,
+{
+    pub(super) fn new(run: &Run) -> Self {
+        let mut rng = StdRng::seed_from_u64(0x12345);
+        let queries = Mat::from_fn(
+            Standard::new(run.num_query_vectors.get(), run.dim.get()).unwrap(),
```
hildebrandmw: This should propagate an error.

diskann-benchmark/perf_test_inputs/multi-vector.json
```diff
@@ -0,0 +1,149 @@
...
+          { "num_query_vectors": 64, "num_doc_vectors": 32,   "dim": 128, "loops_per_measurement": 200, "num_measurements": 50 },
+          { "num_query_vectors": 32, "num_doc_vectors": 32,   "dim": 512, "loops_per_measurement": 50,  "num_measurements": 50 }
+        ]
+      }
+    },
+    {
+      "type": "multi-vector-op",
+      "content": {
+        "element_type": "float32",
+        "isa": "x86-64-v4",
```
hildebrandmw, line 62: Including AVX-512 in the perf test will run into an issue I constantly run into in `diskann-benchmark-simd` and haven't yet fixed and that is including an option to skip if the arch is unsupported rather than failing. Mainly as an ergonomics thing.

Also, running this on a non-AVX-512 supported machine fails at at runtime, i.e.:
```
######################
# Running Job 4 of 8 #
######################

Multi-Vector Operation

               tag: multi-vector-op
      element type: float32
               isa: x86-64-v4
    number of runs: 9

Error: x86-64-v4 not supported: AVX-512 unavailable on this CPU
```
instead of at load or match time. Please bake the check into `Benchmark::try_match` as much as possible so we error out as early as possible.

hildebrandmw (commented):
Thanks Suryansh - this is headed in the right direction, but I'm finding little bits of impedance mismatch in places, for example the `Distance` trait additing additional assertions that should probably go in the `MaxSimKernel` already.

One other comment I had is that I feel a large number of module and struct level documentation adds noise rather than value. For example, explicitly listing structs in a module is something that `rustdoc` already does far better than manual maintenance ever could. I encourage you to try building the docs and see what rustdoc will generate. For this PR, I experimented with stripping down the documentation, particularly when is simply restates the type-signature, and found the result much easier to understand and review. This isn't a blocker for this PR, but I encourage you to consider a "less is more" approach. Terser and more direct documentation means the reader spends less time filtering and more time understanding. It also lowers the chance of documentation going out of date.

diskann-quantization/src/multi_vector/matrix.rs
suri-kumkaran: Added.

diskann-quantization/src/multi_vector/distance/kernel.rs
suri-kumkaran: Removed.

diskann-benchmark/src/backend/multi_vector/driver.rs
suri-kumkaran: 1. Done — dropped `Distance<T>` / `BoxedKernel<T>`. Bench now uses `Box<dyn MaxSimKernel<T>>` from `BoxErase` directly (one vtable hop). Agreed BYOTE itself doesn't get stress-tested here; the bench legitimately wants the default packaging. A non-default `Erase` will come with a real use case.
2. Fixed at the source. Library now enforces the contract; bench has no wrapper assertions left.

diskann-quantization/src/multi_vector/distance/factory.rs
suri-kumkaran: I have tightened the `Refernce` and `Optimised` kernels API.

`MaxSimKernel::compute_max_sim` now returns `Result<(), MaxSimError>`. Size mismatch returns `InvalidBufferLength` on both Prepared and Reference paths — no silent no-ops, no panics. Empty-scores rejection collapsed into the size check (drop `BufferLengthIsZero` entirely since 0==0 is a defined no-op).

diskann-benchmark/src/backend/multi_vector/driver.rs
suri-kumkaran: Done.

diskann-quantization/src/multi_vector/distance/factory.rs
suri-kumkaran: Fixed at the source. Library now enforces the contract; bench has no wrapper assertions left.

diskann-quantization/src/multi_vector/distance/kernel.rs
suri-kumkaran: Documented as `# Errors → InvalidBufferLength`. Both paths return the identical error variant on the identical input. Zero-doc behavior is also defined and consistent: both paths fill `scores` with `f32::MAX` (negated min over empty set)

diskann-quantization/src/multi_vector/distance/factory.rs
suri-kumkaran: Thanks for pointing out.

diskann-benchmark/perf_test_inputs/multi-vector.json
suri-kumkaran: Moved the unavailable-ISA detection to `try_match` so unsupported jobs now fail at job-selection rather than mid-run. Added `MaxSimIsa::is_available()` (library-side, since the enum and its dispatch logic live there) and wired it into `Kernel<T>::try_match` and `description` alongside the existing element-type check. Running `x86-64-v4` on a non-AVX-512 host now surfaces as "ISA unavailable on this CPU" in the registry's mismatch report before any job runs, matching the surrounding `backend/index/scalar.rs` pattern (`+1` per binary mismatch axis).

Left the perf_test `x86-64-v4` entries in place — they're the right thing to run when AVX-512 is available, and now fail informatively when it isn't.

The "skip rather than fail" ergonomics piece is out of scope for this PR — happy to take that up separately if you'd like.

suri-kumkaran:
> Thanks Suryansh - this is headed in the right direction, but I'm finding little bits of impedance mismatch in places, for example the `Distance` trait additing additional assertions that should probably go in the `MaxSimKernel` already.
> 
> One other comment I had is that I feel a large number of module and struct level documentation adds noise rather than value. For example, explicitly listing structs in a module is something that `rustdoc` already does far better than manual maintenance ever could. I encourage you to try building the docs and see what rustdoc will generate. For this PR, I experimented with stripping down the documentation, particularly when is simply restates the type-signature, and found the result much easier to understand and review. This isn't a blocker for this PR, but I encourage you to consider a "less is more" approach. Terser and more direct documentation means the reader spends less time filtering and more time understanding. It also lowers the chance of documentation going out of date.

Thanks for the thorough review @hildebrandmw.

On the impedance mismatch: agreed — I missed these while adopting BYOTE. Pushed the contract into `MaxSimKernel<T>` itself (`Result<(), MaxSimError>`, size-mismatch and zero-doc behavior both defined), and with the library owning it the bench-side `Distance<T>` / `BoxedKernel<T>` indirection had nothing left to do, so I dropped both.

On the documentation: fully agreed. I went through this PR and stripped the noise — dropped the explicit type-listing tables in module docs (rustdoc generates these from re-exports), removed restated-from-signature struct docs, and tightened per-item docs to focus on non-obvious behavior, errors, safety, and design intent. The verbose-docs habit was on me — I leaned too heavily on AI agents without adjusting their output to this codebase's actual style. I'm raising a follow-up PR (#1103) to add a "Documentation" DO/DON'T section to `agents.md` so the rule stops being implicit.

diskann-benchmark/src/backend/multi_vector/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+//!
+//! [`MaxSimIsa`] (library) and [`BenchIsa`] are kept separate so the library
+//! doesn't pin its public API on a serde version or JSON shape. The
+//! benchmark owns its kebab-case JSON layout; the library is serde-agnostic.
+//!
+//! The factory follows the BYOTE ("Bring your own type erasure") pattern
+//! described in [RFC #1068]. Callers that want a different output shape
+//! (chamfer-only closure, batched evaluator, etc.) implement their own
+//! [`Erase<T>`] in place of [`BoxErase`].
+//!
```
arkrishn94: nit: this documentation doesn't seem like it's needed here

diskann-quantization/src/multi_vector/distance/factory.rs
```diff
@@ -0,0 +1,658 @@
...
+
+mod sealed {
+    pub trait Sealed {}
+}
+
+/// Scalar element types accepted by the multi-vector MaxSim factory.
+///
+/// Sealed: external crates cannot add impls. The library ships impls for
+/// `f32` and `half::f16`. Quantized representations (PQ, SQ, packed sub-byte)
+/// do not fit this trait — they carry per-vector codebook/scale state and
```
arkrishn94: Remind me, why do we need to disallow external implementations?

diskann-benchmark/src/backend/multi_vector/driver.rs
```diff
@@ -0,0 +1,261 @@
...
+        );
+        Ok(Self { queries, docs })
+    }
+}
+
+//////////////////////
+// Timing harness   //
+//////////////////////
+
+fn run_loops(run: &Run, body: &mut dyn FnMut()) -> RunResult {
```
arkrishn94: nit: from what I can tell `run_loops` is only used in `run_with_kernel`. Worth consolidating?

arkrishn94 (approved):
Left a few nots and clarifying comments Suryansh. Apart from that, looks good to me.

hildebrandmw (approved):

diskann-quantization/src/multi_vector/distance/factory.rs
suri-kumkaran: Sealing keeps `MaxSimElement` a closed dispatch table over our in-tree kernel/ISA matrix (`F32Kernel`, `F16Entry`, `Prepared`) — external impls have no way to wire into it, and we'd otherwise owe SemVer stability every time a new ISA arm is added to `build`. It also reserves design space for quantized types (PQ/SQ/sub-byte), which need side-band codebook/scale state that the `MatRef<'_, Standard<Self>>` query argument can't carry and so warrant their own factories.

diskann-benchmark/src/backend/multi_vector/mod.rs
suri-kumkaran: Thanks for pointing out, I have trimmed this doc.

diskann-benchmark/src/backend/multi_vector/driver.rs
suri-kumkaran: Make sense, done.

arkrishn94 (approved):

PR #1044, opened 2026-05-08
PQ: tighten dim contract; right-size scratch buffer

wuw92:
## Background

Historically, query buffers came from `AlignedBoxWithSlice`, which silently rounded length up to a multiple of 8 for SIMD alignment. Downstream populate functions therefore had to accept `query.len() >= dim` instead of `query.len() == dim` — pre-PR comment in `TableL2::populate`:

> Alignment means that the size of `query` gets increased ...
> This makes is VERY hard to do error checking on dimension propagation.

With #960 removing `AlignedBoxWithSlice` from the PQ path, the subtree can refactor dim handling along the boundary convert / internal trust idiom.

## Three-layer dim contract

| Layer | Where | Action | Failure |
|---|---|---|---|
| **Boundary (inmem)** | `QueryComputer::new`, `MultiQueryComputer::new`, `DistanceComputer::evaluate_similarity` | Validate `len == dim` | `Result::Err` / `assert_eq!` |
| **Boundary (disk)** | `PQScratch::set` | Validate `len >= dim`, slice `[..dim]` | `Result::Err` on undersize |
| **Internal** | `TableL2/IP/Cosine::{new, populate}` | Trusted, no re-validation | — |
| **Inner kernel** | `preprocess_query`, `populate_chunk_distances_impl`, `direct_distance_impl` | Contract boundary `debug_assert_eq!` | dev/CI panic, release zero-cost; also fail-loud for direct OSS callers |

Boundary methods take `&[f32]`. Quantized inputs are decoded via `VectorRepr::as_f32` once at the caller boundary; the PQ subtree is f32-only internally.

## Why entries take `&[f32]`

`Into<f32>` is per-element. `MinMaxElement<8>` is a single byte that can't decode without the full slice's trailing metadata — it cannot implement `Into<f32>`. Production callers supporting MinMax were therefore always pre-decoding via `VectorRepr::as_f32` and passing `&[f32]` upstream; the previous `<U: Into<f32>>` generic on PQ entries was effectively orphan-only.


diskann-providers/src/model/pq/distance/dynamic.rs
```diff
@@ -67,15 +67,22 @@ where
     ///
     /// Even though PQ does not necessarily preserve the norms of compressed vectors, using L2
     /// for Cosine Normalized seems to work well enough in practice to work as a temporary fix.
-    pub fn new<U>(
+    pub fn new(
         table: T,
         metric: Metric,
-        query: &[U],
+        query: &[f32],
```
hildebrandmw, line 73: I'd support requiring hard equality `query.len() == table.get_dim()`. Yes, this rejects arguments that it previously accepted, but accepting such inputs is a bug to begin with.

harsha-simhadri (approved):

diskann-providers/src/model/mod.rs
```diff
@@ -12,9 +12,9 @@ pub mod pq;
 pub use pq::{
     FixedChunkPQTable, GeneratePivotArguments, MAX_PQ_TRAINING_SET_SIZE, NUM_KMEANS_REPS_PQ,
     NUM_PQ_CENTROIDS, accum_row_inplace, calculate_chunk_offsets_auto, compute_pq_distance,
-    compute_pq_distance_for_pq_coordinates, direct_distance_impl, distance,
-    generate_pq_data_from_pivots_from_membuf, generate_pq_data_from_pivots_from_membuf_batch,
-    generate_pq_pivots, generate_pq_pivots_from_membuf,
+    compute_pq_distance_for_pq_coordinates, distance, generate_pq_data_from_pivots_from_membuf,
+    generate_pq_data_from_pivots_from_membuf_batch, generate_pq_pivots,
+    generate_pq_pivots_from_membuf,
```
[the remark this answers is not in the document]
wuw92: Confirmed — no external callers. Removing intentionally.

diskann-providers/src/model/pq/distance/dynamic.rs
wuw92: Done, I pushed strict `==` to `QueryComputer::new` (returns `Result::Err`),  and `DistanceComputer::evaluate_similarity` (`assert_eq!`).
 I left `PQScratch::set` (the disk-side entry) on `query.len() >= dim` + `[..dim]` slicing for now since the disk path tends to have more external surface.

hildebrandmw (approved):

diskann-disk/src/search/pq/pq_scratch.rs
```diff
@@ -23,13 +23,17 @@ pub struct PQScratch {
     /// This is used to store the pq coordinates of the candidate vectors.
     pub aligned_pq_coord_scratch: Poly<[u8], AlignedAllocator>,
 
-    /// Query scratch buffer stored as `f32`. `set` initializes it by copying/converting the
-    /// raw query values; `PQTable.PreprocessQuery` can then rotate or otherwise preprocess it.
+    /// Query scratch buffer stored as `f32`, sized by the PQ table's logical dimension.
+    /// `set` populates it from a caller-provided `&[f32]`; `PQTable::preprocess_query` can
+    /// then rotate or otherwise preprocess it.
     pub rotated_query: Vec<f32>,
```
suri-kumkaran: nit: Given we removed OPQ, do we still need `rotated_query`?

suri-kumkaran (approved):

diskann-disk/src/search/pq/pq_scratch.rs
wuw92: The buffer itself is still needed: `preprocess_query` mutates it in place (centroid subtract), so we need an owned mutable copy that doesn't touch the caller's slice.
But the name was stale post-OPQ — renamed to `query_scratch`

suri-kumkaran (approved):

PR #1047, opened 2026-05-10
add config examples to use PQ  and SQ indexing and search for wiki-1M with Cohere embeddings

harsha-simhadri:
add config examples to use PQ  and SQ indexing and search for wiki-1M with Cohere embeddings


diskann-benchmark/example/graph-index-product-quantization-wiki1M.json
```diff
@@ -0,0 +1,55 @@
+{
+  "search_directories": [
+  "../big-ann-benchmarks/data/wikipedia_cohere"  ],
```
arrayka, line 3: Minor: In the spirit of self‑service, consider including guidance on how to download the dataset, so that even an AI agent can follow the steps and retrieve the required data.

arrayka (approved):

metajack (approved):

diskann-benchmark/example/graph-index-product-quantization-wiki1M.json
```diff
@@ -0,0 +1,55 @@
...
+  "jobs": [
+    {
+      "type": "graph-index-build-pq",
+      "content": {
+        "index_operation": {
+          "source":{
+            "index-source": "Build",
+            "data_type": "float32",
+            "data": "wikipedia_base.bin.crop_nb_1000000",
```
hildebrandmw, line 12: Would be nice if we could run these through the dry run functionality to ensure they don't get out of date. The one problem would be file-path validation would fail without the existence of the data/query/groundtruth files.

diskann-benchmark/example/graph-index-product-quantization-wiki1M.json
```diff
@@ -0,0 +1,55 @@
...
+            "max_degree": 32,
+            "l_build": 100,
+            "alpha": 1.2,
+            "backedge_ratio": 1.0,
+            "num_threads": 32,
+            "multi_insert": {
+              "batch_size": 128,
+              "batch_parallelism": 32,
+              "intra_batch_candidates": "all"
+            }
```
hildebrandmw, line 24: Do we want to use `multi_insert`, or should we leave this as `null` to do `single_insert`, which will be faster to build.

hildebrandmw (commented):
Thanks. While a pre-built self serve like this is nice, I worry that if we proliferate these kinds of auxiliary files without a CI check for compatibility, they will continually break as we evolve `diskann-benchmark`. Is this just a convenience for not using the `inputs` functionality of `diskann-benchmark` and if so, is there something we can do the to CLI instead to make it more ergonomic?

diskann-benchmark/example/graph-index-product-quantization-wiki1M.json
harsha-simhadri: where do you prefer I add this?

harsha-simhadri: added to benchmark readme. please mark as resolved if this is acceptable

diskann-benchmark/example/graph-index-product-quantization-wiki1M.json
harsha-simhadri: removed

diskann-benchmark/example/graph-index-product-quantization-wiki1M.json
harsha-simhadri: I will look into dry run checks.

The goals of the new files is to illustrate the parameter choices to be used for larger datasets. The params in config files for test datasets are insufficient for larger datasets. If on the other hand you are oppposed to adding these entries here, please let me know and I can drop it, and not spend time on dry run checks

magdalendobson: Relatedly, it looks like there is no example config that builds a PQ index using the example datasets in this repo, although there is one for spherical. Is that something we should change?

hildebrandmw: I think adding examples is great and would really help reduce friction for people trying out the library! My concern is around ensuring these examples are up-to-date, so in the future users don't run into the unfortunate situation where they try an example and it immediately fails for reasons other than the files not being in the correct place.

Apologies if my original comment didn't leave much direction. The main issue with the existing `dry-run` functionality is that file paths are also validated (at least for existence) as part of the `dry-run` check, which makes it harder to use it out-of-the-box.

The issue boils down to [this check here](https://github.com/microsoft/DiskANN/blob/d516da117148c49ab534673ea701f4cae026e3d7/diskann-benchmark-runner/src/files.rs#L49-L60). We want a way to turn it off when running a CI validation run, then we'd be good. The problem is how to turn it off cleanly in a way that is somewhat general. Maybe an additional "--no-environment" flag in addition to `--dry-run` to skip system dependent environment checks and making that available in the `Checker` struct? That would provide a nice place for generally disabling all kinds of system dependent checks for CI purposes.

Issue #1036, opened 2026-05-07
Replace all custom flat scans with the new implementation


arrayka:
1. flat_scan pq.
2. ground truth.
3. PipNN: Local k-NN per leaf
4. knn baseline for benchmarks.

PR #1049, opened 2026-05-11
RFC: integrate PiPNN as an alternative graph-index build algorithm

SeliMeli:
## Summary

This RFC proposes adding **PiPNN** (Pick-in-Partitions Nearest Neighbors, [arXiv:2602.21247](https://arxiv.org/abs/2602.21247)) as a second graph-construction algorithm for DiskANN's disk index, alongside the existing Vamana builder.

PiPNN writes its graph in **Vamana's existing on-disk format** (same headers, node/edge layout, frozen-medoid convention), so the disk writer, PQ codes, and search pipeline are unchanged — search loads PiPNN-built indexes with no code change. The graph **contents themselves** are structurally different from what Vamana would build at the same parameters (the two algorithms construct graphs differently), but recall@k matches at equivalent L. PiPNN is **up to 6.3× faster to build** on our measured workloads (Enron 10M, BigANN 10M). Vamana remains the default and the only algorithm supported for incremental inserts; PiPNN is the proposed faster path for full rebuilds.

## Two-stage integration

- **Stage 1 (this RFC):** Land PiPNN behind a `BuildAlgorithm` selector with `Vamana` as the default. PiPNN is opt-in via a `pipnn` Cargo feature. Existing build sites see no behavior change. Stage 1 defines explicit milestones (M0–M7) gating Stage 2 readiness.
- **Stage 2 (separate proposal, conditional on Stage 1 milestones):** Retire Vamana's full-rebuild path. Vamana remains the implementation for incremental inserts via the hybrid update model — PiPNN for bulk/full rebuilds, Vamana for online inserts, full PiPNN rebuild triggered by recall decay.

## Highlights

- **Disk format and search API are unchanged** — the on-disk graph layout, PQ codes, search pipeline, and public Rust types remain backward-compatible. PiPNN configuration is added under a new tagged-enum variant. Note: the graph **contents** differ from Vamana's by design — equivalence is at the recall@k level, not bytes.
- **Algorithm description**: 4-phase batch builder (Partition / Local k-NN per leaf via GEMM / HashPrune merge / optional final RobustPrune). Detailed in the RFC.
- **Honest gap accounting**: explicit milestones for checkpoint/resume, scalar-quantization parity, label-filtered indexes, three-tier memory mitigation, production validation, and operational readiness — each required before Stage 2 is credible.
- **Memory mitigation**: documented three-tier build path (one-shot / disk-edges / merged-shards) selected by the existing `build_ram_limit_gb` knob, bringing PiPNN's peak RSS to or below Vamana's at a configurable build-time cost.
- **Batch-only** is a paper-confirmed algorithmic property of PiPNN, not just our implementation. The hybrid update model is the workaround.

## Reviewers

Please read the full RFC for trade-offs, benchmark tables, and milestone definitions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,312 @@
+# Integrate PiPNN as an Alternative Graph-Index Build Algorithm
+
+| | |
+|---|---|
+| **Authors** | Weiyao Luo |
+| **Contributors** | DiskANN team |
+| **Created** | 2026-05-11 |
+| **Updated** | 2026-05-11 |
```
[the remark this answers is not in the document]
SeliMeli: Thanks — the header format here intentionally mirrors the existing convention in [`rfcs/00000-template.md`](https://github.com/microsoft/DiskANN/blob/main/rfcs/00000-template.md), which uses the same `| | |` / empty-header pattern. GitHub-flavored Markdown renders empty header cells fine, and changing only this RFC would diverge from the template. If you'd prefer a different table style across all RFCs, that would be a template change (separate small PR) rather than a one-off here.

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,312 @@
...
+### Memory mitigation: three-tier build
+
+For deployments that need PiPNN's build speed but cannot afford its working memory, we reuse the same **`MemoryBudget`** parameter Vamana already uses for sharded builds. When `build_ram_limit_gb` is below a threshold, PiPNN switches to a chunked path that spills HashPrune reservoirs to disk between leaf batches. Measurements on the same dataset as the table above (BigANN 10M):
+
+| Strategy | Peak RSS | Build time | Recall@10 L=50 | Trigger |
+|---|---:|---:|---:|---|
+| **One-shot** (in-memory) | 10.8 GB | 133s | 95.00% | RAM ≥ ~32 GB |
+| **Disk-edges** (per-batch reservoir flush) | 6.4 GB | 126s | 95.00% | RAM 8-32 GB |
+| **Merged shards** (per-shard graph, then merge) | 3.3 GB | 332s | 95.31% | RAM 4-8 GB |
+
```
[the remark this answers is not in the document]
SeliMeli: Fixed in [`943fa74`](https://github.com/SeliMeli/DiskANN/commit/943fa749) — added a paragraph noting that the smaller working set (better cache behavior on HashPrune inserts) plus sequential append-only spills that overlap with compute keep wall-clock roughly equal to the one-shot path despite the I/O.

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,312 @@
...
+        final_prune: bool,           // optional RobustPrune final pass
+        leader_cap: usize,           // hard cap on leaders per level
+        saturate_after_prune: bool,
+    },
+}
+```
+
+`Vamana` is the `Default` so every existing call site that constructs `DiskIndexBuildParameters` without specifying an algorithm keeps the existing behavior.
+
+`DiskIndexBuildParameters` gains a `build_algorithm: BuildAlgorithm` field and a constructor pair: `new` (defaults to Vamana, no PiPNN dep) and `new_with_algorithm` (explicit). The JSON schema for benchmark configs gains an optional `build_algorithm` block that, when present, deserializes via `#[serde(tag = "algorithm")]` into one of the variants above.
```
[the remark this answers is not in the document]
SeliMeli: Good catch — fixed in [`943fa74`](https://github.com/SeliMeli/DiskANN/commit/943fa749). Added an explicit paragraph after the JSON-schema description spelling out the deserialization failure mode (serde `unknown variant 'PiPNN'` parse error) when a PiPNN config is fed to a non-pipnn binary, and noting that configs without `build_algorithm` parse identically across feature combinations so it's not a backward-compatibility regression.

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,312 @@
...
+| Search API (`DiskANNIndex::search`, beam_width, search_list, recall_at, num_nodes_to_cache, search_io_limit, filters API) | unchanged |
+| Public Rust types (`IndexConfiguration`, `DiskIndexWriter`, `DiskIndexBuildParameters`) | additive only (new field with default) |
+| Benchmark JSON config | additive only (new optional `build_algorithm` field) |
+| C/C++ FFI (if any) | unchanged |
+
+Since the produced graph and PQ/SQ artifacts are byte-identical in format, a search-only consumer cannot tell which builder wrote the index.
+
+### Feature gating
+
+- The `diskann-disk` crate gains a `pipnn` Cargo feature. With it disabled, `BuildAlgorithm::PiPNN` does not exist at the type level — no runtime branch, no extra binary size, no dependency on `diskann-pipnn`.
```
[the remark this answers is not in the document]
SeliMeli: Same point as the comment on line 113 — addressed in [`943fa74`](https://github.com/SeliMeli/DiskANN/commit/943fa749) by adding an explicit deserialization-failure-mode paragraph in the JSON schema section above.

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,354 @@
...
+
+### M1 — Feature parity: in-memory build / search
+
+Vamana supports both a **disk-resident** build/search path (via `diskann-disk`) and an **in-memory only** path (via `diskann::graph::index::DiskANNIndex`). PiPNN today only produces graphs handed to `DiskIndexWriter`; an in-mem-only consumer that wants PiPNN's speed has no entry point.
+
+- **Scope:** expose `diskann_pipnn::build_typed` output (`Vec<Vec<u32>>`) as a populated in-memory `DiskANNIndex` so callers can build + search without touching disk.
+- **API:** add `diskann_pipnn::build_into_inmem_index(...)` returning an in-memory index that is read by the existing `DiskANNIndex::search` path unchanged.
+- **Validation:** in-mem search recall on Enron 1M with PiPNN-built graph matches the disk-build + load round-trip recall within noise.
+
+### M2 — Feature parity: checkpoint / resume
```
SeliMeli: checkpoint may need to be put in stage2. currently the streaming checkpoing like vamana dose is not fit PIPNN's batch build

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,354 @@
...
+
+## Summary
+
+Add **PiPNN** (Pick-in-Partitions Nearest Neighbors, arXiv:2602.21247) as a second graph-construction algorithm for DiskANN's disk index. PiPNN produces a graph byte-compatible with Vamana's disk format and search API, at **up to 6.3× lower build time** on the workloads we have measured. Vamana remains the default and the only algorithm supported for incremental inserts; PiPNN is the proposed faster path for full rebuilds.
+
+## Motivation
+
+### Background
+
+DiskANN currently builds the disk index with a single algorithm — **Vamana** (`diskann-disk/src/build/builder/`). Vamana incrementally inserts each point into a graph, running a greedy search + `RobustPrune` for each insertion, producing the on-disk format documented in `diskann-disk/src/storage/`.
```
arrayka: Probably worth noting that today clients update their indices in three main ways:

Incremental — continuously insert and delete vectors in an existing graph.
Full rebuild — rebuild the entire graph from scratch, producing an immutable graph.
Partitioned full rebuild — split points into N clusters, build N separate graphs, then stitch them together with a lightweight merge step to reduce peak memory usage during build.

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,354 @@
...
+
+**PiPNN** (Pick-in-Partitions Nearest Neighbors, arXiv:2602.21247) is a partition-based **batch** graph builder, in contrast to Vamana's **incremental** insert + prune. The construction has four phases:
```
arrayka: Minor: Please provide full link for convinience

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,354 @@
...
+
+1. **Partition** — Randomized Ball Carving (RBC) recursively splits the dataset into small *overlapping* leaf clusters. Each point lands in `fanout` of its nearest cluster leaders at every recursion level, so every point appears in multiple leaves. Recursion stops when a cluster fits a configured leaf-size cap (`c_max`, typically 256–1024 points).
+2. **Local k-NN per leaf** — For each leaf, compute the full pairwise distance matrix in one batched GEMM call, then extract each point's `leaf_k` nearest neighbors inside the leaf. GEMM batching is the source of most of PiPNN's wall-clock advantage over per-point greedy search.
```
arrayka: @arkrishn94, is `Local k-NN per leaf` a good candidate for being implemented using the flat scan proposal we are working on?
A work item to track this discussion: #1036

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,354 @@
...
+3. **HashPrune merge** — Edges from all leaves are merged into a per-point reservoir of bounded size (`l_max`, ~64–128). The pruner is keyed by an LSH **angular bucket** of each candidate neighbor: at most one candidate per bucket is retained, and on collision the closer candidate wins. This produces a diverse short-list per point using O(`l_max`) memory per node and O(1) amortized insert work.
```
arrayka: Is it possible to do `HashPrune merge` in batches (generate candidates for first 10,000 points, the flush adj. lists to disk), generate for next 10,000 points and so on? Would it reduce RAM footprint from `per-point reservoir of bounded size`?

I'm just trying to evaluate if it is possible to divide the work into chunks and run them sequentially to bound RAM footprint.

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,354 @@
...
+
+Vamana's incremental design scales linearly in points × per-insert search cost, which makes full rebuilds expensive at the scales we operate. Measured baselines:
+
+| Dataset | Vamana build time |
+|---|---:|
+| Enron 1M (1.087M × 384, fp16, cosine_normalized) | 70s |
+| BigANN 10M (10M × 128, fp16, squared_l2) | 358s |
+| Enron 10M (10M × 384, fp16, cosine_normalized) | 844s |
+
+Frequent rebuilds (driven by data churn or parameter sweeps) and full rebuilds at 10M-scale and above are the bottleneck. PiPNN's offline benchmarks at matching recall budgets complete the same builds **up to 6.3× faster** while writing the same disk format (full numbers in the Benchmark Results section). This RFC proposes landing PiPNN so teams can opt into faster builds and so we can collect production-relevant signal on whether PiPNN can eventually replace Vamana's full-rebuild path.
```
arrayka: > builds **up to 6.3× faster**

This problem statement should clearly call out the trade-offs, constraints, and what we’re optimizing for.

We should also define concrete hypotheses and validate them.
For example: given a fixed set of workers (with N CPU cores, M RAM, and S SSD per node), PiPNN can improve index build throughput (vectors per minute per worker) by driving close to full CPU and RAM utilization. In contrast, Vamana cannot effectively leverage unused RAM to accelerate the build.

In other words, we should lock available CPU, RAM, SSD and compare build duration with the same available resources.
Let's assume that Vamana needs 6GB to build 10M-vector graph, but PipNN needs 10GB.
Which algorithm is faster if we have:
- 3GB RAM available
- 6GB RAM available
- 10GB RAM available
- 20GB RAM available (can PipNN autotune itself to use all available RAM budget to reduce index build time?)

arrayka: Is it correct to say that `Vamana` offers a way to limit RAM usage during index build, at the cost of slower build time and potentially lower search quality (via partitioned build)? At the same time, it doesn’t seem to provide a mechanism to fully utilize the available RAM budget to improve other metrics, such as build time or quality.

`PiPNN` requires more RAM for an index of the same size, and—similar to Vamana—would rely on partitioned build to keep memory usage bounded.

Is it fair to conclude that while both approaches can keep RAM usage under control, neither really provides a way to leverage the full available RAM budget to optimize for faster build times?

It would be great to explore if `PiPNN` can actually use all available RAM budget to optimize for faster build times.

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,354 @@
...
+
+#### Hybrid update model (Stage 2 direction)
+
+Vamana and PiPNN write the same on-disk graph format, so a graph built by either algorithm can be *read* (and incrementally edited) by either. We exploit this for the production update story:
+
+- **Bulk / full rebuild → PiPNN.** When data churn is large enough to justify a full rebuild, PiPNN is used because it is several times faster than Vamana at this job.
+- **Incremental insert → Vamana.** Between full rebuilds, individual inserts use Vamana's existing greedy-search + RobustPrune insert path. PiPNN's batch design has no natural single-point-insert API and we do not plan to build one.
+- **Quality decay → trigger PiPNN rebuild.** When recall on the live graph degrades past a configured threshold (driven by accumulated incremental inserts), the system schedules a PiPNN full rebuild from the current dataset snapshot.
```
arrayka: I don’t think we expect the graph to degrade enough that a full rebuild from scratch is necessary. DiskANN claims that the graph can remain healthy over time with incremental updates.

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,354 @@
...
+
+#### Two-stage rollout
+
+- **Stage 1 (this RFC):** Land PiPNN behind a build-algorithm selector. Vamana stays default; PiPNN is opt-in. Stage 1 has explicit milestones (in Future Work) that gate readiness for Stage 2.
+- **Stage 2 (separate proposal, conditional on Stage 1 milestones):** Retire the Vamana **full-rebuild** path. Vamana remains the implementation for incremental inserts via the hybrid model above.
+
+### Goals
+
+1. **Algorithm-level pluggability**: introduce a build-algorithm selector to the build pipeline that routes between Vamana (existing) and PiPNN (new). Existing build sites continue to default to Vamana with no behavior change.
+2. **Disk format compatibility**: the PiPNN-built index is byte-compatible with Vamana-built indexes on disk — search, PQ, and storage layouts are unchanged. This is the foundation for the hybrid update model.
```
arrayka: There are no plans to mutate disk index.

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,354 @@
...
+
+`DiskIndexBuildParameters` gains a `build_algorithm: BuildAlgorithm` field and a constructor pair: `new` (defaults to Vamana, no PiPNN dep) and `new_with_algorithm` (explicit). The JSON schema for benchmark configs gains an optional `build_algorithm` block that, when present, deserializes via `#[serde(tag = "algorithm")]` into one of the variants above.
+
+**Deserialization behavior when the `pipnn` feature is disabled**: because `BuildAlgorithm::PiPNN` is gated by `#[cfg(feature = "pipnn")]`, a binary built without the feature does not see that variant. A JSON config containing `"algorithm": "PiPNN"` fed to such a binary fails at parse time with a serde error along the lines of `unknown variant 'PiPNN', expected 'Vamana'`. This is a clear, fail-fast diagnostic — not a backward-compatibility regression. Configs that omit `build_algorithm` (or set `"algorithm": "Vamana"`) parse identically across feature combinations. Documentation alongside the config schema will call this out so users know that PiPNN configs require a PiPNN-enabled build.
```
arrayka: I'm not sure that I understand why index can't be deserialized (loaded) without pipnn feature. If pipnn produces the same graph as Vamana, then the code can load it even if was compiled without pipnn feature.

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,354 @@
...
+
+That delta — roughly **+4.5 GB**, dominated by HashPrune (`10M × 64 × 8 ≈ 5 GB`) plus smaller PiPNN-only working buffers (LSH sketches, partition leaf indices) — is the cost of the batch design and not a bug. It is the working set the algorithm explicitly needs. The next subsection describes the mitigation.
+
+### Memory mitigation: three-tier build
+
+For deployments that need PiPNN's build speed but cannot afford its working memory, we reuse the same **`MemoryBudget`** parameter Vamana already uses for sharded builds. When `build_ram_limit_gb` is below a threshold, PiPNN switches to a chunked path that spills HashPrune reservoirs to disk between leaf batches. Measurements on the same dataset as the table above (BigANN 10M):
+
+| Strategy | Peak RSS | Build time | Recall@10 L=50 | Trigger |
+|---|---:|---:|---:|---|
+| **One-shot** (in-memory) | 10.8 GB | 133s | 95.00% | RAM ≥ ~32 GB |
```
arrayka: Does PipNN accept num_threads as a build parameter? If so, does it impact RAM usage?

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,354 @@
...
+
+- **Scope:** expose `diskann_pipnn::build_typed` output (`Vec<Vec<u32>>`) as a populated in-memory `DiskANNIndex` so callers can build + search without touching disk.
+- **API:** add `diskann_pipnn::build_into_inmem_index(...)` returning an in-memory index that is read by the existing `DiskANNIndex::search` path unchanged.
+- **Validation:** in-mem search recall on Enron 1M with PiPNN-built graph matches the disk-build + load round-trip recall within noise.
+
+### M2 — Feature parity: checkpoint / resume
+
+- **Scope:** add checkpoint/resume to the PiPNN build pipeline using the existing `CheckpointManager` / `ChunkingConfig` infrastructure in `diskann-disk/src/build/chunking/`.
+- **Boundaries:** natural checkpoint points are partition output (`Vec<Leaf>`), per-leaf HashPrune flush, post-extract graph.
+- **Behavior:** matches Vamana's — a killed build resumes from the last checkpoint instead of starting over.
```
arrayka: Let's understand cost-benefit of `checkpoint/resume` feature. Is it really useful today - how often is it used?

rfcs/01049-pipnn-integration.md
SeliMeli: yes, that is what the disk-edge in mem opt trying to do. currently this is generate all edges for dataset and save edges to disk then read chunk to hashprune. But we can also, just write the partition info into disk, then chunk the leaf build+hashprune.

rfcs/01049-pipnn-integration.md
SeliMeli: no, this statement is for the config not for the index file. the cofig this RFC proposed is gated by feature flag and if a pipnn config sent to a binary that dose not built with that flag will fail. The index files built by pipnn or vamana are used by the same search process so there is no special graph that pipnn is building.

rfcs/01049-pipnn-integration.md
SeliMeli: PIPNN's mem usage is mostly dominated by hashprune + dataset, consuming more mem may get better graph quality (larger pruning candidate) but will not bring faster build (if pipnn fit in the mem cap, increase mem barely helps on build speed but with more mem channel and bigger ram BW, pipnn could scale well because of all threads can do ram access quicker)

rfcs/01049-pipnn-integration.md
SeliMeli: Yes — PiPNNConfig.num_threads is a build parameter (default 0 = all logical CPUs, matching Vamana's convention).

Per-thread working buffers: thread-local f32 stripe buffers in partition (~`stripe_kb per thread, typically 16 MB) and thread-local leaf-build scratch (norm/distance matrix, ~c_max² × 4B≈ 256 KB per thread atc_max=256`). Total per-thread overhead ~16-20 MB. For 48 threads that's ~960 MB extra resident vs single-threaded, which is real but small relative to the 5 GB HashPrune reservoir and 2.5 GB dataset

rfcs/01049-pipnn-integration.md
SeliMeli: will move this milestone to stage 2

rfcs/01049-pipnn-integration.md
SeliMeli: yes, but other case rebuilt is still an option like: embedding rotations, schema changes, parameter retuning, batch insert. those cases could be benifit from pipnn as well. Will fix the wording

rfcs/01049-pipnn-integration.md
SeliMeli: the knn build in pipnn is a pair-wise gemm backed (N * N), the scenario is quite different to the flat-scan (1 * N). But we can measure. The flat-scan build knn is more like all-to-all prune option in the paper which is extreamly slow

rfcs/01049-pipnn-integration.md
SeliMeli: will update to rfc

rfcs/01049-pipnn-integration.md
SeliMeli: will fix the wording: only inmem vanana could insert to loaded graph, disk index rebuilt

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,354 @@
...
+2. **Disk format compatibility**: the PiPNN-built index is byte-compatible with Vamana-built indexes on disk — search, PQ, and storage layouts are unchanged. This is the foundation for the hybrid update model.
+3. **Public API compatibility**: the disk-index public API surface (`DiskIndexBuilder::new`, `IndexConfiguration`, `DiskIndexWriter`, JSON config schema) remains backward-compatible. PiPNN configuration is added under a new tagged enum variant.
+4. **Feature-parity milestones**: deliver the Vamana capabilities PiPNN needs for a full-rebuild role in production (see Future Work below).
+5. **Documented memory mitigation**: provide a configuration knob (three-tier build) that brings PiPNN's peak RSS to or below Vamana's at the cost of build time.
+
+## Proposal
+
+### Workspace structure
+
+Add a new crate, `diskann-pipnn`, that depends on the existing `diskann`, `diskann-disk`, `diskann-linalg`, `diskann-vector`, `diskann-quantization`, and `diskann-utils` crates. PiPNN lives outside `diskann-disk` so the core disk path has no compile-time dependency on PiPNN; the disk builder takes a typed `BuildAlgorithm` and only depends on PiPNN behind a feature flag.
```
wuw92: I suppose diskann-pipnn doesn't need symbols from diskann-disk. 
Worth noting diskann-pipnn → diskann-disk (hard) plus diskann-disk →[feature] diskann-pipnn forms a cycle.

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,354 @@
...
+- **Scope:** expose `diskann_pipnn::build_typed` output (`Vec<Vec<u32>>`) as a populated in-memory `DiskANNIndex` so callers can build + search without touching disk.
+- **API:** add `diskann_pipnn::build_into_inmem_index(...)` returning an in-memory index that is read by the existing `DiskANNIndex::search` path unchanged.
+- **Validation:** in-mem search recall on Enron 1M with PiPNN-built graph matches the disk-build + load round-trip recall within noise.
+
+### M2 — Feature parity: checkpoint / resume
+
+- **Scope:** add checkpoint/resume to the PiPNN build pipeline using the existing `CheckpointManager` / `ChunkingConfig` infrastructure in `diskann-disk/src/build/chunking/`.
+- **Boundaries:** natural checkpoint points are partition output (`Vec<Leaf>`), per-leaf HashPrune flush, post-extract graph.
+- **Behavior:** matches Vamana's — a killed build resumes from the last checkpoint instead of starting over.
+- **Validation:** kill-and-resume test on BigANN 10M at three different checkpoint phases; final graph byte-identical to a non-interrupted build given the same seeds.
```
wuw92: PiPNN is a parallel algorithm — byte-identical output across runs would require extra determinism work.
Recall feels like a sufficient validation criterion

rfcs/01049-pipnn-integration.md
SeliMeli: added M6 to provide more decent compare

rfcs/01049-pipnn-integration.md
SeliMeli: yes, the byte-identical is a wrong statement. removed

rfcs/01049-pipnn-integration.md
SeliMeli: fixed wording. there will not be cycle depdency in code.

hildebrandmw (commented):
Most of this PR is dedicated to how PiPNN will integrate into `diskann-disk`, and I do not have the review capacity nor much stake in that side of the integration.

What I am curious about, though, is how we plan to integrate PiPNN into the project as a whole. #856 integrates the algorithm as a separate crate with kind of its own set of abstractions, its own quantization. I have a very strong preference to keep `diskann` homogeneous, even if it means working on lower levels of the stack to build what we need in a more reuseable way (e.g., quantized matrix multiply, which is currently being worked on for multi-vectors). Basically, investing in shared capability where everyone can benefit.

I'm not necessarily asking that this blocks forward progress, but I would like to see milestones that commit to converging on shared abstractions/implementations somewhere in the roll out of late stage 1 or early stage 2. I would love for the PiPNN integration to leave whole project stronger and not just be a faster disk build.

SeliMeli:
> Most of this PR is dedicated to how PiPNN will integrate into `diskann-disk`, and I do not have the review capacity nor much stake in that side of the integration.
> 
> What I am curious about, though, is how we plan to integrate PiPNN into the project as a whole. #856 integrates the algorithm as a separate crate with kind of its own set of abstractions, its own quantization. I have a very strong preference to keep `diskann` homogeneous, even if it means working on lower levels of the stack to build what we need in a more reuseable way (e.g., quantized matrix multiply, which is currently being worked on for multi-vectors). Basically, investing in shared capability where everyone can benefit.
> 
> I'm not necessarily asking that this blocks forward progress, but I would like to see milestones that commit to converging on shared abstractions/implementations somewhere in the roll out of late stage 1 or early stage 2. I would love for the PiPNN integration to leave whole project stronger and not just be a faster disk build.

added M7 — Abstraction convergence as a Stage-2 entry gate in d37bedc7 (https://github.com/SeliMeli/DiskANN/commit/d37bedc7). What diskann-pipnn keeps: partition, BFS scheduler, builder orchestration. What converges:

  - Pruning — final_prune_from_candidates duplicates Vamana's RobustPrune today only because diskann::graph::internal::prune is pub(crate). Expose it (or extract diskann-prune), delete the duplicate. HashPrune co-locates there; whether Vamana adopts it
   is exploration, not a commitment.
  - Quantized GEMM — PiPNN is GEMM-based so its quantization is int8 / fp16 / bf16, not Vamana's SQ1 (different kernel). Consume the multi-vector quantized-GEMM interface; if PiPNN gets there first, contribute it back.
  - Distance(if pipnn needs) — fully delegate to diskann_vector::distance.

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,303 @@
...
+Output: `Vec<Vec<u32>>` adjacency lists, handed to the existing disk writer. PQ training and search are unchanged.
+
+### Problem statement
+
+Vamana full builds at 10M+ are the bottleneck:
+
+| Dataset | Vamana build |
+|---|---:|
+| Enron 1M (384d) | 70s |
+| BigANN 10M (128d) | 358s |
```
hildebrandmw, line 46: Can you help me replicate these results using #856 for the inmem index? If I use the following to compare:
```json
{
  "search_directories": [
    "<snip>"
  ],
  "output_directory": null,
  "jobs": [
    {
      "type": "async-index-build",
      "content": {
        "search_phase": {
          "groundtruth": "sift-1m.gt.bin",
          "num_threads": [
            16
          ],
          "queries": "sift-1m.queries.u8.bin",
          "reps": 5,
          "runs": [
            {
              "recall_k": 10,
              "search_l": [
                10,
                20,
                30,
                40,
                50,
                60,
                70,
                80,
                90,
                100
              ],
              "search_n": 10
            }
          ],
          "search-type": "topk"
        },
        "source": {
          "alpha": 1.2000000476837158,
          "backedge_ratio": 1.0,
          "data": "sift-1m.u8.bin",
          "data_type": "uint8",
          "distance": "squared_l2",
          "index-source": "Build",
          "insert_retry": null,
          "l_build": 50,
          "max_degree": 32,
          "multi_insert": null,
          "num_threads": 16,
          "save_path": null,
          "start_point_strategy": "medoid"
        }
      }
    },
    {
      "type": "async-index-build",
      "content": {
        "search_phase": {
          "groundtruth": "sift-1m.gt.bin",
          "num_threads": [
            16
          ],
          "queries": "sift-1m.queries.u8.bin",
          "reps": 5,
          "runs": [
            {
              "recall_k": 10,
              "search_l": [
                10,
                20,
                30,
                40,
                50,
                60,
                70,
                80,
                90,
                100
              ],
              "search_n": 10
            }
          ],
          "search-type": "topk"
        },
        "source": {
          "alpha": 1.2000000476837158,
          "backedge_ratio": 1.0,
          "data": "sift-1m.u8.bin",
          "data_type": "uint8",
          "distance": "squared_l2",
          "index-source": "Build",
          "insert_retry": null,
          "l_build": 50,
          "max_degree": 32,
          "multi_insert": null,
          "num_threads": 16,
          "save_path": null,
          "pipnn": {
              "algorithm": "PiPNN"
          },
          "start_point_strategy": "medoid"
        }
      }
    }
  ]
}
```
on my laptop, I get 11.2 seconds for index build with the inmem index and 15.6 seconds with PiPNN with worse final recall. Note that I am using the default hyper-parameters for pipnn as I don't have a good idea on how to configure them. The display for PiPNN is as follows:
```
Async Full-Precision Index Build

               tag: async-index-build
              file: <snip>
         data_type: uint8
        max degree: 32
           L-build: 50
             alpha: 1.2
start point strategy: Medoid
    backedge ratio: 1
Using Multi Insert: NO
start_point_strategy: Medoid
     build threads: 16
         Save Path: None
   build algorithm: PiPNN
       pipnn.c_max: 1024
       pipnn.c_min: 256
      pipnn.leaf_k: 3
       pipnn.l_max: 128
    pipnn.replicas: 1

PiPNN build: 1000000 points x 128 dims
  Vector store:  0.059s
PiPNN Build Timing
  LSH sketches:   0.111s
  Partition:      1.777s  (65526 leaves)
  Leaf build:     13.009s  (145781104 edges)
  Graph extract:  0.443s
  Final prune:    0.000s
  Total:          15.382s

  Avg degree: 31.9, Max degree: 32, Isolated: 0
  Edge transfer: 0.162s (medoid=89028)
  Total (incl. store + transfer): 15.602s



Index Build Time: 15.602119s
Vectors Inserted: 1000000
Kind: PiPNN
Insert Latencies:
  average: 15us
      p90: 15us
      p99: 15us
```
Can you provide suggestions on which hyper-parameters to use? Thanks!

hildebrandmw (commented):
Thanks for the turn-around! One quick question for you below about reproducing the performance claims.

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,303 @@
...
+
+> Given a fixed worker (CPU/RAM/SSD), PiPNN delivers higher build throughput than Vamana at matching recall *when its working set fits*. Below that threshold, the three-tier dispatch (one-shot → disk-edges → merged-shards) keeps PiPNN at or under Vamana's RAM footprint.
+
+On BigANN 10M, 16 threads:
+
+| RAM budget | PiPNN strategy | PiPNN build | Vamana build |
+|---:|---|---:|---:|
+| ≥ 12 GB | one-shot | 80–133s | 358s |
+| 6–12 GB | disk-edges | ~126s | 358s |
+| 3–6 GB | merged-shards | ~332s | ~358s (partitioned) |
```
arkrishn94, line 61: Can you elaborate on the `disk-edges` and the `merged-shards` strategies? Not sure if this is mentioned somewhere else

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Hybrid update model (Stage 2 direction)
+
+Both algorithms write the same disk format, so a graph built by either can be loaded and (once in memory) extended by Vamana. Production update story:
+
+- **Full rebuild → PiPNN.** Several times faster than Vamana.
+- **Incremental insert → in-memory Vamana.** Unchanged — applies to the in-memory graph; the disk index file is not mutated in place.
+- **Rebuild triggers.** Embedding rotation, schema/parameter retuning, large batch inserts, or periodic safety rebuilds — not just gradual recall drift. DiskANN's claim that incremental updates keep recall healthy still stands; PiPNN just makes the eventual rebuild cheaper.
+
+This is why we don't need PiPNN to support `insert(point)` — the disk format is the integration point between batch and incremental.
```
arkrishn94, line 73: I don't understand what this section is providing a solution for. Is the key takeaway here that building the graph with one algorithm doesn't preclude us from using another one in the future?

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,303 @@
...
+
+- **Stage 1 (this RFC).** Land PiPNN as an alternative builder for the disk-index full-build path (initial builds *and* rebuilds), behind a `pipnn` Cargo feature. Vamana stays default.
+- **Stage 2 (separate proposal, gated by Stage-1 milestones).** Retire Vamana's full-build path; keep Vamana for in-memory incremental inserts.
+
+In-memory PiPNN build is not in any stage — see "Out of scope".
+
+### Goals (Stage 1)
+
+1. **Pluggable disk builder** — a selector that routes Vamana (default) vs. PiPNN (opt-in). No behavior change at existing call sites.
+2. **Disk-format compatibility** — byte-identical to Vamana's output; search/PQ/storage layouts unchanged.
```
arkrishn94: Is the claim here that PiPNN will produce an identical graph to Vamana?

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,303 @@
...
+
+Recall depends on partition overlap (`fanout`) and reservoir size (`l_max`) — a larger parameter space than Vamana's `R`/`L_build`. Mitigation: keep Vamana as default and ship reference parameter sets per workload class.
+
+## Benchmark Results
+
+Azure `Standard_L16s_v3`, 16 threads, NVMe, `RUSTFLAGS=-C target-cpu=native`.
+
+### Build time
+
+| Dataset | Vamana | PiPNN | Speedup |
```
arkrishn94, line 200: Are the Vamana results here with the spherical/scalar quantizer?

rfcs/01049-pipnn-integration.md
```diff
@@ -0,0 +1,303 @@
...
+
+Stage 1 covers build-from-scratch and full rebuilds with PiPNN. M0 ships in this RFC; M1–M6 are follow-on work, parallelizable where dependencies allow.
+
+### M0 — Skeleton integration (this RFC)
+Crate, `BuildAlgorithm` enum, dispatch behind `pipnn` Cargo feature. JSON config gains optional `build_algorithm`. CI smoke test (SIFT-1M) with `--features pipnn`.
+
+### M1 — Quantization parity
+PiPNN's leaf-build computes pairwise distances via batched GEMM, so its quantization options are constrained by what hardware GEMM units accelerate: **int8** (AVX-512-VNNI, AMX-INT8), **fp16** (AVX-512-FP16), and **bf16** (AVX-512-BF16, AMX-BF16). 1-bit is not in that set — GEMM hardware does multiply-accumulate, not XOR+popcount.
+
+Vamana's existing quantized path is **SQ1 (1-bit) Hamming**, which is a fundamentally different kernel (XOR+popcount, not MAC). PiPNN cannot directly reuse Vamana's SQ1 quantizer for the same speedup story; porting requires adapting the quantization pipeline to a GEMM-compatible format and computing distance through the quantized-GEMM dispatch.
```
arkrishn94, line 240: Why is supporting quantized GEMM in M1? This feels like a nice to have that can come later no? Unless it's needed for RAM savings...

arkrishn94 (commented):
Thanks Weiyao, I've left some clarifying questions. More broadly, I am a bit worried that we're leaning into a largely independent code path for building the graph. Often code paths like this either get abandoned because they are unable to keep up with the main path (incremental Vamana) or they end up playing constant catch up. 

The results are very compelling for PiPNN so I think there is definitely something here for us to learn from and adopt. What I'd love to understand is - 
1. Can we crisply quantify what components of the PiPNN algorithm provide the speed-up over Vamana?  My guess is it's the ability to scale the pruning over independent regions of the graph (thanks to the clustering) + the fast pruning algorithm (HashPrune). 
2. Can we think carefully if there is a way to port some or most of this benefit into the current Vamana algorithm? I don't think you need to do this alone, I can definitely help with this and we can loop in the research folks too. We might come to the conclusion that there is no neat way to consolidate/port the ideas to the existing build path but I'd like to make sure we've explored this before triggering this divergence.

I know this is more work than you anticipated but I believe if we find a way to integrate some of these ideas into the core algorithm and get most of the benefit, it'll allow all consumers to benefit from this innovation and make a proper contribution to the open-source ANN community.

rfcs/01049-pipnn-integration.md
SeliMeli: Hi Mark, with R 32, L 50, pipnn is quite close to vamana. You can use this config:
```
  {
    "max_degree": 32,
    "l_build": 50,
    "alpha": 1.2,
    "distance": "squared_l2",
    "num_threads": 16,
    "pipnn": {
      "algorithm": "PiPNN",
      "c_max": 256,
      "c_min": 16,
      "p_samp": 0.005,
      "fanout": [10, 3],
      "leaf_k": 3,
      "l_max": 50,
      "num_hash_planes": 12,
      "final_prune": true
    }
  }

```
it should match vaman's build speed and recall

And when increase max_degree to 64 L build 72 (pipnn l_max 72), the build time for vamana is 13s and pipnn's remains 7s

rfcs/01049-pipnn-integration.md
SeliMeli: sry, this statement is by mistake. not an byte-identical graph but a quality-identical graph

rfcs/01049-pipnn-integration.md
SeliMeli: no, the result here is fp build

rfcs/01049-pipnn-integration.md
SeliMeli: `Disk-edges` are brefily introduced below, basically we split pipnn into partition and leaf build + hp and graph construct to avoid keep dataset in mem the whole build to reduce mem footprint.
`merged-shards` is just vamana's sharded build

rfcs/01049-pipnn-integration.md
SeliMeli: supporting quantized gemm also unblock using fp16/u8 native gemm, but the proiority could be modified

rfcs/01049-pipnn-integration.md
SeliMeli: yes, basically the two graph are quality identical and disk format identical

SeliMeli:
> Thanks Weiyao, I've left some clarifying questions. More broadly, I am a bit worried that we're leaning into a largely independent code path for building the graph. Often code paths like this either get abandoned because they are unable to keep up with the main path (incremental Vamana) or they end up playing constant catch up.
> 
> The results are very compelling for PiPNN so I think there is definitely something here for us to learn from and adopt. What I'd love to understand is -
> 
> 1. Can we crisply quantify what components of the PiPNN algorithm provide the speed-up over Vamana?  My guess is it's the ability to scale the pruning over independent regions of the graph (thanks to the clustering) + the fast pruning algorithm (HashPrune).
> 2. Can we think carefully if there is a way to port some or most of this benefit into the current Vamana algorithm? I don't think you need to do this alone, I can definitely help with this and we can loop in the research folks too. We might come to the conclusion that there is no neat way to consolidate/port the ideas to the existing build path but I'd like to make sure we've explored this before triggering this divergence.
> 
> I know this is more work than you anticipated but I believe if we find a way to integrate some of these ideas into the core algorithm and get most of the benefit, it'll allow all consumers to benefit from this innovation and make a proper contribution to the open-source ANN community.

totally agree that we should consider port pipnn's speed-up ability to vamana and improve vamana. but at current stage as the rfc described, we can do this in stage 2 or parallel with stage 1 to reaserch on the converge while stage 1 keep focus on pipnn integration

PR #1050, opened 2026-05-11
[diskann-garnet] Implement BIN and Q8 quantizers

### LINKED ISSUE #1055: Implement quantization bootstrap in diskann-garnet
The current RFC for quantization bootstrap proposes handling bootstrapping completely in the provider. The first provider to get this will be diskann-garnet.

### LINKED ISSUE #1059: int8 support in diskann-garnet/vectorset
Add `int8` data type to diskann-garnet/vectorset as it is the second most popular data type.

Currently we see
```
S Q:\DiskANN\vectorset> cargo run --release -- --config .\config.toml.example --data-type int8 ingest --tasks 32 --degree 48 --l-build 300 data.bin                                                                     
error: invalid value 'int8' for '--data-type <DATA_TYPE>'
  [possible values: uint8, float32]
```

### LINKED ISSUE #1088: Add support for i8 vectors
Add i8 support and i8 + BIN quantizer support.

metajack:
This implements the `BIN` and `Q8` quantizers for diskann-garnet.

This PR includes the following:
- Vectors are now stored as `Poly<[u8], AlignOfEight>` instead of `Vec<T>`. This means that the same type can be used for both quantized and full precision vectors.
- A new `GarnetQuantizer` trait is inroduced which allows us to type-erase the different quantizers instead of parameterizing the index/provider types.
- Two new FFI calls were added to diskann-garnet. `insert()` now returns a success flag which can also signal that an associated quantizer is ready for training. `build_quant_table()` is called by Garnet asynchronously after `insert()` signals readiness for training to build quantization tables, and `backfill_quant_vectors` is called when training is complete to quantize previously inserted vectors. These new calls allow Garnet to control the training and backfill threads.
- The accessor is now called `DynamicAccessor` because it dynamically switches between full precision only and quantized only operation depending on availability and readiness of the associated quantizer.
-  A new `visit_used()` call was added to the FSM to facilitate walking all stored vectors.
- The FSM is now lockable, which prevents it reusing IDs during quantization backfill. This means once backfilling starts, new inserts will be assigned IDs above any that need quantization.
- vectorset now supports quantization and distance metric settings, and a concurrency bug with multi-threaded inserts was fixed which caused IDs to be assigned incorrectly.

Disk persistence will be done in a follow along with more extensive testing.

Because of the FFI changes, the version has been bumped to 2.0.0.

### DIFF
--- a/diskann-garnet/src/alloc.rs
+++ b/diskann-garnet/src/alloc.rs
@@ -9,7 +9,7 @@ use std::ptr::NonNull;
 /// Custom allocator that over-aligns to 8 bytes. This is needed since Garnet will hand us byte slices for f32 data
 /// that may be unaligned, so we need an allocator to make owned, aligned byte containers.
 #[derive(Debug, Clone, Copy)]
-pub(crate) struct AlignToEight;
+pub struct AlignToEight;
 
 unsafe impl AllocatorCore for AlignToEight {
     #[inline]
--- a/diskann-garnet/src/alloc.rs
+++ b/diskann-garnet/src/alloc.rs
@@ -9,7 +9,7 @@ use std::ptr::NonNull;
 /// Custom allocator that over-aligns to 8 bytes. This is needed since Garnet will hand us byte slices for f32 data
 /// that may be unaligned, so we need an allocator to make owned, aligned byte containers.
 #[derive(Debug, Clone, Copy)]
-pub struct AlignToEight;
+pub(crate) struct AlignToEight;
 
 unsafe impl AllocatorCore for AlignToEight {
     #[inline]
--- a/diskann-garnet/src/dyn_index.rs
+++ b/diskann-garnet/src/dyn_index.rs
@@ -22,7 +22,7 @@ use std::sync::Arc;
 
 /// Type-erased version of `DiskANNIndex<GarnetProvider>`.
 /// All vector data is passed as untyped byte slices.
-pub trait DynIndex: Send + Sync {
+pub(crate) trait DynIndex: Send + Sync {
     fn insert(&self, context: &Context, id: &GarnetId, data: &[u8]) -> ANNResult<()>;
 
     fn set_attributes(&self, context: &Context, id: &GarnetId, data: &[u8]) -> ANNResult<()>;
--- a/diskann-garnet/src/dyn_index.rs
+++ b/diskann-garnet/src/dyn_index.rs
@@ -7,7 +7,7 @@ use crate::{
     SearchResults,
     garnet::{Context, GarnetId},
     labels::GarnetQueryLabelProvider,
-    provider::GarnetProvider,
+    provider::{DynamicQuantization, GarnetProvider},
 };
 use diskann::{
     ANNResult,
@@ -16,8 +16,7 @@ use diskann::{
     utils::VectorRepr,
 };
 use diskann_providers::{
-    index::wrapped_async::DiskANNIndex,
-    model::graph::provider::{async_::common::FullPrecision, layers::BetaFilter},
+    index::wrapped_async::DiskANNIndex, model::graph::provider::layers::BetaFilter,
 };
 use std::sync::Arc;
 
@@ -55,6 +54,10 @@ pub trait DynIndex: Send + Sync {
     fn internal_id_exists(&self, context: &Context, id: u32) -> bool;
 
     fn external_id_exists(&self, context: &Context, id: &GarnetId) -> bool;
+
+    fn train_quantizer(&self, context: &Context) -> bool;
+
+    fn backfill_quant_vectors(&self, context: &Context, task_idx: usize, task_count: usize);
 }
 
 impl<T: VectorRepr> DynIndex for DiskANNIndex<GarnetProvider<T>> {
@@ -63,7 +66,7 @@ impl<T: VectorRepr> DynIndex for DiskANNIndex<GarnetProvider<T>> {
     /// The data slice here must be aligned to `T` or this will panic.
     fn insert(&self, context: &Context, id: &GarnetId, data: &[u8]) -> ANNResult<()> {
         self.insert(
-            FullPrecision,
+            DynamicQuantization,
             context,
             id,
             bytemuck::cast_slice::<u8, T>(data),
@@ -87,10 +90,10 @@ impl<T: VectorRepr> DynIndex for DiskANNIndex<GarnetProvider<T>> {
     ) -> ANNResult<SearchStats> {
         let query = bytemuck::cast_slice::<u8, T>(data);
         if let Some((labels, beta)) = filter {
-            let beta_filter = BetaFilter::new(FullPrecision, Arc::new(labels.clone()), beta);
+            let beta_filter = BetaFilter::new(DynamicQuantization, Arc::new(labels.clone()), beta);
             self.search(*params, &beta_filter, context, query, output)
         } else {
-            self.search(*params, &FullPrecision, context, query, output)
+            self.search(*params, &DynamicQuantization, context, query, output)
         }
     }
 
@@ -104,14 +107,14 @@ impl<T: VectorRepr> DynIndex for DiskANNIndex<GarnetProvider<T>> {
     ) -> ANNResult<SearchStats> {
         // Look up internal ID
         let iid = self.inner.provider().to_internal_id(context, id)?;
-        let data = self.inner.provider().get_vector(context, iid)?;
+        let data = self.inner.provider().get_full_vector(context, iid)?;
         let data_bytes = bytemuck::cast_slice::<T, u8>(&data);
         self.search_vector(context, data_bytes, params, filter, output)
     }
 
     fn remove(&self, context: &Context, id: &GarnetId) -> ANNResult<()> {
         self.inplace_delete(
-            FullPrecision,
+            DynamicQuantization,
             context,
             id,
             3,
@@ -136,5 +139,15 @@ impl<T: VectorRepr> DynIndex for DiskANNIndex<GarnetProvider<T>> {
 
     fn external_id_exists(&self, context: &Context, id: &GarnetId) -> bool {
         self.inner.provider().vector_id_exists(context, id)
+    }
+
+    fn train_quantizer(&self, context: &Context) -> bool {
+        self.inner.provider().train_quantizer(context)
+    }
+
+    fn backfill_quant_vectors(&self, context: &Context, task_idx: usize, task_count: usize) {
+        self.inner
+            .provider()
+            .backfill_quant_vectors(context, task_idx, task_count);
     }
 }
--- a/diskann-garnet/src/fsm.rs
+++ b/diskann-garnet/src/fsm.rs
@@ -39,14 +39,14 @@ const FAST_SIZE: usize = 1024;
 const FSM_KEY_PREFIX: u32 = u32::from_be_bytes(*b"_fsm");
 
 #[derive(Debug, Error, PartialEq)]
-pub enum FsmError {
+pub(crate) enum FsmError {
     #[error("Garnet operation failed")]
     Garnet(#[from] GarnetError),
     #[error("requested ID is out of range {0}")]
     IdOutOfRange(u32),
 }
 
-pub struct FreeSpaceMap {
+pub(crate) struct FreeSpaceMap {
     /// Garnet callbacks for reading/writing FSM keys
     callbacks: Callbacks,
     /// A flag to signal whether there are free IDs in the FSM.
@@ -67,7 +67,7 @@ pub struct FreeSpaceMap {
 }
 
 impl FreeSpaceMap {
-    pub fn new(ctx: Context, callbacks: Callbacks) -> Result<Self, FsmError> {
+    pub(crate) fn new(ctx: &Context, callbacks: Callbacks) -> Result<Self, FsmError> {
         let has_free_ids = AtomicBool::new(false);
         let fast_free_list = ArrayQueue::new(FAST_SIZE);
         let max_block = RwLock::new(u32::MAX);
@@ -88,7 +88,7 @@ impl FreeSpaceMap {
         let block_key = Self::block_key(0);
         if this
             .callbacks
-            .exists_wid(ctx.term(Term::Metadata), block_key)
+            .exists_wid(&ctx.term(Term::Metadata), block_key)
         {
             this.load_state(ctx)?;
         } else {
@@ -101,11 +101,11 @@ impl FreeSpaceMap {
     }
 
     /// Load all state from Garnet by scanning the FSM blocks.
-    fn load_state(&mut self, ctx: Context) -> Result<(), FsmError> {
+    fn load_state(&mut self, ctx: &Context) -> Result<(), FsmError> {
         let mut max_block_id = 0;
         while self
             .callbacks
-            .exists_wid(ctx.term(Term::Metadata), Self::block_key(max_block_id))
+            .exists_wid(&ctx.term(Term::Metadata), Self::block_key(max_block_id))
         {
             max_block_id += 1;
         }
@@ -119,7 +119,7 @@ impl FreeSpaceMap {
 
             if !self
                 .callbacks
-                .read_single_wid(ctx.term(Term::Metadata), block_key, &mut block)
+                .read_single_wid(&ctx.term(Term::Metadata), block_key, &mut block)
             {
                 break;
             }
@@ -157,20 +157,20 @@ impl FreeSpaceMap {
     }
 
     /// Mark an ID as free.
-    pub fn mark_free(&self, ctx: Context, id: u32) -> Result<(), FsmError> {
+    pub(crate) fn mark_free(&self, ctx: &Context, id: u32) -> Result<(), FsmError> {
         // We don't care about the changed status on free.
         self.mark_id(ctx, id, false).map(|_| ())
     }
 
     /// Mark an ID as occupied.
-    fn mark_used(&self, ctx: Context, id: u32) -> Result<bool, FsmError> {
+    fn mark_used(&self, ctx: &Context, id: u32) -> Result<bool, FsmError> {
         self.mark_id(ctx, id, true)
     }
 
     /// Mark an ID according to value (true = used, false = free).
     /// Side effects only happen if the value actually changed. The return value reflects whether
     /// data changed or not.
-    fn mark_id(&self, ctx: Context, id: u32, used: bool) -> Result<bool, FsmError> {
+    fn mark_id(&self, ctx: &Context, id: u32, used: bool) -> Result<bool, FsmError> {
         if id >= self.next_id.load(Ordering::Acquire) {
             return Err(FsmError::IdOutOfRange(id));
         }
@@ -185,7 +185,7 @@ impl FreeSpaceMap {
         let mut changed = false;
 
         if !self.callbacks.rmw_wid(
-            ctx.term(Term::Metadata),
+            &ctx.term(Term::Metadata),
             block_key,
             BLOCK_SIZE_BYTES,
             |data: &mut [u8]| changed = update_status(used, &mut data[byte_idx], bit_idx),
@@ -211,7 +211,7 @@ impl FreeSpaceMap {
         Ok(changed)
     }
 
-    pub fn is_free(&self, ctx: Context, id: u32) -> Result<bool, FsmError> {
+    pub(crate) fn is_free(&self, ctx: &Context, id: u32) -> Result<bool, FsmError> {
         if id >= self.next_id.load(Ordering::Acquire) {
             return Err(FsmError::IdOutOfRange(id));
         }
@@ -227,7 +227,7 @@ impl FreeSpaceMap {
 
         if !self
             .callbacks
-            .read_single_wid(ctx.term(Term::Metadata), block_key, &mut block)
+            .read_single_wid(&ctx.term(Term::Metadata), block_key, &mut block)
         {
             return Err(FsmError::Garnet(GarnetError::Read));
         }
@@ -240,7 +240,7 @@ impl FreeSpaceMap {
     /// Return a a new ID.
     /// This may be a a fresh ID larger than all the others, or it may be a reused ID that
     /// previously belonged to a deleted element. The returned ID is marked as used.
-    pub fn next_id(&self, ctx: Context) -> Result<u32, FsmError> {
+    pub(crate) fn next_id(&self, ctx: &Context) -> Result<u32, FsmError> {
         if self.can_reuse() && self.has_free_ids.load(Ordering::Acquire) {
             // We retry reusing a freed ID until there are none or we get one and marking it used
             // succeeds in changing the value.
@@ -293,11 +293,11 @@ impl FreeSpaceMap {
         Ok(id)
     }
 
-    pub fn max_id(&self) -> u32 {
+    pub(crate) fn max_id(&self) -> u32 {
         self.next_id.load(Ordering::Acquire).saturating_sub(1)
     }
 
-    pub fn total_used(&self) -> usize {
+    pub(crate) fn total_used(&self) -> usize {
         self.total_used.load(Ordering::Acquire)
     }
 
@@ -320,7 +320,7 @@ impl FreeSpaceMap {
     }
 
     /// Scan the FSM blocks to fill up fast_free_list.
-    fn refill_fast_free_list(&self, ctx: Context) -> Result<bool, FsmError> {
+    fn refill_fast_free_list(&self, ctx: &Context) -> Result<bool, FsmError> {
         // NOTE: We take a write lock to prevent multiple refills happening simultaneously.
         #[allow(clippy::readonly_write_lock)]
         let max_block = self.max_block.write().unwrap();
@@ -337,7 +337,7 @@ impl FreeSpaceMap {
             let block_key = Self::block_key(block_id);
             if !self
                 .callbacks
-                .read_single_wid(ctx.term(Term::Metadata), block_key, &mut block)
+                .read_single_wid(&ctx.term(Term::Metadata), block_key, &mut block)
             {
                 return Err(FsmError::Garnet(GarnetError::Read));
             }
@@ -369,7 +369,7 @@ impl FreeSpaceMap {
     }
 
     /// Ensure enough blocks exist in the FSM to hold `id`.
-    fn expand_to(&self, ctx: Context, id: u32) -> Result<(), FsmError> {
+    fn expand_to(&self, ctx: &Context, id: u32) -> Result<(), FsmError> {
         let (block_id, _, _) = self.indexes_for_id(id);
         let mut max_block = self.max_block.write().unwrap();
         if *max_block == u32::MAX || block_id == *max_block + 1 {
@@ -378,7 +378,7 @@ impl FreeSpaceMap {
 
             if !self
                 .callbacks
-                .write_wid(ctx.term(Term::Metadata), block_key, &block_bytes)
+                .write_wid(&ctx.term(Term::Metadata), block_key, &block_bytes)
             {
                 return Err(FsmError::Garnet(GarnetError::Write));
             }
@@ -396,7 +396,7 @@ impl FreeSpaceMap {
     }
 
     /// Visit each used id in the FSM, invoking f on each id.
-    pub fn visit_used<F>(&self, ctx: Context, mut f: F) -> Result<(), FsmError>
+    pub(crate) fn visit_used<F>(&self, ctx: &Context, mut f: F) -> Result<(), FsmError>
     where
         F: FnMut(u32) -> bool,
     {
@@ -408,7 +408,7 @@ impl FreeSpaceMap {
             let block_key = Self::block_key(block_id);
             if !self
                 .callbacks
-                .read_single_wid(ctx.term(Term::Metadata), block_key, &mut block)
+                .read_single_wid(&ctx.term(Term::Metadata), block_key, &mut block)
             {
                 return Err(FsmError::Garnet(GarnetError::Read));
             }
@@ -442,14 +442,14 @@ impl FreeSpaceMap {
     /// Prevent the reuse of previously deleted IDs.
     /// Each call to this increments a counter, and only once the counter is back to zero
     /// will reuse be allowed again.
-    pub fn lock_reuse(&self) {
+    pub(crate) fn lock_reuse(&self) {
         self.reuse_lock.fetch_add(1, Ordering::AcqRel);
     }
 
     /// Resume reuse of previously deleted IDs.
     /// Each call to this decrements a counter, and only once the counter is back to zero
     /// will reuse be allowed. This returns whether reuse was actually enabled.
-    pub fn unlock_reuse(&self) -> bool {
+    pub(crate) fn unlock_reuse(&self) -> bool {
         let prev = self.reuse_lock.fetch_sub(1, Ordering::AcqRel);
         debug_assert_ne!(prev, 0);
         prev == 1
@@ -487,7 +487,10 @@ mod tests {
     {
         let key = FreeSpaceMap::block_key(block_id);
         let block = store
-            .get(Context(0).term(Term::Metadata).0, bytemuck::bytes_of(&key))
+            .get(
+                Context::new(0).term(Term::Metadata).get(),
+                bytemuck::bytes_of(&key),
+            )
             .unwrap();
         f(&block);
     }
@@ -497,7 +500,7 @@ mod tests {
         let store = Store;
 
         // A fresh FSM should result in full block of all zeroes.
-        let _fsm = FreeSpaceMap::new(Context(0), store.callbacks()).unwrap();
+        let _fsm = FreeSpaceMap::new(&Context::new(0), store.callbacks()).unwrap();
         verify_block(&store, 0, |block| {
             assert_eq!(block.len(), BLOCK_SIZE_BYTES);
             assert_eq!(block.iter().filter(|b| **b != 0).count(), 0);
@@ -508,29 +511,29 @@ mod tests {
     fn basic_next_id() {
         let store = Store;
 
-        let fsm = FreeSpaceMap::new(Context(0), store.callbacks()).unwrap();
+        let fsm = FreeSpaceMap::new(&Context::new(0), store.callbacks()).unwrap();
 
         // A fresh FSM should not have free IDs.
         assert!(!fsm.has_free_ids.load(std::sync::atomic::Ordering::Acquire));
 
         // It should return sequentials IDs starting from 0.
-        assert_eq!(fsm.next_id(Context(0)).unwrap(), 0);
-        assert_eq!(fsm.next_id(Context(0)).unwrap(), 1);
+        assert_eq!(fsm.next_id(&Context::new(0)).unwrap(), 0);
+        assert_eq!(fsm.next_id(&Context::new(0)).unwrap(), 1);
     }
 
     #[test]
     fn basic_delete() {
         let store = Store;
 
-        let fsm = FreeSpaceMap::new(Context(0), store.callbacks()).unwrap();
+        let fsm = FreeSpaceMap::new(&Context::new(0), store.callbacks()).unwrap();
 
         // Deleting a ID out of range should return an error.
-        let res = fsm.mark_free(Context(0), 0);
+        let res = fsm.mark_free(&Context::new(0), 0);
         assert!(res.is_err());
         assert_eq!(res.err().unwrap(), FsmError::IdOutOfRange(0));
 
         for _ in 0u32..64 {
-            let _ = fsm.next_id(Context(0)).unwrap();
+            let _ = fsm.next_id(&Context::new(0)).unwrap();
         }
 
         verify_block(&store, 0, |block| {
@@ -542,8 +545,8 @@ mod tests {
             }
         });
 
-        fsm.mark_free(Context(0), 37).unwrap();
-        fsm.mark_free(Context(0), 9).unwrap();
+        fsm.mark_free(&Context::new(0), 37).unwrap();
+        fsm.mark_free(&Context::new(0), 9).unwrap();
 
         verify_block(&store, 0, |block| {
             assert_eq!(block[0], 0b11111111);
@@ -566,20 +569,20 @@ mod tests {
     fn basic_id_reuse() {
         let store = Store;
 
-        let fsm = FreeSpaceMap::new(Context(0), store.callbacks()).unwrap();
+        let fsm = FreeSpaceMap::new(&Context::new(0), store.callbacks()).unwrap();
 
         for _ in 0u32..64 {
-            let _ = fsm.next_id(Context(0)).unwrap();
+            let _ = fsm.next_id(&Context::new(0)).unwrap();
         }
 
         // Next ID should be 64 since none are free before that.
-        assert_eq!(fsm.next_id(Context(0)).unwrap(), 64);
+        assert_eq!(fsm.next_id(&Context::new(0)).unwrap(), 64);
 
         // After deleting an ID, it should be returned from `next_id()`.
-        fsm.mark_free(Context(0), 37).unwrap();
-        assert_eq!(fsm.next_id(Context(0)).unwrap(), 37);
+        fsm.mark_free(&Context::new(0), 37).unwrap();
+        assert_eq!(fsm.next_id(&Context::new(0)).unwrap(), 37);
         // Once all free IDs are used, fresh ones should be returned.
-        assert_eq!(fsm.next_id(Context(0)).unwrap(), 65);
+        assert_eq!(fsm.next_id(&Context::new(0)).unwrap(), 65);
         // And has_free_ids should now be false.
         assert!(!fsm.has_free_ids.load(Ordering::Acquire));
     }
@@ -588,32 +591,32 @@ mod tests {
     fn basic_recovery() {
         let store = Store;
 
-        let fsm = FreeSpaceMap::new(Context(0), store.callbacks()).unwrap();
+        let fsm = FreeSpaceMap::new(&Context::new(0), store.callbacks()).unwrap();
 
         for _ in 0u32..64 {
-            let _ = fsm.next_id(Context(0)).unwrap();
+            let _ = fsm.next_id(&Context::new(0)).unwrap();
         }
 
-        fsm.mark_free(Context(0), 37).unwrap();
+        fsm.mark_free(&Context::new(0), 37).unwrap();
 
         // Loading FSM from store should recover all the state.
-        let fsm = FreeSpaceMap::new(Context(0), store.callbacks()).unwrap();
+        let fsm = FreeSpaceMap::new(&Context::new(0), store.callbacks()).unwrap();
         assert_eq!(*fsm.max_block.read().unwrap(), 0);
         assert_eq!(fsm.next_id.load(Ordering::Acquire), 64);
         assert!(fsm.has_free_ids.load(Ordering::Acquire));
         assert_eq!(fsm.fast_free_list.len(), 1);
-        assert_eq!(fsm.next_id(Context(0)).unwrap(), 37);
+        assert_eq!(fsm.next_id(&Context::new(0)).unwrap(), 37);
     }
 
     #[test]
     fn dynamic_expansion() {
         let store = Store;
 
-        let fsm = FreeSpaceMap::new(Context(0), store.callbacks()).unwrap();
+        let fsm = FreeSpaceMap::new(&Context::new(0), store.callbacks()).unwrap();
 
         // Asking for more than BLOCK_SIZE_IDS will force another FSM block to be allocated.
         for _ in 0u32..BLOCK_SIZE_IDS as u32 + 1 {
-            let _ = fsm.next_id(Context(0)).unwrap();
+            let _ = fsm.next_id(&Context::new(0)).unwrap();
         }
     }
 }
--- a/diskann-garnet/src/fsm.rs
+++ b/diskann-garnet/src/fsm.rs
@@ -26,7 +26,7 @@
 use crossbeam::queue::ArrayQueue;
 use std::sync::{
     RwLock,
-    atomic::{AtomicBool, AtomicU32, Ordering},
+    atomic::{AtomicBool, AtomicU32, AtomicUsize, Ordering},
 };
 use thiserror::Error;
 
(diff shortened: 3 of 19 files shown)


diskann-garnet/src/provider.rs
```diff
@@ -22,21 +22,40 @@ use diskann::{
...
+        self, DynDistanceComputer, DynQueryComputer, GarnetQuantizer, GarnetQuantizerError,
+    },
 };
 
+thread_local! {
+    /// Thread local flag to detect when we've reached the quantization threshold. This is needed
+    /// to return the correct status at the end of insert() since the provider code that becomes
+    /// aware of the state change cannot directly communicate it back to the caller.
+    pub static QUANTIZER_READY: AtomicBool = const { AtomicBool::new(false) };
+}
```
hildebrandmw: Can we rely on a thread-local for this? What if the same thread is being used to search multiple indexes?

Could this be stored in the `Context` instead?

diskann-garnet/src/provider.rs
```diff
@@ -134,16 +216,50 @@ impl<T: VectorRepr> GarnetProvider<T> {
 
         let fsm = FreeSpaceMap::new(context, callbacks)?;
 
+        let (quantizer, canonical_bytes, all_quantized) = match quant_type {
+            VectorQuantType::NoQuant | VectorQuantType::XPreQ8 => (None, 0, false),
+            VectorQuantType::Invalid => return Err(GarnetProviderError::InvalidQuantizer),
```
hildebrandmw, line 229: Can you help me understand why `VectorQuantType::Invalid` exists?

diskann-garnet/src/provider.rs
```diff
@@ -134,16 +216,50 @@ impl<T: VectorRepr> GarnetProvider<T> {
...
+            VectorQuantType::Q8 => {
+                if TypeId::of::<T>() != TypeId::of::<f32>() {
```
hildebrandmw, line 231: I don't think relying on `TypeId` is a good idea. It feels like ad-hoc specialization and can cause issues if we add `f16` support in the future. I understand that the Redis API doesn't support `f16`, but it has historically proved itself to be useful. Can this be packaged behind a trait? Or use `VectorRepr::as_f32`?

diskann-garnet/src/provider.rs
```diff
@@ -83,21 +152,32 @@ impl From<GarnetProviderError> for ANNError {
 
 diskann::always_escalate!(GarnetProviderError);
 
+/// The Garnet DataProvider implementation.
 pub struct GarnetProvider<T: VectorRepr> {
```
hildebrandmw: Why parameterize by `T` anymore? If we ever end up varying `T`, a bunch of code will get re-monomorphized when what actually changes based on `T` is relatively small.

diskann-garnet/src/quantization.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    fn canonical_bytes(&self) -> usize {
+        Data::<1, GlobalAllocator>::canonical_bytes(self.dim)
+    }
+
+    fn train(
+        &self,
+        metric_type: Metric,
+        data: MatrixView<f32>,
+    ) -> Result<(), GarnetQuantizerError> {
```
hildebrandmw, line 100: Should the whole training loop be done under a lock, with training skipped if the inner quantizer is already initialized?

Otherwise, we risk overwriting the quantizer with a different one when that risk can be mitigated locally.

diskann-garnet/src/quantization.rs
```diff
@@ -0,0 +1,316 @@
...
+/// This quantizer corresponds to `BIN` quantizer in the Redis protocol. It requires hundreds of
+/// vectors (but not thousands) for training. Quantized vectors have 1 bit per dimension plus up
+/// to 6 bytes of overhead.
+pub struct Spherical1Bit {
+    dim: usize,
+    inner: RwLock<Option<spherical::iface::Impl<1, GlobalAllocator>>>,
+}
+
+impl Spherical1Bit {
+    pub fn new(dim: usize) -> Self {
```
hildebrandmw: Should we have the same non-zero check as `MinMax`? Otherwise, failure for zero dim will deferred until the quantizer is trained (though we'd probably fail way before that somewhere else ... hopefully).

diskann-garnet/src/quantization.rs
```diff
@@ -0,0 +1,316 @@
...
+        let mut rng = rand::rng();
+        let transform = Transform::new(
+            TransformKind::DoubleHadamard {
+                target_dim: diskann_quantization::algorithms::transforms::TargetDim::Same,
+            },
+            dim,
+            Some(&mut rng),
+            GlobalAllocator,
+        )?;
+        let grid_scale = Positive::new(1.0).unwrap();
```
hildebrandmw: You can use [`POSITIVE_ONE_F32`](https://github.com/microsoft/DiskANN/blob/d516da117148c49ab534673ea701f4cae026e3d7/diskann-quantization/src/num.rs#L52) if you want to avoid the `uwnrap`.

diskann-garnet/src/quantization.rs
```diff
@@ -0,0 +1,316 @@
...
+
+        Ok(Self {
+            dim: dim.get(),
```
hildebrandmw: We track `dim`? It's available via [`MinMaxQuantizer::dim`](https://github.com/microsoft/DiskANN/blob/d516da117148c49ab534673ea701f4cae026e3d7/diskann-quantization/src/minmax/quantizer.rs#L70-L73)

diskann-garnet/src/quantization.rs
```diff
@@ -0,0 +1,316 @@
...
+            let computer = quantizer
+                .distance_computer(GlobalAllocator)
+                .map_err(|e| GarnetQuantizerError::Alloc(Box::new(e)))?;
+            Ok(GarnetDistanceComputer::new(computer))
+        } else {
+            Err(GarnetQuantizerError::NoQuantizer)
+        }
+    }
+
+    fn query_computer(&self, query: &[f32]) -> Result<GarnetQueryComputer, GarnetQuantizerError> {
```
hildebrandmw, line 152: We might want an additional boolean argument indicating whether or not this is a query computer for graph insert or for general search. For the former, we want to leave open the ability to use the `SameAsData` query layout so distances computed during the search phase of insert are more comparable to pruned distances.

diskann-garnet/src/quantization.rs
```diff
@@ -0,0 +1,316 @@
...
+        self.inner.read().unwrap().is_some()
+    }
+
+    fn required_vectors(&self) -> usize {
+        1000
+    }
+
+    fn canonical_bytes(&self) -> usize {
+        Data::<1, GlobalAllocator>::canonical_bytes(self.dim)
+    }
```
hildebrandmw, line 94: Nit: maybe just `bytes`? No need to leak the implementation details of `diskann-quantization`.

diskann-garnet/src/provider.rs
```diff
@@ -240,6 +377,173 @@ impl<T: VectorRepr> GarnetProvider<T> {
...
+                }
+                Err(_) => return false,
+            };
+
+        // Ensure we don't kick off training twice.
+        match self.training_started.compare_exchange(
+            0,
+            current_time,
+            Ordering::AcqRel,
+            Ordering::Acquire,
```
hildebrandmw: Why are `AcqRel/Acquire` needed? It doesn't look to me like any other variables are being synchronized with this atomic, so isn't `Relaxed` sufficient?

diskann-garnet/src/provider.rs
```diff
@@ -240,6 +377,173 @@ impl<T: VectorRepr> GarnetProvider<T> {
...
+        let max_id = self.fsm.max_id() as usize;
+
+        let task_count = task_count.min(max_id);
+        if task_idx >= task_count {
+            return;
+        }
+
+        let work_count = max_id.div_euclid(task_count);
+        let start_id = (work_count * task_idx) as u32;
+        let end_id = (work_count * (task_idx + 1)).min(max_id) as u32;
```
hildebrandmw: One suggestion would be to use [`partition`](https://github.com/microsoft/DiskANN/blob/d516da117148c49ab534673ea701f4cae026e3d7/diskann/src/utils/async_tools.rs#L282-L302), which will do slightly better load balancing by keeping the chunk size more uniform instead of making the last one small.

**From agent review**: Using `div_euclid` means that we miss items at the end. If `max_id = 10` and `task_count = 3`, the last task processed `6..=9` instead of ever getting to 10.

Also, the use of inclusive ranges means items on the boundary get processed twice.

diskann-garnet/src/provider.rs
```diff
@@ -240,6 +377,173 @@ impl<T: VectorRepr> GarnetProvider<T> {
...
+
+        self.fsm.lock_reuse();
+
+        let mut v = vec![0f32; self.dim];
+        let mut q = vec![0u8; quantizer.canonical_bytes()];
+        for id in start_id..=end_id {
```
hildebrandmw: This entire loop suppresses errors rather than correcting or reporting, why is that?

diskann-garnet/src/provider.rs
```diff
@@ -240,6 +377,173 @@ impl<T: VectorRepr> GarnetProvider<T> {
...
+                .callbacks
+                .write_iid(context.term(Term::Quantized), id, &q)
+            {
+                continue;
+            }
+        }
+
+        let backfill_finished = self.fsm.unlock_reuse();
+
+        if backfill_finished {
```
hildebrandmw, line 542: This isn't a reliable signal: it races. For example, if you have two tasks: the first one gets preempted before acquiring the reuse lock. The second one then acquires the lock, processes data, and then releases the lock. This release will see the counter as zero and think everything is good, but the first task never even started.

In general, the external interaction with the free-space map feels brittle. There are many atomic variables interacting with each other. As another example, checking `can_reuse` in `FreeSpaceMap::next_id` also doesn't feel right. See my other comment.

diskann-garnet/src/fsm.rs
```diff
@@ -214,7 +241,7 @@ impl FreeSpaceMap {
     /// This may be a a fresh ID larger than all the others, or it may be a reused ID that
     /// previously belonged to a deleted element. The returned ID is marked as used.
     pub fn next_id(&self, ctx: Context) -> Result<u32, FsmError> {
-        if self.has_free_ids.load(Ordering::Acquire) {
+        if self.can_reuse() && self.has_free_ids.load(Ordering::Acquire) {
```
hildebrandmw: Is this check correct? `self.can_reuse` can return `true` before any quantization happens, causing this check to pass. Then quantization can be initiated. This means we can be in the loop below **and** doing quantization. I assume this isn't the intended goal since the check exists, but is there a mechanism by which `next_id` broadcasts "I'm running".

diskann-garnet/src/provider.rs
```diff
@@ -287,11 +591,31 @@ impl<T: VectorRepr> SetElement<&[T]> for GarnetProvider<T> {
     ) -> Result<Self::Guard, Self::SetError> {
         let internal_id = self.fsm.next_id(*context)?;
 
+        // Set quantization readiness
+        if let Some(quantizer) = &self.quantizer
+            && !quantizer.is_prepared()
+            && self.fsm.total_used() > quantizer.required_vectors()
+        {
+            QUANTIZER_READY.with(|v| v.store(true, Ordering::Release));
```
hildebrandmw: How does this avoid mayhem when the same thread is used to search another index?

diskann-garnet/src/provider.rs
```diff
@@ -240,6 +377,173 @@ impl<T: VectorRepr> GarnetProvider<T> {
...
+                    if t == 0 {
+                        return false;
+                    }
+                    t
+                }
+                Err(_) => return false,
+            };
+
+        // Ensure we don't kick off training twice.
+        match self.training_started.compare_exchange(
```
hildebrandmw: If training fails, `training_started` is never reset. Should we be able to restart training?

hildebrandmw (commented):
Thanks Jack! It will be great to have quantization land in `diskann-garnet`.

After seeing the 1000-foot view, the things that stand out to me are:
* The somewhat awkward representation of quantizer (`Option<Box<dyn>>` in `GarnetProvider` with `RwLock<Option<_>>` as the implementation).
* Heavy use of conditionals throughout vector retrieval checking for quantization, especially with respect to the start point cache. This also necessitated the working-set check and drain if we switched to quantization part-way through insert.
* Race conditions in the free space map and quantization.

One idea I had regarding the first two points is to lean more heavily into trait objects with something like
```rust
trait Vector {
    // The number of bytes needed per vector
    fn bytes(&self) -> usize;
    fn is_quantized(&self) -> bool;
    // Ther term to provide to `Garnet` for data retrieval.
    fn term(&self) -> Term;
    fn cached(&self, id: u32) -> Option<&[u8]>;
    fn distance_computer(&self) -> Box<dyn _>;
    fn query_computer(&self, ...) -> Box<dyn _>;
}
```
The idea being that both full-precision and quantized vectors can live behind the same facade. Then in the `Provider`, we can have:
```
// Naming and exact structure are just meant to describe the idea
struct GarnetProvider {
    primary: Arc<dyn Vector>,
    quant: Option<Arc<dyn Vector>>,
    ...
}
```
The `Accessor` can then be 
```
struct Accessor<'a> {
    provider: &'a GarnetProvider,
    vectors: &'a dyn Vector,
}
```
With most of the non-reranking vector needs going through `Accessor::vectors`. When no quantization is active, then `vectors = GarnetProvider::primary`, otherwise, `vectors = GarnetProvider::quant`. Then all the vector retrieval code stays pretty much the same without needing to constantly branch on whether or not the vectors are quantized. Further, the issue with the working set goes away because vectors that start their insert as full-precision stay as full-precision for the duration of the insert.

vectorset/src/main.rs
```diff
@@ -407,10 +462,14 @@ async fn ingest<T: Element>(
                                 pipeline.arg(b"XPREQ8");
                             }
                             DataType::Float32 => {
-                                pipeline.arg(b"NOQUANT");
+                                pipeline.arg(quantizer);
```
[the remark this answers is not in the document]
metajack: VADD has a default quantizer, so this is fine.

diskann-garnet/src/provider.rs
metajack: I could store it in context if I wrap add an internal field to context. I suppose that is slightly nicer.

I'm not sure what issue you are envisioning with the threads. The check is between the guts of insert and the return of insert. There's no way for another chunk of code to run on our thread in the meantime. If something were to happen, the bad behavior is limited to returning SUCCESS_TRAINING_READY, which will cause garnet to dispatch train_quantizer(). That getting called a second time is fine, because train_quantizer has detection for whether training has already started and exits if that's the case.

diskann-garnet/src/provider.rs
metajack: Because it exists in Garnet. @kevin-montrose can maybe say more about why.

kevin-montrose: It's just a defensive coding pattern in .NET - generally we'll get everything 0 initialized, but in some of the trickier code in Garnet we don't have definite assignment guarantees.  By reserving 0 in [`VectorQuantType`](https://github.com/microsoft/garnet/blob/185a2c49bc57fa5eb23c1a2c919b3327441ace47/libs/server/Storage/Session/MainStore/VectorStoreOps.cs#L18) as a "should never be used"-value we're less likely to have something half-work by accident.  We're not particularly consistent about following that pattern, but we did it for at least some of the Vector Set related enums.

diskann-garnet/src/provider.rs
hildebrandmw: What happens, though, if there are *multiple* indexes. One has been had quantization bootstrapped and the other has not. If a thread inserts on the bootstrapped index and then runs on the non-boot-strapped index, the thread local will still be set to `quantized`. Does `garnet` not reuse threads for multiple indexes?

diskann-garnet/src/provider.rs
hildebrandmw: I see, make sense! Thanks for the explanation!

diskann-garnet/src/quantization.rs
```diff
@@ -0,0 +1,316 @@
...
+        metric_type: Metric,
+        data: MatrixView<f32>,
+    ) -> Result<(), GarnetQuantizerError> {
+        let mut rng = rand::rng();
+        let quantizer = SphericalQuantizer::train(
+            data.as_view(),
+            TransformKind::DoubleHadamard {
+                target_dim: diskann_quantization::algorithms::transforms::TargetDim::Same,
+            },
+            SupportedMetric::try_from(metric_type).unwrap(),
```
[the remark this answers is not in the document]
metajack: Fixed

diskann-garnet/src/provider.rs
```diff
@@ -240,6 +377,173 @@ impl<T: VectorRepr> GarnetProvider<T> {
...
+
+                // Read the vector into the data matrix.
+                // Note that it's ok to read f32 instead of T here, because this can only get called when T == f32.
+                let row = data.row_mut(row_idx);
+                if !self
+                    .callbacks
+                    .read_single_iid(context.term(Term::Vector), id, row)
+                {
+                    return false;
+                }
```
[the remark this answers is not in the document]
metajack: Fixed.

diskann-garnet/src/provider.rs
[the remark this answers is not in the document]
metajack: Fixed.

diskann-garnet/src/provider.rs
```diff
@@ -782,75 +1277,162 @@ impl<'a, 'b, T: VectorRepr> SearchPostProcess<FullAccessor<'a, T>, &'b [T], Garn
...
+                        &mut v,
+                    ) {
+                        bad_read = true;
+                    }
+
+                    Some((
+                        n.id,
+                        f.evaluate_similarity(query, bytemuck::cast_slice::<u8, T>(&v)),
+                    ))
+                }
```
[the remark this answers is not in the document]
metajack: Fixed.

diskann-garnet/src/lib.rs
```diff
@@ -466,6 +483,34 @@ where
...
+/// # Safety
+///
+/// FFI
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn build_quant_table(context: u64, index_ptr: *const c_void) -> bool {
+    let index = unsafe { &*index_ptr.cast::<Index>() };
+    let ctx = Context(context);
+
+    index.inner.train_quantizer(&ctx)
+}
```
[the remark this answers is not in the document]
metajack: Fixed.

rfcs/00000-quantizer-bootstrap.md
```diff
@@ -0,0 +1,198 @@
...
+   strategy for Phase Two.
+2. **Quantization Backfill**: During Phase Two, previously inserted vectors will
+   need to be quantized. As background jobs are a performance concern, how
+   exactly this is accomplished must be customizable by the data provider.
+
+### Switching Strategies
+
+DiskANN already has the ability to run multiple strategies including hybrid full
+precision and quantized ones. These strategies represent the high level intent,
+but the data provider can choose alternate implementions depending on the data
```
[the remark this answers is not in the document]
metajack: Fixed.

rfcs/00000-quantizer-bootstrap.md
```diff
@@ -0,0 +1,198 @@
...
+
+#### Insertion and Deletion
+
+Insertion and deletion can remain largely the same in the data provider
+implementation. Inserts will need to write vectors, mappings, attributes, et al
+into storage, and can track the current phase to gate writes to quantized
+vectors. The search portion of these operations will return different objects
+depending on the current phase.
+
+For example, consider a `DataProvder::set_element()` implementation:
```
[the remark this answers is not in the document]
metajack: Fixed.

rfcs/00000-quantizer-bootstrap.md
```diff
@@ -0,0 +1,198 @@
...
+
+```rust
+struct ExampleProvider {
+     // other fields omitted
+     quantizer: Option<Quantizer>,
+}
+
+impl SetElement<[f32]> for ExampleProvider {
+     // associated types ommitted
```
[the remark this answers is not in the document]
metajack: Fixed.

rfcs/00000-quantizer-bootstrap.md
```diff
@@ -0,0 +1,198 @@
...
+        context: &Self::Context,
+        id: &Self::ExternalId,
+        element: &[T],
+     ) -> Result<Self::Guard, Self::SetError> {
+          let internal_id = self.new_id()?;
+          self.write_vector(context, internal_id, element)?;
+          self.set_internal_map(internal_id, id)?;
+          self.set_external_map(id, internal_id)?;
+
+          // Quantize and storage quant vector if we have a quantizer.
```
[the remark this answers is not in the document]
metajack: Fixed.

rfcs/00000-quantizer-bootstrap.md
```diff
@@ -0,0 +1,198 @@
...
+          if let Some(quantizer) = self.quantizer {
+               let qv = quantizer.quantize(element)?;
+               self.write_quant_vector(context, internal_id, element)?;
+          } else {
+               // This function will check if we are ready for Phase Two, and if so, do or schedule the quantizer intialization.
```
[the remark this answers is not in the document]
metajack: Fixed.

rfcs/00000-quantizer-bootstrap.md
```diff
@@ -0,0 +1,198 @@
+# Quantization Bootstrapping
+
+| | |
+|---|---|
+| **Authors** | Jack Moffitt |
```
harsha-simhadri: Should we break this file into what is general for the entire repo and specifics for Garnet?

vectorset/src/main.rs
```diff
@@ -161,6 +169,50 @@ enum DataType {
     Float32,
 }
 
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ValueEnum)]
+enum Quantizer {
+    None,
+    Bin,
```
harsha-simhadri, line 177: Add comments to clarify the quantizers they use in the back end.
Bin --> Spherical quantizer?
Q8 -->

vectorset/src/main.rs
```diff
@@ -161,6 +169,50 @@ enum DataType {
...
+}
+
+impl ToRedisArgs for Quantizer {
+    fn write_redis_args<W>(&self, out: &mut W)
+    where
+        W: ?Sized + redis::RedisWrite,
+    {
+        let q = match self {
+            Quantizer::None => b"NOQUANT".as_slice(),
+            Quantizer::Bin => b"BIN".as_slice(),
```
harsha-simhadri, line 207: What is the path to configuring specifics of Spherical Quantizer in the future. 1-bit may not be enough for many datasets, especially lower dimension ones.

vectorset/src/loader.rs
```diff
@@ -60,43 +51,40 @@ impl<T: bytemuck::Pod> DatasetLoader<T> {
...
-            {
-                let mut f = self.file.lock().await;
-
-                first_id = f.1;
-                if f.1 >= self.num_vectors {
-                    buffer.clear();
-                    return Ok((0, first_id));
-                }
-
-                buffer.resize(BATCH_SIZE * self.dim, T::zeroed());
```
harsha-simhadri, line 75: lets wire up load to the library's load.

diskann-garnet/src/alloc.rs
```diff
@@ -9,7 +9,7 @@ use std::ptr::NonNull;
 /// Custom allocator that over-aligns to 8 bytes. This is needed since Garnet will hand us byte slices for f32 data
 /// that may be unaligned, so we need an allocator to make owned, aligned byte containers.
 #[derive(Debug, Clone, Copy)]
-pub(crate) struct AlignToEight;
+pub struct AlignToEight;
```
harsha-simhadri: IS this used outside the crate?

diskann-garnet/src/lib.rs
```diff
@@ -375,6 +377,10 @@ fn interpret_vector<'a>(
     Some(v)
 }
 
+const INSERT_FAIL: u8 = 0;
```
harsha-simhadri: could this be enum INSERT_RESULT?

diskann-garnet/src/test_utils.rs
```diff
@@ -203,7 +204,14 @@ mod tests {
...
+        let provider = GarnetProvider::<u8>::new(
+            dim,
+            VectorQuantType::NoQuant,
+            Metric::L2,
+            max_degree,
+            callbacks,
+            ctx,
+        );
 
         // Provider should be created successfully
```
harsha-simhadri, line 216: Any quantization-based tests here?

diskann-garnet/src/quantization.rs
```diff
@@ -0,0 +1,317 @@
...
+    }
+}
+
+impl GarnetQuantizer for Spherical1Bit {
+    fn is_prepared(&self) -> bool {
+        self.inner.read().unwrap().is_some()
+    }
+
+    fn required_vectors(&self) -> usize {
+        1000
```
harsha-simhadri, line 89: Should this 1000 be part of spherical quantizer code, so others can reuse this. We should also point to experiments that conclude this is enough.

diskann-garnet/src/quantization.rs
```diff
@@ -0,0 +1,317 @@
...
+    fn is_prepared(&self) -> bool;
+    /// Returns the number of vectors needed before the quantizer can be trained
+    fn required_vectors(&self) -> usize;
+    /// Returns the size of a quantized vector
+    fn canonical_bytes(&self) -> usize;
+    /// Train the quantizer.
+    /// Each row of the matrix will be a vector
+    fn train(&self, metric: Metric, data: MatrixView<f32>) -> Result<(), GarnetQuantizerError>;
+    /// Quantize a vector
+    fn compress(&self, v: &[f32], into: &mut [u8]) -> Result<(), GarnetQuantizerError>;
```
harsha-simhadri, line 51: Why is the source vector limited to [f32]? We do plan to support other types like [i8], right?

diskann-garnet/src/quantization.rs
```diff
@@ -0,0 +1,317 @@
...
+            self,
+            Opaque::new(a),
+        )
+        .unwrap()
+    }
+}
+
+/// 8-bit scalar quantizer using MinMax
+///
+/// This quantizer requires no training at all and is usable immediately on the first first. Each
```
harsha-simhadri: first first -> first vector?

diskann-garnet/src/quantization.rs
```diff
@@ -0,0 +1,317 @@
...
+        Ok(Self(inner))
+    }
+}
+
+impl DynQueryComputer for MinMax8BitQueryComputer {
+    fn evaluate_similarity(&self, a: &[u8]) -> f32 {
+        let a = diskann_providers::common::MinMax8::from_bytes(a);
+        self.0.evaluate_similarity(a)
+    }
+}
```
harsha-simhadri, line 307: Unit tests for these trait impls would be useful

diskann-garnet/src/provider.rs
```diff
@@ -83,21 +152,32 @@ impl From<GarnetProviderError> for ANNError {
...
+    /// Tracks whether quantization backfill is complete.
+    all_quantized: AtomicBool,
+    /// Tracks whether training has already started.
+    training_started: AtomicU64,
     id_buffer_pool: ObjectPool<AdjList>,
     filtered_ids_pool: ObjectPool<Vec<u32>>,
+    quant_buffer_pool: ObjectPool<Vec<u8>>,
     neighbor_cache: DashMap<u32, Vec<u32>, foldhash::fast::RandomState>,
-    start_point_cache: DashMap<u32, Vec<T>, foldhash::fast::RandomState>,
+    start_point_cache: DashMap<u32, Poly<[u8], AlignToEight>, foldhash::fast::RandomState>,
```
harsha-simhadri, line 177: some comments on what these variables are meant for would be helpful

diskann-garnet/src/provider.rs
```diff
@@ -240,6 +377,189 @@ impl<T: VectorRepr> GarnetProvider<T> {
...
+                false
+            }
+        }
+    }
+
+    /// Bulk quantize previously inserted vectors.
+    ///
+    /// This function will be invoked on multiple threads. The total number of tasks and the ID of
+    /// the current task are given as inputs.
+    pub fn backfill_quant_vectors(&self, context: &Context, task_idx: usize, task_count: usize) {
```
harsha-simhadri: IS this run as a blocking call when count threshold is reached, or on a seperate thread?

diskann-garnet/src/quantization.rs
hildebrandmw: Using `f32` as data type to pass across the trait object boundary is the right choice for now. `i8` data would almost surely need to be upconverted to `f32` for most quantizers, so it keeps this part of the interface simpler.

diskann-garnet/src/quantization.rs
hildebrandmw: Not blocking.

diskann-garnet/src/provider.rs
metajack: Garnet dispatches these on a spearate thread pool.

diskann-garnet/src/dyn_index.rs
```diff
@@ -102,16 +105,25 @@ impl<T: VectorRepr> DynIndex for DiskANNIndex<GarnetProvider<T>> {
...
+                context,
+            )?;
+
         // Look up internal ID
         let iid = self.inner.provider().to_internal_id(context, id)?;
-        let data = self.inner.provider().get_vector(context, iid)?;
-        let data_bytes = bytemuck::cast_slice::<T, u8>(&data);
-        self.search_vector(context, data_bytes, params, filter, output)
+        let data = rt.block_on(accessor.get_element(iid))?;
+        self.search_vector(context, &data, params, filter, output)
```
hildebrandmw: Does this break when using quantization? A quantized vector is retrieved from the accessor but `search_vector` is expecting a full-precision vector?

diskann-garnet/src/provider.rs
```diff
@@ -230,42 +361,209 @@ impl<T: VectorRepr> GarnetProvider<T> {
...
         }
 
-        Ok(v)
+        let work_count = (max_id + 1) / task_count; // will be >= 1
+        let start_id = (work_count * task_idx) as u32;
+        let end_id = (work_count * (task_idx + 1)).min(max_id + 1) as u32;
+
+        self.fsm.lock_reuse();
+
+        let mut v = vec![0f32; self.dim];
```
hildebrandmw: Does this still hold when using `u8`/`i8`?

diskann-garnet/src/provider.rs
```diff
@@ -230,42 +361,209 @@ impl<T: VectorRepr> GarnetProvider<T> {
...
+
+        let max_id = self.fsm.max_id() as usize;
+
+        let task_count = task_count.min(max_id + 1);
+        if task_idx >= task_count {
+            return;
         }
 
-        Ok(v)
+        let work_count = (max_id + 1) / task_count; // will be >= 1
```
hildebrandmw: Use `.div_ceil` - otherwise we lose stuff at the end.

diskann-garnet/src/provider.rs
```diff
@@ -601,61 +1067,106 @@ impl<'a, T> Reborrow<'a> for Escape<T> {
...
+                {
+                    if let Some(guard) = self.provider.start_point_quant_cache.get(&id) {
+                        e.insert(Escape((&**guard).into()));
+                    }
+                } else if let Entry::Vacant(e) = set.entry(id) {
+                    if let Some(guard) = self.provider.start_point_cache.get(&id) {
+                        e.insert(Escape((&**guard).into()));
+                    }
+                } else {
+                    continue;
```
hildebrandmw, line 1196: Do we want to error here, or is this protecting against a potential race with the quantized start point getting set concurrently?

diskann-garnet/src/fsm.rs
```diff
@@ -301,12 +332,12 @@ impl FreeSpaceMap {
 
         let mut has_free_ids = false;
         let mut id = 0u32;
+        let mut block = vec![0u8; BLOCK_SIZE_BYTES];
         'scan: for block_id in 0..*max_block {
```
hildebrandmw: Is `max_block` inclusive or exclusive? The loop in `visit_used` implies inclusive, so shouldn't the upper bound be `=*max_block`?

hildebrandmw (approved):
I'm still worried for the potential for race conditions between backfill and the free-space map. Basically, `FreeSpaceMap::next_id` can be running concurrently with quantization - the check at the beginning is not sufficient for mutual exclusion. This can result in IDs being sent out that never get quantized which will explode when accessing these vectors in the future. I understand the pressure to start experimenting but would appreciate some cycles in a followup working through this interaction if it is deemed sufficiently unlikely to block development.

diskann-garnet/src/provider.rs
metajack: It did not, but I have since fixed it.

diskann-garnet/src/provider.rs
metajack: This has been fixed.

diskann-garnet/src/provider.rs
metajack: This is now fixed. Garnet handles the conversions for us now.

diskann-garnet/src/provider.rs
metajack: I need to call `T::as_f32()` in several places. I could replace that with a match statement on the type, but you didn't want dynamic type checks? Happy to make a change here if you think we should.

diskann-garnet/src/quantization.rs
metajack: It does operate under a lock. There is the training_started atomic, which once sets prevents training from retriggering. Or is there some situation where that is insufficient?

diskann-garnet/src/provider.rs
metajack: There is currently no way to signal failure here or report errors (other than writing stuff to stderr). I'll handle that in a followup. For now we make a best effort and if a quantizer can't be trained we run in full precision mode indefinitely.

diskann-garnet/src/provider.rs
metajack: I think I've fixed this. The reuse_lock is not a boolean, and instead of unreliably tracking completion with reference count, we increment at atomic that keeps track of how many invididual backfill jobs have completed, and only unlock once that number is the expected one.

So the FSM interaction is now just a two calls, one to lock_ruse() and one to unlock_reuse(), and the caller must track when it is the right time to do those things. FSM just has to do the once check before assigning ids whether that atomic is true or false.

diskann-garnet/src/fsm.rs
metajack: I can't tell if I've fixed this or not. reuse_lock starts true, meaning we can reuse ids. This happens berfore quantization starts. Once we have enough vectors to train a quantizer, we build quantization tables and set reuse_lock to false. Now every insert will quantize the vector, and we know that every vectors with ID greather than max_id when the reuse_lock was set to false will be quantized. When we backfill we only have to backfill up to that spot as nothing below it will be quantized. Once backfill is complete, we set reuse_lock to true, and now ids can be reused again, which is fine because all vectors are quantized in the entire index and any newly inserted ones will also be quantized.

diskann-garnet/src/provider.rs
metajack: This was fixed.

diskann-garnet/src/provider.rs
metajack: This was fixed.

rfcs/00000-quantizer-bootstrap.md
metajack: This RFC is not part of this PR. The commit is included because this PR is stacked on another.

vectorset/src/main.rs
metajack: Done.

vectorset/src/main.rs
metajack: We would add more quantizer flags for things like 2/4 bit. For the other spherical quantizer parameters I don't know. Adding more top level quantizers probably doesn't scale, but we'd need more complex changes to the redis protocol since a single flag for the quantizer is all that is currently supported.

diskann-garnet/src/alloc.rs
metajack: Fixed all the visibility modifiers. This is shared library, so it doesn't much matter, but it never hurts to do it right.

diskann-garnet/src/quantization.rs
metajack: This has been fixed with the i8 support.

diskann-garnet/src/quantization.rs
metajack: I will have a followup PR with tests.

diskann-garnet/src/provider.rs
metajack: Fixed.

diskann-garnet/src/quantization.rs
metajack: Fixed.

vectorset/src/loader.rs
metajack: There is already an issue for this.

diskann-garnet/src/lib.rs
metajack: Fixed.

diskann-garnet/src/dyn_index.rs
metajack: Indeed. I'm not sure why I used such a complicated path to get an accessor. This is fixed now by directly constructing the accessor and setting quantization=false.

diskann-garnet/src/provider.rs
metajack: Fixed.

diskann-garnet/src/provider.rs
metajack: Why would this be an error? If the `Entry`s are occupied, then the point is already there.

Or do you mean the (currently omitted) else clauses of the start point cache fetches? I can make those an error.

diskann-garnet/src/fsm.rs
metajack: Good catch.

JordanMaples (approved):
lgtm

diskann-garnet/src/provider.rs
hildebrandmw: We can defer. Just having removed a type parameter like this to remove compilation time in another implementation, it's something I'm a bit primed to see, especially since we're using type-erasure anyways for the internal element type.

diskann-garnet/src/fsm.rs
hildebrandmw: I think my concern is this: A thread can enter `next_id` and check `can_reuse`, see that it can, and enter the loop body. While it's in the loop body thinking it can reuse IDs, backfill can trigger. Now backfill and `next_id` under the assumption it can reuse IDs are running concurrently running. Can this cause an issue?

metajack:
I've added a lot more documentation so that it's clear what the atomics are doing. They aren't really interacting very much.

Garnet may invoke training several times due to concurrent inserts, so that call must ensure training only gets actually executed successfully once. `training_started` handles this.

Once training is completed, a quantizer will exist and `insert()` will automatcally quantize any new vectors. Once we start backfilling, we want all new vectors to have new IDs so that we constrain the backfill range. To achieve this, we have a new lock on ID reuse in the FSM. This gets locked when backfill starts and only gets unlocked once the final backfill task is complete. To track completion, we have an atomic counter `backfills_completed` that increments as each backfill task completes its assigned range of backfills. Once that counter is at the `task_counts, we quantize the start points and unlock the ID ranges and then set the index to operate in quantized mode `all_quantized`.

diskann-garnet/src/fsm.rs
metajack: Ok, I think that is a tricky ordering.  I believe that is easily fixed by changing reuse_lock to a `Mutex<bool>` instead of the atomic.

So the issue would be that we're in an `insert()` call after backfill has started, or at least around the same time. But if backfill starts, then we have quantizer, which means the vector will be quantized immediately. So the race would have to be that train_quantizer() is still running and hasn't filled in the quantization table, and insert() progresses past the check for the quantizer, and somehow in that span we finish training and start backfill and progress through the backfill past the ID in question.

I'll put up the fix shortly.

metajack: This is fixed. I ended up using a RwLock so inserts can still be completely concurrent, but still exclusive to the reuse lock flipping state.

diskann-garnet/src/quantization.rs
metajack: If you prefer I can switch training_started to RwLock as well. Maybe I'll do that so RAII handles the error cases.

metajack: I did this, turned out to make the code simpler.

diskann-garnet/src/fsm.rs
```diff
@@ -363,6 +471,62 @@ impl FreeSpaceMap {
...
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Signal that new IDs should be quantized.
+    pub(crate) fn enable_quantization(&self) {
+        let guard = self.next_id.read().unwrap();
```
hildebrandmw: This needs to be `self.next_id.write()` instead to ensure all data writes are complete.

diskann-garnet/src/provider.rs
```diff
@@ -312,22 +653,45 @@ impl<T: VectorRepr> SetElement<&[T]> for GarnetProvider<T> {
...
+        }
 
         let insert = || -> Result<(), Self::SetError> {
             self.callbacks
-                .write_iid(context.term(Term::Vector), internal_id, element)
+                .write_iid(&context.term(Term::Vector), internal_id.id(), element)
                 .then_some(())
                 .ok_or(GarnetError::Write)?;
+            if let Some(quantizer) = &self.quantizer
+                && !internal_id.should_quantize()
```
hildebrandmw: Should this be `&& internal_id.should_quantize()`, without the `!`?

diskann-garnet/src/fsm.rs
```diff
@@ -240,34 +316,51 @@ impl FreeSpaceMap {
...
+        {
+            let mut guard = self.next_id.write().unwrap();
+            id = *guard;
+            *guard += 1;
 
             // Make sure we have enough FSM blocks for this id.
             let (block_id, _, _) = self.indexes_for_id(id);
             let max_block = { *self.max_block.read().unwrap() };
             if block_id > max_block {
                 self.expand_to(ctx, id)?;
```
hildebrandmw, line 336: There is one more potential deadlock - calling `expand_to` acquires the write lock for `max_block` after the write lock for `next_id` is held, but earlier in the code, `refill_fast_free_list` goes `max_block` -> `next_id` in the other direction. We could drop the guard earlier, but there's still another race (see my other comment).

diskann-garnet/src/fsm.rs
```diff
@@ -240,34 +316,51 @@ impl FreeSpaceMap {
...
-            changed = self.mark_used(ctx, id)?;
         }
 
-        Ok(id)
+        // This must be outside the next_id write lock.
+        self.mark_used(ctx, id)?;
+
+        Ok(ReuseGuard::new(
+            id,
+            self.quantization_enabled.load(Ordering::Acquire),
```
hildebrandmw: There's a race here still. This reads `quantization_enabled` before acquiring the `next_id` read lock, but `enable_quantization` sets `quantization_enabled` **after** acquiring the `next_id` write lock. So we can have a situation where thread A is in between reading `quantization_enabled = false` and acquiring the read guard when thread B enables quantization, which could immediately start back-filling without thread A ever having written the full-precision data.

This coupled with the two locks on `next_id` and `max_block` being a potential footgun for deadlocks, I think there's one thing we can do to fix the problems:

```rust
struct FreeSpaceMap {
    ...
    barrier: RwLock<Barrier>,
    
}

struct Barrier {
    max_id_for_back_fill: usize,
    should_quantize: bool,
}

struct ReuseGuard<'a> {
    id: u32,
    guard: RwLockReadGuard<'a, Barrier>,
}

impl<'a> ReuseGuard<'a> {
    ...
    pub(crate) should_quantizer(&self) -> bool {
        self.guard.should_quantize
    }
}

// In free-space-map
fn next_id(...) -> Result<ReuseGuard<'a>, FsmError> {
    let guard = barrier.read().unwrap();
    // fsm stuff
    Ok(ReuseGuard { id, guard })
}

fn enable_quantization(&self) {
    // By acquiring the write barrier, we know:
    // 1. No one is trying to mint new IDs.
    // 2. All data for current IDs has been written.
    let mut guard = barrier.write().unwrap();
    *guard.quantization_enabled = true;
    *guard.max_id_for_backfill = // read next-id atomically.
}
```

This does a few things:
1. We can keep `next_id` atomic - it is only updated in a non-constructor way in the `next_id` function, which will be protected by `barrier`.
2. With `next_id` atomic, we don't have to keep acquiring a lock every time we read it.
3. Guaranteed mutual exclusion of ID minting and switching to "quantized" mode.

hildebrandmw (commented):
Sorry - a few more issues I found. I think it's worth keeping `next_id` as an atomic but introducing a proper `RwLock` around the quantization-enabled + max backfill ID. I think that shores up the rest of the potential concurrency problems.

In general, I'd say try to avoid spreading signals across multiple atomics because weird things can happen that are very hard to reason about.

diskann-garnet/src/provider.rs
metajack: No, because we only want to set_quantizer_ready() before quantization starts. That means we have enough vectors to train. So shoudl_quantize() will be false. If should_quantize() is true, we've already trained a quantizer and finished backfill.

diskann-garnet/src/fsm.rs
metajack: Good suggestions. I've added this.

hildebrandmw (approved):

hildebrandmw (approved):

hildebrandmw (approved):

diskann-garnet/src/quantization.rs
metajack: Fixed.

diskann-garnet/src/quantization.rs
metajack: Fixed.

diskann-garnet/src/provider.rs
metajack: No longer relevant.

diskann-garnet/src/quantization.rs
metajack: Can address this in a followup. Since there is no main trait for quantizers, it's not clear how we'd do this except by having a constant or documenting it.

hildebrandmw (approved):

PR #1067, opened 2026-05-14
Simplify the `DataProvider` contract for graph search

hildebrandmw:
Simplify graph search.

## What

At a high level, this PR collapses `Accessor`, `BuildQueryComputer`, `ExpandBeam`, `SearchExt`, and `AsNeighbor/NeighborAccessor` into a single `SearchAccessor` trait. Along the way, we have the following changes:

* Removes the following traits:
  - `Accessor`: For search, `SearchAccessor::expand_beam` is now used. The indexing layer now has no notion of element types. For insert, `workingset::Fill` is used exclusively to get elements.
  - `BuildQueryComputer`: Computing query distances has been fused into `SearchAccessor`.
  - `SearchExt`: Merged into `SearchAccessor`.
  - `ExpandBeam`: Merged into `SearchAccessor`.
  Additionally, search accessors no longer need to define `DelegateNeighbor`.
* **SearchStrategy**: This gains a lifetime, removes the lifetime from the `SearchAccessor` associated type, and passes the query to `search_accessor`. These changes allow the constructed accessor to **borrow** the query if so desired.
* **DefaultSearchStrategy/DefaultPostProcessor**: Also get the new lifetime to cooperate with `SearchStrategy`.
* **InsertStrategy**: Also gets a lifetime because why not (and also to be compatible with `SearchStrategy`).
* **SearchPostProcess**: Drops the `QueryComputer` argument and now only requires `HasId` on the accessor argument.
* The default implementation of `workingset::Fill` for `workingset::Map` has been removed. This means that users now need to implement `Fill` manually if they were relying on the provided implementation. A new `Map::fill` synchronous method has been added to make this process as easy as possible.

One consequence of fusing everything into `SearchAccessor` is that structs used for `SearchAccessors` can no longer be easily shared with those used as `PruneAccessors`. This causes a split between accessors used for search and those used for pruning. However, with this PR, the traits needed for insert `PruneAccessors` and `SearchAccessors` are disjoint. In practice, I did not find it very hard to make this split.

## Why

The whole `Accessor` -> `BuildQueryComputer` -> `ExpandBeam` originated from earlier patterns in the library of building up indexing algorithms from tiny pieces, enabling a user to implement a small set of methods and get the whole indexing algorithm for free.

This breaks down, though, when we consider performance: the more information we provide a backend database about the operation we want to perform, the better the database can optimize. This lead to `on_elements_unordered` (bulk element retrieval, allowing for example prefetching), `distances_unordered` (bulk retrieval with a concrete distance computations), and finally `ExpandBeam`. Of these, the latter has proved by far the most useful, being used by `diskann-disk` and `diskann-garnet` to specialize the algorithm to their preferred data access pattern.

I've been exploring how to make filtered search more of a first-class citizen to allow us to batch together predicate queries and potentially fuse predicate evaluation with data access. I was halfway through writing a parallel Accessor/Computer/ExpandBeam/Strategy hierarchy for predicate aware search (and struggling with expressing the more complicated type relationships required), when I realized that all we really needed were a few slightly different flavors of `ExpandBeam`. Instead of building a complex nest of interrelated traits to express these subtly different `ExpandBeam` methods, why not just require users to implement the necessary `ExpandBeam`? The required semantics are not that complicated and it gives the implementor **full** control of the data access.

## Reviewing Order

I've broken this PR into commits that focus on one change at a time. Each commit should compile and pass tests on its own.

1. Lifting the lifetimes into `{Search/Insert}Strategy`. 
2. Small cleanup off some test code that was relying on `Accessor`.
3. Remove the provided `Fill<Map<_>>` implementation.
4. Fuse distance computations into the `Accessor` trait. This is the biggest commit and results in the splitting of search and prune accessors. Note that the state of the code after this commit is a bit goofy, but it's setup for the next commit.
5. Add the `SearchAccessor` trait and collapse the inter-related search traits.
6. Documentation cleanup.
7. More documentation cleanup.

### Removed traits

 - `Accessor` — element-wise access is no longer required.
 - `BuildQueryComputer` — fused into `SearchAccessor`.
 - `SearchExt` — superseded.
 - `ExpandBeam` — superseded.
 - Search accessors no longer need to implement neighbor delegation.

### What to implement instead

A single SearchAccessor with three required methods:

```rust
async fn starting_points(&self) -> ANNResult<Vec<Self::Id>>;
async fn start_point_distances<F>(&mut self, f: F) -> ANNResult<()>
    where F: FnMut(Self::Id, f32) + Send;
async fn expand_beam<Itr, P, F>(&mut self, ids: Itr, pred: P, on_neighbors: F) -> ANNResult<()>
    where Itr: Iterator<Item = Self::Id> + Send,
          P: HybridPredicate<Self::Id> + Send + Sync,
          F: FnMut(Self::Id, f32) + Send;
```

`expand_beam` is now responsible for distance computation; the query is provided once at accessor construction.

### Splitting search and prune accessors

Search and prune now have orthogonal responsibilities. If your old accessor served both, split it:

 - Search accessor: implements `SearchAccessor`.
 - Prune accessor: implements `HasElementRef` + `BuildDistanceComputer` + `AsNeighborMut` + `workingset::Fill`.

`PruneStrategy::prune_accessor` produces the latter.

### SearchStrategy lifetime

`SearchStrategy<'a, Provider, T>` now has a lifetime so accessors can borrow the query:
```rust
fn search_accessor(
    &'a self,
    provider: &'a Provider,
    context: &'a Provider::Context,
    query: T,
) -> Result<Self::SearchAccessor, Self::SearchAccessorError>;
```
`InsertStrategy` also has a lifetime and is now passed by reference to DiskANNIndex::insert.

### `Fill<Map<...>>` blanket impl removed

If you previously relied on the blanket `Fill<Map<_>>` for accessors, you must implement `Fill` directly. The new `Map::fill` helper makes this a few lines:
```rust
fn fill<'a, Itr>(&'a mut self, set: &'a mut WorkingSet, itr: Itr)
    -> impl SendFuture<Result<Self::View<'a>, Self::Error>>
    where Itr: ExactSizeIterator<Item = Self::Id> + Clone + Send + Sync, Self: 'a
{
    let view = set.fill(itr, |id| self.fetch(id).allow_transient("..."));
    std::future::ready(view)
}
```

### Migration checklist

 - [ ]  Delete old `Accessor/BuildQueryComputer/SearchExt/ExpandBeam` impls.
 - [ ]  Implement `SearchAccessor` (search path) and a separate prune accessor (build path).
 - [ ]  Implement `Fill` directly on the prune accessor using Map::fill.
 - [ ]  Update `SearchStrategy`/`InsertStrategy` impls for the new 'a lifetime.
 - [ ]  Pass `&strategy` to `DiskANNIndex::insert`.


metajack:
get_element is basically unused in garnet, except for implementing VEMB (return an vector) and other debugging-style functions. 

This sounds like a very promising direction, and I am very supportive.

JordanMaples (approved):
lgtm

diskann/src/graph/glue.rs
```diff
@@ -192,133 +310,30 @@ where
...
 
     /// Construct and return the search accessor.
-    fn search_accessor<'a>(
+    fn search_accessor(
         &'a self,
         provider: &'a Provider,
         context: &'a Provider::Context,
-    ) -> Result<Self::SearchAccessor<'a>, Self::SearchAccessorError>;
+        query: T,
+    ) -> Result<Self::SearchAccessor, Self::SearchAccessorError>;
```
arkrishn94, line 335: I'm noticing that lifting the lifetime to `SearchStrategy` is introducing a lifetime on all upstream objects generic over it also - e.g.`Search` and `Knn`. 

If the primary goal is to allow the accessor to borrow the query, can this be done by keeping the lifetime on the associated type `SearchAccessor` and the method `search_accessor`?

harsha-simhadri (approved):

diskann/src/graph/glue.rs
```diff
@@ -192,133 +310,30 @@ where
...
-
 /// A search strategy for query objects of type `T`.
 ///
 /// This trait should be overloaded by data providers wishing to extend
 /// (search)[`crate::graph::DiskANNIndex::search`].
-pub trait SearchStrategy<Provider, T>: Send + Sync
+pub trait SearchStrategy<'a, Provider, T>: Send + Sync
 where
     Provider: DataProvider,
+    T: 'a,
```
arrayka: It seems like `T: 'a` bound is redundant everywhere in this file.
Can we allow 'a outlive T?

diskann/src/graph/workingset/mod.rs
```diff
@@ -41,9 +41,9 @@
 //! * Allocate in `Accessor`: Another approach is to allocate scratch space directly in `Self`
 //!   for the elements in `itr`. This works since [`Fill::View`] borrows from `Self`.
 //!
-//! * Passthrough: If random access into the [`Accessor`] is synchronous and fast, then an
-//!   implementation of [`View`] can simply reach through the [`Accessor`], use zero-sized
-//!   types for the `WorkingSet`, and avoid allocation entirely.
+//! * Passthrough: If random access is synchronous and fast, then an implementation of
+//!   [`View`] can simply reach through the `Accessor`, use zero-sized types for the
```
arkrishn94: nit: missed deleting this `Accessor` reference.

diskann-providers/src/model/graph/provider/async_/inmem/full_precision.rs
```diff
@@ -102,218 +102,249 @@ where
...
+                    self.provider.base_vectors.prefetch_hint(id.into_usize());
+                }
+
+                for (i, id) in id_buffer.iter().enumerate() {
+                    // Prefetch `lookahead` iterations ahead as long as it is safe.
+                    if lookahead > 0 && i + lookahead < len {
+                        self.provider
+                            .base_vectors
+                            .prefetch_hint(id_buffer[i + lookahead].into_usize());
+                    }
```
arkrishn94, line 332: Only cause I want to learn: can you help me understand what lines 326 - 332 are doing?

arkrishn94 (commented):
Thanks Mark, this is looking great. Just left one comment about lifetimes in `SearchStrategy` (and corresponding `InsertStrategy`). 

Not specifically for this PR, but this change reminds me it might be useful to call out the entry points for the major traits and structs for graph somewhere in the module documentation (or more central place)? Today the (super well-written) documentation for the major components is spread between glue.rs, provider.rs, index.rs and workingset modules iiuc. Might be useful for folks unfamiliar with this to know where to start.

diskann/src/graph/glue.rs
hildebrandmw: Great find! You are right. Now I'm questioning everything. 😟

diskann/src/graph/glue.rs
hildebrandmw: This can't really be done by having a lifetime on the `SearchAccessor` GAT while still allowing the accessor to borrow from the query. Take for example the following:
```rust
trait SearchStrategy<Provider, T> {
    type SearchAccessor<'a>: SearchAccessor;
    fn search_accessor<'a>(&'a self, provider: &'a Provider, query: T) -> Self::SearchAccessor<'a>;
}

impl<T> SearchStrategy<MyProvider, &[T]> for MyStrategy {
    type SearchAccessor<'a> = MyAccessor<'a>;
    fn search_accessor<'a>(&'a self, provider: &'a Provider, query: &[T]) -> Self::SearchAccessor<'a> {
        // How do I borrow from `query`?
        //
        // The trait is generic over any lifetime for the reference, which could
        // potentially be shorter than `'a` and there's no way (that I know of) to 
        // tie the lifetime in the trait to `'a` on the method.
    }
}
```
Using `where T: 'a` on the `search_accessor` method ends up being pretty awkward, and the lifetime in the `SearchAccessor` GAT usually ends up with sneaky implied `'static` bounds when involved in HRTBs, which is almost fine except that the disk index uses a non-`'static` search strategy.

Also, forcing taking `T` as a reference like we used to doesn't work because it's not always possible to materialize `T` **as** a reference. One of our integrations uses a non-reference type everywhere to reduce monomorphizations, and anything multi-vector can't be a reference either.

diskann/src/graph/workingset/mod.rs
hildebrandmw: Thanks!

diskann-providers/src/model/graph/provider/async_/inmem/full_precision.rs
hildebrandmw: Yeah, when you are processing a list of neighbors, you want to use a [software prefetch](https://en.wikipedia.org/wiki/Cache_prefetching) for the data we'll be accessing for distance computations (this is really important for performance). Fetching just the next iteration ahead often doesn't provide enough time for the data to work its way into the L1 cache, so we instead prefetch the data several iterations ahead with the hope that when we actually need the data, it is ready and waiting.

diskann/src/graph/index.rs
```diff
@@ -2097,19 +2074,19 @@ where
     }
 
     /// Execute a search with an explicit post-processor parameter.
-    pub fn search_with<S, T, O, OB, P, PP>(
-        &self,
+    pub fn search_with<'a, S, T, O, OB, P, PP>(
```
arrayka, line 2077: NIT: Maybe we should rename OB to OO so the generic parameters read “A STOP” :) (just kidding)

diskann/src/graph/glue.rs
arkrishn94: Got it, thanks for explaining.

arkrishn94 (approved):

diskann/src/graph/index.rs
hildebrandmw: At some point - we can probably simplify this as well. But a job for another day.

PR #1069, opened 2026-05-14
Fix recall computation for fewer than k groundtruth results

### LINKED ISSUE #1073: Fix recall computation when fewer than k results are present

magdalendobson:
Groundtruth for filtered datasets uses the `.rangeres` format, since a filtered query may have few or no results. However, at the time of recall computation for top-k search, fewer than `k` results causes an error. This PR patches that error and computes recall for each point using the following paradigm: 

1. If the point has `0 < k' <= k` groundtruth results, recall is the number of correctly reported points divided by `k'`. 
2. If the point has no groundtruth results, it is completely excluded from the recall calculation.

The average recall is then the sum of all recalls divided by the number of *non-zero* results. The points with zero results do not affect recall. The reasoning for this is that in a dataset with majority-null results for filtered queries, we want recall to capture our performance on the queries that actually have results, rather than being artificially high when we always correctly report that these points have no results.

Since this error will no longer trigger even for a regular top-k search in recall computation, I have pushed the error for not enough groundtruth results to when the groundtruth in the regular bin format is actually read. This is an improvement since a failure will now trigger before an index build and search, making a better experience for the user. I leave some of the instances of `NotEnoughGroundTruth` inside recall computation as an extra layer of failsafe, but they could also be removed if reviewers feel we should completely remove them.

Along the way I get rid of the `minimum` and `maximum` values in the recall stats, since after some discussion with the team it appears that no one uses them.


magdalendobson:

diskann-benchmark/src/utils/datafiles.rs
```diff
@@ -114,6 +114,17 @@ pub(crate) fn load_groundtruth(path: BinFile<'_>) -> anyhow::Result<Matrix<u32>>
...
+
+    if let Some(expected_k) = k {
+        if groundtruth.ncols() != expected_k {
+            return Err(anyhow::anyhow!(
+                "Each row of groundtruth must have length {} (got {})",
+                expected_k,
+                groundtruth.ncols()
+            ));
+        }
+    }
```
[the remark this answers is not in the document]
magdalendobson: @copilot apply changes based on [this feedback](https://github.com/microsoft/DiskANN/pull/1069#discussion_r3243985480)

diskann-benchmark-core/src/recall.rs
```diff
@@ -22,10 +22,6 @@ pub struct RecallMetrics {
     pub num_queries: usize,
     /// The average recall across all queries.
     pub average: f64,
```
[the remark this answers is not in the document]
magdalendobson: @copilot apply changes based on [this feedback](https://github.com/microsoft/DiskANN/pull/1069#discussion_r3243985534)

diskann-benchmark-core/src/recall.rs
```diff
@@ -186,76 +182,75 @@ where
...
-    let maximum = recall_values.iter().max().unwrap_or(&0);
+        recall_values.push(recall);
+    }
 
-    // We explicitly check that each groundtruth row has at least `recall_k` elements.
-    let div = recall_k * nrows;
-    let average = (total as f64) / (div as f64);
+    // Compute the average recall
+    let total: f64 = recall_values.iter().sum();
+    let average = (total) / (num_nonzero as f64);
```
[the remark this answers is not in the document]
magdalendobson: @copilot apply changes based on [this feedback](https://github.com/microsoft/DiskANN/pull/1069#discussion_r3243985611)

diskann-benchmark-core/src/recall.rs
```diff
@@ -575,18 +571,90 @@ mod tests {
...
+                }
+            }
+
+            let recall = knn(&groundtruth, None, &results, 10, 10, false).unwrap();
+            assert_eq!(recall.num_queries, 10);
+            assert!((recall.average - 1.0).abs() < 1e-10);
+        }
+
+        // All queries have zero groundtruth: should return average = 0.0 (not NaN/inf).
+        {
```
harsha-simhadri: can all these code blocks be split into independent tests?

diskann-benchmark/src/utils/datafiles.rs
```diff
@@ -114,6 +114,17 @@ pub(crate) fn load_groundtruth(path: BinFile<'_>) -> anyhow::Result<Matrix<u32>>
     let mut groundtruth = Matrix::<u32>::new(0, num_points, dim);
     let groundtruth_slice: &mut [u8] = bytemuck::cast_slice_mut(groundtruth.as_mut_slice());
     file.read_exact(groundtruth_slice)?;
+
+    if let Some(expected_k) = k {
+        if groundtruth.ncols() < expected_k {
```
harsha-simhadri, line 119: what if `gt.ncols() > expected>k`? is that ok?

harsha-simhadri (approved):

diskann-benchmark-core/src/recall.rs
```diff
@@ -186,76 +183,79 @@ where
...
-        {
-            let distances_row = distances.row(i);
-            if distances_row.len() > recall_k - 1 && gt_row.len() > recall_k - 1 {
-                let last_distance = distances_row[recall_k - 1];
-                for (d, g) in distances_row.iter().zip(gt_row.iter()).skip(recall_k) {
-                    if *d == last_distance {
-                        this_groundtruth.insert(g.clone());
-                    } else {
-                        break;
+        // groundtruth does not have to be fixed-size, so we compute recall_k for this row based on its gt length
```
hildebrandmw: Small nit, but it really makes a difference: Please wrap variable identifiers in backticks. E.g.,
```rust
// `groundtruth` does not have to be fixed-size,
// so we compute `recall_k` for this row based on its
// actual length.
```
For readers, this instantly disambiguates when something is meant as a variable versus prose.

diskann-benchmark-core/src/recall.rs
```diff
@@ -186,76 +183,79 @@ where
...
+            num_nonzero += 1;
+
+            // Populate the groundtruth using the top-k
+            this_groundtruth.clear();
+            this_groundtruth.extend(gt_row.iter().take(this_recall_k).cloned());
+
+            // If we have distances, then continue to append distances as long as the distance
+            // value is constant
+            if let Some(distances) = groundtruth_distances
+                && this_recall_k > 0
```
hildebrandmw: The `this_recall_k` check here is redundant.

diskann-benchmark-core/src/recall.rs
```diff
@@ -186,76 +183,79 @@ where
...
+            {
+                let distances_row = distances.row(i);
+                if distances_row.len() > this_recall_k - 1 && gt_row.len() > this_recall_k - 1 {
```
hildebrandmw: Should we instead error if the length of `distances_row` is not the same as `gt_row`? Seems like we should expect these two to have the same length.

diskann-benchmark-core/src/recall.rs
```diff
@@ -186,76 +183,79 @@ where
...
-                let last_distance = distances_row[recall_k - 1];
-                for (d, g) in distances_row.iter().zip(gt_row.iter()).skip(recall_k) {
-                    if *d == last_distance {
-                        this_groundtruth.insert(g.clone());
-                    } else {
-                        break;
+        // groundtruth does not have to be fixed-size, so we compute recall_k for this row based on its gt length
+        let this_recall_k = gt_row.len().min(recall_k);
+
+        let recall = if this_recall_k > 0 {
```
hildebrandmw: Seems like if `this_recall_k == 0`, we could simply `continue`. Then at the end, the computation of the `average` doesn't need to trak `num_nonzero` independently because we'll only push valid items to the `recall_values`. Then we could save a level of indentation :laugh:

diskann-benchmark/src/utils/datafiles.rs
magdalendobson: Yes, this is fine, during the recall computation only the first 'k' elements would be used.

diskann-benchmark-core/src/recall.rs
magdalendobson: Done

diskann-benchmark-core/src/recall.rs
magdalendobson: True, I created an error type for this and check explicitly and error.

diskann-benchmark-core/src/recall.rs
magdalendobson: Good point

diskann-benchmark-core/src/recall.rs
magdalendobson: Good point, removed

diskann-benchmark-core/src/recall.rs
magdalendobson: Yeah, fine by me, I split them up

magdalendobson:
@copilot resolve the merge conflicts in this pull request

hildebrandmw (commented):
I had a thought - should this be behind a configuration option? (maybe ditching `allow_insufficient_results` since I think that should just always be the case).

My concern is that if we do a normal k-nearest-neighbors search and request recall at 100, but the groundtruth only has 10 nearest neighbors, we will silently get recall computed for 10 nearest neighbors instead of the expected 100. Basically, we need a way to opt-out of this behavior.

Here is my proposal:

1. We ditch the `allow_insufficient_results` flag. Algorithms can't always ensure this and it feels like something that callers should enforce anyways.
2. Add the following enum as an argument:
   ```rust
   pub enum GroundTruthMode {
       Fixed,
       Flexible,
   }
   ```
   In `Fixed` mode, we ensure we have at least `recall_k` entries, and `Flexible` mode allows what's happening in this PR.

   Using an `enum` makes updating the callsites easier and more explicit: we aren't changing one `bool` for another.
3. Update `diskann_benchmark::utils::compute_multiple_recalls` to do one upfront check that the results have at least `recall_n` entries.
4. Update the KNN searches to use `GroundTruthMode::Fixed` and the filtered KNN searches to use `GroundTruthMode::Flexible`

magdalendobson:
> I had a thought - should this be behind a configuration option? (maybe ditching `allow_insufficient_results` since I think that should just always be the case).
> 
> My concern is that if we do a normal k-nearest-neighbors search and request recall at 100, but the groundtruth only has 10 nearest neighbors, we will silently get recall computed for 10 nearest neighbors instead of the expected 100. Basically, we need a way to opt-out of this behavior.
> 
> Here is my proposal:
> 
> 1. We ditch the `allow_insufficient_results` flag. Algorithms can't always ensure this and it feels like something that callers should enforce anyways.
> 2. Add the following enum as an argument:
>    ```rust
>    pub enum GroundTruthMode {
>        Fixed,
>        Flexible,
>    }
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    In `Fixed` mode, we ensure we have at least `recall_k` entries, and `Flexible` mode allows what's happening in this PR.
>    Using an `enum` makes updating the callsites easier and more explicit: we aren't changing one `bool` for another.
> 3. Update `diskann_benchmark::utils::compute_multiple_recalls` to do one upfront check that the results have at least `recall_n` entries.
> 4. Update the KNN searches to use `GroundTruthMode::Fixed` and the filtered KNN searches to use `GroundTruthMode::Flexible`

I believe this is resolved now

diskann-benchmark-core/src/recall.rs
```diff
@@ -121,6 +116,12 @@ pub trait RecallCompatible: Eq + Hash + Clone + std::fmt::Debug {}
 
 impl<T> RecallCompatible for T where T: Eq + Hash + Clone + std::fmt::Debug {}
 
+#[derive(Copy, Clone)]
+pub enum GroundTruthMode {
+    Fixed,
+    Flexible,
+}
```
hildebrandmw, line 126: Can you derive `Debug` as well and add a minimal doc-string the this struct/variants?

diskann-benchmark-core/src/recall.rs
```diff
@@ -130,15 +131,15 @@ impl<T> RecallCompatible for T where T: Eq + Hash + Clone + std::fmt::Debug {}
 /// the last values of each entry of `results`. Values will be counted towards the recall if
 /// they have the same distance as the last ordered candidate.
 ///
-/// If `allow_insufficient_results`, an error will not be given if an entry in `results`
+/// Note that an error will NOT be given if an entry in `results`
 /// has fewer than `recall_n` candidates.
```
hildebrandmw, line 138: Please update the docstring to describe the behavior of `ground_truth_mode`.

hildebrandmw (commented):
Thanks Magdalen - two small comments then looks good to me!

diskann-benchmark-core/src/recall.rs
magdalendobson: Yep, done

diskann-benchmark-core/src/recall.rs
magdalendobson: Done

hildebrandmw (approved):

hildebrandmw (approved):

PR #1077, opened 2026-05-15
[benchmark] Simply `Inputs` more.

hildebrandmw:
Move the `Input` trait in `diskann-benchmark-runner` over to a strongly typed representation and remove `Any` from the public API. This finishes the large work in streamlining the benchmark API. The new interface is
```rust
pub trait Input: Sized + std::fmt::Debug + 'static {
    type Raw: serde::de::DeserializeOwned + serde::Serialize;
    fn tag() -> &'static str;
    fn from_raw(raw: Self::Raw, checker: &mut Checker) -> anyhow::Result<Self>;
    fn serialize(&self) -> anyhow::Result<serde_json::Value>;
    fn example() -> Self::Raw;
}
```
with the big changes being the introduction of the `Raw` associated type, changing `try_deserialize` for `from_raw`, adding `serialize` (for custom serialization) and `example` returning `Self::Raw` instead of raw `serde_json::Value`. The idea behind `Raw` is that it is used to replace the `CheckDeserialization` trait and allows decoupling of an unvalidated raw deserialized struct from the final validated object.

This PR also removes the `CheckDeserialization` trait.

# Suggested Reviewing Order
* `diskann-benchmark-runner`:
  - `src/input.rs`: This is where the bulk of the actual changes are. The now removed custom `Any` type is moved as an internal implementation detail in this module and is no longer user facing.
  - `src/files.rs`: The file path validation for `InputFile` is moved from `CheckDeserialization` to an inherent `resolve` method.
  - `src/test/dim.rs`: A mechanical change where `Raw = Self` is used for the `Input` definitions.
  - `src/test/typed.rs`: `TypeInput` is tweaked to use a separate type as the raw input to ensure there is some coverage there.
  - `src/any.rs`: Removed.
  - The rest of the changes are mostly mechanical updates of import paths.
* `diskann-benchmark`:
  - `src/inputs.mod.rs`: Update the `as_input!` macro to the new form. Now that we aren't relying on `CheckDeserialization`, I switched over benchmarks to using ad-hoc `validate` inherent methods. This is somewhat of a stopgap to preserve the existing checks that mutate in-place.
  - `README.md`: Update some out-of-date examples and descriptions.
  - The rest of the changes are moving calls to `check_deserialization` to `validate`.
* `diskann-benchmark-simd`: Simple port to the new `Input` trait.


harsha-simhadri (approved):

diskann-benchmark-runner/src/input.rs
```diff
@@ -79,64 +77,123 @@ impl std::fmt::Debug for Registered<'_> {
...
+    pub(crate) struct Any {
+        any: Box<dyn RuntimeAny>,
+    }
 
-impl<T> Wrapper<T> {
-    pub(crate) const INSTANCE: Self = Self::new();
+    impl Any {
+        pub(crate) fn new<T>(input: T) -> Self
+        where
+            T: Input + std::fmt::Debug + 'static,
```
suri-kumkaran: nit: Input already requires `Debug + 'static`, so `Debug + 'static` is redundant

diskann-benchmark-runner/src/input.rs
```diff
@@ -79,64 +77,123 @@ impl std::fmt::Debug for Registered<'_> {
...
+        // reflection
+        fn as_any(&self) -> &dyn std::any::Any;
+        fn type_name(&self) -> &'static str;
     }
-    fn type_name(&self) -> &'static str {
-        std::any::type_name::<T>()
+
+    impl<T> DynInput for Wrapper<T>
+    where
+        T: Input + std::fmt::Debug + 'static,
```
suri-kumkaran: nit: Input already requires Debug + 'static, so Debug + 'static is redundant

suri-kumkaran (approved):

diskann-benchmark-runner/src/input.rs
hildebrandmw: Yup, fixed.

diskann-benchmark-runner/src/input.rs
hildebrandmw: Yup, fixed.

suri-kumkaran (approved):

PR #1079, opened 2026-05-16
Support file-based serialize/deserialize for DiskANN indices using Save/Load traits

### LINKED ISSUE #737: Document and extend index serialization format

suhasjs:

#### Reference Issues/PRs
Initial attempt at #737

Introduces two new traits (`diskann_record::save::Save` and `diskann_record::load::Load`)and a new create `diskann-record` to support file-based serialization + deserialization. This PR also implements the traits for `DiskANNIndex` (and all nested structs) for a successful prototype.
Sample output can be found [here](https://github.com/microsoft/DiskANN/tree/mhildebr/saveload/diskann-record/sample_output).
#### Any other comments?
This PR is meant to be in a draft state until we're happy with the abstractions and the results.


diskann-record/src/save/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+    fn save(&self, context: Context<'_>) -> Result<Record<'_>>;
+
+    /// Return the variant tag for enum types. Default: `None` (struct).
+    ///
+    /// Enum implementations must return `Some(variant_name)` for every variant.
+    /// The framework writes this into the manifest as `$variant` and enforces on
+    /// load that the tag's presence matches the corresponding [`Load::IS_ENUM`].
+    fn variant(&self) -> Option<Cow<'_, str>> {
+        None
+    }
```
hildebrandmw: After thinking about this for longer than I care to admit, I think we can get rid of enum handling entirely at the trait level. If we consider three cases:
```rust
enum UnitEnum {
    A,
    B,
    C,
}

enum InlineStruct {
    A(f32),
    B {
       x: String,
       y: f32,
    }
}

enum Struct {
    A(A),
    B(B),
}

struct A {
    field: f32,
}

struct B {
    x: String,
    y: f32,
}
```

The serialized representations in JSON could be
```json
{
    "$version": "0.0.0",
    "A": null,
}

{
    $version": "0.0.0",
    "B": {
        "x": "hello world",
        "y": 1,
    }
}

{
    "$version": "0.0.0",
    "A": {
        "$version": "0.0.1",
        "0": 1,
    }
}
```
For versions 1 and 2, there is just a top level version, which makes sense since the structure of these enums evolves in lockstep with the outer enum. For Version 3, this allows nested objects in enums to have distinct versions naturally.

diskann-record/src/load/error.rs
```diff
@@ -0,0 +1,117 @@
...
+            },
+        }
+    }
+}
+
+#[derive(Debug)]
+enum ErrorInner {
+    Light(Kind),
+    Heavy(anyhow::Error),
+}
```
hildebrandmw, line 99: While I still really like the Light/Heavy idea, if we're honest it's probably overkill and will likely not get implemented ever.

Instead, the main goal I think is to differentiate crucial errors vs "recoverable" errors, where the latter is used for a probing API. In that context, we'd have something like:
```rust
struct Error {
    inner: anyhow::Error,
    recoverable: bool,
}
```
Most of the constructors would imply critical errors, with probing APIs using an additional (`_recoverable`) API. This keeps normal code simple and the probing API possible and reasonably efficient.

diskann-record/src/load/context.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    pub(super) fn read(&self, key: &str) -> Result<Reader<'_>> {
+        let key_as_path: &Path = key.as_ref();
+        if !self.files.contains(key_as_path) {
+            return Err(Error::from(error::Kind::MissingFile).context(format!(
+                "handle references file {:?} which is not registered in the manifest",
+                key,
+            )));
+        }
```
hildebrandmw, line 77: Probably worth adding a 
```rust
if key.contains("..") || key_as_path.is_absolute() {
   return Err(/* ... */);
}
```
to keep files inside the directory.

diskann-providers/src/storage/record_shim.rs
```diff
@@ -0,0 +1,394 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+//! Adapters that expose a single in-hand [`Read`]/[`Write`] target as a
```
hildebrandmw, line 6: Should we go ahead and start updating the exiting code to use `io::Read/io::Write` directly so we don't need this thing?

diskann-record/src/save/value.rs
```diff
@@ -0,0 +1,291 @@
...
+        }
+    }
+
+    pub fn contains_key(&self, key: &str) -> bool {
+        self.record.contains_key(key)
+    }
+
+    pub fn get(&self, key: &str) -> Option<&Value<'a>> {
+        self.record.get(key)
+    }
```
hildebrandmw, line 213: In the vein of moving enums to handling themselves, if we add:
```rust
impl<'a> Record<'a> {
    fn keys(&self) -> Keys<'_> {
        Keys::new(self.record.key())
    }
}

struct Keys<'a> {
    keys: std::collections::hash_map::Keys<'a, Cow<'a, str>, Value<'a>>,
}

impl<'a> Iterator for Keys<'a> {
    // iterate
}

impl ExactSizeIterator for Keys<'_> {}
```
Then enums could have a way of inspecting the keys for loading.

hildebrandmw (commented):
Thanks Suhas! My comments this round are mainly about enum handling. I think we can do this more cleanly by having enums "handle themselves" rather than lifting it into a somewhat fragile interaction in the `Save`/`Load` traits.

Other than that, the code inside `diskann-record` could use considerable hardening and testing.

Do we want to plan out how plugins like VFS would work in this PR, or save that for the future? I ask because VFS is used pretty heavily in tests, so having a replacement that's compliant with this new system will be required. And it will be a good stress test that we can expand the backend serialization to other formats in the future.

diskann-record/src/load/error.rs
suhasjs: It did seem overkill. I've removed it in the latest commit.

diskann-record/src/save/mod.rs
suhasjs: `$variant` is a little more explicit in how it handles enums. But since most of our enums fall into one of the three variants you described, I've gone ahead and removed `$variant` handling.

diskann-record/src/load/context.rs
suhasjs: Good catch. Added it.

diskann-providers/src/storage/record_shim.rs
suhasjs: By 'existing code', do you mean all code that uses `Storage{Read/Write}Provider` traits? Not sure I understood your ask here.

hildebrandmw: Most of the code that currently takes `Storage{Read/Write}Provider` does so by taking the provider and a file name and immediately opening the necessary file. These could all be pretty easily migrated to taking `std::io::Read` and `std::io::Write` directly, moving the file open up one level and avoiding the need for this hacky one-off provider.

diskann-record/HANDOFF.md
```diff
@@ -0,0 +1,220 @@
...
+  destructured), drop the first argument: `save_fields!(context, [weights])` reads from
+  the local bindings.
+- `load_fields!(object, [x, y, inner, vector: Handle])` — loads named fields with optional
+  type annotations when inference fails. Works identically for structs and enum variants.
+
+### Reserved Keys
+
+Keys starting with `$` are reserved for infrastructure (`$version`, `$variant`,
+`$handle`). Since Rust identifiers can't start with `$`, the `save_fields!` macro is
+inherently safe — no runtime check needed in the macro path.
```
arrayka, line 90: 1. This struct can be created without any issues:
```
pub struct Test {
    version: String,
    handle: String,
    variant: String,
}
```

2. The choice of programming language shouldn’t determine the field names in a public contract.

diskann-record/sample_output/manifest.json
```diff
@@ -0,0 +1,120 @@
...
+        "$version": {
+          "major": 0,
+          "minor": 0,
+          "patch": 0
+        },
+        "$variant": "L2"
+      },
+      "aux_vectors": {
+        "$version": {
+          "major": 0,
```
arrayka: version object repeats everywhere. It will bloat this file. Is it a concern? Should we consider tracking version at high level objects only?

diskann-record/HANDOFF.md
```diff
@@ -0,0 +1,220 @@
...
+    Number(Number),       // Copy enum: U64, I64, F64
+    String(Cow<'a, str>),
+    Bytes(Cow<'a, [u8]>),
+    Array(Vec<Value<'a>>),
+    Object(Versioned<'a>),
+    Handle(Handle),
+}
+```
+
+The lifetime `'a` allows the save path to borrow from the data being saved (zero-copy
```
arrayka, line 42: Is there a measurable benefit of introducing a lifetime here? I would say that config load/save is outside of hot path or at least its contribution is negligible to overall RAM/CPU usage, since config is tiny.

diskann-record/src/load/context.rs
```diff
@@ -0,0 +1,398 @@
...
+
+#[derive(Debug, serde::Deserialize)]
+struct FileRepr {
+    files: HashSet<PathBuf>,
+    value: save::Value<'static>,
+}
+
+impl ContextInner {
+    pub(super) fn new(metadata: &Path, dir: &Path) -> Result<Self> {
+        let file = std::fs::File::open(metadata).map_err(|e| {
```
arrayka, line 48: we should not code directly against the file system.  Instead, use Storage{Read/Write}Provider abstraction. It allows to write directly to remote store and provides better testability of code.

diskann-record/HANDOFF.md
```diff
@@ -0,0 +1,220 @@
...
+  exposes `Save::variant() -> Option<Cow<'_, str>>` (default `None` = struct), and
+  `save_fields!` has a two-argument form (`save_fields!(context, [...])`) for use inside
+  enum match arms after destructuring. Load side adds `Load::IS_ENUM` (default `false`)
+  plus `Object::variant()`. The blanket `Loadable` impl strictly enforces tag presence:
+  loading a tagged record as a struct yields `UnexpectedVariant`, loading an untagged
+  record as an enum yields `MissingVariant`.
+
+## Remaining Work
+
+### Value::Bytes Wiring
```
arrayka, line 117: are we planning to store bytes in a configuration file?

hildebrandmw (commented):
A few more drive-by comments.

diskann-record/sample_output/manifest.json
```diff
@@ -0,0 +1,120 @@
...
+    "vectors.bin"
+  ],
+  "value": {
+    "config": {
+      "pruned_degree": 16,
+      "prune_kind": {
+        "$version": {
+          "major": 0,
+          "minor": 0,
+          "patch": 0
```
JordanMaples: I think major/minor/patch is overkill. Each object should just have a single version number that is not tied to the overall version of diskann.

diskann/src/graph/config/mod.rs
```diff
@@ -695,6 +695,179 @@ impl Builder {
...
+                max_minibatch_par,
+                intra_batch_candidates,
+                saturate_after_prune,
+                experimental_insert_retry,
+            ]
+        ))
+    }
+}
+
+impl diskann_record::load::Load<'_> for Config {
```
arrayka, line 829: Config::load bypasses its own Builder.

Config::load directly constructs Self { pruned_degree, max_degree, ... } from deserialized fields — bypassing the Builder which enforces invariants like 
max_degree >= pruned_degree and alpha >= 1.0.

A corrupted or hand-edited manifest with `max_degree: 4, pruned_degree: 64`
would load successfully and produce a `Config` that `Builder` would have
rejected, silently causing excessive pruning or algorithmic misbehavior.

The loader should feed values through Builder or at minimum call a validation method.

## Open Question: Should deserialization bypass safety checks?

- **Option A — Yes, separate `::load()`:** Accept a dedicated load path that
  skips the builder. Risk: two constructors accepting the same parameters in
  different forms (struct fields vs. serialized DTO), with only one enforcing
  invariants.

- **Option B — No, chain through the constructor:** The domain type exposes an
  API to create itself *and* an API to export the values needed to recreate
  itself. A separate adapter is responsible for persisting those values
  (including versioning and format mapping) and calling the validated
  constructor on load. This keeps the contract between DTO and public API
  explicit.

diskann-providers/src/model/graph/provider/async_/simple_neighbor_provider.rs
```diff
@@ -445,4 +499,97 @@ mod tests {
...
     }
+
+    /////////////////////////////////
+    // diskann-record round-trips //
+    /////////////////////////////////
+
+    type TestNeighborProvider = SimpleNeighborProviderAsync<u32>;
+
+    fn round_trip_helper(provider: &TestNeighborProvider) -> TestNeighborProvider {
+        let dir = tempfile::tempdir().expect("tempdir");
```
arrayka, line 510: this test should work against VFS

diskann-providers/src/model/graph/provider/async_/fast_memory_vector_provider.rs
arrayka, line 256: what is the future of SaveWith/LoadWith?

arrayka, line 209: We need to understand why we add responsibility of saving data to a file to the fast memory provider, when it already implements GetData trait needed by storage::bin::save_to_bin().

diskann-record/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+description.workspace = true
+authors.workspace = true
+documentation.workspace = true
+license.workspace = true
+edition = "2024"
+
+[dependencies]
+anyhow.workspace = true
+serde = { workspace = true, features = ["derive"] }
+serde_json.workspace = true
```
arrayka, line 13: serde_json propagates to diskann and diskann-vector crates.

Serialization should belong in an adapter layer, not in domain crates:
- Domain types stay format-agnostic; adapters own the format.
- Inner crates must not depend on serialization frameworks.

`diskann-vector` (Tier 1) and `diskann` (Tier 3) now depend on
`diskann-record`. Every domain type below has `impl Save`/`impl Load` directly
on it, embedding format-specific logic (constructing Records, inserting
`Value::Null` tags, calling `single_key()`, etc.).

| Crate / file                                  | Type(s)                                          | Lines       |
|-----------------------------------------------|--------------------------------------------------|-------------|
| `diskann-vector/src/distance/metric.rs`       | `Metric`                                         | 112–151     |
| `diskann/src/graph/config/mod.rs`             | `Config`, `PruneKind`, `IntraBatchCandidates`    | 712–869     |
| `diskann/src/graph/config/experimental.rs`    | `InsertRetry`                                    | 82–125      |
| `diskann/src/graph/index.rs`                  | `DiskANNIndex<DP>`                               | 2948–2987   |

If `diskann-record` were swapped for another format, all of these files — plus
the 6 provider-side files — would need modification (10+ files across 3 crates).

Example: Metric (a Tier 1 type in diskann-vector) now has impl Save and impl Load directly on the domain type, pulling a serialization framework into a foundation crate. Same for Config, PruneKind, IntraBatchCandidates, InsertRetry, and DiskANNIndex in the core diskann crate.

**Recommendation:** Move all `impl Save/Load` out of `diskann-vector` and
`diskann` into a dedicated adapter crate (e.g. `diskann-record-adapters`) or
consolidate them in `diskann-providers`. The adapter depends on both the domain
types and `diskann-record`; the domain crates keep zero serialization
dependencies.

diskann/src/graph/config/mod.rs
arrayka: My guiding principles during this review:

1. **Keep code simple.** Extend functionality by adding new code, not modifying
   existing code.

2. **Separate concerns clearly.**
   - Domain types must support save/restore of their state so it can be
     recreated later.
   - *How* that state is persisted (format, versioning, I/O) is not the
     domain type's responsibility.

Let's take Disk-index checkpointing as an example:

```rust
// Domain type owns state, exposes it through public API
let builder = DiskIndexBuilder::new(config);
let offset = builder.save();                    // returns resumption state
let builder = DiskIndexBuilder::new_from_offset(config, offset);  // validates on entry
```

**Responsibilities are clean:**
- `DiskIndexBuilder` can stop and return the metadata needed to resume (`offset`). Restore process is still safe, because offset is validated.
- The **caller** decides how to persist `config` and `offset` (format, versioning, storage).

**The proposed design would require adding next methods to the builder:**

```rust
// Domain type now also owns serialization mapping
builder.save(SaveContext) -> Value       // new: format-specific logic on domain type
DiskIndexBuilder::load(Object) -> Self   // new: bypasses new_from_offset() validation
```

This asks the domain type to handle serialization mapping and
versioning/backward-compatibility — responsibilities that belong in an external
adapter.

The adapter should deserialize, then call
`DiskIndexBuilder::new_from_offset(config, offset)` — reusing the validated
constructor.

diskann-providers/src/model/graph/provider/async_/fast_memory_vector_provider.rs
suhasjs: Plan to do away with those traits once we check in #1188's successor.

diskann-record/sample_output/manifest.json
suhasjs: Reduced to just `major.minor`. Single version number, while cleaner, makes it hard to discern (to a reader) what changes correspond to 'minor'. Having `major.minor` makes it clear: version 1.x can read/load all 1.y (y <= x).

diskann-record/HANDOFF.md
suhasjs: Removed `Bytes` from the current version. Did not seem very useful + we were storing it as an array of UInt8 (confusing, must cast to `Bytes` explicitly).

diskann-record/src/load/context.rs
suhasjs: New version adds a VFS abstraction (requires `io::Read + io::Seek`)

diskann-record/sample_output/manifest.json
suhasjs: Need per-object tracking for independent evolution.

PR #1082, opened 2026-05-18
Turn on stack protection for the diskann-garnet NuGet build

metajack:
Garnet's release automation requires that stack protectors be enabled, so this sets RUSTFLAGS to set `stack-protectors=all` and uses RUSTC_BOOTSTRAP=1 to for the stable compiler into nightly mode.

### DIFF
--- a/.github/workflows/publish-diskann-garnet-nuget.yml
+++ b/.github/workflows/publish-diskann-garnet-nuget.yml
@@ -26,6 +26,8 @@ jobs:
             artifact-name: linux
             lib-path: target/release/libdiskann_garnet.so
           - os: windows-latest
+            env:
+              EXTRA_RUSTFLAGS: control-flow-guard
             artifact-name: windows
             lib-path: target/release/diskann_garnet.dll
             pdb-path: target/release/diskann_garnet.pdb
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-diskann-garnet-nuget.yml
```diff
@@ -35,6 +35,9 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Build diskann-garnet (release)
+        env:
+          RUSTC_BOOTSTRAP: 1
+          RUSTFLAGS: -Z stack-protector=all
```
hildebrandmw: Setting `RUSTFLAGS` this way overrides the settings in `.cargo/config.toml`, which will probably break SIMD.

`RUSTC_BOOTSTRAP` is a little spooky - are we using this to see if this passes the internal gate?

For internal consumption, we should be using the appropriate approved toolchain.

metajack: I've fixed RUSTFLAGS now. Good catch.

As for the toolchain, I wanted to keep using the declared release toolchain, as opposed some random nightly one since we are making a release build. RUSTC_BOOTSTRAP allows us to use nightly features of that toolchain.

hildebrandmw (approved):
Let's see what happens I guess.

JordanMaples (approved):

suhasjs (approved):

PR #1097, opened 2026-05-21
Refactor `DiskIndexSearcher::flat_search` to use batching

hildebrandmw:
Refactor `DiskIndexSearcher::flat_search` to use the bulk `pq_distances` method instead of one-by-one computation. This does two things:

1. The bulk method is presumably a little more efficient.
2. This moves the implementation away from `Accessor`/`BuildQueryComputer` to help with #1067.

The small tricky bit is deriving the maximum batch size from the size of scratch. To that end, I added a private accessor for `PQScratch` to return this bound.

### DIFF
--- a/diskann-disk/src/search/provider/disk_provider.rs
+++ b/diskann-disk/src/search/provider/disk_provider.rs
@@ -920,25 +920,35 @@ where
     where
         OB: search_output_buffer::SearchOutputBuffer<(u32, Data::AssociatedDataType)> + Send,
     {
+        // We process IDs up to `BATCH_SIZE` elements at a time.
+        const BATCH_SIZE: usize = 32;
+        let mut id_buffer = Vec::with_capacity(BATCH_SIZE);
+
         let provider = self.index.provider();
         let mut accessor = strategy
             .search_accessor(provider, &DefaultContext)
             .into_ann_result()?;
-        let computer = accessor.build_query_computer(query).into_ann_result()?;
 
         let mut best = NeighborPriorityQueue::new(neighbors_before_reranking);
         let mut cmps = 0u32;
 
-        let num_points = provider.num_points as u32;
-        for id in 0..num_points {
-            if vector_filter(&id) {
-                let element = accessor.get_element(id).await.into_ann_result()?;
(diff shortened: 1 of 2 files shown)


diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -924,21 +924,31 @@ where
...
-                best.insert(Neighbor::new(id, dist));
-                cmps += 1;
+        let mut iter = (0..provider.num_points as u32).filter(vector_filter);
+        loop {
+            id_buffer.clear();
+            id_buffer.extend(iter.by_ref().take(batch_size));
+
+            if id_buffer.is_empty() {
+                break;
             }
```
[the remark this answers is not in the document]
hildebrandmw: ooph, good catch. This really should have been a construction invariant on the pq-scratch.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -924,21 +924,42 @@ where
...
+
+            accessor.pq_distances(&id_buffer, |dist, id| best.insert(Neighbor::new(id, dist)))?;
```
arkrishn94, line 956: Do we know if this will cause a large allocation if the `batch_size` is large (~1M)?

arkrishn94 (approved):
Thanks Mark, just had one question about whether there is any change to allocation in the search path.

harsha-simhadri (approved):

diskann-disk/src/search/provider/disk_provider.rs
hildebrandmw: Construction of the PQ scratch has already paid for the allocation. I wouldn't worry about it too much, though, because the size of the PQ scratch is limited by the graph degree.

hildebrandmw:
> Thanks Mark, just had one question about whether there is any change to allocation in the search path.

There shouldn't be. If there is, then the normal beam expansion also has an allocation. I believe the PQ scratch is sized such that all the necessary buffers are already allocated. Plus, the scratch is stored in an `ObjectPool` to further amortize the allocation. Basically, there should be no more allocations than there already are.

PR #1099, opened 2026-05-21
Significantly speed up bitmap computation

### LINKED ISSUE #1074: Make bitmap computation faster

magdalendobson:
# Introduction

Bitmap computation in diskann-label-filter is unacceptably slow. Currently, with a 1 million size slice of yfcc and a 10k query set, computing the query bitmaps takes 43.10 seconds. With just a 100K slice of the caselaw dataset and a 10k query set, computing the bitmaps takes 6.03 seconds. This was making it hard to run experiments on filtered search algorithms for the full sizes of these datasets.

Speeding up the bitmap computation is conceptually simple. Instead of iterating over every base label for every query filter, we compute an inverted index for each label type, which maps the label value to the documents with the same value. Then, at query time, we query the inverted index for the relevant label values, and compose the resulting sets as necessary to find the documents satisfying the entire filter expression. At a high level, that is what this PR does.

# Lower level details

The overall workflow of the main function, `compute_query_bitmaps`, is as follows:
1. Check whether the query expression contains any `ASTExpr::Not` clauses. If so, default to the existing slow path. This is because we don't store the document universe for each label, and thus can't compute the complement of an arbitrary bitset.
2. Otherwise, move to the fast path.
3. Flatten the base labels so that nested values map to a single string (e.g. the JSON string {"car": {"color":"red", "make":Mazda"}} would be transformed to {"car.color":red, "car.make":"Mazda}), and re-organize as a hash map of labels to values.
4. For each label, compute either an inverted index (strings and bools) or an B-tree (ints and floats) depending on its type. 
5. At query time, use either the inverted index or the B-tree to produce a bitset for each `CompareOp` in the clause, and then compose them with AND and OR as needed to produce the final bitset.

We also add a utility to `diskann-label-filter` for computing the specificity of a set of query filters with respect to a base set, outputting some statistics on it, and optionally outputting the individual specificity values to a file for further processing.

## Inverted Index
The inverted index maps each label value, converted to a string, to a bitset containing the doc ids corresponding to that value.

## B-Tree
For simplicity, the B-tree implementation converts integers to floats before inserting so that we don't have to deal with two different types of B-tree. The performance of this piece of code isn't sensitive enough that it makes sense to differentiate, but this could be changed in the future. 

The B-tree maps collections of ids to vectors instead of bitsets, because concatenating vectors is much cheaper than extending bitsets, and potentially many vectors would be concatenated during a range query.

# Timings

Returning to the earlier discussion of timings, for the 1 million size slice of yfcc and a 10k query set, computing the query bitmaps now takes .6 seconds. For the 100K slice of the caselaw dataset and a 10k query set, computing the bitmaps now takes 1.728 seconds.


diskann-tools/src/utils/compute_bitmap.rs
```diff
@@ -0,0 +1,994 @@
...
+impl Ord for OrderedFloat {
+    fn cmp(&self, other: &Self) -> Ordering {
+        // By construction, we know the partial comparison will succeed.
+        // Return `Eq` if it doesn't for better code-gen.
+        self.0.partial_cmp(&other.0).unwrap_or(Ordering::Equal)
+    }
+}
+pub enum QueryAccelerator {
+    InvertedIndex(HashMap<AttributeValue, BitSet>),
+    BTree(BTreeMap<OrderedFloat, Vec<usize>>),
```
hildebrandmw: How much slower is a bitset compared to `Vec<usize>`?

hildebrandmw (changes requested):
Performance improvement of bitmap calculation is really needed to unblock serious work on the filtered algorithms, so I have a vested interest in seeing something like this merged ASAP.

That said, there is a lot of work needed to fully productionize it. As examples:
1. The conversion of `i64` to `f64` is an imprecise conversion. Similarly `u64` would have to be treated separately from `i64`, which is currently isn't in `AttributeValue`.
2. I'm not convinced the handling of "." as a separator is robust: what happens if a user string contains a "."? For example, currently `{"a.b": 1}` and `{"a": {"b": 1}}` get lowered the same, creating ambiguity. 
3. Probably want to use `RoaringBitSet` instead of `BitSet`.
4. This adds `anyhow` as a low-level library error type, which is not a great fit.
5. This *also* unconditionally adds `rayon` as a dependency of `diskann-label-filter` and doesn't provide a caller with a clear ability to opt-out.
6. Copy-paste in `eval_query_using_accelerators` that could be factored out.
7. Many of the helper structs made public that don't necessarily need to be.
This is probably fine for the datasets we are testing on where everything is "nice", but I have concerns about the overall correctness in the presence of corner cases.
8. Minor, but the PR description says R-tree when the implementation uses a B-tree.

To unblock algorithmic work, though, what if we do the following:

1. Put this behind an "experimental" feature flag and in an "experimental" module to clearly indicate that things can go wrong.
2. Gate the new dependencies on this features to avoid unconditionally adding dependencies (in particularly, `anyhow` and `rayon`).
3. Enable this feature in benchmarks to unblock filtered work.

magdalendobson:
> Performance improvement of bitmap calculation is really needed to unblock serious work on the filtered algorithms, so I have a vested interest in seeing something like this merged ASAP.
> 
> That said, there is a lot of work needed to fully productionize it. As examples:
> 
> 1. The conversion of `i64` to `f64` is an imprecise conversion. Similarly `u64` would have to be treated separately from `i64`, which is currently isn't in `AttributeValue`.
> 2. I'm not convinced the handling of "." as a separator is robust: what happens if a user string contains a "."? For example, currently `{"a.b": 1}` and `{"a": {"b": 1}}` get lowered the same, creating ambiguity.
> 3. Probably want to use `RoaringBitSet` instead of `BitSet`.
> 4. This adds `anyhow` as a low-level library error type, which is not a great fit.
> 5. This _also_ unconditionally adds `rayon` as a dependency of `diskann-label-filter` and doesn't provide a caller with a clear ability to opt-out.
> 6. Copy-paste in `eval_query_using_accelerators` that could be factored out.
> 7. Many of the helper structs made public that don't necessarily need to be.
>    This is probably fine for the datasets we are testing on where everything is "nice", but I have concerns about the overall correctness in the presence of corner cases.
> 8. Minor, but the PR description says R-tree when the implementation uses a B-tree.
> 
> To unblock algorithmic work, though, what if we do the following:
> 
> 1. Put this behind an "experimental" feature flag and in an "experimental" module to clearly indicate that things can go wrong.
> 2. Gate the new dependencies on this features to avoid unconditionally adding dependencies (in particularly, `anyhow` and `rayon`).
> 3. Enable this feature in benchmarks to unblock filtered work.

I'll address your individual comments separately, but one thing this makes me wonder is if we should actually move this piece of code to `diskann-tools`. It's true that unlike most of the rest of the library, this is geared towards internal users who are computing groundtruth/benchmarking/etc. This would resolve your concern about adding `rayon` and `anyhow` to `diskann-label-filter`, which I agree has drawbacks. To me this would be significantly preferable for the user experience than an experimental feature.

diskann-tools/src/utils/compute_bitmap.rs
magdalendobson: It was quite a lot slower--I experimented with Bitsets and found it unacceptable. Did you want to see explicit timings?

magdalendobson:
> Performance improvement of bitmap calculation is really needed to unblock serious work on the filtered algorithms, so I have a vested interest in seeing something like this merged ASAP.
> 
> That said, there is a lot of work needed to fully productionize it. As examples:
> 
> 1. The conversion of `i64` to `f64` is an imprecise conversion. Similarly `u64` would have to be treated separately from `i64`, which is currently isn't in `AttributeValue`.

With the move to `diskann-tools`, is this still something you would like to see handled?

> 2. I'm not convinced the handling of "." as a separator is robust: what happens if a user string contains a "."? For example, currently `{"a.b": 1}` and `{"a": {"b": 1}}` get lowered the same, creating ambiguity.

Doesn't our existing code already explicitly treat those two expressions as the same? E.g. https://github.com/microsoft/DiskANN/blob/e3139b4838f20f1ffa36a600ef1fce403e68f92f/diskann-label-filter/src/parser/query_parser.rs#L50

> 3. Probably want to use `RoaringBitSet` instead of `BitSet`.

After the move to `diskann-tools`, since the existing drivers use `usize` for vector ids, I don't think it makes sense to move to `RoaringBitSet`.

> 4. This adds `anyhow` as a low-level library error type, which is not a great fit.

Resolved by the move to `diskann-tools`. 

> 5. This _also_ unconditionally adds `rayon` as a dependency of `diskann-label-filter` and doesn't provide a caller with a clear ability to opt-out.

Resolved by the move to `diskann-tools`. 

> 6. Copy-paste in `eval_query_using_accelerators` that could be factored out.

Resolved in latest edits.

> 7. Many of the helper structs made public that don't necessarily need to be.
>    This is probably fine for the datasets we are testing on where everything is "nice", but I have concerns about the overall correctness in the presence of corner cases.

In latest edits I made the helper functions and structs private.

> 8. Minor, but the PR description says R-tree when the implementation uses a B-tree.

Resolved.


> 
> To unblock algorithmic work, though, what if we do the following:
> 
> 1. Put this behind an "experimental" feature flag and in an "experimental" module to clearly indicate that things can go wrong.
> 2. Gate the new dependencies on this features to avoid unconditionally adding dependencies (in particularly, `anyhow` and `rayon`).
> 3. Enable this feature in benchmarks to unblock filtered work.

diskann-tools/src/utils/compute_bitmap.rs
hildebrandmw: Nah, that's okay. Just curious.

hildebrandmw (approved):

hildebrandmw:
> > Performance improvement of bitmap calculation is really needed to unblock serious work on the filtered algorithms, so I have a vested interest in seeing something like this merged ASAP.
> > That said, there is a lot of work needed to fully productionize it. As examples:
> > 
> > 1. The conversion of `i64` to `f64` is an imprecise conversion. Similarly `u64` would have to be treated separately from `i64`, which is currently isn't in `AttributeValue`.
> 
> With the move to `diskann-tools`, is this still something you would like to see handled?
> 
> > 2. I'm not convinced the handling of "." as a separator is robust: what happens if a user string contains a "."? For example, currently `{"a.b": 1}` and `{"a": {"b": 1}}` get lowered the same, creating ambiguity.
> 
> Doesn't our existing code already explicitly treat those two expressions as the same? E.g.
> 
> https://github.com/microsoft/DiskANN/blob/e3139b4838f20f1ffa36a600ef1fce403e68f92f/diskann-label-filter/src/parser/query_parser.rs#L50
> 
> > 3. Probably want to use `RoaringBitSet` instead of `BitSet`.
> 
> After the move to `diskann-tools`, since the existing drivers use `usize` for vector ids, I don't think it makes sense to move to `RoaringBitSet`.
> 
> > 4. This adds `anyhow` as a low-level library error type, which is not a great fit.
> 
> Resolved by the move to `diskann-tools`.
> 
> > 5. This _also_ unconditionally adds `rayon` as a dependency of `diskann-label-filter` and doesn't provide a caller with a clear ability to opt-out.
> 
> Resolved by the move to `diskann-tools`.
> 
> > 6. Copy-paste in `eval_query_using_accelerators` that could be factored out.
> 
> Resolved in latest edits.
> 
> > 7. Many of the helper structs made public that don't necessarily need to be.
> >    This is probably fine for the datasets we are testing on where everything is "nice", but I have concerns about the overall correctness in the presence of corner cases.
> 
> In latest edits I made the helper functions and structs private.
> 
> > 8. Minor, but the PR description says R-tree when the implementation uses a B-tree.
> 
> Resolved.
> 
> > To unblock algorithmic work, though, what if we do the following:
> > 
> > 1. Put this behind an "experimental" feature flag and in an "experimental" module to clearly indicate that things can go wrong.
> > 2. Gate the new dependencies on this features to avoid unconditionally adding dependencies (in particularly, `anyhow` and `rayon`).
> > 3. Enable this feature in benchmarks to unblock filtered work.

Thanks Magdalen, most of this gets resolved by moving it to `diskann-tools`, which is a better fit.

diskann-tools/src/bin/compute_specificities.rs
```diff
@@ -0,0 +1,105 @@
...
+use std::env;
+use std::fs::File;
+use std::io::Write;
+use std::process;
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    if args.len() != 3 && args.len() != 4 {
+        eprintln!(
+            "Usage: {} <base_label_file> <query_label_file> [specificity_output_file]",
```
harsha-simhadri: Would prefer input intake with argparse. This is error prone.

diskann-tools/src/utils/compute_bitmap.rs
```diff
@@ -0,0 +1,978 @@
...
+};
+use diskann_label_filter::{ASTExpr, CompareOp};
+use rayon::prelude::*;
+use std::cmp::Ordering;
+use std::collections::BTreeMap;
+use std::collections::HashMap;
+use std::mem::discriminant;
+use std::ops::Bound::{Excluded, Included, Unbounded};
+
+struct NotNonNan;
```
harsha-simhadri, line 25: could you provide some explanation for what a `NotNonNan` is?

diskann-tools/src/utils/compute_bitmap.rs
```diff
@@ -0,0 +1,978 @@
...
+}
+
+enum QueryAccelerator {
+    InvertedIndex(HashMap<AttributeValue, BitSet>),
+    BTree(BTreeMap<OrderedFloat, Vec<usize>>),
+}
+
+fn check_for_disallowed_operators(query_expr: &ASTExpr) -> bool {
+    match query_expr {
+        ASTExpr::Not(_) => true,
```
harsha-simhadri: should the check for disallowed operators be more central than here?(Such as where the syntax is parse and validated)

diskann-tools/src/utils/compute_bitmap.rs
```diff
@@ -0,0 +1,978 @@
...
+                let b = eval_query_using_accelerators(e, query_accelerators)?;
+                acc = Some(match acc {
+                    None => b,
+                    Some(acc_b) => acc_b.union(&b).collect(),
+                });
+            }
+            Ok(acc.unwrap_or_else(BitSet::new))
+        }
+        ASTExpr::Not(_) => Err(anyhow::anyhow!(
+            "NOT operator is not supported when using query accelerators"
```
harsha-simhadri: why not?

diskann-tools/src/utils/compute_bitmap.rs
```diff
@@ -0,0 +1,978 @@
...
+                    }
+                }
+            } else {
+                // if field not present, return an empty bitset
+                Ok(BitSet::new())
+            }
+        }
+    }
+}
+
```
harsha-simhadri, line 296: Coudl this be designed with "query_accelerator" as a trait with multiple concrete implementations?

harsha-simhadri (commented):
posted some questions inline. thanks

diskann-label-filter/src/utils/flatten_utils.rs
```diff
@@ -72,7 +72,7 @@ impl Document {
     }
 
     /// Configurable version that uses FlattenConfig
-    /// For example, with config.separator=".": {"details": {"color": "red"}} becomes {".details.color": "red"}
+    /// For example, with config.separator="/": {"details": {"color": "red"}} becomes {"/details/color": "red"}
```
sampathrg, line 75: I was just curious about why the default behavior of this API changed. Is it wrong? If this behavior is needed, you could also use `flatten_json_pointers_with_separator`. Same question for the change below.

diskann-tools/src/utils/ground_truth.rs
```diff
@@ -35,25 +36,18 @@ pub fn read_labels_and_compute_bitmap(
...
-        .map(|(_query_id, query_expr)| {
-            let mut bitmap = BitSet::new();
-            for base_label in base_labels.iter() {
-                if eval_query_expr(query_expr, &base_label.label) {
-                    bitmap.insert(base_label.doc_id);
-                }
-            }
-            bitmap
-        })
-        .collect();
```
sampathrg, line 54: I was experimenting with something to make this faster, my changes are [here](https://github.com/microsoft/DiskANN/commit/ed8ffd7e7369921ac31b71c7ee27414617a14da3#diff-ad3a5ad44780967537011048873cc13be903e8b8e996dd0b7243a447e09a193a). 

What I'm trying to in those changes is to is reuse the expression filtering apis in diskann-label-filter and trying to improve time taken. I was seeing about 35x speedup for this part. The filtering apis can't be used as it is today in main and will need some work, some of the changes to fix performance issues with the api are in the same commit.

We can perhaps discuss which approach is better. The reason to reuse the filtering apis in diskann-label-filter is that we can have one code that takes care of filtering and is performant. On the other hand, compute-ground-truth may not have the same constraints as the code that needs to handle filtering during search so we may want to build something else. I'm not sure about the latter part though

magdalendobson: Thanks! I agree that reusing code would have been somewhat preferable, but a concern I have with your approach is that numeric ranges are disallowed and accelerating them would probably require a separate data structure. That is something that is really needed in the short-term to unblock work on the caselaw dataset, so it's something we need to have incorporated into our benchmarking code ASAP (like this week). Are you planning on also augmenting the inverted index implementation to support numeric ranges?

diskann-label-filter/src/utils/flatten_utils.rs
magdalendobson: This was a mistake from some experimental code. Removed now.

diskann-tools/src/utils/compute_bitmap.rs
magdalendobson: Yes, I made this change now.

diskann-tools/src/bin/compute_specificities.rs
magdalendobson: Done, thanks

diskann-tools/src/utils/compute_bitmap.rs
magdalendobson: Added some comments. It's used to throw an error if we try to create the `OrderedFloat` type on a Nan. We need to exclude Nan so that we can build a Btree over float values.

diskann-tools/src/utils/compute_bitmap.rs
magdalendobson: I've removed this check as we now support `Not` operators.

diskann-tools/src/utils/compute_bitmap.rs
magdalendobson: I re-wrote this code to support Not operators in the end. They were not supported because they required computing a document universe over all the labels contained in each sub-expression. After your comment I decided to add a `universe()` function to the `QueryAccelerator` trait so we could support this operator.

diskann-tools/src/utils/ground_truth.rs
magdalendobson: Coming back here, now that this function supports the `Not` operator I think this is another big difference between this code and the filtering API.

magdalendobson:
> posted some questions inline. thanks

I believe these are all now resolved, would you mind approving if you agree?

diskann-tools/src/utils/ground_truth.rs
sampathrg: > but a concern I have with your approach is that numeric ranges are disallowed and accelerating them would probably require a separate data structure. That is something that is really needed in the short-term to unblock work on the caselaw dataset

you're right, the current approach doesn't support range filters because it encodes <attribute name, attribute value> pairs as integers. It will require some rework to get it to work correctly. Since it's urgent it would make sense to submit this and revisit this later.

One thing I was wondering was about tracking performance regressions for this piece of code. If someone were to change the compute_bitmap part, are there benchmarks that keep track of the baseline performance? 

>now that this function supports the Not operator I think this is another big difference between this code and the filtering API.

The `Not` operator is [supported](https://github.com/microsoft/DiskANN/blob/e3139b4838f20f1ffa36a600ef1fce403e68f92f/diskann-label-filter/src/inline_beta_search/predicate_evaluator.rs#L82) in the filter expression.

diskann-tools/src/utils/compute_bitmap.rs
```diff
@@ -0,0 +1,1088 @@
...
+            }
+            global_label_set.insert(key.clone(), value.clone());
+        }
+    }
+    Ok(global_label_set)
+}
+
+fn compute_query_accelerator(
+    key: String,
+    value: AttributeValue,
```
sampathrg: These two values can be passed as immutable reference. There are no usages of these that require the function to own the values. This will avoid the `clone` at the callsite.

sampathrg (approved):

diskann-tools/src/utils/compute_bitmap.rs
magdalendobson: Thanks for pointing that out, fixed

harsha-simhadri (approved):

diskann-tools/src/utils/ground_truth.rs
magdalendobson: Marking resolved for the moment, since this improvement is urgent and we really need to include the numeric filters. Let's keep in touch when your changes land and see if we want to refactor

PR #1101, opened 2026-05-25
RFC - Beta Filter For Disk Search

dyhyfu:
## Summary

Adds beta-biased filtering on the disk search path by introducing a `SearchPlan` enum that replaces the existing `(vector_filter, is_flat_search)` parameter pair on `searcher.search()`. The new enum is hierarchical (`SearchPlan { FlatScan, Graph(GraphMode) }`), makes invalid combinations unrepresentable, and gives future filter algorithms a single extension point without further growing the public `search()` signature.

## Motivation

- **Beta-biased graph search is unavailable on the disk path.** The in-memory side has a `BetaFilter` strategy; the disk side has no equivalent.
- **A raw closure can't carry `beta`.** Bolting it on as a third `Option<f32>` parameter creates meaningless combinations (`is_flat_search=true` + `beta=Some(_)`) the type system can't reject.
- **No clean integration point for future filter algorithms** (`MultihopSearch` exists in `diskann` but the disk API has no way to select it).

## What's in the RFC

- `SearchPlan { FlatScan { filter: Option<Predicate> }, Graph(GraphMode) }` + `GraphMode { Unfiltered, PostFilter(p), BetaFilter { predicate, beta } }`.
- `Predicate = Box<dyn Fn(u32) -> bool + Send + Sync>` — closure-based.
- Single `GraphMode` match site in `search_strategy()` that projects `(predicate, beta)`; downstream `DiskAccessor` and `RerankAndFilter` read those fields directly — adding a new `GraphMode` variant touches one place.
- Invalid combinations (beta on flat scan, beta without a predicate) unrepresentable by construction.
- Zero allocation and zero per-iteration overhead preserved on the no-filter graph path.

## Caller migration

| Today `(vector_filter, is_flat_search)` | New `plan` |
|---|---|
| `(None, false)` | `SearchPlan::graph()` |
| `(None, true)` | `SearchPlan::flat()` |
| `(Some(p), false)` | `SearchPlan::graph_with(GraphMode::post_filter(p))` |
| `(Some(p), true)` | `SearchPlan::flat_filtered(p)` |
| (new capability) | `SearchPlan::graph_with(GraphMode::beta_filter(p, β))` |

## Known follow-ups (deferred)

- Disk's `GraphMode::BetaFilter` applies bias + hard post-filter; in-memory `BetaFilter` only biases. Future work to align.

See `rfcs/01101-disk-beta-filter.md` for the full design.


rfcs/01101-disk-beta-filter.md
```diff
@@ -0,0 +1,538 @@
...
+
+2. **Adding beta breaks the existing orthogonality between `is_flat_search` and `vector_filter`.** Today the two flags are genuinely orthogonal: all four `(is_flat_search, vector_filter)` combinations are meaningful, and the benchmark exposes them as independent inputs ([diskann-benchmark/src/inputs/disk.rs:83-85](../diskann-benchmark/src/inputs/disk.rs)). But beta only exists in graph search — it biases beam expansion, and a flat scan has no beam. Adding beta as a third independent parameter creates meaningless combinations (e.g. `is_flat_search=true` + `beta=Some(_)`) that the type system can't reject.
+
+3. **No integration point for future graph algorithms.** `MultihopSearch` already exists in `diskann/src/graph/search/multihop_search.rs`, but the disk API has no way to select it. Each new algorithm would mean another boolean flag.
+
+This design replaces both flags with a single `plan: SearchPlan` enum. Each of the five supported configurations is one explicit variant or constructor; future graph algorithms slot in as new `GraphMode` variants without changing the public `search()` signature; the compiler enforces exhaustiveness at every dispatch site.
```
suri-kumkaran: nit: There’s quite a bit of repetition across these sections — can we make them a bit more concise?

rfcs/01101-disk-beta-filter.md
```diff
@@ -0,0 +1,538 @@
...
+    search_list_size,
+    beam_width,
+    SearchPlan::graph_with(GraphMode::beta_filter(
+        move |id| active_ids.contains(id),
+        0.5,  // beta: bias beam toward matching vectors
+    )),
+)?;
+```
+
+**Benchmark CLI**: the `--is_flat_search` flag is removed from the disk-index benchmark input schema. The benchmark layer constructs `SearchPlan` from `(is_flat_search, vector_filters_file)` once at the boundary and passes only `SearchPlan` to `searcher.search()`. JSON input schemas that still carry `is_flat_search` should be migrated as a separate task (config-only, no code dependency after the backend update).
```
suri-kumkaran: I think we can migrate the JSON input in one go as well, since it’s only used in CI testing and doesn’t have any other significant users.

rfcs/01101-disk-beta-filter.md
```diff
@@ -0,0 +1,538 @@
...
+#### 4.2 `search_strategy` projection — the single `GraphMode` match site
+
+`search_strategy` is the **only** place that introspects `GraphMode`'s variants. One exhaustive match produces `(predicate, beta)`; the strategy carries them as fields. Every downstream consumer reads `strategy.predicate` and `strategy.beta` — no further variant matching anywhere.
+
+```rust
+fn search_strategy<'a>(
+    &'a self,
+    query: &'a [Data::VectorDataType],
+    plan: &'a SearchPlan,
+) -> DiskSearchStrategy<'a, ...> {
```
suri-kumkaran, line 179: I feel it will be valuable to add the new modified `DiskSearchStrategy` struct here in RFC.

suri-kumkaran (commented):
Thanks for working through this — I’m aligned with the design. I do feel we can trim some of the redundant or overly explanatory parts of the RFC and leave a few of the implementation details for the implementation PRs.

rfcs/01101-disk-beta-filter.md
```diff
@@ -0,0 +1,538 @@
...
+
+#### 2.1 The `Predicate` type alias
+
+The disk path identifies filtered IDs through a closure, not a trait. There is no `IdFilter` trait, no `BitmapProvider` wrapper, and no `QueryLabelProvider<u32>` dependency.
+
+```rust
+// diskann-disk/src/search/filter_parameter.rs
+// (Moved from build/configuration/ — this is a search-time concept, not build-time.)
+
+pub type Predicate = Box<dyn Fn(u32) -> bool + Send + Sync>;
```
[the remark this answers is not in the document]
suri-kumkaran: As written, `Predicate` defaults to `+ 'static` (the default object lifetime bound for `Box<dyn Trait>` in alias position), and the constructors require `F: Fn(u32) -> bool + Send + Sync + 'static`. That's strictly tighter than today's `VectorFilter<'a, Data>`, which accepts non-`'static` closures — and the in-tree benchmark caller relies on this:

```rust
 // diskann-benchmark/src/backend/disk_index/search.rs:264-272
 zipped.for_each_in_pool(pool.as_ref(),
     |(((((q, vf), id_chunk), dist_chunk), stats), rc)| {
         let vector_filter = ...
             Some(Box::new(move |vid: &u32| vf.contains(vid))
                 as Box<dyn Fn(&u32) -> bool + Send + Sync>);
         match searcher.search(q, ..., vector_filter, ...) { ... }
 ```

 Here `vf: &RoaringBitmap` is borrowed from `vector_filters: Vec<RoaringBitmap>` (reused across the outer L sweep), so the closure is not `'static`. After this change it won't compile as written.

 I think `'static`-only is still the right call — parameterizing `Predicate<'a>` would propagate a lifetime through `SearchPlan`, `GraphMode`, every constructor, the `search()` signature, and every future `GraphMode` variant. The actual benefit is small: per-query bitmaps built fresh just `move`-by-value into the closure (zero overhead, closure becomes `'static`); the one prebuilt-and-reused case (the benchmark) is a 1-line hoist (`Vec<Arc<RoaringBitmap>>` outside the L loop, `Arc::clone` into each closure — a few ns per query, invisible against search latency). And since `Predicate` requires `Send + Sync` anyway, stack-borrowed non-`Send` data was never an option.

 But the choice should be made explicit rather than silent. Two suggested edits:

 **§8 "Backward Compatibility"** — add:

 > `Predicate` is `'static` — callers passing closures that borrow non-`'static` data must own the data into the closure (`move` an owned `RoaringBitmap`, or hold backing data in `Arc<_>` and clone the `Arc` into the closure). This is a deliberate simplification over today's `VectorFilter<'a, Data>`: borrowed-predicate flexibility did not justify a lifetime parameter on `SearchPlan` / `GraphMode` / `Predicate` and every future `GraphMode` variant. The one in-tree caller affected — the benchmark's `vector_filters: Vec<RoaringBitmap>` reused across the L sweep — migrates by hoisting to `Vec<Arc<RoaringBitmap>>` once before the loop and cloning the `Arc` into each per-query closure.

 **§2.4 migration table** — annotate the `(Some(p), false)` and `(Some(p), true)` rows with a footnote: *"identical behavior when `p` is `'static`; borrowed-predicate callers need to own their data into the closure — see §8."*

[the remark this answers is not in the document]
Comment (suri-kumkaran) [reply rfcs/01101-disk-beta-filter.md]: Hey @wuw92, I'd love to get your perspective on this since you have a better feel for the actual usage patterns of this feature in the wild.

What are your thoughts on strictly enforcing `'static` versus parameterizing with a lifetime (`'a`)?

* **Going with `'static`:** Gives us a significantly cleaner API. We avoid bubbling a lifetime parameter up through `SearchPlan`, `GraphMode`, sibling/nested enums, and the public API layer. The downside is that callers *must* own the `Predicate` data. If they need to share a bitmap across multiple queries, they are forced to use an `Arc`.
* **Going with `'a` (Alternative):** Preserves full flexibility for zero-cost shared predicates (like our in-tree benchmark does today), but at the cost of a much noisier and more complex API surface everywhere the graph mode is touched.

My working assumption is that most real-world use cases are generating per-query bitmaps in single-threaded searches anyway, making it trivial to just `move` ownership into the closure.

Do you think the ergonomic win of a cleaner, lifetime-free API justifies the `Arc`/ownership tax for the shared cases?

File: rfcs/01101-disk-beta-filter.md [hunk 5]
```diff
@@ -0,0 +1,265 @@
...
+
+## Proposal
+
+### Types
+
+```rust
+// diskann-disk/src/search/filter_parameter.rs
+// (Moved from build/configuration/ — search-time concept, not build-time.)
+
+pub type Predicate<'a> = Box<dyn Fn(u32) -> bool + Send + Sync + 'a>;
```
suri-kumkaran, line 53: **Consider keying `Predicate` on external IDs, with translation inside the disk crate.**

Today the predicate is invoked with internal IDs at all three sites (`flat_search`, `pq_distances` callback, `RerankAndFilter`). It only works because `DiskProvider::to_internal_id`/`to_external_id` are identity (disk_provider.rs:93-124). External-keyed is the more correct contract:

- Matches the `DataProvider` abstraction; doesn't leak the positional internal representation.
- Matches in-memory `BetaFilter` (keyed on `DP::ExternalId` via `QueryLabelProvider`) — closes one corner of the divergence the Future Work item flags.
- Future-proof: non-identity translation (sparse IDs, deletion compaction, u64→u32) stays disk-internal; callers don't break silently.

**Cost in the identity case is negligible.** `to_external_id` is `Ok(id)`; with `#[inline]` LLVM folds it to the value. The three call sites already return `ANNResult` and propagate via `?`, so the only added cost is a tag-check on the `Result` discriminant — which the inliner constant-folds when it can see the `Ok(_)` body, and which is in any case branch-predicted free in a loop doing PQ distance work. No caller changes today (their bitmaps are already keyed on the IDs they see in search outputs).

Caveat: the "external" contract stays doc-level unless we newtype `ExternalId` — separate refactor; flipping the contract now is the pragmatic step.

rfcs/01101-disk-beta-filter.md
```diff
@@ -0,0 +1,265 @@
...
+- **Additional `GraphMode` variants in this change.** The enum is built for extension, but each new variant should land with its own consumer integration, tests, and (for traversal modifiers) recall data.
+
+## Trade-offs
+
+N/A
+
+## Benchmark Results
+
+N/A
+
```
suri-kumkaran, line 295: nit: We can omit these NA sections.

rfcs/01101-disk-beta-filter.md
```diff
@@ -0,0 +1,265 @@
...
+    ProviderFactory: VertexProviderFactory<Data>,
+{
+    io_tracker: IOTracker,
+    query: &'a [Data::VectorDataType],
+
+    // === Changed ===
+    // Removed: vector_filter: &'a VectorFilter<'a, Data>
+    // Added: projected from `plan` once in search_strategy()
+    predicate: Option<&'a Predicate<'a>>,
+    beta: Option<f32>,
```
suri-kumkaran: The two independent `Option` fields admit a structurally-invalid `(None, Some(β))` state, prevented today only by the exhaustive match in `search_strategy`. A sum type at the strategy layer makes it unrepresentable *and* extends better than nesting `beta` inside a `FilterConfig` struct:

```rust
enum FilterMode<'a> {
    None,
    Filter(&'a (dyn Fn(u32) -> bool + Send + Sync + 'a)),
    BetaFilter {
        predicate: &'a (dyn Fn(u32) -> bool + Send + Sync + 'a),
        beta: f32,
    },
    // Future: Multihop { predicate, depth }, LabelBeta { labels, beta }, ...
}
```

A `FilterConfig { predicate, beta: Option<f32> }` works for today's set but bakes in "one predicate + modifiers hang off as Options" — degrades the moment a future modifier needs different supplementary data or no predicate at all. The sum type lets each variant carry exactly what it needs, same pattern §Key design decisions already commits to at `GraphMode`, extended one layer down.

(Also folded in: each variant carries `&dyn Fn` directly instead of `&Box<dyn Fn>` — one fewer indirection at every predicate call. `SearchPlan` still owns the `Box`; the strategy projects via `predicate.as_ref()` in the match.)

Consumers stay short:
```rust
// pq_distances
if let FilterMode::BetaFilter { predicate, beta } = self.filter {
    if predicate(id) { distance *= beta; }
}
// RerankAndFilter
match &self.filter {
    FilterMode::None => None,
    FilterMode::Filter(p) | FilterMode::BetaFilter { predicate: p, .. } => Some(*p),
}
```
Exhaustiveness forces every future variant through both sites.

suri-kumkaran (commented):
This is getting close!

rfcs/01101-disk-beta-filter.md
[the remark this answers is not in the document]
wuw92: `'static` carries a hidden cost that depends on what the Filter closure actually captures. Since DiskANN is a public library, callers will show up in shapes we can't fully predict:
  | Filter closure | `'static` cost |
  |---|---|
  | Self-contained, pre-built (owned bitmap loaded per request) | Low — one `Arc::new` at request entry. |
  | Dispatches to external state (service field, DB / KV lookup, per-request context on the stack) | High — forces every such dependency into `Arc<...>`, an API-driven structural change to the caller's data  model. Sometimes unsatisfiable when the resource isn't owned by the caller.|

Given DiskANN's library position, worth keeping `'a` to stay neutral on caller architecture?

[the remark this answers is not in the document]
suri-kumkaran: I feel that this particular API is specific to `DiskSearch` and your team is the major client so I feel we can tailor it according to that for now. I am fine with anything :)

rfcs/01101-disk-beta-filter.md
```diff
@@ -0,0 +1,324 @@
...
+| (not expressible) | `SearchPlan::graph_with(GraphMode::beta_filter(p, β)?)` |
+
+Boundary code in the benchmark and tools constructs `SearchPlan` once and passes it down. The `--is_flat_search` benchmark flag and the corresponding `is_flat_search` field in benchmark JSON schemas (`diskann-benchmark/example/*.json`, `diskann-benchmark/perf_test_inputs/*.json`) are removed in the same change — no external consumers depend on the old schema, so no migration grace period is needed.
+
+### ID convention — fix the internal/external mismatch in `RerankAndFilter`
+
+Current state:
+
+- **Input predicate is keyed on `ExternalId`** by `DataProvider` contract.
+- **`Index::flat_search`** is correct — calls `to_external_id` before invoking the predicate ([diskann/src/graph/index.rs](../diskann/src/graph/index.rs)).
```
suri-kumkaran: There's no `flat_search` in `diskann/src/graph/index.rs` (grep confirms — only `disk_provider.rs` has one). I think this might be a conflation with `diskann::flat::FlatIndex::knn_search` in `diskann/src/flat/index.rs` (the module from RFC 00983-flat-search) — that's a separate, parallel abstraction that the disk path doesn't delegate to.

The disk's own `flat_search` (`disk_provider.rs:912`) is a standalone implementation, and it invokes the predicate with positional/internal IDs directly:

```rust
let mut iter = (0..provider.num_points as u32).filter(vector_filter);
```

No `to_external_id` call anywhere in the disk flat_search path (confirmed: grep for `to_external_id` in `disk_provider.rs` returns only the trait definition and one unit test). The `0..num_points` idiom looks generic but is the positional/internal space by construction — there's no abstraction between integer counting and positional storage. Identity translation masks this today; under any future non-identity provider, flat-scan and graph paths would diverge silently.

So all three sites (`flat_search`, `pq_distances`, `RerankAndFilter::post_process`) are currently incorrect — this RFC needs to fix all three.

Suggested edits:
- Remove the "Sites already correct" bullet (line 269).
- Add `flat_search` to "Sites updated in the implementation PR" (line 262):
  > - **`flat_search`** (the disk-local one in `disk_provider.rs`) — call `to_external_id` on each id in `(0..num_points)` before invoking the predicate.

rfcs/01101-disk-beta-filter.md
```diff
@@ -0,0 +1,324 @@
...
+- **`RerankAndFilter::post_process`** — call `to_external_id` on each candidate before invoking the predicate.
+- **`pq_distances`** (new code path for `FilterMode::BetaFilter`) — call `to_external_id` on each id before invoking the predicate.
+
+**Sites already correct:**
+
+- **`Index::flat_search`** — already converts to external before invoking.
+
+**Cost** in the identity case is negligible: `to_external_id` returns `Ok(id)`; `#[inline]` lets LLVM fold it, leaving only a `Result` tag-check that the inliner constant-folds when it sees the `Ok(_)` body. In the PQ-distance hot loop this is branch-predicted free.
+
+**Caller impact:** none. Bitmaps that callers build from prior search results are already keyed on external IDs by the `DataProvider` contract — they were the *intended* input all along; this RFC just makes every invocation site honor it.
```
suri-kumkaran: nit: Slight overstatement. True today *because of identity translation* — once a future non-identity provider lands, callers building bitmaps against output IDs stay correct, but any caller building bitmaps against positional indices would break. The §Non-Goals on newtyping `ExternalId` covers why we can't enforce this at compile time; worth softening the line here to match:

 > Caller impact: none today (identity translation). Going forward, callers must build predicates over external IDs — the contract this RFC establishes. Bitmaps built from prior search outputs are external-keyed by construction.

rfcs/01101-disk-beta-filter.md
```diff
@@ -0,0 +1,324 @@
...
+
+1. **Beta-biased graph search is unavailable on the disk path.**
+
+2. **The current API shape can't carry beta cleanly.** A raw closure has no place to attach `β`. Threading `beta: Option<f32>` as a fourth parameter creates orthogonality problems: beta is only meaningful in graph search, and `(vector_filter, is_flat_search, beta)` as three independent inputs admits meaningless combinations (`is_flat_search=true` + `beta=Some(_)`) the type system can't catch.
+
+3. **No clean integration point for future filter algorithms.** 
+
+### Goals
+
+1. Add beta-biased disk graph search as a first-class capability.
```
suri-kumkaran, line 42: Goal 2 from the previous draft ("Provide a single extension point for future filter algorithms without changing the public `search()` signature") was dropped, but the body still extensively justifies that property (§Key design decisions bullet 1, §Extensibility table). If the trim was for terseness, fine; if it was inadvertent, worth restoring — it's load-bearing in the §Trade-offs reasoning.

rfcs/01101-disk-beta-filter.md
wuw92: nit: `flat_search` has been moved to `disk_provider.rs`

suri-kumkaran (approved):
Aligned on the design. ID convention is a clean fix and the `FilterMode<'a>` sum type makes the invariants compiler-enforced.

On `Predicate<'a>` vs `Predicate` (`'static`-only): fine either way — lifetime preserves borrow flexibility, `'static`-only is simpler. Author's call.

Approving once the inline comments are resolved.

rfcs/01101-disk-beta-filter.md
```diff
@@ -0,0 +1,324 @@
...
+| `(Some(p), true)` | `SearchPlan::flat_filtered(p)` |
+| (not expressible) | `SearchPlan::graph_with(GraphMode::beta_filter(p, β)?)` |
+
+Boundary code in the benchmark and tools constructs `SearchPlan` once and passes it down. The `--is_flat_search` benchmark flag and the corresponding `is_flat_search` field in benchmark JSON schemas (`diskann-benchmark/example/*.json`, `diskann-benchmark/perf_test_inputs/*.json`) are removed in the same change — no external consumers depend on the old schema, so no migration grace period is needed.
+
+### ID convention — fix the internal/external mismatch in `RerankAndFilter`
+
+Current state:
+
+- **Input predicate is keyed on `ExternalId`** by `DataProvider` contract.
```
wuw92: Input predicate is based on `InternalId`, you may look at imem BetaFilter and multihop implementation.

rfcs/01101-disk-beta-filter.md
wuw92: Both `BetaFilter`  and `MultihopSearch`  take `QueryLabelProvider<DP::InternalId>` — their predicates run  in the **internal-id** space. The only `to_internal_id` / `to_external_id` translations in `index.rs` happen at the insert/delete API boundary, not on the search path.
We can keep `Predicate` keyed on internal-id here too.

rfcs/01101-disk-beta-filter.md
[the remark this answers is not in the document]
dyhyfu: Discussed with @wuw92, we decide to keep the lifetime param.

rfcs/01101-disk-beta-filter.md
dyhyfu: Checking latest code, flat_search has been moved to disk_provider.rs and is using internal_id as input param. The "ID convension" part will be removed and `Predicate` will be keyed on internal_id.

suri-kumkaran: The motivation for external-keyed predicates was broader than the `flat_search` audit — that was a symptom, not the core argument. External-keyed is still the right contract because:

 - Matches the `DataProvider` abstraction; doesn't leak the positional internal representation.
 - Matches in-memory `BetaFilter` (keyed on `DP::ExternalId` via `QueryLabelProvider`) — closes one corner of the divergence the Future Work item flags.
 - Future-proof: non-identity translation (sparse IDs, deletion compaction, `u64`→`u32`) stays disk-internal; callers don't break silently.

Practically: how do external clients construct a predicate over internal IDs? They only see external IDs — from search outputs, ingest, the `DataProvider` contract. Asking them to maintain a bitmap over a space they can't observe pushes implementation details out as caller responsibility, and silently couples every consumer to today's identity-translation invariant.

The fix is small and inlines to a no-op under identity translation. Worth keeping in the RFC.

wuw92: `DiskProvider` doesn't really expose an Internal/External contract today. Both types are pinned to `u32`, `VectorIdType = u32` is a hard `where` bound on the impl (`disk_provider.rs:95-103`), I'd suggest dropping the ID-convention change from this RFC.
```rust
  impl<Data> DataProvider for DiskProvider<Data>
  where Data: GraphDataType<VectorIdType = u32>,
  {
      type InternalId = u32;
      type ExternalId = u32;

      fn to_internal_id(&self, _: &DefaultContext, gid: &u32) -> Result<u32, _> { Ok(*gid) }
      fn to_external_id(&self, _: &DefaultContext, id: u32)   -> Result<u32, _> { Ok(id) }
  }
```
For Practical, external clients implements to_internal_id / to_external_id on their provider, so they define the translation and own both sides. The question isn't whether they can observe internal id — it's which side of the boundary is the natural place to do the translation.

For the 1:N domain model (which is what our team has: external = doc_id, internal = vector_id, one doc-id → many vector-ids), "caller side" is a feasible option.

- DataProvider's ExternalId ↔ InternalId is 1:1 by trait signature; a 1:N domain doesn't fit through it.
- So we keep the doc_id ↔ {vector_id} map outside the library, for example in a filtered scenario:
  1. Filtered doc-ids → vector-id bitmap (outside DiskANN).
  2. DiskANN filter-search returns vector-ids.
  3. Vector-ids → doc-ids (outside DiskANN).

suri-kumkaran: Two separate things getting mixed up — the **outer** mapping (domain ↔ vector IDs) and the **inner** consistency (which ID space DiskANN uses internally).

**Outer mapping (1:N)**: agreed, this lives outside the library. `DataProvider::ExternalId ↔ InternalId` is 1:1 by signature, so doc-id ↔ {vector-id} has to be maintained caller-side. The flow you describe (filtered doc-ids → vector-id bitmap → DiskANN search → vector-ids → doc-ids) is the only way it can work. No disagreement there.

**Inner consistency (the RFC's change)**: by the time DiskANN receives the predicate, it's already in vector-id space (= external ID). The question is whether DiskANN invokes that predicate with `ExternalId` or `InternalId`. Today three sites — `RerankAndFilter::post_process`, `pq_distances`, `flat_search` — pass internal directly. This is a latent bug: it produces wrong results the moment `DiskProvider` ever uses non-identity translation (e.g., for deletion compaction or sparse IDs added in place to the existing provider — no fork required).

Re: "`DiskProvider` doesn't expose the distinction today" — true, and that's exactly why the bug is invisible. The trait exists precisely so future in-place changes don't break consumers. Fixing now is essentially free (translation inlines to a no-op under identity). Leaving it means the next person extending `DiskProvider` has to remember to also audit every predicate site, with no compile-time help.

Want to keep the ID-convention section in this RFC — it pre-existing-bug we surface and fix here. Fix lands in the impl PR.

dyhyfu:

dyhyfu:
Close this PR since we are not planning on proceeding beta filter on disk index.

PR #1102, opened 2026-05-25
RFC: Query Planner for Per-Filter-Bitmap Disk Search

YuanyuanTian-hh:


rfcs/01102-query-planner.md
```diff
@@ -0,0 +1,479 @@
...
+
+## Summary
+
+Adds a lightweight **query planner** that automatically selects between flat scan and beta-filtered graph search on the disk path based on the filter bitmap's **matching count** and **match rate**. The planner sits between the caller and `search_internal()`, takes a bitmap of allowed vector IDs plus the matching count, and produces the appropriate `SearchPlan`. Callers no longer choose a search strategy manually — the planner adapts to the actual data distribution.
+
+## Motivation
+
+### Background
+
+When a query is scoped to a specific filter category, the caller extracts a list of items matching that category, maps them to DiskANN's internal vector IDs, and constructs a bitmap. The caller wraps the bitmap in a `Predicate` closure and passes it to the disk searcher.
```
suri-kumkaran: Line 17 makes the caller responsible for mapping to internal IDs:
> maps them to DiskANN's internal vector IDs, and constructs a bitmap

Once the beta-filter RFC flips `Predicate` to be keyed on external IDs (with translation inside DiskANN), the planner should follow the same contract — the caller just builds a bitmap of vertex IDs as seen from search outputs, no internal-ID mapping at the call site.

Suggested edit to line 17: drop the "internal vector IDs" phrasing. Bitmap is over external IDs; DiskANN translates internally at the predicate boundary.

rfcs/01102-query-planner.md
```diff
@@ -0,0 +1,479 @@
...
+    }
+
+    /// Plan and produce a `SearchPlan` with the appropriate predicate wiring.
+    ///
+    /// The caller provides the bitmap (as an `Arc<RoaringBitmap>`) and the
+    /// matching count. The planner selects the strategy and constructs the
+    /// `SearchPlan` with the closure already wired to the bitmap.
+    pub fn plan_search(
+        &self,
+        bitmap: Arc<RoaringBitmap>,
```
suri-kumkaran: **`plan_search` is pinned to `RoaringBitmap`, which contradicts the beta-filter RFC it builds on.**

Line 219:
```rust
pub fn plan_search(&self, bitmap: Arc<RoaringBitmap>, matching_count: u64) -> SearchPlan
```

But line 48 in this RFC's own §Prerequisites says:
> The `Predicate` closure can wrap any bitmap type (`RoaringBitmap`, `BitSet`, `HashSet<u32>`)

The whole point of `Predicate = Box<dyn Fn(u32) -> bool>` is bitmap-agnosticism — callers using `BitSet`, `HashSet<u32>`, or any custom backing structure can't use the planner as written.

Cleaner shape — take the predicate abstraction the beta-filter RFC already chose, with `matching_count` as a separate caller-provided hint (predicates aren't size-aware, so the count must come from outside regardless):

```rust
pub fn plan_search<F>(&self, predicate: F, matching_count: u64) -> SearchPlan
where F: Fn(u32) -> bool + Send + Sync + 'static + Clone,
 ```

The planner clones the predicate into whichever `SearchPlan` arm it picks. Bitmap implementation stays the caller's choice; the planner only cares about "is there a predicate and how many points does it match."

File: rfcs/01102-query-planner.md [hunk 3]
```diff
@@ -0,0 +1,479 @@
...
+    ) -> SearchPlan {
+        let strategy = self.plan(matching_count);
+        let beta = self.config.beta;
+
+        match strategy {
+            QueryStrategy::FlatSearch => {
+                let bm = bitmap.clone();
+                SearchPlan::FlatScan {
+                    filter: Some(Box::new(move |id| bm.contains(id))),
+                }
```
suri-kumkaran: `plan_search` builds variants via direct struct construction (lines 228-230, 234-237):
```rust
SearchPlan::FlatScan { filter: Some(Box::new(move |id| bm.contains(id))) }
SearchPlan::Graph(GraphMode::BetaFilter { predicate: Box::new(...), beta })
```

This bypasses the constructors (`SearchPlan::flat_filtered`, `GraphMode::beta_filter`) that the beta-filter RFC introduced precisely to centralize predicate boxing and `β ∈ (0, 1]` validation. As written, `QueryPlannerConfig::beta` is unvalidated — a caller passing `beta = 2.0` produces a plan that *increases* distances on matching vectors (semantic inversion), silently.

Use the constructors and propagate the `Result`:
```rust
QueryStrategy::FlatSearch =>
    Ok(SearchPlan::flat_filtered(move |id| bitmap.contains(id))),
 QueryStrategy::BetaFilter =>
    GraphMode::beta_filter(move |id| bitmap.contains(id), beta)
        .map(SearchPlan::graph_with),
```

`plan_search` returns `Result<SearchPlan, BetaError>`. Validation discipline established by the beta-filter RFC stays intact through the planner.

rfcs/01102-query-planner.md
```diff
@@ -0,0 +1,479 @@
...
+│                 │  │    → hard-removes non-matching   │
+└────────┬────────┘  └──────────────┬───────────────────┘
+         │                          │
+         └──────────┬───────────────┘
+                    ▼
+            SearchResult returned
+            to caller
+```
+
+### 3. Strategy Selection
```
suri-kumkaran, line 111: `plan_search` only produces `FlatScan` or `BetaFilter` — `GraphMode::PostFilter` and `Graph::Unfiltered` (and any future `GraphMode` variant) are unreachable through the planner.

Two questions worth answering explicitly:

1. **Why is `PostFilter` never selected?** The beta-filter RFC keeps it as a first-class mode. If it's strictly dominated by `BetaFilter` at every (matching_count, match_rate) cell on the benchmark surface, say so — and consider deprecating it from the underlying API. If it's not dominated, the planner is missing a case.

2. **What's the extension story?** When future graph algorithms land as new `GraphMode` variants (the beta-filter RFC explicitly leaves room — `Multihop`, etc.), how does the planner pick them up? Today every new variant needs a new threshold derivation and a new `QueryStrategy` arm. Worth at least sketching: does each `GraphMode` self-describe its viable (matching_count, match_rate) region, with the planner aggregating? Or does each new variant land with a planner update?

Even a brief §Future Work note ("planner currently covers two of three modes; PostFilter dominance to be validated") would close the gap. Right now a reader can't tell whether the omission is intentional or oversight.

suri-kumkaran, line 111: The current decision logic has a lower bound on flat scan (`matching_count ≤ 200K` → flat) but no upper bound. Consider: 10M index at 20% match rate. matching_count = 2M, match_rate < 25%, so the planner picks flat scan.

From Experiment 5: on a 10M index, flat scan at 1M matching vectors is ~970ms; at 5M it's ~3.4s. 2M matching extrapolates to ~1.5–2s. Latency tracks matching_count almost linearly — at 2M it's already 5× the ~315ms beta plateau.

Real workloads will have a latency budget. If a caller's budget is, say, 500ms, the planner shouldn't route to flat scan when matching_count is large enough that flat scan blows past the budget — even if the recall would be better. Accept the recall hit; meet the SLA.

A symmetric `max_brute_force_matching_count` (or, more precisely, a latency-budget-aware threshold) closes this gap:

```rust
if matching_count <= matching_count_threshold       → FlatScan
else if matching_count >= max_brute_force_count     → BetaFilter   // (new)
else if match_rate <= flat_search_threshold         → FlatScan
else                                                → BetaFilter
```

Today's use cases may not hit it (smaller indexes), but the decision logic should cover the full surface so larger indexes don't need a re-derivation. Even a conservative default (e.g. `max_brute_force_count = 1M`) caps worst-case flat-scan latency at ~1s on a 10M index.

suri-kumkaran (commented):
Overall the planner is a solid addition — the benchmark methodology is thorough, the hybrid threshold is well-justified against the data, and the routing layer cleanly composes with the beta-filter RFC's `SearchPlan` shape. Leaving inline comments on a few design points worth resolving before this lands.

One general note: several of the code examples carry small impl-detail issues (unnecessary `Arc::clone()` where ownership transfer would do, direct struct construction, hard-pinned bitmap type, etc.). These are distracting at the RFC stage — readers spend cycles on whether the code is right, rather than whether the *design* is right. Happy to handle these in the impl PR; the RFC just needs to nail the design decisions that need consensus.

rfcs/01102-query-planner.md
YuanyuanTian-hh: Fixed — dropped the 'internal vector IDs' phrasing.

rfcs/01102-query-planner.md
YuanyuanTian-hh: Fixed — The planner is bitmap-agnostic — callers wrap any bitmap type into a closure before calling. Bounds match the beta-filter RFC's constructor signatures exactly.

rfcs/01102-query-planner.md
YuanyuanTian-hh: Fixed — now uses `SearchPlan::flat_filtered()` and `GraphMode::beta_filter()` constructors.

rfcs/01102-query-planner.md
YuanyuanTian-hh: Good catch — the omission is intentional, not an oversight.

**Why PostFilter is not selected:** `PostFilter` applies the same hard post-filter as `BetaFilter` but without the beta bias on PQ distances. It is strictly dominated — `BetaFilter` does everything `PostFilter` does plus biases beam traversal toward matching vectors, improving recall at high match rates. At low match rates where beta bias hurts (the 2–10% dip zone), the planner routes to `FlatScan` instead of `PostFilter`, since flat scan has ~100% recall. There is no (matching_count, match_rate) cell in the benchmark data where `PostFilter` outperforms both `FlatScan` and `BetaFilter`.

**Extension story:** Each new `GraphMode` variant lands with its own threshold derivation and a new `QueryStrategy` arm — the planner update is part of the variant's integration. Self-describing viable regions is an interesting future direction but over-engineers the current two-strategy planner. Will add a Future Work note.

YuanyuanTian-hh: Added `max_brute_force_count` (default: 1M) as a third threshold:

\\\
if matching_count <= 200K           → FlatScan
else if matching_count >= 1M        → BetaFilter   // NEW
else if match_rate <= 25%           → FlatScan
else                                → BetaFilter
\\\

Validated across 7 index sizes (1M–10M, dim=384). Where flat scan crosses 500ms:

{32DE80A4-347B-4FD0-B3CE-1AE03183EC38}

suri-kumkaran: Do we know that 500ms is an acceptable latency for the use case?, but given it's configurable - the RFC is unblocked.

rfcs/01102-query-planner.md
```diff
@@ -0,0 +1,534 @@
...
+        QueryStrategy::BetaFilter
+    }
+}
+```
+
+- `MATCHING_COUNT_THRESHOLD` default: **200,000** — flat scan is cheap when few vectors match
+- `MAX_BRUTE_FORCE_COUNT` default: **1,000,000** — caps flat scan latency at ~500–700ms
+- `FLAT_SEARCH_THRESHOLD` default: **0.25** (25%) — avoids the beta recall dip zone (2–10%)
+
+### 4. API
```
suri-kumkaran, line 134: This RFC builds on `beta_filter_rfc` but diverges from its design philosophy ("make invalid states unrepresentable," typed errors, fallible only at the boundary constructor). Worth aligning the two so they compose cleanly.

rfcs/01102-query-planner.md
```diff
@@ -0,0 +1,534 @@
...
+    pub fn validate(&self) -> Result<(), &'static str> {
+        if !(self.beta > 0.0 && self.beta <= 1.0) {
+            return Err("beta must be in (0, 1]");
+        }
+        if self.matching_count_threshold >= self.max_brute_force_count {
+            return Err("matching_count_threshold must be < max_brute_force_count");
+        }
+        Ok(())
+    }
+}
```
suri-kumkaran, line 201: The all-`pub` fields + separate `validate()` means invariants aren't enforced by construction — easy to build an invalid config and never validate it. Prefer a single checked constructor/builder that returns a validated config, mirroring `GraphMode::beta_filter()` being the only constructor.

rfcs/01102-query-planner.md
```diff
@@ -0,0 +1,534 @@
...
+            flat_search_threshold: 0.25,
+            beta: 0.5,
+        }
+    }
+}
+
+impl QueryPlannerConfig {
+    /// Validate the configuration. Returns an error if beta is not in (0, 1]
+    /// or if matching_count_threshold >= max_brute_force_count.
+    pub fn validate(&self) -> Result<(), &'static str> {
```
suri-kumkaran, line 192: `Result<_, &'static str>` is stringly-typed and unmatchable by callers. It's also inconsistent with the sibling RFC (`BetaError` via `thiserror`) and isn't a convention in this repo. Use a typed `thiserror` error.

Similarly at other places as well.

rfcs/01102-query-planner.md
```diff
@@ -0,0 +1,534 @@
...
+        }
+    }
+
+    /// Plan and produce a `SearchPlan` with the appropriate predicate wiring.
+    ///
+    /// The caller provides a predicate closure (wrapping any bitmap type) and
+    /// the matching count. The planner selects the strategy and constructs the
+    /// `SearchPlan` with the predicate wired in.
+    ///
+    /// Returns `Err` if beta validation fails (e.g. beta not in (0, 1]).
```
suri-kumkaran, line 251: `beta` is validated in `config.validate()` and again in `GraphMode::beta_filter()` inside `plan_search()`. That second check is the *only* reason `plan_search` returns `Result` — and it's unreachable once `new()` validated.
Validate beta once (e.g. a `Beta` newtype) and make `plan_search` infallible.

rfcs/01102-query-planner.md
```diff
@@ -0,0 +1,534 @@
...
+/// The strategy selected by the query planner.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum QueryStrategy {
+    /// Brute-force linear scan with hard filter.
+    FlatSearch,
+
+    /// Beta-biased graph search with post-filter.
+    BetaFilter,
+}
+```
```
suri-kumkaran, line 148: This is a third enum mirroring a subset of `GraphMode`, needing a hand-maintained mapping. Future Work even notes new arms must land in lockstep with `GraphMode` variants — two enums kept in sync by hand. Reconsider whether this needs to exist vs. returning into the existing `SearchPlan`/`GraphMode` hierarchy.

rfcs/01102-query-planner.md
```diff
@@ -0,0 +1,534 @@
...
+
+#### 4.2 `QueryPlannerConfig`
+
+```rust
+pub struct QueryPlannerConfig {
+    /// Matching-count threshold. If matching_count <= this value, always use
+    /// flat scan — flat scan is cheap for small result sets regardless of
+    /// total index size.
+    /// Default: 200_000
+    pub matching_count_threshold: u64,
```
suri-kumkaran, line 158: `matching_count_threshold` (lower bound) vs `max_brute_force_count` (upper bound) read as unrelated, but both bound the same quantity. Something like `flat_scan_min_count` / `flat_scan_max_count` makes the relationship obvious.

rfcs/01102-query-planner.md
```diff
@@ -0,0 +1,534 @@
...
+    pub fn plan(&self, matching_count: u64) -> QueryStrategy {
+        if matching_count <= self.config.matching_count_threshold {
+            return QueryStrategy::FlatSearch;
+        }
+
+        if matching_count >= self.config.max_brute_force_count {
+            return QueryStrategy::BetaFilter;
+        }
+
+        let match_rate = matching_count as f64 / self.total_points as f64;
```
suri-kumkaran, line 237: `matching_count / total_points` with `total_points == 0` yields `NaN`, and `NaN <= threshold` is `false`, so it silently routes to `BetaFilter`. 
`new()` should reject `total_points == 0`.

rfcs/01102-query-planner.md
suri-kumkaran, line 1: The whole RFC reads quite verbose, not just the data. §5 (Threshold Derivation) and the Benchmark Results section largely restate the same numbers, and the "All experiments: beta=0.5..." preamble is duplicated (419 ≈ 324) — but the prose sections above (Motivation, Design Overview, Strategy Selection, Decision Flow, Trade-offs) also repeat the three-threshold rationale several times over. Worth a pass to make it crisp  throughout: state each threshold + its rationale once, drop the repeated restatements, and fold the bulk benchmark tables into a collapsed appendix so the design intent isn't buried.

suri-kumkaran (commented):
Thanks for this - routing logic and threshold derivation are well-argued. Two pieces of high-level feedback:

The RFC reads quite **verbose throughout** - the three-threshold rationale is restated several times across Motivation, Design Overview, Strategy Selection, Decision Flow, and Trade-offs, and Benchmark Results largely repeat the same numbers. Worth a pass to state each point once and fold the bulk tables into an appendix, so the design intent isn't buried.

My other comments are all on the **API surface** and share one theme: the RFC builds on `beta_filter_rfc` but drifts from its principles ("make invalid states unrepresentable", typed `thiserror` errors, fallible only at the boundary constructor). Since the two are meant to compose, I'd like the planner to adopt the same conventions. Details inline -mostly around `validate()`/config construction, error types, and the `QueryStrategy` enum. None block the core idea; just want the public API to land clean since it's an extension point.

PR #1106, opened 2026-05-26
add bf_tree benchmark infrastructure

### LINKED ISSUE #1090: Diskann-benchmark should support Bf-tree and other providers
Allow diskann-benchmark to use non in-mem providers. Bf-tree would be a good example to add.

JordanMaples:
## Summary

Adds benchmark support for the bf_tree storage backend — both full-precision and spherical-quantized, in static (build-once) and streaming (insert/delete/search) modes.

## What's included

**Core crate fix (`diskann/src/graph/index.rs`):**
- **VisitedAndTopK ordering fix:** `inplace_delete_inner` now pre-fetches the delete element *before* calling `delete()`, so that `VisitedAndTopK` can use the deleted vector as a search query. This is critical for hard-delete providers (bf-tree) where the data is erased on delete. Soft-delete providers (in-mem) are unaffected since data remains accessible either way. All three delete methods (OneHop, TwoHopAndOneHop, VisitedAndTopK) continue to work for both provider types.

**Benchmark variants (4 tags):**
 - `graph-index-bftree-full-precision-f32` — static build + search
 - `graph-index-bftree-stream-full-precision-f32` — streaming with runbook
 - `graph-index-build-bftree-spherical-quantization` — static spherical (1/2/4-bit)
 - `graph-index-stream-bftree-spherical-quantization` — streaming spherical (1/2/4-bit)

**Bug fixes in diskann-bftree:**
 1. **Hard delete with VisitedAndTopK support:** bf-tree now performs true hard deletes via `Delete::delete`. The ordering fix in the core crate ensures the delete element is captured before erasure, enabling all three inplace_delete methods to work correctly.
 2. **Max-degree mismatch in append_vector:** The old `append_vector` would resize the neighbor list to `self.dim` and write the count at `dim - 1`, but `set_neighbors` wrote a variable-length buffer. This inconsistency caused bf-tree page fragmentation and potential read errors when the two code paths produced different value sizes for the same key. Fix: both paths now write a fixed-size dim-element buffer using the same format (`|neighbors|padding|count|`).
 3. **Reusable buffer in hot path:** `set_neighbors` and `append_vector` previously allocated on every call. Replaced with a caller-provided `&mut [u8]` scratch buffer, eliminating per-call heap allocations.

**Benchmark infrastructure:**
 - Expose bf_tree internal config (buffer sizes, fanout, promotion rates) as JSON input fields with validation
 - Shared helpers: `bftree_parameters_from()`, `run_streaming()` in dedicated util module
 - `GenericStats::empty` replaced with `Option<GenericStats>` for maintain stats (bf_tree has no separate drop_deleted/release phases)
 - Minimum circular buffer size validated early with a friendly error message (must be ≥ 2× leaf_page_size)

## Example configs

See `diskann-benchmark/example/graph-index-bftree*.json` for ready-to-run configurations targeting test_data.

## Tests on CosmosDBDevBox VM (16 vCPU 64GB RAM 2,048GB SSD)
Results from MSTuring-1M

**Full Precision streaming:**
```
 max_degree=64, L_build=128, alpha=1.2, 4GB bf_tree buffers, 4 build threads, 2 search threads

┌───────┬────────┬─────────┬───────────────────┬──────────────────────────┬────────────────────┐
│ Stage │ Op     │ Vectors │ Avg / p99 Latency │ Recall@10 (L=32 / L=200) │ QPS (L=32 / L=200) │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 1     │ Insert │ 1M      │ 11.9ms / 17.9ms   │ —                        │ —                  │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 2     │ Search │ —       │ —                 │ 82.8% / 97.0%            │ 460 / 99           │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 3     │ Delete │ 500K    │ 16.3ms / 20.9ms   │ —                        │ —                  │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 4     │ Search │ —       │ —                 │ 84.2% / 97.9%            │ 474 / 97           │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 5     │ Insert │ 500K    │ 15.2ms / 19.6ms   │ —                        │ —                  │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 6     │ Search │ —       │ —                 │ 83.9% / 97.4%            │ 430 / 92           │
└───────┴────────┴─────────┴───────────────────┴──────────────────────────┴────────────────────┘
```

**Spherical 2-bit streaming:**
```
 max_degree=64, L_build=128, alpha=1.2, 4GB bf_tree buffers, 4 build threads, 2 search threads

┌───────┬────────┬─────────┬───────────────────┬──────────────────────────┬────────────────────┐
│ Stage │ Op     │ Vectors │ Avg / p99 Latency │ Recall@10 (L=32 / L=200) │ QPS (L=32 / L=200) │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 1     │ Insert │ 1M      │ 4.8ms / 7.6ms     │ —                        │ —                  │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 2     │ Search │ —       │ —                 │ 74.7% / 94.0%            │ 1167 / 255         │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 3     │ Delete │ 500K    │ 6.9ms / 9.3ms     │ —                        │ —                  │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 4     │ Search │ —       │ —                 │ 73.4% / 94.1%            │ 1103 / 233         │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 5     │ Insert │ 500K    │ 6.0ms / 8.6ms     │ —                        │ —                  │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 6     │ Search │ —       │ —                 │ 75.1% / 94.3%            │ 1053 / 229         │
└───────┴────────┴─────────┴───────────────────┴──────────────────────────┴────────────────────┘
```

**Spherical 4-bit streaming:**
```
 max_degree=64, L_build=128, alpha=1.2, 4GB bf_tree buffers, 4 build threads, 2 search threads

┌───────┬────────┬─────────┬───────────────────┬──────────────────────────┬────────────────────┐
│ Stage │ Op     │ Vectors │ Avg / p99 Latency │ Recall@10 (L=32 / L=200) │ QPS (L=32 / L=200) │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 1     │ Insert │ 1M      │ 4.8ms / 7.7ms     │ —                        │ —                  │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 2     │ Search │ —       │ —                 │ 82.3% / 96.7%            │ 1127 / 253         │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 3     │ Delete │ 500K    │ 6.9ms / 9.5ms     │ —                        │ —                  │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 4     │ Search │ —       │ —                 │ 82.4% / 97.3%            │ 1095 / 231         │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 5     │ Insert │ 500K    │ 6.3ms / 9.1ms     │ —                        │ —                  │
├───────┼────────┼─────────┼───────────────────┼──────────────────────────┼────────────────────┤
│ 6     │ Search │ —       │ —                 │ 83.1% / 97.1%            │ 1019 / 217         │
└───────┴────────┴─────────┴───────────────────┴──────────────────────────┴────────────────────┘
```

**Comparison across bit widths at L=200:**
```
┌─────────────────┬────────┬─────┬────────────────┐
│ Variant         │ Recall │ QPS │ Insert Latency │
├─────────────────┼────────┼─────┼────────────────┤
│ Full Precision  │ 97.0%  │ 99  │ 11.9ms         │
├─────────────────┼────────┼─────┼────────────────┤
│ Spherical 4-bit │ 96.7%  │ 253 │ 4.8ms          │
├─────────────────┼────────┼─────┼────────────────┤
│ Spherical 2-bit │ 94.0%  │ 255 │ 4.8ms          │
└─────────────────┴────────┴─────┴────────────────┘
```

The configuration used in the above tests was enough to keep all of the data in memory and didn't overflow to disk. I also tested full-precision using 32mb of cb size and experienced a 2x performance degradation due to the disk lookups.

## CI Notes

CI reports an increase in IR size of 11%, which is over the 5% allowed in the CI file. This is largely unavoidable given the amount of code added here, so that CI gate should be ignored for this PR.


diskann-benchmark/src/inputs/bftree.rs
```diff
@@ -0,0 +1,529 @@
...
+            &self.quant_store_config,
+        )
+    }
+
+    pub(crate) fn validate(&mut self, checker: &mut Checker) -> anyhow::Result<()> {
+        self.build.validate(checker)?;
+        self.search_phase.validate(checker)?;
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
JordanMaples: added validation

diskann-benchmark/src/inputs/bftree.rs
```diff
@@ -0,0 +1,529 @@
...
+            &self.quant_store_config,
+        )
+    }
+
+    pub(crate) fn validate(&mut self, checker: &mut Checker) -> anyhow::Result<()> {
+        self.build.validate(checker)?;
+        self.search_phase.validate(checker)?;
+        self.runbook_params.validate(checker)?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
JordanMaples: added validation

diskann-bftree/src/neighbors.rs
```diff
@@ -170,29 +173,31 @@ impl<I: VectorId> NeighborProvider<I> {
...
+        debug_assert!(total_len <= MAX_NEIGHBOR_BUF_BYTES);
 
-        // Insert the assembled (K, V) pair into bf-tree
-        self.adjacency_list_index.insert(key, value);
+        let mut buf = [0u8; MAX_NEIGHBOR_BUF_BYTES];
+        buf[..neighbor_list_edges_in_byte.len()].copy_from_slice(neighbor_list_edges_in_byte);
+        buf[neighbor_list_edges_in_byte.len()..total_len]
+            .copy_from_slice(neighbor_list_len_in_byte);
+
+        self.adjacency_list_index.insert(key, &buf[..total_len]);
```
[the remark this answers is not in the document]
JordanMaples: added some validation and error logic

diskann-bftree/src/provider.rs
```diff
@@ -1020,24 +1079,40 @@ where
...
+        let v = match result {
+            Ok(()) => Ok(Opaque::new(&self.element)),
+            Err(RankedError::Transient(_)) => {
+                let cache = self.provider.quant_delete_cache.lock().unwrap();
+                if let Some(cached) = cache.get(&id) {
+                    self.element.copy_from_slice(cached);
+                    Ok(Opaque::new(&self.element))
+                } else {
+                    Err(result.unwrap_err())
+                }
```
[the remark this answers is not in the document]
JordanMaples: it does compile though.

diskann-bftree/src/provider.rs
```diff
@@ -851,25 +887,42 @@ where
...
+            Ok(()) => Ok(&*self.element),
+            Err(RankedError::Transient(_)) => {
+                // Check delete_cache for recently-deleted vectors still needed for pruning.
+                let cache = self.provider.delete_cache.lock().unwrap();
+                if let Some(cached) = cache.get(&id) {
+                    self.element.copy_from_slice(cached);
+                    Ok(&*self.element)
+                } else {
+                    Err(result.unwrap_err())
+                }
```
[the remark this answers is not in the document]
JordanMaples: it does compile though

JordanMaples:
@copilot re-evaluate please

diskann-benchmark/src/backend/index/bftree/spherical.rs
```diff
@@ -0,0 +1,457 @@
...
+    where
+        P: plugins::Plugin<BfTreeSQProvider, SearchPhase, Strategy<common::Quantized>> + 'static,
+    {
+        self.search.register(plugin);
+        self
+    }
+}
+
+/// Construct a type-erased `Poly<dyn Quantizer>` for the given NBITS.
+macro_rules! make_quantizer_poly {
```
hildebrandmw: This need not be a macro:
```rust
fn new_quantizer<const NBITS: usize>(
    quantizer: SphericalQuantizer,
) -> Result<Poly<dyn Quantizer>, AllocatorError>
where
    spherical_iface::Impl<NBITS>: spherical_iface::Constructible + Quantizer,
{
    let imp = spherical_iface::Impl::<NBITS>::new(quantizer)?;
    diskann_quantization::poly!(Quantizer, imp, GlobalAllocator)
}
```

diskann-benchmark/src/backend/index/bftree/spherical.rs
```diff
@@ -0,0 +1,459 @@
...
+
+        let training_time = start.elapsed();
+        writeln!(output, "Training time: {:.2}s", training_time.as_secs_f64())?;
+
+        // 2. Dispatch on num_bits to create the type-erased quantizer.
+        let quantizer_poly = match input.num_bits().get() {
+            1 => make_quantizer_poly!(1, quantizer),
+            2 => make_quantizer_poly!(2, quantizer),
+            4 => make_quantizer_poly!(4, quantizer),
+            n => anyhow::bail!("Unsupported num_bits: {n}"),
```
hildebrandmw: We can add this check to `try_match`? The goal of `try_match` is to catch input issues like this as early as possible, so a user doesn't queue up a list of benchmarks just to have them fail mid-run.

I guess this is added as part of input validation, so maybe not entirely needed, but is still a nice safety mechanism?

diskann-bftree/src/provider.rs
```diff
@@ -196,6 +198,13 @@ where
     // Graph configuration parameters for persistence
     //
     pub(crate) graph_params: Option<GraphParams>,
+
+    // During `inplace_delete`, the algorithm first removes vectors from storage via
+    // `Delete::delete`, then prunes the neighbors of the deleted node which requires
+    // reading the deleted vector back through the accessor. We cache vectors here
+    // before deletion so the accessor can fall back to the cache on a `Transient` error.
```
hildebrandmw: I don't think is the right way to handle this. The issue is multi-insert wanting to access the adjacency list, right? I think there are several better options:
1. In `bf-tree`, we can delete the **data**, but not the adjacency lists. That way inplace delete will continue to work. This is a local fix.
2. Adjust the semantics of inplace delete or how it's called in benchmarks to actually delete the vector only after inplace delete runs so the algorithm still has adjacency list information?

In general, if we see a mismatch like this, we should focus on not working around the problem.

diskann-bftree/src/neighbors.rs
```diff
@@ -20,6 +20,11 @@ use diskann::{
 use super::ConfigError;
 use crate::TestCallCount;
 
+/// Stack buffer size for neighbor serialization. Supports up to 512 neighbors
+/// with u32 IDs (512 * 4 = 2048 bytes). Any practical ANN workload uses
+/// max_degree well below this limit.
+const MAX_NEIGHBOR_BUF_BYTES: usize = 2048;
```
hildebrandmw: This is a *huge* stack allocation. Is this showing up in profiles?

We already create auxiliary accessors for neighbor accesses - can't we use a `Vec` inside those as the scratch space? The accessors are already reused across calls and so we can amortize the `Vec` allocation.

diskann-benchmark/ADDING_PROVIDERS.md
```diff
@@ -0,0 +1,67 @@
...
+---
+
+## Files to Create
+
+| File | Purpose |
+|------|---------|
+| `src/inputs/<provider>.rs` | Input structs (JSON schema), `Display`, `Checker`, `Example` impls |
+| `src/backend/index/<provider>/mod.rs` | Module root, `register_benchmarks()`, shared helpers |
+| `src/backend/index/<provider>/full_precision.rs` | Static + streaming FP benchmarks |
+| `src/backend/index/<provider>/spherical.rs` | Quantized benchmarks (if applicable) |
```
hildebrandmw: Some of these seem pretty prescriptive for BF-tree in particular? Can this be trimmed down to more general instructions? Or better yet, folded into the README?

diskann-benchmark/ADDING_PROVIDERS.md
```diff
@@ -0,0 +1,67 @@
...
+| `Cargo.toml` | Add optional dependency on your provider crate |
+| `src/inputs/mod.rs` | Feature-gated `pub(crate) mod <provider>` |
+| `src/backend/index/mod.rs` | Feature-gated `mod <provider>` + call `register_benchmarks()` |
+
+---
+
+## Structural Checklist
+
+### Input structs
+- [ ] Reuse `graph_index::IndexBuild` for common graph fields (max_degree, L_build, alpha, etc.)
```
hildebrandmw: We want to be careful with saying "always use `IndexBuild`. It comes with a lot of configuration which new backend may not always implement. This is problematic because it means the benchmarks accept inputs that aren't completely used and are thus misleading.

diskann-benchmark/ADDING_PROVIDERS.md
```diff
@@ -0,0 +1,67 @@
...
+  - `maintain` — provider-specific maintenance (cache clearing, consolidation, etc.)
+- [ ] Wrap in `Managed<T, StreamStats>` (handles slot management, GT translation, maintenance scheduling)
+- [ ] Implement the `Benchmark` trait for the streaming entry point
+
+### Registration
+- [ ] Choose descriptive tag strings (e.g., `graph-index-<provider>-full-precision-f32`)
+- [ ] Feature-gate with `#[cfg(feature = "...")]`
+
+### Monomorphization
+- [ ] Prefer runtime dispatch (match on literals) over const generics for variant parameters
```
hildebrandmw: This is not quite right. Matching on literals *does* effectively dispatch to const-generics. The integration of spherical quantization here could have been done with const generics as long as the `Benchmark` implementations immediately create the quantizer and then invoke general routines to avoid code bloat.

diskann-benchmark/ADDING_PROVIDERS.md
```diff
@@ -0,0 +1,67 @@
...
+- [ ] Check IR growth: `cargo llvm-lines --package diskann-benchmark --all-features --release`
+- [ ] Each new provider type adds ~150-300K IR lines from `DiskANNIndex<DP>` instantiation — this is expected
+
+---
+
+## Notes
+
+- `Managed` calls `maintain()` when deletions exceed `consolidate_threshold × active_slots`
+- `StreamStats` has variants for each operation; use `GenericStats::noop(...)` for operations your provider doesn't need
```
hildebrandmw: See my other not on `GenericStats::noop`.

diskann-benchmark/src/inputs/mod.rs
```diff
@@ -51,4 +54,15 @@ macro_rules! as_input {
     };
 }
 
+// This constant is used to ensure that summaries of graph-index related jobs properly have
+// their field descriptions aligned.
+const PRINT_WIDTH: usize = 18;
```
hildebrandmw, line 60: We really need to move away from benchmarks manually managing spacing ...

diskann-benchmark/src/inputs/graph_index.rs
```diff
@@ -662,6 +652,40 @@ impl IndexBuild {
...
+    pub(crate) fn l_build(&self) -> usize {
+        self.l_build
+    }
+
+    pub(crate) fn num_threads(&self) -> usize {
+        self.num_threads
+    }
+
+    // this is accessed in the bftree flow
+    #[allow(dead_code)]
```
hildebrandmw: Please avoid `dead_code` - it makes things harder to maintain. If it's not needed - it's not needed.

diskann-benchmark/src/inputs/bftree.rs
```diff
@@ -0,0 +1,628 @@
...
+pub(crate) struct BfTreeStoreConfig {
+    /// Size of the circular buffer (in-memory write cache) in bytes.
+    pub(crate) cb_size_byte: usize,
+
+    /// Size of leaf pages in bytes.
+    pub(crate) leaf_page_size: usize,
+
+    /// Maximum record size that can be stored in the circular buffer.
+    #[serde(default)]
+    pub(crate) cb_max_record_size: Option<usize>,
```
hildebrandmw, line 33: I recommend not using `default` for these and always requiring the fields to be present with `#[serde(deserialize_with = "Deserialize::deserialize")]` (if the fields are options). The benchmark runs are meant to record all the information we can about a run. If a silent default is applied, we have no record of what that default value actually was.

diskann-benchmark/src/inputs/bftree.rs
```diff
@@ -0,0 +1,628 @@
...
+        Self::default()
+    }
+}
+
+impl std::fmt::Display for BfTreeStoreConfig {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write_field!(f, "cb_size_byte", self.cb_size_byte)?;
+        write_field!(f, "leaf_page_size", self.leaf_page_size)?;
+        if let Some(v) = self.cb_max_record_size {
+            write_field!(f, "cb_max_record_size", v)?;
```
hildebrandmw, line 160: Along with my other comment, we should at least record what the default values are, even if they aren't explicitly set by the user (though they probably should be).

diskann-benchmark/src/inputs/bftree.rs
```diff
@@ -0,0 +1,628 @@
...
+}
+
+as_input!(BfTreeFullPrecisionBuild);
+
+#[derive(Debug, Serialize, Deserialize)]
+pub(crate) struct BfTreeFullPrecisionBuild {
+    build: IndexBuild,
+    search_phase: SearchPhase,
+    #[serde(default)]
+    vector_store_config: Option<BfTreeStoreConfig>,
```
hildebrandmw, line 219: I really think it's a good idea to not use defaults here. It's more verbose, but also means we have more information.

diskann-benchmark/src/backend/index/streaming/stats.rs
```diff
@@ -108,6 +108,22 @@ impl GenericStats {
             latencies: percentiles::compute_percentiles(&mut latencies)?,
         })
     }
+
+    /// Construct an empty stats entry (e.g., when maintenance is skipped).
+    #[cfg(feature = "bftree")]
+    pub(crate) fn empty(kind: Cow<'static, str>) -> Self {
```
hildebrandmw: Is there a way to avoid this kind of `empty` representation? The generic stats was written in a way that was intentionally pretty hard to construct with out data, and Rust already has a way of representing optional data with `Option`. I get that this is to reuse `StreamStats`, but maybe there's a better way?

diskann-benchmark/src/backend/index/bftree/spherical.rs
```diff
@@ -0,0 +1,459 @@
...
+    };
+
+    let consolidate_threshold: f32 = input.runbook_params().consolidate_threshold;
+    let data = datafiles::load_dataset::<f32>(datafiles::BinFile(input.build().data()))?;
+    let queries = Arc::new(datafiles::load_dataset::<f32>(datafiles::BinFile(
+        &topk.queries,
+    ))?);
+
+    let max_points =
+        ((max_points_arg as f32) * (1.0 + 2.0 * consolidate_threshold)).ceil() as usize;
```
hildebrandmw: I believe this exists to work around limitations in delete tracking for the inmem index. Since bf-tree is using "hard deletes" instead, I think we can potentially avoid needing this extra overhead. Happy to talk about it offline.

diskann-benchmark/src/backend/index/bftree/spherical.rs
```diff
@@ -0,0 +1,459 @@
...
+////////////////////////
+// Streaming BfTree SQ //
+////////////////////////
+
+use std::{borrow::Cow, num::NonZeroUsize};
+
+use diskann::graph::InplaceDeleteMethod;
+use diskann::utils::ONE;
+use diskann_benchmark_core::{recall::Rows, streaming::executors::bigann};
+use diskann_utils::views::MatrixView;
```
hildebrandmw: Does it make sense to either group all imports together at the top or split this into two files?

hildebrandmw (commented):
Thanks Jordan, this is looking good!

diskann-benchmark/src/backend/index/bftree/spherical.rs
JordanMaples: I'll nuke that from orbit

diskann-benchmark/src/inputs/graph_index.rs
JordanMaples: This was causing clippy to complain when not using the bftree feature. I'll look for another way around it.

diskann-benchmark/src/inputs/mod.rs
JordanMaples: This I just lifted up out of another file.

diskann-benchmark/src/backend/index/bftree/spherical.rs
JordanMaples: good call. I'll add the check to try_match

diskann-benchmark/ADDING_PROVIDERS.md
JordanMaples: yeah, let me un-bftree this and a more generic version into the readme.

diskann-benchmark/src/backend/index/bftree/spherical.rs
JordanMaples: I'm going to split out spherical and FP into two separate files.

diskann-bftree/src/neighbors.rs
JordanMaples: This was something that I was experimenting with and forgot to revert. I'll use the reusable vec buffer though.

diskann-benchmark/src/backend/index/streaming/stats.rs
JordanMaples: removed

magdalendobson:
I'm not sure I fully understand the bug in deletion. Why do we need to keep the deleted vector around for pruning its neighbors? If it's in someone's neighbor list after deletion, it should be ignored during pruning. Is there some difference in how bf-tree handles this versus how in-mem providers handle this that meant deletes needed to be handled differently?

diskann-benchmark/example/graph-index-bftree-spherical-stream.json
```diff
@@ -0,0 +1,94 @@
...
+                    "l_build": 128,
+                    "start_point_strategy": "medoid",
+                    "alpha": 1.2,
+                    "backedge_ratio": 1.0,
+                    "num_threads": 4
+                },
+                "search_phase": {
+                    "search-type": "topk",
+                    "queries": "query100K.fbin",
+                    "groundtruth": "msturing-gt-1M",
```
magdalendobson: For the sake of keeping the examples directory a bit tighter, maybe we don't need an example for streaming + spherical + bf-tree, since we already have separate examples of streaming and spherical.

diskann-benchmark/src/inputs/bftree.rs
```diff
@@ -0,0 +1,680 @@
...
+// ─── BfTree Store Configuration ───────────────────────────────────────────────
+
+/// Configuration for a single bf_tree store instance.
+///
+/// Required fields control memory sizing before data spills to disk.
+/// Optional fields tune internal behavior and default to bf_tree's defaults.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct BfTreeStoreConfig {
+    /// Size of the circular buffer (in-memory write cache) in bytes.
+    pub(crate) cb_size_byte: usize,
```
magdalendobson, line 26: iirc bf-tree has a minimum circular buffer size of 8192 bytes. Should we enforce that here?

diskann-benchmark/src/backend/index/bftree/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+    registry.register(
+        "graph-index-bftree-stream-full-precision-f32",
+        full_precision_streaming::StreamingFullPrecision::<f32>::new(),
+    )?;
+
+    const SQ_NAME: &str = "graph-index-build-bftree-spherical-quantization";
+
+    registry.register(SQ_NAME, spherical::BfTreeSpherical::new().search(Topk))?;
+
+    const SQ_STREAM: &str = "graph-index-stream-bftree-spherical-quantization";
```
magdalendobson: Do we need to create separate name variables to pass clippy or something?

diskann-benchmark/src/backend/index/bftree/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+    },
+};
+use crate::inputs::graph_index::DynamicRunbookParams;
+
+mod full_precision;
+mod full_precision_streaming;
+mod spherical;
+mod spherical_streaming;
+
+/// Run a streaming benchmark using the given runbook parameters.
```
magdalendobson: I'm not sure this function belongs in `mod.rs`, can we move it to its own file with a more descriptive name, like streaming_utils or similar?

magdalendobson:
I think posting some experiments with a more complicated runbook would be a good idea, it looks like all the experimental results here are from `simple_runbook.yaml` which has just a few steps and might not be sufficient to catch bugs or perf issues. If you like I am happy to provide some suggestions for runbooks.

diskann-benchmark/src/backend/index/bftree/spherical.rs
```diff
@@ -0,0 +1,202 @@
...
+/// monomorphization.
+pub(super) struct BfTreeSpherical {
+    search: Plugins<BfTreeSQProvider, SearchPhase, Strategy<Quantized>>,
+}
+
+impl BfTreeSpherical {
+    pub(super) fn new() -> Self {
+        Self {
+            search: Plugins::new(),
+        }
```
magdalendobson, line 56: Can some of these general helpers and structs be moved to their own file? Then we could have one file for spherical utils etc., one for generic search, and one for streaming. Would make it easier to understand the structure and navigate

magdalendobson (changes requested):
I left some cosmetic comments but overall the structure and code look good to me. For the streaming I have a couple of concerns:
1) I would like to see experiments with a more complicated runbook to make sure we have caught any perf or correctness issues. I would especially want to make sure we've tested `replace` adequately.
2) I am not sure I understand why encountering deleted nodes during pruning and throwing a `Transient` error is a bug. This seems like intended behavior and the transient error can be appropriately ignored. If we instead leave the deleted vectors in the graph until the end of deletion and take them into account during pruning of other nodes, we're just wasting work by computing distances to them anyway. Can you explain the reasoning here if I am wrong?

Thanks!

diskann-benchmark/src/backend/index/bftree/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+    writeln!(
+        output,
+        "{}",
+        crate::backend::index::streaming::stats::Summary::new(results.iter().map(|r| r.inner()))
+    )?;
+
+    Ok(results)
+}
+
+pub(crate) fn register_benchmarks(registry: &mut Registry) -> anyhow::Result<()> {
```
magdalendobson, line 27: Can we model the tag names based off https://github.com/microsoft/DiskANN/blob/68cc3c4ddf930f491e9cb4298837a342860d85bd/diskann-benchmark/src/backend/index/benchmarks.rs#L95-L111? It's a bit confusing that one is "dynamic" and the other is "stream"

diskann-benchmark/src/backend/index/bftree/mod.rs
JordanMaples: Nah, I'll remove these and just put them inline.

diskann-benchmark/src/backend/index/bftree/mod.rs
JordanMaples: a streaming_utils would work. I'll shuffle it over.

diskann-benchmark/src/backend/index/bftree/mod.rs
JordanMaples: I'll rename these all in a follow-up pr to align with each other

diskann-benchmark/src/backend/index/bftree/spherical.rs
JordanMaples: I've already largely refactored all of this out in a follow-up pr. I just didn't want to overwhelm this pr with changes.

diskann-benchmark/src/inputs/bftree.rs
JordanMaples: i can add that check.

JordanMaples:
> I'm not sure I fully understand the bug in deletion. Why do we need to keep the deleted vector around for pruning its neighbors? If it's in someone's neighbor list after deletion, it should be ignored during pruning. Is there some difference in how bf-tree handles this versus how in-mem providers handle this that meant deletes needed to be handled differently?

In-mem marks the vectors as deleted but keeps them around to be used during the inplace delete as the query point for finding replacement neighbors. bf_tree's design decision to hard delete would break this flow, which is why storing them temporarily is necessary. The bug I'm addressing was introduced during the bftree migration out of providers and into its own crate and we opted to go for the hard deletes.

diskann-benchmark/src/backend/index/bftree/full_precision.rs
```diff
@@ -0,0 +1,143 @@
+use std::{io::Write, sync::Arc};
```
hildebrandmw, line 5: Don't forget your header. We should add a CI lint for this.

diskann-bftree/src/provider.rs
```diff
@@ -196,6 +197,9 @@ where
     // Graph configuration parameters for persistence
     //
     pub(crate) graph_params: Option<GraphParams>,
+
+    // Nodes marked for deletion. Vector data is removed in `flush_deletes()`.
+    pending_deletes: Mutex<Vec<u32>>,
```
hildebrandmw: I'm still trying to understand this. I think we need to fix `inplace_delete` instead of hack around. I'm not sure how `inplace_delete` is even supposed to work at all since it calls `delete` and then immediately accesses the deleted element. For now, using the `onehop` or `onehop_and_twohop` should avoid fetching the element we just deleted.

TLDR: I propose we don't bend over to make soft-deletes work and instead use the inplace-delete methods that don't require soft deletes. Then we can work on getting rid of the whole concept of a soft delete.

diskann-bftree/src/neighbors.rs
```diff
@@ -208,21 +216,32 @@ impl<I: VectorId> NeighborProvider<I> {
...
-            let value = cast_slice::<I, u8>(&neighbor_list);
-            self.adjacency_list_index.insert(key, value);
+
+            // Build the value into the scratch buffer:
+            // |neighbor_0|...|neighbor_n|padding(Invalid)|...|nbr_count|
+            // Total size is always self.dim elements to avoid bf-tree page fragmentation.
+            let id_size = std::mem::size_of::<I>();
+            let total_len = self.dim * id_size;
+
+            let mut buf = self.buf.lock().unwrap();
```
hildebrandmw: When I was referring to scratch allocation, I meant passing in this scratch buffer as an argument. The various accessors can then allocate the buffer and pass it into these calls.

Having a central buffer behind a lock introduces a serialization bottleneck that isn't needed.

diskann-benchmark/src/inputs/graph_index.rs
```diff
@@ -662,6 +652,39 @@ impl IndexBuild {
 
         Ok(())
     }
+
+    pub(crate) fn data_type(&self) -> DataType {
+        self.data_type
+    }
```
hildebrandmw, line 658: I agree in principle with moving these behind accessors. However, the larger goal I think should be towards not interacting with inputs piecemeal and instead having inputs do coarser grained things, like return a whole `diskann::graph::Config`.

The reasoning is that these piecemeal type operations can very easily lead to benchmarks trying to reuse inputs from other benchmarks and just "forgetting" about arguments, leading them to accept inputs they don't honor.

I'm not really sure what the right solution is outside of engineering discipline.

diskann-benchmark/src/backend/index/streaming/stats.rs
```diff
@@ -20,8 +20,10 @@ pub(crate) enum StreamStats {
     Replace(BuildStats),
     Delete(GenericStats),
     Maintain {
-        drop_deleted: GenericStats,
-        release: GenericStats,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        drop_deleted: Option<GenericStats>,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        release: Option<GenericStats>,
```
hildebrandmw: Making these optional is not the move. It forces us into an awkward situation of "knowing" that "None" means bf-tree, but even in that case, the time spent in bf-tree's flushing isn't accounted for. Either make this an enum that is more explicit about the backend, or duplicate the struct to match what's more precisely needed.

diskann-benchmark/src/backend/index/bftree/spherical_streaming.rs
```diff
@@ -0,0 +1,291 @@
...
+    ))?);
+
+    // Train the spherical quantizer.
+    let m: diskann_vector::distance::Metric = input.build().distance().into();
+    let pre_scale = match input.pre_scale() {
+        Some(&v) => v.try_into()?,
+        None => diskann_quantization::spherical::PreScale::None,
+    };
+
+    let quantizer = diskann_quantization::spherical::SphericalQuantizer::train(
```
hildebrandmw, line 236: Maybe not necessary for this PR, but sequences like this are what I imagine we can use the inputs more directly for: Instead of pulling out all the configuration from the inputs, we can instead provide the input with raw data and let *it* construct the quantizer. This makes it much more difficult to forget about fields. Anyways, food for thought.

Doubly so that this is basically reimplemented in the static benchmarks.

hildebrandmw (commented):
Thanks - I think this is revealing issues with our `inplace_delete` API that need to get resolved. I'd say stick with hard deletes and avoid the inplace delete method that requires soft-deletes for now.

diskann-benchmark/src/backend/index/bftree/spherical_streaming.rs
JordanMaples: I can tackle this in the follow-up PR that I have planned to refactor the benchmarks

diskann-benchmark/src/backend/index/bftree/full_precision.rs
JordanMaples: I'll fix that

diskann-bftree/src/neighbors.rs
JordanMaples: ahh gotcha, i'll do that approach instead.

diskann-bftree/src/provider.rs
JordanMaples: we're doing hard deletes now and at least for the moment visited and top k will cause a crash. I'll chat offline to prevent this from happening in prod

diskann-bftree/src/provider.rs
JordanMaples: latest push has a fix for the delete problem.

diskann-benchmark/src/inputs/mod.rs
JordanMaples: We can fix this in a follow-up pr

diskann-benchmark/src/inputs/bftree.rs
JordanMaples: if they're not set by the user, the defaults are now reported in both the json and on the terminal

diskann-benchmark/src/backend/index/bftree/spherical.rs
JordanMaples: removed and replaced with runbook size + start point count

diskann-bftree/src/provider.rs
JordanMaples: latest has a proposed fix for inplace_delete that fixes the topk

diskann-benchmark/src/backend/index/streaming/stats.rs
JordanMaples: removed the optional nature of this. let me know what you think of the new approach

magdalendobson:
Could you add your new json examples as integration tests to https://github.com/microsoft/DiskANN/blob/main/diskann-benchmark/src/main.rs?

Now we have a runbook in our test data directory, we should have a norm that all example jsons should be wired up as integration tests.

magdalendobson:
> Could you add your new json examples as integration tests to https://github.com/microsoft/DiskANN/blob/main/diskann-benchmark/src/main.rs?
> 
> Now we have a runbook in our test data directory, we should have a norm that all example jsons should be wired up as integration tests.

And to add on to this, perhaps you could modify the README to mention adding example jsons as integration tests when you mention creating them?

magdalendobson:
Thank you for bearing with the weirdness of soft deletes and slot recycling, and for dealing with the learning curve of big-ann-benchmarks and runbooks! 

Now that the integration tests are present and you validated with a longer runbook, I am happy with this PR.

magdalendobson (approved):

diskann/src/graph/internal/prune.rs
```diff
@@ -22,6 +23,27 @@ pub(crate) struct Options {
     /// maximum degree by greedily adding skipped neighbors from the original candidate
     /// pool.
     pub(in crate::graph) force_saturate: bool,
+    /// Handling of transient errors encountered during vector retrieval.
+    pub(in crate::graph) transient_handling: TransientHandling,
```
hildebrandmw: Don't we already have a means of communicating transient failure with the transient error types? Why push down the logic and invent this new machinery when we can change the `escalate/allow_transient` calls higher up?

I'm pretty sure we can achieve this instead by simply changing the `escalate` call in `add_edge_and_prune` to an `allow_transient`.

diskann-bftree/src/neighbors.rs
```diff
@@ -208,21 +216,31 @@ impl<I: VectorId> NeighborProvider<I> {
...
+            let total_len = self.dim * id_size;
+
+            buf[..total_len].fill(0);
+
+            // Copy existing neighbors into the buffer
+            let neighbors_bytes = cast_slice::<I, u8>(&neighbor_list);
+            buf[..neighbors_bytes.len()].copy_from_slice(neighbors_bytes);
+
+            // Write the count at the last position
+            let count_id: I = I::from_usize(nbr_count).expect("Fails to cast usize as VectorId");
```
hildebrandmw: One thing to consider here is that since we're working with raw bytes, there's no *need* to store the count via `VectorId` round-trip. It could be stored as `u16` or `u32` directly and avoid this potentially fallible dance.

diskann-bftree/src/neighbors.rs
```diff
@@ -236,6 +254,77 @@ impl<I: VectorId> NeighborProvider<I> {
...
+    pub(crate) fn scratch(&self) -> NeighborScratch<'_, I> {
+        let buf_size = self.dim * std::mem::size_of::<I>();
+        NeighborScratch {
+            provider: self,
+            buf: vec![0u8; buf_size],
+        }
+    }
+}
+
+pub struct NeighborScratch<'a, I>
```
hildebrandmw: Small nit: The general vocabulary of this library would call this `NeighborAccessor`.

diskann-benchmark/README.md
```diff
@@ -505,3 +505,67 @@ error reporting in the event of a dispatch fail much easier for the user to unde
 
 Refer to implementations within the benchmarking framework for what some of this may look like.
 
+### Adding a Storage Provider
```
hildebrandmw: IMO, a checklist like this is pretty fragile long-term, particularly with respect to file-paths and names. I'd suggest paring it down to a minimum, or simply stating "refer to bftree" as a reference or something to help against bit-rot.

diskann-benchmark/src/inputs/bftree.rs
```diff
@@ -0,0 +1,708 @@
...
+
+    /// If true, only use the in-memory circular buffer (no disk pages).
+    #[serde(deserialize_with = "Deserialize::deserialize")]
+    pub(crate) cache_only: Option<bool>,
+}
+
+impl BfTreeStoreConfig {
+    /// Minimum circular buffer size in bytes found in bf-tree's example configs.
+    /// bf-tree's default cb_size_byte is 32MB.
+    const MIN_CB_SIZE_BYTE: usize = 8192;
```
hildebrandmw: Is there a way to retrieve this constant from the `bftree` dependency directly? Just to ensure it doesn't go out-of-date?

diskann-benchmark/src/inputs/bftree.rs
```diff
@@ -0,0 +1,708 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+use std::num::{NonZero, NonZeroUsize};
```
hildebrandmw, line 5: There is a lot of duplication in this file between the spherical and full-precision. I understand the need (basically: example printing :cry:). But in an ideal world, we could have like a `quantizer` enum that lived next to the build/streaming settings. Then in `Benchmark::try_match`, we can switch on the enum to report success or failure. It would be nice for that to be in this PR, but I don't know if it's worth it without better support for struct exploration from `diskann-benchmark-runner`.

hildebrandmw (commented):
Thanks Jordan - almost there. My biggest concern is the code added to `diskann`, which I think could be greatly simplified by just changing the ranked error call in `add_edge_and_prune` to an `allow_transient`.

diskann-benchmark/src/inputs/bftree.rs
JordanMaples: they don't expose this. Magdalen mentioned it but I didn't see anything in the docs.

Digging through their code, I found this:
```
        // Circular buffer size checks
        // In cache-only mode, the circular buffer size >= 4 * leaf_page_size
        // This is because during page split, two full sized mini-page need to be in memory
        // Note that 2 * leaf_page_size only guarantees one full sized mini-page due to AllocMeta
        //
        // In non cache-only mode, the circular buffer size >= 2 * leaf_page_size
        // As at most one full sized leaf page is required
```

So, something more sophisticated than a raw byte count should be used for validation

diskann-benchmark/src/inputs/bftree.rs
JordanMaples: we can maybe deal with some of this duplication in the follow up pr I have

diskann-benchmark/README.md
JordanMaples: removing changes to this file

diskann-bftree/src/neighbors.rs
JordanMaples: i'll change it

diskann/src/graph/internal/prune.rs
JordanMaples: rolling back to pre-this

diskann-benchmark/src/inputs/bftree.rs
JordanMaples: added

diskann/src/graph/internal/prune.rs
```diff
@@ -152,6 +153,21 @@ where
     }
 }
 
+impl<I> From<ListError<I>> for ANNError
```
hildebrandmw: Please remove this impl - `ListError` intentionally doesn't implement conversion to `ANNError` to ensure its uses go through `RankedError` and don't get mistakenly escalated.

diskann/src/graph/index.rs
```diff
@@ -1647,6 +1643,21 @@ where
                 .data_provider
                 .to_internal_id(context, id)
                 .escalate("id translation for `inplace_delete` must succeed")?;
+
+            // For VisitedAndTopK, we must capture the delete element *before* erasing
+            // the vector data, since it uses the deleted vector as a search query.
+            // This is especially necessary in hard-delete providers, such as the bf-tree.
+            // soft-delete providers, like in-mem don't mind the ordering as much.
```
hildebrandmw: Can we remove provider-specific mentions from comments? Mentioning hard vs soft deletes is sufficient.

hildebrandmw (commented):
Two quick comments, then we'll be good to go.

diskann/src/graph/index.rs
JordanMaples: kk removing the last line of the comment and everything after the common on the third line.

diskann/src/graph/internal/prune.rs
JordanMaples: will remove

hildebrandmw (approved):

PR #1109, opened 2026-05-27
[minmax] Add grid scaling to recompress API

arkrishn94:
The original implementation of the recompression API for minmax didn't contain support to scale the grid during recompression. This was fine since it was being used by consumers at 4-bitrate only; now there is demand to test this with 2-bits for which scaling down the grid is necessary to improve recall. 

Ran benchmarks to make sure results from benchmarking is identical to compressing from f32 directly. 

### 4-bit

| Pipeline | g=1.00 | g=0.90 | g=0.60 |
|---|---:|---:|---:|
| f32 → 4 (direct) | 0.9001 | 0.9084 | 0.8093 |
| 8 → 4 (recompress) | 0.9001 | 0.9061 | 0.8088 |

### 2-bit

| Pipeline | g=1.00 | g=0.90 | g=0.60 |
|---|---:|---:|---:|
| f32 → 2 (direct) | 0.5970 | 0.6370 | 0.6952 |
| 8 → 2 (recompress) | 0.5970 | 0.6375 | 0.6950 |


diskann-quantization/src/minmax/recompress.rs
```diff
@@ -217,20 +255,26 @@ mod recompress_tests {
...
+        MinMaxQuantizer: for<'a, 'b> CompressInto<&'a [f32], DataMutRef<'b, N>>,
         Recompressor: for<'a, 'b> CompressInto<DataRef<'a, N>, DataMutRef<'b, M>, Output = ()>,
     {
         let distribution = Uniform::new_inclusive::<f32, f32>(-1.0, 1.0).unwrap();
         let quantizer = MinMaxQuantizer::new(
             Transform::Null(NullTransform::new(NonZeroUsize::new(dim).unwrap())),
             Positive::new(1.0).unwrap(),
         );
-        let recompressor = Recompressor;
+        let recompressor = Recompressor::new(Positive::new(grid_scale).unwrap());
```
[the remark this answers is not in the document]
arkrishn94: There can be slight mismatches between the direct compression versus the recompression due to information loss from the first compression. So comparing this way doesn't hold up. The math, unit tests for random vectors and recall results should be enough evidence that the logic here is correct.

diskann-quantization/src/minmax/recompress.rs
```diff
@@ -352,7 +403,9 @@ mod recompress_tests {
             Transform::Null(NullTransform::new(NonZeroUsize::new(dim).unwrap())),
             Positive::new(1.0).unwrap(),
         );
-        let recompressor = Recompressor;
+
+        // SAFETY: I'm positive that 1.0 is positive.
+        let recompressor = Recompressor::new(unsafe { Positive::<f32>::new_unchecked(1.0) });
```
arrayka: Remove unsafe code from test. It's not perf sensitive.

diskann-quantization/src/minmax/recompress.rs
```diff
@@ -323,7 +373,8 @@ mod recompress_tests {
 
     #[test]
     fn test_dimension_mismatch_error() {
```
arrayka, line 380: See codecov report. Make the change coverage>90%.

diskann-quantization/src/minmax/recompress.rs
arrayka, line 219: is it possible that recompress produces float NaN or Inf? If so, is it a bug?

diskann-quantization/src/minmax/recompress.rs
arkrishn94: Weird. 3/6 lines it's showing are in the testing code.

diskann-quantization/src/minmax/recompress.rs
hildebrandmw: https://github.com/microsoft/DiskANN/blob/e2dc9a0a0a79ef21fd94ef1140e42d987d3eb1ee/diskann-quantization/src/num.rs#L52

diskann-quantization/src/minmax/recompress.rs
arkrishn94: If original is not NaN or Inf then this won't produce NaN or Inf. And original is designed to not produce NaN or Inf - or at least throw an error if it does.

fbv4_results.md
```diff
@@ -0,0 +1,33 @@
+## Recall@10 on enron-email-1M-fbv4 (1,087,932 × 384, fp16, IP, 1000 queries)
```
arrayka: Remove this .md file? or move to RFC folder or wiki.

diskann-quantization/src/minmax/recompress.rs
```diff
@@ -14,6 +15,13 @@ use thiserror::Error;
 /// This struct provides functionality to further compress MinMax quantized
 /// vectors from a source bitrate `N` to a target bitrate `M` for `N` > `M`.
 ///
+/// A positive `grid_scale` is used to (optionally) tighten or widen the
+/// reconstruction range that the target codes span before re-quantization.
+/// `grid_scale == 1.0` preserves the source reconstruction range exactly
+/// (legacy behavior). `grid_scale < 1.0` narrows the range, dedicating more
+/// resolution to interior values at the cost of clipping the extremes;
+/// `grid_scale > 1.0` widens the range (coarser steps, no clipping).
```
arrayka, line 23: It feels like we should throw an error if  `grid_scale > 1.0`, we should allow only (0;1] values.
Does passing grid_scale > 1 have any value?

arrayka (approved):

doliawu (approved):

diskann-quantization/src/minmax/recompress.rs
arkrishn94: It's not quite an error in my opinion. There's nothing wrong with using `grid_scale > 1.0`, it just won't give results as good as `grid_scale = 1.0`

diskann-quantization/src/minmax/recompress.rs
```diff
@@ -37,12 +45,30 @@ use thiserror::Error;
...
+/// let recompressor = Recompressor::new(Positive::new(1.0).unwrap());
 /// let mut encoded_4 = Data::<4>::new_boxed(4);
-/// Recompressor.compress_into(encoded_8.reborrow(), encoded_4.reborrow_mut()).unwrap();
+/// recompressor.compress_into(encoded_8.reborrow(), encoded_4.reborrow_mut()).unwrap();
 /// ```
 #[derive(Debug, Clone, Copy)]
-pub struct Recompressor;
+pub struct Recompressor {
+    grid_scale: Positive<f32>,
+}
```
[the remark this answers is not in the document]
arkrishn94: This is intentional. Will bump version soon

PR #1114, opened 2026-05-28
Ensure range results are sorted after the second round.

hildebrandmw:
When performing range expansion, nodes in the range are not added in order. This can lead to the final result of range search not being sorted by distance, which I think is a little surprising.

This PR fixes that by sorting the `in_range` buffer after the second search round. The out-of-order-ness can be seen in the old baseline, which this PR updates to the new sorted order.


magdalendobson:
When this code was originally written, leaving the results unsorted was intentional, for a couple of reasons. The first is that it just doesn't matter for correctness, and unlike top-k search we don't get the results in sorted order for free. Furthermore, the number of results within the range is not bounded, so the sort could potentially be very expensive. 

I am open to the argument that callers might expect the results to be in sorted order and be confused when they are not, but is that better to fix with documentation?

hildebrandmw:
Good to know. I didn't know if this was intentional or not. Since it is, I can change this PR to explicitly document that results are within the bounds, but not necessarily sorted.

harsha-simhadri (approved):

PR #1127, opened 2026-06-02
Add a small streaming runbook and groundtruth to test_data

magdalendobson:
Currently we don't have a way to benchmark streaming algorithms using the existing test data. This PR adds a streaming runbook and groundtruth for the 256-point slice of sift that already exists in `test_data`. It also updates the example dynamic index in `diskann-benchmark` to use these files, and to be able to run correctly. This will help the existing and future dynamic benchmarks stay in sync with any changes, and allow us to run small tests.


JordanMaples (approved):

diskann-benchmark/example/graph-index-dynamic.json
```diff
@@ -1,26 +1,26 @@
 {
   "search_directories": [
-    "../big-ann-benchmarks/data/MSTuringANNS",
-    "../big-ann-benchmarks/neurips23/runbooks"
+    "test_data/disk_index_search"
```
hildebrandmw, line 3: Care to wire this up to the [integration tests](https://github.com/microsoft/DiskANN/blob/62341f3c6ae557a8ee97b1286e352d7bb9b0b86d/diskann-benchmark/src/main.rs#L123) to prevent regression?

magdalendobson: Done! This also gave the opportunity to remove some repeated code :)

diskann-benchmark/src/main.rs
```diff
@@ -254,12 +254,10 @@ mod tests {
         }
     }
 
-    // Graph Index Build Integration Test.
-    #[test]
-    fn graph_index_integration() {
+    fn run_integration_test_from_json_path(json_string: &str) {
```
hildebrandmw: All callers of this seem to first construct a `serde_json::Value` from a file, then turn it into a string, then parse that string again. Couldn't we have this take `serde_json::Value` by-value and save an intermediate step? E.g.
```rust
fn run_integration_test(mut raw: serde_json::Value)
```

Also - I wouldn't be opposed to just calling this `run_integration_test`.

hildebrandmw (commented):
Love the code simplification! Just one minor comment to avoid redundant work in the tests.

diskann-benchmark/src/main.rs
magdalendobson: Ok, this should be resolved now

JordanMaples (approved):

hildebrandmw (approved):

PR #1128, opened 2026-06-02
[RFC] What filtered search algorithms should DiskANN support?

### LINKED ISSUE #1135: Post evaluation results and RFC with recommendations

magdalendobson:
This PR adds an RFC on which filtered search algorithms the DiskANN repo should support.


rfcs/01128-filtered-algorithms.md
```diff
@@ -0,0 +1,102 @@
+# Filtered Search Algorithms in DiskANN
+
+| | |
+|---|---|
+| **Authors** | Magdalen Manohar |
+| **Created** | 2026-06-02 |
+
```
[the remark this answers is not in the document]
magdalendobson: Bad suggestion

rfcs/01128-filtered-algorithms.md
```diff
@@ -0,0 +1,102 @@
...
+This RFC presents an empirical evaluation of the existing algorithms and makes recommendations to keep two algorithms and close/deprecate the other filtered search algorithms.
+
+### Overview of Existing Filtered Algorithms
+
+#### Inline Filtered Search
+
+Inline filtered search is a simple baseline which I introduced to sanity-check the other filtered search algorithms. It conducts a standard graph search with the only additional step of maintaining a separate queue of every predicate-satisfying element seen so far, and returning the closest $L_{search}$ predicate-satisfying elements at the end of the search. 
+
+The branch implementing inline filtered search is [here](https://github.com/microsoft/DiskANN/blob/users/magdalen/inline-filter/diskann/src/graph/search/inline_filter_search.rs).
```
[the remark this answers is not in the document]
magdalendobson: Fixed

rfcs/00000-filtered-algorithms.md
```diff
@@ -0,0 +1,102 @@
...
+
+The code for two-queue search is found in [this PR](https://github.com/microsoft/DiskANN/pull/929).
+
+##### Adaptive L Search (977)
+
+Adaptive L search runs a filtered search in the following way: for each query, it runs a standard search until the search has performed 1000 distance computations. Then, it computes what fraction of the points seen so far satisfy the filter predicate, and scales the L_search parameter up accordingly. See these lines for the exact scaling parameters. It only performs the adaptive scaling at one point during the search, so L_search is capped at 16 times the original value.
```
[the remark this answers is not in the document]
magdalendobson: Fixed

harsha-simhadri:
Could this be in the wiki? The plots are a quite large addition to the repo.

magdalendobson:
> Could this be in the wiki? The plots are a quite large addition to the repo.

Do we want an RFC that points to benchmarks in the wiki then? Because folks were pretty clear that they wanted an RFC on filtered search.

magdalendobson:
> > Could this be in the wiki? The plots are a quite large addition to the repo.
> 
> Do we want an RFC that points to benchmarks in the wiki then? Because folks were pretty clear that they wanted an RFC on filtered search.

I have moved the benchmark discussion section to the DiskANN wiki: https://github.com/microsoft/DiskANN/wiki/Evaluation-of-Filtered-Search-Algorithms

rfcs/01128-filtered-algorithms.md
```diff
@@ -0,0 +1,68 @@
...
+
+The branch implementing inline filtered search is [here](https://github.com/microsoft/DiskANN/blob/users/magdalen/inline-filter/diskann/src/graph/search/inline_filter_search.rs). 
+
+#### Beta Search
+
+Beta search is conceptually very simple. It sets a value $\beta \in (0,1]$, and for a point $p$ encountered during a graph search that satisfies the query filter, the raw distance between the query and $p$ is multiplied by $\beta$. Thus the search is biased towards points which satisfy the filter.
+
+The code for beta search is found [here](https://github.com/microsoft/DiskANN/blob/main/diskann-providers/src/model/graph/provider/layers/betafilter.rs).
+
+#### Multi-hop Search
```
arrayka: Is it two-hop or multi-hop? How many extra expansions are allowed for a non-matching node? Is it hardcoded to one?

rfcs/01128-filtered-algorithms.md
```diff
@@ -0,0 +1,68 @@
...
+
+The code for multi-hop search is found [here](https://github.com/microsoft/DiskANN/blob/main/diskann/src/graph/search/multihop_search.rs).
+
+#### Two-Queue Search (PR #929)
+
+Two-queue search maintains a queue of neighbors satisfying the filter predicate (size k*p), where p is a multiplicative factor set by the user, and a separate, unbounded size queue of the best neighbors found so far, regardless of predicate. The search proceeds as normal with the larger queue, adding any results satisfying the predicate to the filtered queue. The search terminates for one of four reasons: (1) when the closest unexplored node in the regular queue is further away from the query than the furthest node in the filter-satisfying queue, (2) when no candidates remain to visit, (3) the number of hops exceeds a user-set maximum, or (4) the QueryVisitDecision returns a termination.
```
arrayka: Consider making (4) more clear:
- (4) The filter callback explicitly asked the search to stop via QueryVisitDecision::Terminate.

rfcs/01128-filtered-algorithms.md
```diff
@@ -0,0 +1,68 @@
...
+
+The code for two-queue search is found in [this PR](https://github.com/microsoft/DiskANN/pull/929).
+
+#### Adaptive L Search (PR #977)
+
+Adaptive L search runs a filtered search in the following way: for each query, it runs a standard search until the search has performed 1000 distance computations. Then, it computes what fraction of the points seen so far satisfy the filter predicate, and scales the L_search parameter up accordingly. See [these lines](https://github.com/microsoft/DiskANN/pull/977/changes#diff-0ed5dd0ab0fa4906e3aa6e0c77d6b381f2a364b4d64df85d81224f609104388eR274-R285) for the exact scaling parameters. It only performs the adaptive scaling at one point during the search, so L_search is capped at 16 times the original value.
```
arrayka: 1. Does L_search expansion happens only once?
2. If "match_rate ≥ 50%   → 1× L (no change, most nodes match", does it mean that the search is done? It feels like this threshold should be configurable to control trade-off between latency and precision.

rfcs/01128-filtered-algorithms.md
```diff
@@ -0,0 +1,68 @@
...
+## Benchmark Results
+
+To avoid adding large files to the main repo, the presentation and discussion of benchmark results is contained in a [DiskANN Wiki page](https://github.com/microsoft/DiskANN/wiki/Evaluation-of-Filtered-Search-Algorithms).
+
+## Proposal
+
+Based on the benchmarking results and their analysis, I propose the following actions:
+1. Move inline filtering to the main repo as a new filtered search algorithm, with the adaptive-L subroutine an option that can be enabled.
+2. Deprecate beta-filtered search.
+3. Retain multi-hop filtered search.
```
arrayka, line 64: 1. Can we define clear decision-making logic for clients to choose the right algorithm based on the target specificity mode (high, medium, or low)?
2. Doesn’t inline brute-force become competitive in high-specificity mode?
3. Should we suggest clients use query planner to switch between algorithms based on estimated filter specificity?

rfcs/01128-filtered-algorithms.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Goals
+
+The goal is to align on at most two filtered search algorithms to remain in the main branch of the DiskANN repository, based on performance evaluation of current candidates.
+
+## Benchmark Results
+
+To avoid adding large files to the main repo, the presentation and discussion of benchmark results is contained in a [DiskANN Wiki page](https://github.com/microsoft/DiskANN/wiki/Evaluation-of-Filtered-Search-Algorithms).
+
+## Proposal
```
arrayka, line 59: Do we assume that generating a bitmap (or evaluating is_match(vector): bool) is negligible compared to the cost of graph traversal?
What if the cost of generating a bitmap for the whole index is significant, or the cost of is_match(vector): bool is comparable to — or even higher than — the cost of distance computation? In that case, do we need alternative algorithms?

rfcs/01128-filtered-algorithms.md
```diff
@@ -0,0 +1,68 @@
...
+
+Based on the benchmarking results and their analysis, I propose the following actions:
+1. Move inline filtering to the main repo as a new filtered search algorithm, with the adaptive-L subroutine an option that can be enabled.
+2. Deprecate beta-filtered search.
+3. Retain multi-hop filtered search.
+4. Close the PR with two-queue search.
```
arrayka: Does close mean "abandon"?

rfcs/01128-filtered-algorithms.md
magdalendobson: 1) Yes, it happens only once
2) It doesn't necessarily mean that the search is done if the match rate is greater than 50%. It just means that the search will continue with the same `L_search` until the normal termination criteria is met. Regarding the threshold for increasing `L_search`, whatever you set as the initial `L_search` value will let you control the tradeoff of accuracy and precision pretty well already. Once the initial inline search PR lands, I can experiment with modifying the scaling threshold and add it as a configurable option if it appears that you can get meaningfully better results by configuring it, if that sounds good to you.

rfcs/01128-filtered-algorithms.md
magdalendobson: 1) I want to leave decisions of query planning (in the sense of defaulting to an inverted index, or brute-force search) out of this RFC and focus only on algorithms that assume the query will use a graph index. I can add some guidelines for which algorithm to choose
2) I'm not sure what you mean by "inline brute force." Do you mean using the inline search algorithm without adaptive L? If so, I would agree that's the best algorithm to use when specificity is high (ie high match rate)
3) I think this is mostly addressed in my answer to (1), but I don't suggest that the query planner dispatch to different graph algorithms.

rfcs/01128-filtered-algorithms.md
magdalendobson: Yes

rfcs/01128-filtered-algorithms.md
magdalendobson: Regarding your first question, we don't assume that it's negligible. Every search algorithm evaluated here, with the exception of multi-hop search, performs one filter evaluation for each distance comparison. Unless you want to just do a regular search and post-filter, that is probably the best you can do in terms of minimizing filter evaluations--or at least we don't have an algorithm that is better.

Multi-hop search allows you to trade off more filter evaluations for fewer distance comparisons. This is one of the main reasons we're retaining it in addition to inline search--in scenarios like multi-vector search, it may be really important to be able to make this tradeoff.

rfcs/01128-filtered-algorithms.md
magdalendobson: Right now it's hardcoded to one. I named it multi-hop to leave open the possibility of additional extra expansions.

rfcs/01128-filtered-algorithms.md
```diff
@@ -0,0 +1,68 @@
...
+
+The code for adaptive L search was originally contributed in [this PR](https://github.com/microsoft/DiskANN/pull/977). [This branch](https://github.com/microsoft/DiskANN/tree/users/magdalen/two-queue-adaptive-l) integrates it into benchmark and keeps up-to-date with the main branch.
+
+### Goals
+
+The goal is to align on at most two filtered search algorithms to remain in the main branch of the DiskANN repository, based on performance evaluation of current candidates.
+
+## Benchmark Results
+
+To avoid adding large files to the main repo, the presentation and discussion of benchmark results is contained in a [DiskANN Wiki page](https://github.com/microsoft/DiskANN/wiki/Evaluation-of-Filtered-Search-Algorithms).
```
wuw92, line 57: Curious about the latency side: distance comparisons and hops are good proxies when distance compute dominates, but that depends on the provider. On a disk-backed provider a single vector or neighbor fetch can easily outweigh a distance compare, and  `is_match` may itself be a non-trivial cost (e.g., inverted-index lookup vs. inline bit check).

Would it be worth a rough cost model like:

`T_query ≈ n_dist · C_dist + n_vec · C_vec + n_neighbor · C_neighbor + n_match · C_match`

Each algorithm has a different `(n_*)` signature; each provider sets the `(C_*)`. Then we can see when the ranking flips as one term starts dominating, and make the recommendation explicit about which provider regime it's optimizing for.

magdalendobson: I added some rough rules of thumb about the comparative numbers of match checks, neighbor fetches, and distance computations (which will be the same number as vector fetches) for multi-hop search compared to other forms of search for the same recall. I am hesitant to make an explicit formula since we only have two datasets and I don't think that is enough data points for an explicit prediction for multi-hop (the other algorithms will follow the same cost model as regular graph search, with one match evaluation per distance computation).

arrayka (approved):
LGTM. Approving, assuming the QPS and latency comparison between Adaptive L and the beta filter for disk search scenarios looks promising.

rfcs/01128-filtered-algorithms.md
```diff
@@ -0,0 +1,80 @@
...
+
+Adaptive L search runs a filtered search in the following way: for each query, it runs a standard search until the search has performed 1000 distance computations. Then, it computes what fraction of the points seen so far satisfy the filter predicate, and scales the L_search parameter up accordingly. See [these lines](https://github.com/microsoft/DiskANN/pull/977/changes#diff-0ed5dd0ab0fa4906e3aa6e0c77d6b381f2a364b4d64df85d81224f609104388eR274-R285) for the exact scaling parameters. It only performs the adaptive scaling at one point during the search, so L_search is capped at 16 times the original value. In the future PR which integrates Adaptive L search into main, we will make the number of samples and the maximum scaling factor configurable parameters. Whether to also allow the specificity cutoffs for scaling to be configurable is deferred to future experiments.
+
+The code for adaptive L search was originally contributed in [this PR](https://github.com/microsoft/DiskANN/pull/977). [This branch](https://github.com/microsoft/DiskANN/tree/users/magdalen/two-queue-adaptive-l) integrates it into benchmark and keeps up-to-date with the main branch.
```
sampathrg: This link is broken. Did you mean to point to [this](https://github.com/microsoft/DiskANN/tree/users/magdalen/inline-with-adaptive-l/) branch?

magdalendobson: Thanks! You're right, fixed

rfcs/01128-filtered-algorithms.md
```diff
@@ -0,0 +1,80 @@
...
+3. Retain multi-hop filtered search.
+4. Abandon the PR with two-queue search.
+
+## Advice for Users
+
+Next we provide advice for library users to choose a filtered algorithm based on their specific scenario and knowledge of their dataset's characteristics.
+
+1. Use inline filtered search when (a) you know that your query set has high specificity, or (b) if you have any other reason to want to control the $L_{search}$ parameter directly.
+2. Use the adaptive L feature of inline search when you have a query set with varied specificity across queries and you do not wish to configure the $L_{search}$ individually across queries, or you do not know the specificity of your queries.
+3. Use multi-hop filtering if you wish to trade off more predicate evaluations for fewer distance comparisons. This may be especially relevant for large vectors or for expensive distance functions such as Chamfer distance.
```
suri-kumkaran: Should we also call out that the multi hop algorithm reads more graph nodes, specifically more adjacency lists? This may be important for workloads where adjacency list reads are expensive, such as disk based graph search.

rfcs/01128-filtered-algorithms.md
```diff
@@ -0,0 +1,80 @@
...
+
+The code for adaptive L search was originally contributed in [this PR](https://github.com/microsoft/DiskANN/pull/977). [This branch](https://github.com/microsoft/DiskANN/tree/users/magdalen/two-queue-adaptive-l) integrates it into benchmark and keeps up-to-date with the main branch.
+
+### Goals
+
+The goal is to align on at most two filtered search algorithms to remain in the main branch of the DiskANN repository, based on performance evaluation of current candidates.
+
+## Benchmark Results
+
+To avoid adding large files to the main repo, the presentation and discussion of benchmark results is contained in a [DiskANN Wiki page](https://github.com/microsoft/DiskANN/wiki/Evaluation-of-Filtered-Search-Algorithms).
```
suri-kumkaran, line 57: Can we add a few more details about the benchmarking setup, such as the recall@k values used (for example, k = 10 or 100)? It would also be helpful to mention that the benchmarks were run using in memory graph providers, so graph access costs were not dominated by storage I/O.

rfcs/01128-filtered-algorithms.md
```diff
@@ -0,0 +1,80 @@
...
+
+The code for adaptive L search was originally contributed in [this PR](https://github.com/microsoft/DiskANN/pull/977). [This branch](https://github.com/microsoft/DiskANN/tree/users/magdalen/two-queue-adaptive-l) integrates it into benchmark and keeps up-to-date with the main branch.
+
+### Goals
+
+The goal is to align on at most two filtered search algorithms to remain in the main branch of the DiskANN repository, based on performance evaluation of current candidates.
+
+## Benchmark Results
```
suri-kumkaran, line 55: Should we also consider using a flat scan for high specificity queries? Many of our use cases are bitmap based, where we already know the filter match set, which could make a flat scan a viable option.

rfcs/01128-filtered-algorithms.md
```diff
@@ -0,0 +1,80 @@
...
+
+This RFC presents an empirical evaluation of the existing algorithms and makes recommendations to keep two algorithms and close/deprecate the other filtered search algorithms.
+
+### Overview of Existing Filtered Algorithms
+
+In this section we provide an overview of existing filtered algorithms. Of particular note is that beta search, inline filtered search, two-queue search, and adaptive L search all perform one predicate evaluation per distance computation. Multi-hop search performs more predicate evaluations than distance comparisons, so it may be a good choice when distance computations are expensive and predicate evaluations are cheap. Currently we do not have any algorithms that perform *fewer* predicate evaluations than distance computations, aside from a naive post-filtering. 
+
+#### Inline Filtered Search
+
+Inline filtered search is a simple baseline which I introduced to sanity-check the other filtered search algorithms. It conducts a standard graph search with the only additional step of maintaining a separate queue of every predicate-satisfying element seen so far, and returning the closest $L_{search}$ predicate-satisfying elements at the end of the search.
```
suri-kumkaran, line 23: This is similar to a paged search that keeps fetching pages until $K$ predicate-satisfying elements are found, but strictly better: post-filtering a paged search only sees the top-by-distance window and discards matches the traversal walked past, requiring deeper (more expensive, more approximate) paging to recover them. Inline filtered search instead keeps every match it visits in a single pass. The two converge when the predicate is unselective, but inline filtered search is the stronger baseline as selectivity increases.

rfcs/01128-filtered-algorithms.md
magdalendobson: Under the section on advice for users, I mention use of a query planner for cases like this, but evaluating it or making specific recommendations is outside the scope of this RFC.

rfcs/01128-filtered-algorithms.md
magdalendobson: You're right, within our team and within the [draft PR](https://github.com/microsoft/DiskANN/pull/1131) on adaptive search, we've been discussing that this could be an alternative for some existing use cases of paged search.

rfcs/01128-filtered-algorithms.md
magdalendobson: I have updated the Wiki post with a link to the json config files used for each experiment, and also mention the in-memory setup in the section on benchmarks.

rfcs/01128-filtered-algorithms.md
magdalendobson: I changed the language to read "more predicate evaluations and more adjacency list fetches"

rfcs/01128-filtered-algorithms.md
suri-kumkaran: Given that these algorithms are quite similar and share some common components, we should consider which pieces can be exposed at a more granular level. Different clients could then use those pieces through providers, with or without their own query planners, depending on their scenarios.

I understand this is out of scope for this PR. Is there another RFC or PR where this is being discussed?

rfcs/01128-filtered-algorithms.md
suri-kumkaran: Nit: I think it would be useful to mention recall 10@20 somewhere in the wiki.

suri-kumkaran (approved):
Thanks for the great work. The RFC recommendations around deprecating beta filter, adding inline filtered search, and adding adaptive L search make sense to me.

One follow up implementation question is how we should expose these similar and overlapping algorithms at the right level of granularity, so clients can reuse them through providers, with or without their own query planners, depending on their use cases.

rfcs/01128-filtered-algorithms.md
magdalendobson: Thanks, I made this and the other edits to the wiki page that you requested

randybird (approved):

rfcs/01128-filtered-algorithms.md
magdalendobson: There isn't another RFC which discusses this at the moment. There is a PR adding the [filtered accessor](https://github.com/microsoft/DiskANN/pull/1141) and the PR adding [inline search with adaptive L](https://github.com/microsoft/DiskANN/pull/1131). Agree that there are some questions about how different pieces compose that we will need to answer soon--e.g. can adaptive L replace paged search in some cases, and should inline search really be its own codepath as opposed to just a special case of normal search.

PR #1130, opened 2026-06-03
[diskann-garnet] Fix handling of missing quant vectors during delete()

metajack:
Deleting an ID from the index removes several different keys associated with a vector: the internal and external ID mappings, the vector and quant vector data, the neighbor data, and the attributes.

For each of these associated keys, we returned an error from `delete()` whenever the key was not found, but for attributes and quant vectors, these errors are benign since it is normal for those keys to be missing. It is cheaper to attempt to delete a key than to check whether it exists and then delete it if it does.

A corresponding test for delete is added, although other tests would have also caught this had the `Store` test storage utility not unconditionally trued true for deletes. It now better models Garnet's behavior where deletes to non-existent keys returns false.

### DIFF
--- a/diskann-garnet/src/provider.rs
+++ b/diskann-garnet/src/provider.rs
@@ -730,13 +730,18 @@ impl<T: VectorRepr> Delete for GarnetProvider<T> {
         let mut ok = true;
         ok &= self.callbacks.delete_iid(&context.term(Term::ExtMap), id);
         ok &= self.callbacks.delete_eid(&context.term(Term::IntMap), gid);
-        ok &= self
+
+        // It is not an error to fail deleting attributes; they may not exist.
+        let _: bool = self
             .callbacks
             .delete_eid(&context.term(Term::Attributes), gid);
+
         // NOTE: Commented out until DiskANN fixes accessing neighbor data post-delete.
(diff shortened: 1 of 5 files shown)


diskann-garnet/src/provider.rs
```diff
@@ -733,10 +733,13 @@ impl<T: VectorRepr> Delete for GarnetProvider<T> {
...
             .callbacks
             .delete_eid(&context.term(Term::Attributes), gid);
+
         // NOTE: Commented out until DiskANN fixes accessing neighbor data post-delete.
         //ok &= self.callbacks.delete_iid(context.term(Term::Neighbors), id);
         ok &= self.callbacks.delete_iid(&context.term(Term::Vector), id);
-        ok &= self
+
+        // It is not an error to fail deleting quantized terms; they may not exist yet.
+        let _ = self
```
hildebrandmw: one small suggestion: if you type the return value like
```rust
let _: bool = self.callbacks.delete_iid(...);
```
then you will get a compiler error in the future if the return type for `delete_iid` ever changes for some reason and we need to revisit the "ignore the return value" decision.

metajack: Good suggestion. Fixed.

hildebrandmw (approved):

JordanMaples (approved):

PR #1131, opened 2026-06-03
Inline filtered search with adaptive L

### LINKED ISSUE #1136: Merge new filtered search algorithms to main

magdalendobson:
This PR implements the recommendation in the [filtered search RFC](https://github.com/microsoft/DiskANN/pull/1128) to implement inline filtering with the adaptive L method as an optional addition.


diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,425 @@
...
+    fn from(err: AdaptiveLSearchError) -> Self {
+        Self::new(ANNErrorKind::IndexError, err)
+    }
+}
+
+/// Adaptive L for greedy filtered search.
+#[derive(Debug, Clone)]
+pub struct AdaptiveL {
+    pub sample_count: usize,
+    pub scale_factor: f64,
```
hildebrandmw: Usual feedback applies: making these fields public invalidates the invariants established by the constructor. I.E., please make these private.

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,425 @@
...
+}
+
+impl AdaptiveL {
+    /// Create a new adaptive L.
+    pub fn new(sample_count: usize, scale_factor: f64) -> Result<Self, AdaptiveLSearchError> {
+        if scale_factor < 1.0 {
+            return Err(AdaptiveLSearchError::ScaleFactorLessThanOne);
+        }
```
hildebrandmw, line 55: Is there a practical upper-bound on the scale factor?

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,425 @@
...
+
+/// Parameters for pure greedy filtered search.
+///
+/// All nodes participate in greedy navigation regardless of filter match.
+/// Matched results are tracked separately and returned as final output.
+/// Early termination is controlled by the `QueryLabelProvider` callback.
+#[derive(Debug)]
+pub struct InlineSearch<'q, InternalId> {
+    /// Base graph search parameters.
+    pub inner: Knn,
```
hildebrandmw, line 84: The implementation doesn't actually use `Knn::k_value()`. If that is intentional, can we not use `Knn` and instead just pull out the relevant fields that are actually used?

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,425 @@
...
+    /// Label evaluator for determining node matches and early termination.
+    pub label_evaluator: &'q dyn QueryLabelProvider<InternalId>,
+    /// Adaptive L for the search.
+    pub adaptive_l: Option<AdaptiveL>,
```
hildebrandmw, line 88: Thoughts on making this a full `Algo` enum instead of an `Option` to give us room to expand in the future?

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,425 @@
...
+                )
+                .await
+                .into_ann_result()?;
+
+            Ok(stats.finish(result_count as u32))
+        }
+    }
+}
+
+/// Inline filtered search: a standard k-NN search with an additional step
```
hildebrandmw: Can this doc-comment go on the `InlineSearch` struct instead? This is the documentation that will actually show up in rust-doc, versus this which is on an internal method.

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,425 @@
...
+
+    // Helper to build the final stats from scratch state.
+    let make_stats = |scratch: &SearchScratch<I>| InternalSearchStats {
+        cmps: scratch.cmps,
+        hops: scratch.hops,
+        range_search_second_round: false,
+    };
+
+    // Initialize search state if not already initialized.
+    // This allows paged search to call this function multiple times
```
hildebrandmw: This isn't being used in paged search? Shouldn't this always be empty?

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,425 @@
...
+            let decision = query_label_evaluator.on_visit(neighbor);
+
+            match decision {
+                QueryVisitDecision::Accept(accepted) => {
+                    // All nodes go into scratch.best for navigation,
+                    // matched nodes also go into matched_results for final output.
+                    scratch.best.insert(neighbor);
+                    matched_results.insert(accepted);
+                    if adaptive_l.is_some() {
+                        sample_matched += 1;
```
hildebrandmw: micro-opt: it's probably faster to unconditionally increment `sample_matched`. While it's only useful if `adaptive_l.is_some()`, checking `adaptive_l` every time introduces an extra branch.

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,425 @@
...
+            );
+            if new_l > l_search {
+                scratch.resize(new_l);
+            }
+        }
+    }
+
+    // Replace scratch.best with only the matched results
+    // so that post_process returns the right candidates.
+    scratch.best = matched_results;
```
hildebrandmw: Whoa ... can we instead return `matched_results` directly from this function?

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,425 @@
...
+///   specificity ≥ 50%   → 1× L (no change, most nodes match)
+///   10% ≤ specificity < 50%  → 2× L
+///   specificity < 10%   → log-scale: 2^(1 - log10(specificity))
+///     specificity = 0.1  (10%)   → 2× L
+///     specificity = 0.01 (1%)    → 4× L
+///     specificity = 0.001 (0.1%) → 8× L
+///   0 matches in sample → `max_multiplier`× L (maximum expansion)
+///
+/// Clamped to [1×, max_multiplier] range.
+fn compute_adaptive_l(base_l: usize, visited: u32, matched: u32, max_multiplier: f64) -> usize {
```
hildebrandmw: Why are these `u32` instead of `usize`, out of curiosity.

diskann-benchmark/src/inputs/graph_index.rs
```diff
@@ -244,6 +244,63 @@ impl MultiHopSearchPhase {
...
+pub(crate) struct AdaptiveL {
+    pub(crate) sample_count: NonZeroUsize,
+    pub(crate) scale_factor: f64,
+}
+
+impl AdaptiveL {
+    pub(crate) fn validate(&self, _checker: &mut Checker) -> Result<(), anyhow::Error> {
+        if self.scale_factor <= 0.0 {
+            return Err(anyhow::anyhow!("scale_factor must be positive"));
+        }
```
hildebrandmw: Can you instead validate by constructing the actual `AdaptiveL` struct and using its error? That save replication of work here and eliminates the posibility of drift.

diskann-benchmark/src/inputs/graph_index.rs
```diff
@@ -244,6 +244,63 @@ impl MultiHopSearchPhase {
...
+#[derive(Debug, Serialize, Deserialize)]
+pub(crate) struct InlineSearchPhase {
+    pub(crate) queries: InputFile,
+    pub(crate) query_predicates: InputFile,
+    pub(crate) groundtruth: InputFile,
+    pub(crate) reps: NonZeroUsize,
+    pub(crate) data_labels: InputFile,
+    pub(crate) num_threads: Vec<NonZeroUsize>,
+    pub(crate) runs: Vec<GraphSearch>,
+    pub(crate) adaptive_l: Option<AdaptiveL>,
```
hildebrandmw, line 271: Can you use a `#[serde(deserialize_with = Deserialize::deserialize)]` here to force this to be "null" in the corresponding JSON rather than defaulting to `None` when the field is ommitted?

hildebrandmw (commented):
Thanks Magdalen! In addition to my inline comments - can you also add some [integration tests](https://github.com/microsoft/DiskANN/blob/v0.53.0/diskann/src/graph/test/cases/mod.rs) exercising the functionality here? These go a surprisingly long way towards protecting the algorithm.

Also, can we bikeshed `InlineSearch` a little? Maybe `FilteredSearch`? Or `InlineFilteredSearch`? Not a huge deal, but `InlineSearch` seems a little opaque to me.

diskann-benchmark-core/src/search/graph/inline.rs
```diff
@@ -0,0 +1,138 @@
...
+                buffer,
+            )
+            .await?;
+
+        Ok(super::knn::Metrics {
+            comparisons: stats.cmps,
+            hops: stats.hops,
+        })
+    }
+}
```
hildebrandmw, line 138: Small request: can you add some tests here like the other tests in `diskann-benchmark-core`?

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Hmm it depends on what `L_search` is set to so not really. I mean I guess 10000 would be too high, but we don't bound `L_search` this way so I think it would be better to not cap the user.

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: The reason for leaving it as `Knn` is that while `k` isn't explicitly used in the algorithm, it's being used to enforce that `L` is not allowed to be less than `k`, which is still desirable here. Otherwise we'd be leaving it to the caller to perform that check, which might be confusing since we perform it for the caller in other cases.

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Not a huge proponent of this idea since hypothetical future variants might want to mix and match with adaptive L rather than plug and play. Seems like we should just leave this to future work if needed.

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Multi-hop search does this also even though it isn't used in paged search. I thought the custom was to write the algorithm this way so it can be used with paged search in the future. Should we instead remove it from both algorithms?

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Wouldn't this make it so you have to write a custom post-processor every time you use inline filtering? Is there a big downside to replacing `scratch.best`?

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Honestly I just took that function from the original PR and didn't think much about it. I changed them to `usize`, you're right it makes more sense.

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Done

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Done

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Resolved

diskann-benchmark/src/inputs/graph_index.rs
magdalendobson: Done

diskann-benchmark/src/inputs/graph_index.rs
magdalendobson: Done

diskann/src/graph/search/inline_filter_search.rs
hildebrandmw: I don't understand - why would we care that `k` is not less than `L` is `k` is not being used? The caller can already pass an output buffer with less than `k` entries, so nothing is actually being enforced.

diskann/src/graph/search/inline_filter_search.rs
hildebrandmw: There's nothing keeping us from keeping the `AdaptiveL` struct and using that in the enum like
```rust
pub enum Kind {
    Standard,
    AdaptiveL(AdaptiveL),
}
```

diskann/src/graph/search/inline_filter_search.rs
hildebrandmw: Yeah, we should remove it from both algorithms until it's actually needed by paged search. Otherwise, it's just confusing.

diskann/src/graph/search/inline_filter_search.rs
hildebrandmw: Post process takes an iterator, not a scratch. So, you can return `matched_results` and pass an iterator of those contents to post-process.

The downside to using `scratch.best` is mainly that it constantly churns the allocation in scratch.

magdalendobson: Resolved

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: In our side channel conversation, we discovered that k_value() is almost never used and we should just refactor the struct for all the algorithms. Deferred to future PR to notify folks working on diverse search.

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Marking resolved.

magdalendobson:
> Thanks Magdalen! In addition to my inline comments - can you also add some [integration tests](https://github.com/microsoft/DiskANN/blob/v0.53.0/diskann/src/graph/test/cases/mod.rs) exercising the functionality here? These go a surprisingly long way towards protecting the algorithm.
> 
> Also, can we bikeshed `InlineSearch` a little? Maybe `FilteredSearch`? Or `InlineFilteredSearch`? Not a huge deal, but `InlineSearch` seems a little opaque to me.

I've added some integration test. In addition to throwing some of the existing cases in multihop at it, I also designed a test that should return different results depending on whether or not the adaptive L feature is enabled.

Regarding the naming, I named it `InlineSearch` to be consistent with `MultihopSearch` (ie, multihop search isn't called `MultihopFilteredSearch` either). Should we maybe rename both to mention filtering?

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,414 @@
...
+/// scaled up in the following manner:
+///   specificity ≥ 50%    → 1× L (no change, most nodes match)
+///   10% ≤ specificity < 50% → 2× L
+///   specificity < 10%    → log-scale: 2^(-log10(specificity))
+///     specificity = 10%  (100/1000) → 2× L
+///     specificity = 1%   (10/1000)  → 4× L
+///     specificity = 0.1% (1/1000)   → 8× L
+///   and so on up to a pre-set maximum multipler
+#[derive(Debug)]
+pub struct InlineSearch<'q, InternalId> {
```
harsha-simhadri: not a blocker for this PR, but how does this compose with range search and diverse search functionalities?

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,414 @@
...
+    scratch: &mut SearchScratch<I>,
+    search_record: &mut SR,
+    query_label_evaluator: &dyn QueryLabelProvider<I>,
+    adaptive_l: Option<AdaptiveL>,
+) -> ANNResult<(InternalSearchStats, Vec<Neighbor<I>>)>
+where
+    I: VectorId,
+    A: ExpandBeam<T, Id = I> + SearchExt,
+    SR: SearchRecord<I> + ?Sized,
+{
```
harsha-simhadri, line 177: would we want to allow a paginated API For this function?

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: From an algorithmic perspective, inline filtering (without any adaptation) will compose seamlessly with any other type of search, because it's just adding the extra step of checking match and keeping track of matched elements. However, from the perspective of our codebase as written right now, it would require new function signatures that accept a `LabelProvider`. 

Thinking about how adaptive L in particular would compose with range search, we could use it within the initial graph search before deciding whether to move on to the unbounded part of range search. Similarly we could compose it with diverse search to increase `L_search` when few matching candidates are found.

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Hmm, that's a good question. Using pagination *and* adaptive L seems like a strange thing to do. Adapting `L_search` is morally pretty similar to pagination, since it extends the search for longer when few matched results are found. If there was a user who wanted to use pure inline filtering and control length of the search via pagination only, on the other hand that seems reasonable.

diskann/src/graph/search/multihop_search.rs
```diff
@@ -188,16 +188,12 @@ where
...
-                scratch.visited.insert(id);
-                scratch.best.insert(Neighbor::new(id, distance));
-            })
-            .await?;
-    }
+    accessor
+        .start_point_distances(|id, distance| {
+            scratch.visited.insert(id);
+            scratch.best.insert(Neighbor::new(id, distance));
+        })
```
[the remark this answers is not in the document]
magdalendobson: Bad suggestion, we only need to check emptiness if a paginated API is present for the search implementation

diskann/src/graph/search/inline_filter_search.rs
harsha-simhadri: The user might be in a position where the first search did not yield enough results that pass filters? In such a case, is our advise to retry with larger L?

diskann/src/graph/search/inline_filter_search.rs
harsha-simhadri: ok, could you please document these to-dos as issues

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: This would be the advice if the user is sure that the number of results they are requesting actually exist.

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Done: https://github.com/microsoft/DiskANN/issues/1139

harsha-simhadri (approved):

diskann/src/graph/search/inline_filter_search.rs
hildebrandmw: Still in support of exposing as an enum here.

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,404 @@
...
+
+            match decision {
+                QueryVisitDecision::Accept(accepted) => {
+                    // All nodes go into scratch.best for navigation,
+                    // matched nodes also go into matched_results for final output.
+                    scratch.best.insert(accepted);
+                    matched_results.push(accepted);
+                    sample_matched += 1;
+
+                    sample_matched += 1;
```
hildebrandmw: Did this happen twice?

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,404 @@
...
+/// scaled up in the following manner:
+///   specificity ≥ 50%    → 1× L (no change, most nodes match)
+///   10% ≤ specificity < 50% → 2× L
+///   specificity < 10%    → log-scale: 2^(-log10(specificity))
+///     specificity = 10%  (100/1000) → 2× L
+///     specificity = 1%   (10/1000)  → 4× L
+///     specificity = 0.1% (1/1000)   → 8× L
+///   and so on up to a pre-set maximum multiplier
+#[derive(Debug)]
+pub struct InlineSearch<'q, InternalId> {
```
hildebrandmw: `InlineSearch` is really quite a vague name. Can we call it like `InlineFilteredSearch` or something a little more descriptive?

diskann/src/graph/test/cases/inline.rs
```diff
@@ -0,0 +1,326 @@
...
+        search_output_buffer,
+        test::provider as test_provider,
+        test::synthetic::Grid,
+    },
+    test::tokio::current_thread_runtime,
+};
+
+use super::multihop::{BlockAndAdjust, EvenFilter, build_1d_provider, setup_grid_index};
+
+// Topology (3 levels below the start):
```
hildebrandmw, line 58: These tests still use the older style "looks kind of okay" approach to tests rather than using the more rigorous baseline tests that our other algorithms have moved to. This makes it significantly harder to refactor with confidence and also kind of forces us into a regime where the search sizes are pretty small.

For example, none of these integration test trigger the low-match regime of the adaptive L algorithm, so we aren't really protecting the behavior there.

Adding baselines should be relatively straightforward and greatly improves the quality of algorithm tests.

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,404 @@
...
+        cmps: scratch.cmps,
+        hops: scratch.hops,
+        range_search_second_round: false,
+    };
+
+    accessor
+        .start_point_distances(|id, distance| {
+            scratch.visited.insert(id);
+            scratch.best.insert(Neighbor::new(id, distance));
+        })
```
hildebrandmw, line 204: What happens if start points don't match the predicate? Or do match the predicate?

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,404 @@
...
+        .await?;
+
+    // Pre-allocate with good capacity to avoid repeated allocations
+    let mut one_hop_neighbors = Vec::with_capacity(max_degree_with_slack);
+
+    // Matched results tracked separately — scratch.best contains all nodes
+    // for greedy navigation, matched_results contains only filter-matching nodes.
+    let mut matched_results = Vec::new();
```
hildebrandmw: Should this be a `NeighborPriorityQueue`, or at least some other data structure that puts an upper-bound on its size?

diskann-benchmark/src/inputs/graph_index.rs
```diff
@@ -244,6 +244,70 @@ impl MultiHopSearchPhase {
...
+#[derive(Debug, Serialize, Deserialize)]
+pub(crate) struct InlineSearchPhase {
+    pub(crate) queries: InputFile,
+    pub(crate) query_predicates: InputFile,
+    pub(crate) groundtruth: InputFile,
+    pub(crate) reps: NonZeroUsize,
+    pub(crate) data_labels: InputFile,
+    pub(crate) num_threads: Vec<NonZeroUsize>,
+    pub(crate) runs: Vec<GraphSearch>,
+    #[serde(deserialize_with = "require_present_option")]
```
hildebrandmw: Use just `#[serde(deserialize_with = "Deserialize::deserialize"]` is sufficient.

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -0,0 +1,404 @@
...
+///     specificity = 0.1% (1/1000)   → 8× L
+///   and so on up to a pre-set maximum multiplier
+#[derive(Debug)]
+pub struct InlineSearch<'q, InternalId> {
+    /// Base graph search parameters.
+    pub inner: Knn,
+    /// Label evaluator for determining node matches and early termination.
+    pub label_evaluator: &'q dyn QueryLabelProvider<InternalId>,
+    /// Adaptive L for the search.
+    pub adaptive_l: Option<AdaptiveL>,
```
hildebrandmw, line 88: The other comment may have gotten lost, so I'm going to plug once more that I think this should be an enum from the get go.

hildebrandmw (commented):
Thanks Magdalen, my main concern here is getting in decent test coverage using the baseline testing infrastructure. It's more robust than ad-hoc metrics since it can store and validate a lot more information.

Also to bike shed, I *do* think `InlineSearch` is a bit vague. Maybe `InlineFilteredSearch`? Or `FilteredKnn`?

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: I'm not sure I agree. The main reason I disagree is that morally what is going on here is that we have inline filtering with an optional addition. If that changes later on, we can do an enum, but I think this is most accurate to my understanding of the algorithm.

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: I got a 7-10% gain in QPS from pushing results to a vector and sorting once at the end instead of using a `NeighborPriorityQueue`, which is why I chose not to use it. If you have alternative ideas I am happy to experiment with them!

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Done. I may follow up and rename MultihopSearch to MultihopFilteredSearch. It seems confusing to have "filter" in the name for one filtered algorithm and not another.

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Yes, bad Copilot

diskann/src/graph/test/cases/inline.rs
magdalendobson: I added tests with baselines, including a test with specificity as low as .1%. Do they look good to you now?

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: I modified this code to check whether start points match, since I realize per our previous conversation that they are  sometimes allowed to be returned as results. We plan to get rid of the termination option soon, so I am not bothering to address it for start points here.

diskann/src/graph/search/inline_filter_search.rs
hildebrandmw: Agree do disagree I guess :smile:

diskann/src/graph/test/cases/inline.rs
hildebrandmw: If you comment out the unit tests for the adaptive-L, you will find the low-specificity tests that were added still never trigger the code in `compute_daptive_l` outside of the initial preamble (e.g., `if matched == 0 || visited == 0`). From what I can tell, even though a filter with a low selection percent are used, the filter is selecting for low IDs. Since the query is at [10, 10, 10] and the grid is constructed so lower coordinates have lower IDs, this actually means the selection criteria is crossed before any matches are actually found and it's always using the max multiplier.

You can test this yourself by running
```
cargo llvm-cov nextest --html --package diskann --cargo-profile ci
```
and opening `./target/llvm-cov/html/index.html` and navigating to `inline_filter_search.rs`.

To actually test the algorithm, the initial exploration needs to see enough matched nodes before the decision point that the piecewise heuristic is actually triggered.

diskann/src/graph/search/inline_filter_search.rs
hildebrandmw: Marking as resolved.

diskann/src/graph/test/cases/inline.rs
```diff
@@ -0,0 +1,626 @@
...
+    l: usize,
+    no_adapt_result_count: usize,
+    no_adapt_cmps: usize,
+    no_adapt_hops: usize,
+    no_adapt_ids: Vec<u32>,
+    adapt_result_count: usize,
+    adapt_cmps: usize,
+    adapt_hops: usize,
+    adapt_ids: Vec<u32>,
+}
```
hildebrandmw: In addition to my other comment about not triggering the piecewise functionality of adaptive-L, I think the tests here could use one more cleanup pass. Here is my suggestion:

1. If we want to leave ourselves open to expanding in the future, using one baseline per flavor (one for fixed-L, one for adaptive-L) instead of merging into a single struct makes this much easier. Additionally, `run_inline_on_grid` could then return an `InlineBaseline` directly instead of a mega-tuple.

   It's okay to go nuts with generating multiple baselines per test.

2. Always include IDs and distances in baselines. This is very cheap extra protection. Basically, the more you throw in there, the better.

3. Probably worth also using a baseline for the three-level tests as that's cheap and captures strictly more information.

hildebrandmw (approved):

diskann/src/graph/test/cases/inline.rs
hildebrandmw: Resolved.

diskann/src/graph/search/inline_filter_search.rs
hildebrandmw: Good to know - thanks!

diskann/src/graph/test/cases/inline.rs
hildebrandmw: Resolved :smile:

PR #1134, opened 2026-06-05
[multi-vector] Detect L1/L2 cache sizes for matrix kernels

### LINKED ISSUE #1062: Add support for reading L1/L2 cache sizes for different platforms(Windows, Linux, MacOS) for our efficient cache aware multi-vector distance functions.
We might be able to take some reference from - https://github.com/ciresnave/hardware-query/blob/main/src/cpu.rs

We have the type `TileBudget` type which we need to wire by reading cache sizes from the machine. 

Think more on how we can do minimal system calls and reuse the cache sizes across all distance computations.

wuw92:

#### Reference Issues/PRs

Builds on the matrix-kernel design introduced in #1368.

Detect L1d/L2 capacities at runtime for matrix-kernel tile budgets, memoized once per process. Keep the existing 75% L1d / 50% L2 budget policy and the 48,000 B / 1,250,000 B fallback estimates.

Adapted from [gemm-common 0.18.2](https://github.com/sarah-quinones/gemm/blob/a9d8ed36e51ba039ffe8cb2512925202612e19fd/gemm-common/src/cache.rs), retaining its probe order and shared-cache normalization:

| Platform | Probe order |
|---|---|
| Linux | sysfs, then `lscpu`, then CPUID where supported |
| Apple platforms | `hw.perflevel0.*` sysctl, then CPUID where supported |
| Other supported x86 targets, including Windows | CPUID |
| Unsupported targets or no usable probe result | Existing DiskANN fallback estimates |

The local adaptation avoids bringing in gemm-common's `pulp` and other support dependencies for non-linalg consumers.

##### Differences from gemm-common

- Retain only L1d/L2 detection and the metadata it needs; omit L3, GEMM-specific tuning, and the unused Apple `physicalcpu` query.
- Require both capacities to be nonzero, trying the next source on incomplete results instead of filling defaults during probing. Invalid `lscpu` numeric values fail the probe rather than panic.
- Use standard-library `OnceLock` and direct `libc` sysctl calls instead of `once_cell` and `sysctl`, keeping dependencies minimal.
- Include `physical_line_partitions()` in Intel's capacity formula, as specified by [Intel SDM Table 21-14](https://cdrdv2-public.intel.com/922477/253665-092-sdm-vol-1.pdf#page=460); upstream omits this factor.

#### Any other comments?

Verified successful cache detection in CI, without fallback. Observed values before tile-budget scaling:

| Runner | Architecture | L1d (bytes) | L2 (bytes) | Test log |
|---|---|---:|---:|---|
| `windows-latest` | x86-64 | 32,768 | 524,288 | [Passed](https://github.com/microsoft/DiskANN/actions/runs/35484552321/job/106008664396) |
| `ubuntu-latest` | x86-64 | 16,384 | 262,144 | [Passed](https://github.com/microsoft/DiskANN/actions/runs/35484552321/job/106008664354) |
| `ubuntu-24.04-arm` | ARM64 | 65,536 | 1,048,576 | [Passed](https://github.com/microsoft/DiskANN/actions/runs/35484552321/job/106008664337) |


diskann-quantization/src/matrix_kernels/cache.rs
```diff
@@ -0,0 +1,390 @@
...
+                for cache in params {
+                    if !matches!(cache.cache_type(), CacheType::Data | CacheType::Unified) {
+                        continue;
+                    }
+                    let size = match cache.level() {
+                        1 => &mut l1d,
+                        2 => &mut l2,
+                        _ => continue,
+                    };
+                    *size = cache.associativity()
```
hildebrandmw: I think this needs to be divided by `cache.max_cores_for_cache()`?

hildebrandmw (commented):
Two high-level comments. 

First, it would be great to include an auxiliary binary (like how `diskann-inmem` [exposes some internals](https://github.com/microsoft/DiskANN/blob/a6825ad12c21b710452c352b0d61cad09281cb89/diskann-inmem/Cargo.toml#L46-L49) for testing) that prints out the detected cache sizes. This would make it a lot easier to debug on various platforms.

Second, we need to be careful with respect to hyper threading. Currently, the `sysfs` path will assume that all hyper-threads will divide the physical L1/L2 cache sizes by the number of hyper threads. The CPUID path does not do this correction. However, for matrix kernels, it is probably best to assume that a well written kernel will saturate the execution units with a single logical core and thus hyper-threading should probably not be used. So for this PR, it might be best to report the smallest physical cache instance capacity (to tackle heterogeneous systems) and report the number of logical sharers. On highly heterogeneous Apple systems, this is overly conservative. But without thread pinning ... I'm not sure how much better we'd be able to do better. Also, CPUID only works for the current CPU, so smallest physical cache size enumeration won't work reliably. I say we just pick whatever CPUID happens to throw at us and revisit some ability to change the decision at run time.

PR #1138, opened 2026-06-06
Consolidate insert into `PruneAccessor`

hildebrandmw:
The insert followup to #1067, consolidating the following traits:
* `diskann::provider::{DelegateNeighbor, AsNeighbor, AsNeighborMut}`: Neighbor accessor delegation in this manner is no longer needed.
* `diskann::provider::{HasElementRef, BuildDistanceComputer}`: Consolidated into `PruneAccessor`.
* `diskann::graph::workingset::{Fill, AsWorkingSet}`: Consolidated into `PruneAccessor`

Into a single `PruneAccessor` trait:
```rust
pub trait PruneAccessor: HasId + Send + Sync {
    type Neighbors<'a>: provider::NeighborAccessorMut<Id = Self::Id>
    where
        Self: 'a;

    type ElementRef<'a>;

    type View<'a>: for<'x> workingset::View<Self::Id, ElementRef<'x> = Self::ElementRef<'x>>
        + Send
        + Sync
    where
        Self: 'a;

    type Distance<'a>: for<'x, 'y> DistanceFunction<Self::ElementRef<'x>, Self::ElementRef<'y>, f32>
        + Send
        + Sync
    where
        Self: 'a;

    /// Replaces neighbor delegation
    fn neighbors(&mut self) -> Self::Neighbors<'_>;

    /// Replaces `workingset::Fill`.
    fn fill<Itr>(
        &mut self,
        itr: Itr,
    ) -> impl SendFuture<ANNResult<(Self::View<'_>, Self::Distance<'_>)>>
    where
        Itr: ExactSizeIterator<Item = Self::Id> + Clone + Send + Sync;
```
For insert/prune, the accessor has two jobs:
1. Neighbor access for manipulating datasets. The new trait uses a `neighbors` method for delegation mainly because it's a pattern we use quite heavily already: implement one `NeighborAccessor` and have all the other accessors use that.
2. Gathering item for pruning and enabling distance computation. This is done via `fill`, which now returns both a `View` and the distance computer. I think this should still remain as a method with a view because it scopes the duration for which pruning is being performed, allowing implementation to acquire locks or otherwise enter some critical region for a duration shorter than the lifetime of the whole accessor.

## Interaction with the Working Set
The working set is used to cache entries across prunes to reduce trips to the backing provider. With this change, the working set and distance computers become part of the accessor, which adds a *little* bloat to the accessors, but simplifies how these objects get along. 

Without an external working-set, `PruneStrategy` becomes much simpler (now it's like `SearchStrategy` with just two associated types and a single method. `MultiInsertStrategy` is slightly changed with `finish` still returning an opaque `Seed` but with the `PruneAccessor` being constructed via `seeded_prune_accessor`. This is the replacement for the `AsWorkingSet` transformation.

## Suggested Review Order
* `diskann/src/graph/glue.rs`: The new `PruneAccessor` trait
* `diskann/src/provider.rs`: Simplification to provider-related traits.
* `diskann/src/graph/workingset/`: Simplification to workingset API.
* `diskann/src/graph/index.rs`: Updated indexing algorithm to take a `PruneAccessor` rather than a strategy + provider + context.
* `diskann-providers/`: Mostly mechanical clean-up.
* `diskann-bftree/`: Mostly mechanical clean-up.
* `diskann-garnet/`: Mostly mechanical, with one simplification. Since the workingset is now part of the accessor, we can no longer run into situations where the working set contains full-precision vectors but the accessor wants quantized vectors, resulting in some simplification.

### DIFF
--- a/diskann/src/provider.rs
+++ b/diskann/src/provider.rs
@@ -55,40 +55,10 @@
 //!
 //!   The [`HasId`] trait provides a common base-trait for these related concepts, which
 //!   ensures implementers only need to define and constrain it once.
-//!
-//! * [`BuildDistanceComputer`]: A sub-trait of [`HasElementRef`] that allows for
-//!   random-access distance computations on the retrieved elements.
-//!
-//! # Neighbor Delegation
-//!
-//! Graph operations require both data access (distances, elements) and graph
-//! access (neighbor retrieval and mutation). These concerns are typically served
-//! by different backing stores and need not vary together.
-//!
-//! For example, search and pruning use different data accessors (optimized for
-//! query distances vs. pairwise element distances), but both share the same
-//! graph topology.
-//!
-//! Rather than requiring every data accessor to manually forward
-//! [`NeighborAccessor`] methods, the [`DelegateNeighbor`] trait routes neighbor
-//! access to a single shared implementation.
-//!
-//! Data accessor types should implement [`DelegateNeighbor`] to return a
-//! [`NeighborAccessor`]. This automatically provides [`AsNeighbor`] and
-//! [`AsNeighborMut`] (the latter when the delegate implements
-//! [`NeighborAccessorMut`]).
-//!
-//! Algorithms requiring graph access should accept `&mut T` where
-//! `T: AsNeighbor` or `T: AsNeighborMut`, which provides blanket
-//! implementations of [`NeighborAccessor`] and [`NeighborAccessorMut`]
-//! for `&mut T`.
 
 use std::ops::Deref;
 
-use diskann_vector::DistanceFunction;
-use sealed::{BoundTo, Sealed};
-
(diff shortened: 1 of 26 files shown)


harsha-simhadri (approved):

JordanMaples (approved):
lgtm

diskann/src/provider.rs
```diff
@@ -366,45 +336,9 @@ where
...
-        + Sync;
-
-    /// Build the random-access distance computer for this accessor.
-    ///
-    /// This method is expected to be relatively cheap to invoke and implementations are
-    /// encouraged to make this method infallible.
-    fn build_distance_computer(
-        &self,
-    ) -> Result<Self::DistanceComputer, Self::DistanceComputerError>;
-}
```
arkrishn94, line 403: Nice!

arkrishn94 (approved):
Love the consolidation here! Makes the build side traits so much easier to understand.

PR #1141, opened 2026-06-09
Add a `FilteredAccessor` for filtered search.

### LINKED ISSUE #1175: Overhaul the DataProvider APIs around filtering
Remove QueryLabelProvider and implement FilterAccessor

hildebrandmw:
Introduce a `FilteredAccessor` as a filter aware version of `SearchAccessor`.

## Motivation
Our current hook for performing filtered search relies on passing a `QueryLabelProvider` externally to all call sites. This has the following disadvantages:

* The caller must somehow know exactly the ID space of the provider.
* The `QueryLabelProvider` is hard-coded to use the accessor's internal ID, which may not always be the best fit depending on context.
* `QueryLabelProvider` evaluations always live behind a trait object and are just queried one at a time.
* The search accessor has no knowledge of the filter.
* Splitting the label check with distance computations opens a time-of-check, time-of-use window that can be tricky to resolve.

## How
This PR boils down to the following enums and trait:
```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
#[repr(transparent)]
pub struct Accept<T>(T);

#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
#[repr(transparent)]
pub struct Reject<T>(T);

#[derive(Debug, Clone, Copy)]
pub enum Decision<T> {
    Accept(Accept<T>),
    Reject(Reject<T>),
}

pub trait FilteredAccessor: HasId + Send + Sync {
    fn start_point_distances<F>(
        &mut self,
        f: F,
    ) -> impl std::future::Future<Output = ANNResult<()>> + Send
    where
        F: FnMut(Decision<Self::Id>, f32) + Send;

    fn expand_beam_filtered<Itr, P, F>(
        &mut self,
        ids: Itr,
        pred: P,
        on_neighbors: F,
    ) -> impl std::future::Future<Output = ANNResult<()>> + Send
    where
        Itr: Iterator<Item = Self::Id> + Send,
        P: HybridPredicate<Self::Id> + Send + Sync,
        F: FnMut(Decision<Self::Id>, f32) + Send;

    fn expand_beam_accept_only<Itr, P, F>(
        &mut self,
        ids: Itr,
        pred: P,
        on_neighbors: F,
    ) -> impl std::future::Future<Output = ANNResult<()>> + Send
    where
        Itr: Iterator<Item = Self::Id> + Send,
        P: HybridPredicate<Accept<Self::Id>> + Send + Sync,
        F: FnMut(Accept<Self::Id>, f32) + Send;

    fn terminate_early(&mut self) -> bool {
        false
    }

    fn num_starting_points(&self) -> impl std::future::Future<Output = ANNResult<usize>> + Send {
        // provided
    }
}
```
The existing multi-hop search is rewritten to use `FilteredAccessor` and #1131 can be easily rewritten this way as well. This makes the `FilteredAccessor` "filter-aware" and allows the stated problems to be solved (the TOCTOU potential is really pushed down to the accessor, but that's the layer that knows the concurrency story so it's in a much better place to solve it anyways).

To reduce code churn, I added a new module `diskann::graph::ext::labeled` that is the new home for `QueryLabelProvider` and introduces a wrapping `FilteredAccessor` and associated strategy. The functionality in this module is intended as a compatibility layer for code that is currently relying on `QueryLabelProvider`.

## Two new `expand_*` methods

Unfortunately, the semantics of filtered search for `MultiHopFilteredSearch` basically requires the use of two different `expand_*` methods. This is because the accept-only path *really* needs only accepted items to be passed to `pred.eval_mut`. This PR models that by bespoke `Accept`/`Reject` types and because of this nuance, two methods seemed like the most robust way of expressing this.

## Bug Fixes in MultiHop
With this change, it was clear that multi-hop search was slightly buggy with respect to start point handling: start points were treated as always fulfilling the predicate. This PR goes ahead and fixes that bug by tracking rejected start points explicitly and filtering them out in the post-process call.

## What's up with `QueryVisitDecision`
This PR gets rid of `QueryVisitDecision`. It is conflating a large number of unrelated decisions.
* It was almost never used properly: no code outside of the core `diskann` ever bothered with it, leading to semantic issues with, for example, beta-filtered search never calling `on_visit`.
* Multi-hop search wasn't even using it right when adding items from `two_hop_neighbors`. To actually use it right would require a redundant interrogation of the `QueryLabelProvider`.
* The decision to terminate is weird - that decision belongs with the accessor.
* Tweaking distance is sketchy and not clear if it's intended to influence navigation or simply be recorded and passed to post process. Upcoming ID refactors will help restore this functionality *and* make it clear what the intent is.

Dropping this extra functionality means we can ditch the multi-hop tests specifically testing this functionality.

## Suggested Review Order
* Start with the first commit. Since it seemed cleanest to still use `SearchStrategy`, this commit relaxes the `SearchAccessor` trait bound on `SearchStrategy::SearchAccessor` to allow the disjoint `FilteredAccessor` to join the party. This commit is the one responsible for most of the churn in random parts of the code.
* `diskann/src/graph/glue.rs`: The new `FilteredAccessor` and associated enums.
* `diskann/src/graph/search/multihop_search.rs`: Refactored multi-hop search to use the new trait.
* `diskann/src/graph/ext/labeled.rs`: The adaptor layer that allows code currently written for `QueryLabelProvider` to keep working.
* `diskann/test/generated/graph/test/cases/multihop/callback_filtering_grid.json` was deleted since it's corresponding test was deleted. We no longer support distance tweaking.


diskann-benchmark-core/src/search/graph/multihop.rs
```diff
@@ -62,7 +62,7 @@ where
         index: Arc<graph::DiskANNIndex<DP>>,
         queries: Arc<Matrix<T>>,
         strategy: Strategy<S>,
-        labels: Arc<[Arc<dyn graph::index::QueryLabelProvider<DP::InternalId>>]>,
+        labels: Arc<[Arc<dyn labeled::QueryLabelProvider<DP::InternalId>>]>,
```
suri-kumkaran, line 65: We are still using `QueryLabelProvider` with `DP::InternalId`, are we planning to remove it in a follow-up PR?

hildebrandmw: I wasn't planning on making this change. This PR allows filtered search implementations without using `QueryLabelProvider` whatsoever. A decent chunk of internal code assumes internal ID on the `QueryLabelProvider` (though I'm hoping in the near future all notion of an "external" vs "internal" ID will go away). The intent is to enable providers like the disk index to hook into filtering as they see fit without resorting to trait gymnastics on the `QueryLabelProvider`.

diskann/src/graph/glue.rs
```diff
@@ -245,7 +247,188 @@ pub trait SearchAccessor: HasId + Send + Sync {
...
+            let set: std::collections::HashSet<_> = self
+                .starting_points()
+                .await?
+                .into_iter()
+                .map(|v| v.into_inner())
+                .collect();
+
+            Ok(move |id| !set.contains(&id))
+        }
+    }
```
magdalendobson: I vote we get rid of is_not_start_point. It seems very wrong to have this be a method on `FilteredAccessor`, or really a method of any accessor.

diskann/src/graph/glue.rs
```diff
@@ -245,7 +247,188 @@ pub trait SearchAccessor: HasId + Send + Sync {
...
+
+    /// Return the number of starting points.
+    fn num_starting_points(&self) -> impl std::future::Future<Output = ANNResult<usize>> + Send {
+        self.starting_points()
+            .map(|result: ANNResult<_>| result.map(|v: Vec<_>| v.len()))
+    }
```
magdalendobson: Similarly, do we really need all these special functions on starting points? They don't seem to really save many lines of code, and they're not any special optimizations either. Could we consider getting rid of `num_starting_points` and maybe even `start_point_distances`?

diskann/src/graph/glue.rs
```diff
@@ -245,7 +247,188 @@ pub trait SearchAccessor: HasId + Send + Sync {
...
+/// Filter decision for [`FilteredAccessor`].
+///
+/// Both variants carry the item `T` since rejected items are useful for graph navigation.
+#[derive(Debug, Clone, Copy)]
+pub enum Decision<T> {
+    /// The item satisfies the filter criteria.
+    Accept(T),
+    /// The item does not satisfy the filter criteria.
+    Reject(T),
+}
```
magdalendobson, line 323: Just thinking this through--one thing folks in MSRI have discussed experimenting with is algorithms that use "partial" matches (e.g. maybe one label in an AND expression matches but the other doesn't). This could be modified in the future to return some kind of match score without breaking too much down the line, right?

hildebrandmw: mmmmmm. Do you know what that would look like from an algorithmic side?

diskann/src/graph/glue.rs
hildebrandmw: I'm fine getting rid of it.

diskann/src/graph/glue.rs
hildebrandmw: Perhaps. One thing to consider is that implementing these as provided definitions doesn't really cost much. With hooking into `start_points_distances`, there is a bit of a chicken and egg problem: knowing the number of start points helps when sizing the neighbor priority queue, which is then pushed into from `start_point_distances`. I'm in favor of paring these down, but perhaps as a different PR?

diskann/src/graph/ext/labeled.rs
```diff
@@ -0,0 +1,489 @@
...
+}
+
+impl<P, I> glue::PredicateMut<I> for Wrapper<'_, P, I>
+where
+    P: glue::PredicateMut<I>,
+    I: VectorId,
+{
+    fn eval_mut(&mut self, item: &I) -> bool {
+        self.inner.eval_mut(item) && self.labels.is_match(*item)
+    }
```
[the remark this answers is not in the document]
hildebrandmw: Not correct. Items basically get one chance to be "in" or "out". Mutating the inner predicate means that we can avoid evaluating the actual filter unless absolutely necessary.

diskann/src/graph/glue.rs
```diff
@@ -822,3 +1006,43 @@ where
...
+        assert!(a.is_accept());
+        assert_eq!(a.into_inner(), 6);
+        let r = Decision::Reject(3).map(|x| x * 2);
+        assert!(r.is_reject());
+        assert_eq!(r.into_inner(), 6);
+
+        // as_ref borrows without consuming
+        let d = Decision::Accept(vec![1, 2, 3]);
+        assert_eq!(d.as_ref().into_inner(), &[1, 2, 3]);
+        drop(d); // original still valid
```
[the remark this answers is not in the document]
hildebrandmw: Thanks.

diskann/src/graph/glue.rs
magdalendobson: This would look like making traversal decisions (whether to add something to `scratch.best`) based on partial match information rather than binary yes/no.

diskann/src/graph/glue.rs
magdalendobson: Fair enough, maybe we could not support `is_not_start_point` in this new accessor since that one is really quite extra.

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -155,50 +146,63 @@ where
...
     let beam_width = search_params.beam_width().get();
     let l_search = search_params.l_value().get();
 
-    // Helper to build the final stats from scratch state.
-    let make_stats = |scratch: &SearchScratch<I>| InternalSearchStats {
-        cmps: scratch.cmps,
-        hops: scratch.hops,
-        range_search_second_round: false,
-    };
+    // // Helper to build the final stats from scratch state.
```
harsha-simhadri: remove commented portion

harsha-simhadri (approved):

diskann/src/graph/ext/labeled.rs
```diff
@@ -0,0 +1,473 @@
...
+}
+
+impl<A> HasId for FilteredAccessor<'_, A>
+where
+    A: HasId,
+{
+    type Id = A::Id;
+}
+
+impl<'a, A> glue::FilteredAccessor for FilteredAccessor<'a, A>
```
arrayka, line 162: impl glue::FilteredAccessor for FilteredAccessor is confusing. 
What about naming the struct `LabelFilteredAccessor` or `QueryLabelFilteredAccessor`?

diskann/src/graph/glue.rs
arrayka: If the only purpose of `starting_points():Vec<Decision<T>>` is to provide Vec.len(), then the current default implementation feels suboptimal. Calling num_starting_points() ends up allocating a vector on the heap, running filters for each starting point, and then discarding the resulting Decisions.

In practice, most clients should be able to return the count of starting points directly, without additional Vec allocation or filter evaluation.

arrayka (approved):

diskann/src/graph/ext/labeled.rs
```diff
@@ -0,0 +1,473 @@
...
+        ids: Itr,
+        pred: P,
+        mut on_neighbors: F,
+    ) -> ANNResult<()>
+    where
+        Itr: Iterator<Item = Self::Id> + Send,
+        P: glue::HybridPredicate<Self::Id> + Send + Sync,
+        F: FnMut(Decision<Self::Id>, f32) + Send,
+    {
+        match kind {
```
metajack: I'm having trouble following this.

In the ExpansionKind::All case, we call decide (which calls is_match()). The the AcceptOnly case we wrap the predicate and then unconditional return Accept on everything. This appears to be because Wrapper folds the is_match into the eval.

I think this deserves a comment to lay out how the filter matching is happening. It's a little bit obtuse due to each branch handling it differently and with the wrapper types.

metajack (approved):

diskann-benchmark-core/src/search/graph/multihop.rs
suri-kumkaran: Sorry, I missed that. I thought this was in a general location, but it is actually in benchmark, so we can clean it up separately.

diskann/src/graph/glue.rs
hildebrandmw: Agreed. This is why the `num_starting_points` method exist to skip the unnecessary allocation. I vote we keep the `starting_points` method for now, but we can simplify down to just `starting_point_distances` and `num_starting_points` in a bit.

diskann/src/graph/glue.rs
hildebrandmw: Coming back to this, I think there is a channel to do this in the future. Would you be open to leaving the decision as binary for now and relaxing it in the future when the idea of how to use fuzzy matching will be used?

diskann/src/graph/ext/labeled.rs
hildebrandmw: Thanks Jack - I realized that this is way more obtuse than I originally thought (and I got it wrong originally as well). Because of these nuances, I've split the two traversal modes into two independent methods, with the `AcceptOnly` side taking an `Accept` struct directly where needed.

diskann/src/graph/ext/labeled.rs
hildebrandmw: It's a bit confusing, but I find there are often naming issues that arise like this. The `FilteredAccessor`'s main job is to implement the `FilteredAccessor` trait. It is *in* the `labeled` module and not reexported, so can be used as `labeled::FilteredAccessor`. The problem I have with adding prefixes is that you end up with something like `labeled::LabelFilterAccessor`, but where do you stop? You could to `graph::labeled::LabelGraphFilterAccessor` etc.

I'm slowly settling on a pattern where I make the module path part of the identity of the type, but recognize I could go over-board on this sometimes. If you feel strongly about it, I'm happy to change.

diskann/src/graph/search/inline_filter_search.rs
hildebrandmw: Good catch - removed.

diskann/src/graph/glue.rs
hildebrandmw: Know what, we might as well start this off on the right foot. I removed `num_starting_points`  and it's better for it. Thanks!

diskann/src/graph/ext/labeled.rs
```diff
@@ -0,0 +1,481 @@
...
+    }
+}
+
+impl<P, I> glue::PredicateMut<I> for Wrapper<'_, P, I>
+where
+    P: glue::PredicateMut<Accept<I>>,
+    I: VectorId,
+{
+    fn eval_mut(&mut self, item: &I) -> bool {
+        // Oreder must be `label` -> `inner` because we have to know an ID is accepted before
```
magdalendobson: Sorry to be picky but small typo here

arkrishn94 (approved):

diskann/src/graph/ext/labeled.rs
```diff
@@ -0,0 +1,481 @@
...
+        Next: glue::SearchPostProcess<Self::NextAccessor, T, O> + Sync,
+    {
+        next.post_process(&mut accessor.inner, query, candidates, output)
+    }
+}
+
+/// A [`glue::HybridPredicate`] that uses an additional [`QueryLabelProvider`] to skip
+/// computing distances to items that do not satisfy the predicate.
+#[derive(Debug)]
+struct Wrapper<'a, P, I>
```
arkrishn94: nit: does it make sense to name this something more instructive?

diskann/src/graph/ext/labeled.rs
hildebrandmw: Thanks!

diskann/src/graph/ext/labeled.rs
hildebrandmw: Gave `EvalFilter` a try. What do yah think?

magdalendobson (approved):

PR #1145, opened 2026-06-10
[VectorId] Remove Id conversion bounds and traits

arkrishn94:
This PR removes the conversion traits/bounds of `VectorId` to and from scalar like types. This is part of an effort to loosen the constraints we impose on Ids.

**For reviewers**: Review in the following order of changes mentioned. 

## `VectorId` bounds

Removed all the `VectorIdTryFrom`, `Into*` and `FromPrimitive` trait bounds on `VectorId`. This removes the constraint that internal ids need to be able to be converted to and from `usize`. 

Along with that, this removes the following traits entirely from `diskann/src/utils/utils.rs`: 
- traits `VectorIdTryFrom<T>`, `TryIntoVectorId<T>`
- methods `vector_id_try_from`, `try_into_vector_id`
- helpers `vecid_from_u32`, `vecid_from_usize`
- error type `IdConversionError<const, F, T>` + alias `ErrorToVectorId<F, T>`
- their tests

`diskann-label-filter` needs `IntoUsize` bound on the ids since they are used as keys in the roaring treemap. This bound on the Id is added to: 
- `RoaringAttributeStore<IT>` (roaring treemap keys must be `u64`)
- `InlineBetaStrategy` (propagates the above through `DocumentProvider`)
- `QueryBitmapEvaluator` / `BitmapFilter` (bitmap membership is keyed by `usize`)

## Algorithm changes

- **`DiskANNIndex::prune_range`**: now takes `impl IntoIterator<Item = DP::InternalId> + Send` instead of `Range<DP::InternalId>`. Caller explicitly constructs the iterator for its specific provider's Id type.
- **`InmemIndexBuilder::final_prune`** for `disk-index`: signature unchanged externally (`Range<u32>`); now constructs the `u32` range with `as u32` at the trait-object boundary.

## Provider specializations

- **`SimpleNeighborProviderAsync<I>`** was written as generic over an `I : VectorId`. The generic is now removed since it is only ever instantiated with `I = u32`. Propagated this change to all in-mem providers. 
- Similar change to **`bftree::VectorProvider<T, I: VectorId = u32>`**. `starting_points()` becomes infallible (`Vec<u32>` instead of `Result<Vec<I>, ErrorToVectorId<…>>`).

## `bftree::NeighborProvider`

Removed the `AsKey` trait since it was a wrapper used to convert from `usize` and `u32` to raw bytes.


diskann/src/utils/vector_id.rs
```diff
@@ -18,13 +17,6 @@ pub trait VectorId:
     + std::cmp::Ord
     + std::hash::Hash
     + TypeStr
-    + VectorIdTryFrom<u32>
-    + VectorIdTryFrom<u64>
-    + VectorIdTryFrom<usize>
-    + TryIntoInteger<u32>
-    + Into<u64>
-    + IntoUsize
-    + FromPrimitive
```
hildebrandmw, line 27: Incredible!

diskann-bftree/src/lib.rs
```diff
@@ -50,7 +50,7 @@ impl From<ConfigError> for ANNError {
     }
 }
 
-trait AsKey {
+pub trait AsKey {
```
hildebrandmw: Why did this need to become public? Instead of using this trait, can't the implied `bytemuck::Pod` from `VectorId` be used internally instead to invoke `bytemuck::bytes_of`?

diskann-bftree/src/neighbors.rs
```diff
@@ -87,8 +87,7 @@ impl<I: VectorId> NeighborProvider<I> {
         let mut guard = neighbors.resize(self.dim);
 
         // Serialize the key, vector_id, into a byte string, &[u8]
-        let i = vector_id.into_usize();
-        let key = i.as_key();
+        let key = vector_id.as_key();
```
hildebrandmw: See my comment on `AsKey` - we can use `bytemuck::Pod` instead to invoke `bytes_of`.

hildebrandmw: This is technically a larger change than it first appears as this will invalidate any previously saved `bf-trees`. That said, the layout was already changed this week with the reordering of the adjacency lists and no one's using bf-tree.

diskann-bftree/src/neighbors.rs
```diff
@@ -324,6 +320,27 @@ where
...
+/// This is the encoding used for the trailing length cell in the neighbor-list value;
+/// see [`NeighborProvider::set_neighbors`] for the full layout. The decode side is
+/// [`read_len`].
+fn len_cell<I: bytemuck::Pod>(len: u32) -> I {
+    const { assert!(std::mem::size_of::<I>() >= 4) };
+
+    let mut cell = I::zeroed();
+    bytemuck::bytes_of_mut(&mut cell)[..4].copy_from_slice(&len.to_le_bytes());
+    cell
+}
```
hildebrandmw: These names are a little confusing, I would suggest something like `len_to_cell` and `cell_to_len`, but there's the deeper issue of the post-monomorphization error. However, we can side-step all of this entirely. BF-tree is already completely serializing the layout of the adjacency list. Why are we still trying to masquerade the length through `I`? We can store the raw `u32` bytes directly.

diskann-disk/src/build/builder/build.rs
```diff
@@ -675,7 +675,7 @@ where
 
                 match vector_data {
                     Some((i, (vector, _))) => {
-                        let id = vecid_from_usize(i)?;
+                        let id = i as u32;
```
hildebrandmw: This does loosen the safety here. Before the conversion was checked -now it's not. Is that a concern?

diskann/src/graph/index.rs
```diff
@@ -2852,20 +2852,20 @@ where
 
     /// Prune all nodes in the graph.
     ///
-    /// This is used as the final step of graph construction.
-    pub fn prune_range<S>(
+    /// This is used as the final step of graph construction. `ids` yields the
+    /// [`DataProvider::InternalId`] of every point to prune.
```
hildebrandmw, line 2816: Nit: while we're here, we might as well make the docstring correct
```rust
/// Prune all nodes in `ids`.
```

hildebrandmw (commented):
I support this change! Two general categories of feedback. First, can we just store the `u32` length directly in `diskann-bftree` and remove the conversion to `I` entirely? We're already going to bytes.

Second: this change adds unchecked `as u32` conversions where-as they were checked before. I'd prefer to keep these as checked conversions to be defensive.

diskann-bftree/src/lib.rs
arkrishn94: Removed and converted to bytes directly using `bytemuck::bytes_of`

diskann-providers/src/model/graph/provider/async_/simple_neighbor_provider.rs
```diff
@@ -82,15 +78,17 @@ impl<I: VectorId> SimpleNeighborProviderAsync<I> {
...
         list[0..neighbors.len()].copy_from_slice(neighbors);
 
-        // Lint: neighbor lists won't overflow the VectorIdType
-        #[allow(clippy::unwrap_used)]
-        {
-            list[self.graph.dim() - 1] = neighbors.len().try_into_vector_id().unwrap();
-        }
+        // The assertion above guarantees `neighbors.len() < self.graph.dim()`, which
+        // means it fits in a `u32` (graph dim is sized in `u32` anyway).
+        list[self.graph.dim() - 1] = neighbors.len() as u32;
```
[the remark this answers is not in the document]
arkrishn94: Skipping the check since `neighbors.len()` won't exceed `u32::MAX` in practice.

diskann-bftree/src/neighbors.rs
```diff
@@ -232,10 +229,10 @@ impl<I: VectorId> NeighborProvider<I> {
             let neighbors_bytes = cast_slice::<I, u8>(&neighbor_list);
             buf[..neighbors_bytes.len()].copy_from_slice(neighbors_bytes);
 
-            // Write the count at the last slot as a raw u32 (no VectorId round-trip needed)
-            let count: u32 = nbr_count as u32;
+            // Write the length cell at the last slot
+            let cell: I = len_cell(nbr_count as u32);
             let count_offset = (self.dim - 1) * id_size;
-            buf[count_offset..count_offset + 4].copy_from_slice(bytes_of(&count));
+            buf[count_offset..count_offset + id_size].copy_from_slice(bytemuck::bytes_of(&cell));
```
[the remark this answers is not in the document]
arkrishn94: Same as above: neighbor list length won't exceed `u32::MAX` in practice.

diskann-disk/src/build/builder/build.rs
arkrishn94: Made the conversion fallible, thanks for flagging.

diskann-benchmark/src/utils/filters.rs
arrayka, line 1: Consider adding a similar intro at the beginning of the PR description:

## Why VectorId conversion bounds existed before

The `VectorId` trait required `VectorIdTryFrom<u32>`, `VectorIdTryFrom<usize>`, `Into<u64>`, `IntoUsize`, and `FromPrimitive` because the algorithm code freely converted between ID types and integers:

1. **`prune_range`** took `Range<DP::InternalId>`. To iterate it, the code converted `start.into(): u64`, iterated `start..end` as a `u64` range, then converted back with `id.try_into_vector_id()`. This forced `Into<u64>` + `VectorIdTryFrom<u64>` on every `VectorId`.

2. **`SimpleNeighborProviderAsync<I>`** was generic over `I: VectorId`, propagating all conversion bounds through the generic parameter — even though it was only ever instantiated with `u32`.

3. Helper functions like `vecid_from_u32` and `vecid_from_usize` used `FromPrimitive` to convert raw integers into IDs at various points.

## Why they're redundant now

1. **`prune_range`** now takes `impl IntoIterator<Item = DP::InternalId>` — the caller constructs the iterator for their specific ID type. No conversions inside the algorithm. The `u64` range trick is gone.

2. **`SimpleNeighborProviderAsync`** is de-genericized to always use `u32`. No generic `I: VectorId`, so no conversion bounds needed.

3. **`bftree::VectorProvider`** similarly drops the generic ID, making `starting_points()` infallible (`Vec<u32>` instead of `Result<Vec<I>, ErrorToVectorId<…>>`).

4. Places that genuinely need integer conversion (like `diskann-label-filter` for roaring bitmap keys) add `IntoUsize` **locally** on specific structs, rather than burdening the global `VectorId` trait.

## In short

The bounds existed because the algorithm mixed ID types with integer arithmetic. The PR pushes conversions to the edges (callers/providers), so the core `VectorId` trait only needs identity-like bounds (`Eq`, `Ord`, `Hash`, `Copy`, `Send`, `Sync`).

diskann-bftree/src/neighbors.rs
```diff
@@ -245,8 +241,7 @@ impl<I: VectorId> NeighborProvider<I> {
 
     pub fn delete_vector(&self, vector_id: I) -> ANNResult<()> {
         // Serialize the key, vector_id, into a byte string, &[u8]
-        let i = vector_id.into_usize();
-        let key = i.as_key();
+        let key = bytemuck::bytes_of(&vector_id);
```
arrayka, line 240: We’re changing key serialization from “convert `i` to `usize`, then serialize” to “serialize `i` directly.”  
Just to double check - Is it safe? My assumption is that it should be, as long as the new encoding does not reduce key cardinality or introduce collisions relative to the previous path.

arrayka (approved):

diskann-bftree/src/neighbors.rs
arkrishn94: `bf-tree` folks seem fine with it based on #1150. They're always using `u32` as the vector id, so the 4 most significant bytes of the key are always 0 when converting to `usize`

diskann-benchmark/src/utils/filters.rs
arkrishn94: Will do.

hildebrandmw (approved):

PR #1146, opened 2026-06-10
Fix handling of deletes and ID minting during concurrent insert/delete

metajack:
Garnet cluster testing discovered a race condition with ID minting. The issue was:

1. Thread A insert does have a free ID so bumps next_id to mint a new ID id0. That ID is now valid, but unused.
2. Other threads delete and insert simultaneously, causing free IDs to be fully consumed.
3. Thread B insert detects there might be a free ID, but the fast free list is empty, so it triggers a refill. It sees that id0 is unused so it adds it to the fast free list. Thread B then writes vectors data to id0.
4. Thread A marks id0 as used and then writes vector data to id0.

Because there is a time period between minting the ID and marking it used, it is possible to give the ID out twice.

One possible fix is to reorder the operations so that we mark IDs used before bumping the next_id atomic, however, this requires a manual CAS loop. Instead, I combined max_block and next_id into the same RwLock so that these operations can be more simple and explicitly controlled.

In addition, there was another race during delete where an ID is marked free during delete, but that ID may be recycled before the function exits. To address this, we the mapping and attributes in delete() first, which prevents the vector from being returned from search, and then mark the ID deleted at the end.

### DIFF
--- a/diskann-garnet/src/fsm.rs
+++ b/diskann-garnet/src/fsm.rs
@@ -20,8 +20,8 @@
 
 use crossbeam::queue::ArrayQueue;
 use std::sync::{
-    RwLock, RwLockReadGuard,
-    atomic::{AtomicBool, AtomicU32, AtomicUsize, Ordering},
+    Mutex, RwLock, RwLockReadGuard, RwLockWriteGuard,
+    atomic::{AtomicBool, AtomicUsize, Ordering},
 };
 use thiserror::Error;
 
@@ -66,6 +66,12 @@ struct Barrier {
     quantization_enabled: bool,
 }
 
+struct IdMinter {
+    next_id: u32,
+    max_block: u32,
+    buffer: Vec<u8>,
+}
+
 /// The free space map manages an ID pool in Garnet to track which IDs are being
 /// used. This is necessary to reuse deleted vectors IDs, which is needed due to
 /// the small ID space of u32.
@@ -88,16 +94,16 @@ pub(crate) struct FreeSpaceMap {
     has_free_ids: AtomicBool,
     /// A queue of previously deleted IDs to prevent excessive reads to the FSM
     fast_free_list: ArrayQueue<u32>,
-    /// The maximum block ID stored in the FSM
-    max_block: RwLock<u32>,
-    /// The next ID that will be minted if reusing previously deleted IDs is unavailable
-    next_id: AtomicU32,
+    /// Controls minting new IDs and expanding the FSM blocks
+    id_minter: RwLock<IdMinter>,
     /// The total number of IDs marked used in the FSM
     total_used: AtomicUsize,
     /// Controls when ID reuse for deleted IDs is enabled.
     reuse_enabled: AtomicBool,
     /// Quantization backfill related parameters that must be synchronized.
     barrier: RwLock<Barrier>,
+    /// Refill lock, to prevent multiple fast free list refills happening concurrently.
+    refill_lock: Mutex<()>,
 }
 
 impl FreeSpaceMap {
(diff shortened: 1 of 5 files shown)


diskann/src/graph/index.rs
```diff
@@ -1620,6 +1624,11 @@ where
 
             self.drop_adj_list(&mut accessor, vector_id).await?;
 
+            self.data_provider
+                .release(context, vector_id)
+                .await
+                .escalate("`inplace_delete` requires a successful release")?;
```
hildebrandmw: Question for you: we are in all liklihood getting rid of at least `release` and likely `delete` as well since they are almost never used properly. `diskann-garnet` relying on this is probably not a great idea. Would it be more robust long-term for `diskann-garnet` to invoke whatever is needed by release externally after a successful inplace-delete call?

diskann-garnet/src/provider.rs
```diff
@@ -804,32 +799,44 @@ impl<T: VectorRepr> Delete for GarnetProvider<T> {
...
             .delete_iid(&context.term(Term::Quantized), id);
 
+        // Mark the ID free in the FSM.
+        if let Err(e) = self.fsm.mark_free(context, id) {
+            return future::ready(Err(e.into()));
+        };
+
         if !ok {
             return future::ready(Err(GarnetError::Delete.into()));
         }
```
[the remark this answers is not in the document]
metajack: This is benign. The current order creates stale entries which cannot be returned by searches (the mappings are deleted in delete()). During an insert that reuses that ID, it is possible that the vector ends up in the search results though, since vector data is written first, before neighbor data, and the mappings may get set again before the traversal is finished allowing the vector to be returned. This would cause a brief and small recall penalty, but is otherwise ok.

The other order means that we will leak a vector which will never be fully deleted. This seems a strictly worse outcome.

diskann-garnet/src/fsm.rs
```diff
@@ -256,14 +260,29 @@ impl FreeSpaceMap {
...
+        self.mark_id_unchecked(ctx, id, used)
+    }
+
     /// Return whether a given ID is free.
     pub(crate) fn is_free(&self, ctx: &Context, id: u32) -> Result<bool, FsmError> {
-        if id >= self.next_id.load(Ordering::Acquire) {
+        let id_minter = self.id_minter.read().unwrap();
+        if id >= id_minter.next_id {
             return Err(FsmError::IdOutOfRange(id));
         }
```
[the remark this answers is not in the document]
metajack: Fixed.

diskann-garnet/src/fsm.rs
```diff
@@ -212,20 +222,14 @@ impl FreeSpaceMap {
         self.mark_id(ctx, id, true)
     }
 
-    /// Mark an ID according to value (true = used, false = free).
+    /// Mark an ID according to value (true = used, false = free), but don't check that its in range.
```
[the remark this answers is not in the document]
metajack: Fixed.

diskann-garnet/src/provider.rs
```diff
@@ -1101,7 +1108,7 @@ impl<'a, T: VectorRepr> SearchPostProcess<DynamicAccessor<'a, T>, &[T], GarnetId
         for n in candidates {
             let id = match accessor.provider.to_external_id(accessor.context, n.id) {
                 Ok(id) => id,
-                Err(e) => return future::ready(Err(e)),
+                Err(_) => continue, // vector got deleted; skip
             };
```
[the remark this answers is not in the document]
metajack: Fixed.

diskann/src/graph/index.rs
metajack: I'm not sure what to do here. We shouldn't depend on the provider's behavior being correct only when the outermost use of DiskANN also has provider knowledge and does extra things. So I think while I agree that delete/release probably aren't the long term solution, I don't think handling this outside the DataProvider is better. I'd prefer to leave this as it is as anyone who tries to refactor delete/release will immediately see how its used and have to come up with a full solution. That will be hidden information if we put release logic outside DiskANN's provider chain.

hildebrandmw: I don't agree. If `diskann-garnet` takes a dependency on calling `release` here (and would be the only one as far as I know actually relying on `release` being called), whoever comes along to refactor `release` won't really know the true nature of the dependency (especially since there aren't tests in `diskann-garnet` to flag this) and this could easily break.

I'm getting convinced that `diskann` making `delete/release` calls isn't really a contract that can be implemented correctly for a lot of providers. Given that, I think it's much less likely to break if garnet makes the call manually after inplace-delete completes.

metajack: If there are data correctness guarantees that the provider needs to make, the provider needs to be able to make them. It cannot rely on the external caller upholding them.

This means that we cannot put this logic outside of top-level API calls to DiskANNIndex, of which `inplace_delete()` is one. So whatever cleanup needs to happen needs to be somewhere inside DiskANNIndex.

I agree with you that the original semantics of delete/release don't make any sense, but since they are not actually used, it seems reasonable to me to just document them as something more akin to prepare_delete() and finish_delete(). It seems such a post-delete hook is definitely needed in the DataProvider API somewhere.

Now we could choose to hoist this cleanup into the driver as a workaround temporarily, but my claim is that doing this won't be obviously noticed during cleanup of delete/release when that actually happens. Using release() leaves breadcrumbs to the actual problem, since we only have one provider that does this.

metajack: This is now removed entirely. We delete as much as we can in delete() (except for neighbors which is a documented bug in DiskANN), and no longer use release().

Mark and I considered the races that were left and convinced ourselves that they result in benign issues which we accept as part of not stopping the world for mutation in the index.

hildebrandmw (approved):

JordanMaples (approved):

PR #1150, opened 2026-06-12
[bftree] Bug with writing length to neighbor list

arkrishn94:
In the process of loosening bounds related to `VectorId` in #1145, I came across some errors in the way the neighbor list was being constructed in the `NeighborProvider` in `diskann-bftree`, 

1. The reading and writing of neighbor lists happens through an allocated buffer in the provider. The calculation of the offset at which to write the length of the neighbor list was being done using the size of the id type `std::mem::size_of::<I>()`, but the number of bytes being written at the end was always 4 (i.e. a `u32`). This was not causing an issue since all tests only use `u32` ids. 
2. Another bug is that in `set_neighbors`, the length was not being written to the tail of the neighbor list but immediately after the bytes storing the neighor ids. 

This PR fixes both by consolidating the set neighbor calls and using a single path in `set_neighbors_internal` - the array being written is now always `(self.max_degree() + 1) * size_of::<I>()` bytes long and always contains the length in little-endian `u32` in the last `size_of::<I>()` bytes. Read path checks to make sure the read length is exactly `max_degree + 1`.

### DIFF
--- a/diskann-bftree/src/neighbors.rs
+++ b/diskann-bftree/src/neighbors.rs
@@ -96,7 +96,7 @@ impl<I: VectorId> NeighborProvider<I> {
                 // If found, then re-construct the neighbor list (valid neighbors only)
                 if read_size > 0 {
                     // A retrieved neighbor list should be exactly max_degree + 1 length
-                    if read_size != (self.max_degree() + 1) * std::mem::size_of::<I>() as u32 {
+                    if read_size as usize != self.dim * std::mem::size_of::<I>() {
                         return Err(ANNError::log_index_error(
                             "Retrieved neighbor list is not expected length = max degree + 1",
                         ));
@@ -140,20 +140,21 @@ impl<I: VectorId> NeighborProvider<I> {
 
     /// Internal function for setting the neighbors for a vector id.
     ///
-    /// Each key is a `VectorId`, written in bytes. The value is a neighbor list
-    /// of length exactly `self.max_degree() + 1`. Specifically, an array of
-    /// exactly `n` neighbors is written as :  
+    /// Each key is a `VectorId`, written in bytes. The value is a
+    /// neighbor list of length exactly `self.dim`. Specifically,
+    /// an array of exactly `n` neighbors is written as :  
     /// ```text
     /// |  I0  | ... |  In  | padding |  ...  | [n; 0] |
     ///     
     /// -----------------------------------------------
     ///                     |
-    ///            self.max_degree() + 1
+    ///                 self.dim
     /// ```
-    /// where `[n; 0]` represents the u32 count `n` byte-packed into a type `I`,
-    /// and 'padding' indicates unfilled empty slots.
+    /// where `[n; 0]` represents the u32 count `n` byte-packed
+    /// into a type `I`, and 'padding' indicates unfilled empty slots.
     ///
-    /// Note: assuming all neighbors in the input list, 'neighbors', are valid.
+    /// Note: assuming all neighbors in the input list, 'neighbors',
+    /// are valid.
     fn set_neighbors_internal(
         &self,
         vector_id: I,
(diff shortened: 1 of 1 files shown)


diskann-bftree/src/neighbors.rs
```diff
@@ -76,19 +75,26 @@ impl<I: VectorId> NeighborProvider<I> {
         Self::new(max_degree, adjacency_list_index)
     }
 
+    /// The length of a single `value` in the tree
+    ///
+    /// The length in bytes of the values read and set
+    /// into the bf-tree is always `max_degree + 1` I-cells long.
+    fn value_bytes(&self) -> usize {
```
doliawu: `num_value_bytes` or `value_size`?

diskann-bftree/src/neighbors.rs
```diff
@@ -98,41 +104,20 @@ impl<I: VectorId> NeighborProvider<I> {
...
 
-                    // A retrieved neighbor list should not be longer than the max degree
-                    if (read_size as usize) > (std::mem::size_of::<I>() * self.dim) {
-                        return Err(ANNError::log_index_error(
-                            "Retrieved neighbor list is longer than the max degree",
-                        ));
-                    }
+                    // The last I-cell stores the neighbor count (see `cell_to_len`).
+                    //
+                    // SAFETY: we know this will not panic since
```
doliawu, line 107: Kind of weird seeing unsafe note but no unsafe keyword used.

diskann-bftree/src/neighbors.rs
```diff
@@ -98,41 +104,20 @@ impl<I: VectorId> NeighborProvider<I> {
...
-
-                    // The specified list length must be smaller than the retrieved data length
-                    if (read_size as usize) < (std::mem::size_of::<I>() * (nbr_count + 1)) {
-                        return Err(ANNError::log_index_error(
-                            "The length of the retrieved neighbor list is shorter than the specified length",
-                        ));
-                    }
-
-                    guard.finish(nbr_count);
+                    guard.finish(count as usize);
```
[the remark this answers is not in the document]
arkrishn94: Good point, added the check.

diskann-bftree/src/neighbors.rs
```diff
@@ -155,45 +140,77 @@ impl<I: VectorId> NeighborProvider<I> {
...
             return Err(ANNError::log_index_error(
-                "The provided neighbor list is longer than the max degree",
+                "The provided buffer is not long enough",
             ));
         }
 
-        // Serialize the key, vector_id, into a byte string, &[u8]
-        let i = vector_id.into_usize();
-        let key = i.as_key();
+        let buf = cast_slice_mut::<u8, I>(&mut buf[..self.value_bytes()]);
```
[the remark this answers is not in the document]
arkrishn94: `Buf` is now a `Vec<I>`

diskann-bftree/src/neighbors.rs
```diff
@@ -83,56 +82,40 @@ impl<I: VectorId> NeighborProvider<I> {
...
+        // &[u8], in the format described in [`Self::set_neighbors_internal`].
         let value = cast_slice_mut::<I, u8>(&mut guard);
         match self.adjacency_list_index.read(key, value) {
             bf_tree::LeafReadResult::Found(read_size) => {
                 // If found, then re-construct the neighbor list (valid neighbors only)
                 if read_size > 0 {
-                    // A non-empty neighbor list should at least contain one entry for the list length
-                    if (read_size as usize) < std::mem::size_of::<I>() {
+                    // A retrieved neighbor list should be exactly max_degree + 1 length
+                    if read_size != (self.max_degree() + 1) * std::mem::size_of::<I>() as u32 {
```
metajack: If we have `self.dim` for this, we should use it consistently.

metajack (approved):

diskann-bftree/src/neighbors.rs
arkrishn94: Removed this since it was not needed :)

JordanMaples (approved):

diskann-bftree/src/neighbors.rs
arkrishn94: Makes sense, fixed.

diskann-bftree/src/neighbors.rs
arkrishn94: Yeah that's fair. Didn't know how else to flag it so just went with this p

JordanMaples (approved):
lgtm

PR #1151, opened 2026-06-12
Remove `k_value` from `Knn`

magdalendobson:
When preparing the inline filter search PR I noticed that the `k_value` in `Knn` is actually not used anymore, since the output buffer is treated as the source of truth for `k`. This PR removes that field. In order to do this, since diversity search made use of it, we add a field for it specifically in diversity search, and construct a wrapper struct for `Knn` in diskann-benchmark-core for validation of parameters and use in size hinting. This moves the `k` parameter out of places it is not used and directly into the places where it is used.


harsha-simhadri (approved):

diskann-benchmark-core/src/search/graph/knn.rs
```diff
@@ -133,6 +134,51 @@ where
...
+        ANNError::opaque(err)
+    }
+}
+
+/// A wrapper for the [`graph::search::Knn`] struct that also includes the `k` value.
+#[derive(Debug, Copy, Clone)]
+pub struct KnnWrapper {
+    k_value: NonZeroUsize,
+    pub knn: graph::search::Knn,
+}
```
hildebrandmw, line 264: Can we change this to
```rust
pub struct KnnParams {
    num_neighbors: NonZeroUsize,
    knn: graph::search::Knn,
}

impl KnnParams {
  pub fn new(num_neighbors: NonZeroUsize, knn: graph::search::Knn) -> Result<Self, Error>;

  pub fn num_neighbors(&self) -> NonZeroUsize;
  pub fn knn(&self) -> &graph::search::Knn;  
```

Reasons:
* A `pub` field for `knn` undoes the validation check in the constructor.
* Taking `graph::search::Knn` directly allows us to use a configurable beam width and simplifies the error.
* `num_neighbors` is IMO slightly clearer.

diskann/src/graph/misc.rs
```diff
@@ -51,13 +79,20 @@ where
     pub fn new(
         diverse_attribute_id: usize,
         diverse_results_k: usize,
+        original_k_value: usize,
```
hildebrandmw: Bikeshed: maybe `total_k_value` or `total_nearest_neighbors` instead of `original_k_value`?

diskann/src/graph/misc.rs
```diff
@@ -2,6 +2,15 @@
...
  */
+#[cfg(feature = "experimental_diversity_search")]
+mod imports {
+    pub(super) use crate::{ANNError, ANNErrorKind};
+    pub(super) use std::num::NonZeroUsize;
+    pub(super) use thiserror::Error;
+}
+
+#[cfg(feature = "experimental_diversity_search")]
+use imports::*;
```
hildebrandmw: This is an ... odd way of structuring this. This would be a lot cleaner if the parameters were grouped with the implementation of `experimental_diversity_search` rather than spread around. Can we do that shuffle in this PR?

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -910,7 +909,13 @@ where
                 &mut result_output_buffer,
             ))?
         } else {
-            let knn_search = Knn::new(k, l, beam_width)?;
+            if search_list_size < k_value as u32 {
+                return Err(ANNError::message(
+                    diskann::ANNErrorKind::IndexError,
+                    "search list size must be at least as large as the number of results requested",
+                ));
+            }
```
hildebrandmw, line 1207: If we're going to throw an error here - does it make sense to move this to be the first check that happens? That way, all code-paths get to benefit and we can bail earlier if something looks wrong?

hildebrandmw (commented):
Thanks Magdalen - I support this movement. Just a few requests on my end.

diskann/src/graph/misc.rs
```diff
@@ -2,6 +2,15 @@
  * Copyright (c) Microsoft Corporation.
  * Licensed under the MIT license.
  */
+#[cfg(feature = "experimental_diversity_search")]
+mod imports {
+    pub(super) use crate::{ANNError, ANNErrorKind};
+    pub(super) use std::num::NonZeroUsize;
+    pub(super) use thiserror::Error;
+}
```
hildebrandmw: A request: About 90% of the awkwardness with this import is because `DiverseSearchParams` doesn't belong here - it belongs in the `diverse_search` module. Do you mind moving it there? Alternatively, I can open a mini-PR to do that.

diskann-benchmark-core/src/search/graph/knn.rs
hildebrandmw: Alternatively, the name `Knn` could work too since it's in a separate module, but in this case I think `KnnParams` is a little less overloaded. Maybe we could call it `Params` and access it as `knn::Params`? Any one of these works, but I'm not a fan of `KnnWrapper`.

magdalendobson: Yep, renamed

diskann/src/graph/misc.rs
magdalendobson: Good point, renamed

diskann/src/graph/misc.rs
magdalendobson: Yep, done

diskann-disk/src/search/provider/disk_provider.rs
magdalendobson: Moved earlier

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -1059,6 +1059,12 @@ where
         let mut associated_data =
             vec![Data::AssociatedDataType::default(); return_list_size as usize];
 
+        if search_list_size < return_list_size {
+            return Err(ANNError::message(
```
hildebrandmw: Small not: use `diskann_error!` for consistency.

hildebrandmw (approved):

hildebrandmw (approved):

PR #1158, opened 2026-06-12
Concurrency hardening for BfTree parallel workloads

JordanMaples:
# Problem

`NeighborProvider::append_vector` performs a read-modify-write cycle (read existing neighbors → append → write back) without synchronization. Under concurrent mutation, the last writer wins and silently drops the other's edges. Stress testing shows edge loss under multi-thread contention on a single vertex.

Additionally, the dual-store `set_element` wrote quant before full-precision, meaning if the full-precision write failed after quant succeeded, it would leave a quantized ghost with no backing data. And `delete` did not clean up the quant store entry.

# Solution

## Striped `Mutex` table for per-vertex synchronization

- `StripedLocks`: a fixed-size table of `Mutex<()>` stripes (`diskann-bftree/src/locks.rs`).
- Stripe count is derived from hardware parallelism: `(cpus * 4).next_power_of_two().max(64)` — constant memory regardless of dataset size, suitable for billion-scale workloads.
- Vertex IDs map to stripes via Fibonacci multiply-shift hashing (`floor(2^64 / phi)`), keeping false contention negligible.
- Each mutating op acquires exactly **one** stripe lock for the target vertex, does its work, and drops the guard — no nested locking, so deadlock is structurally impossible.

### Why striped instead of per-vertex locks?

BfTree's value proposition is spilling to disk for datasets exceeding RAM. At 1B vectors, per-vertex locks would consume gigabytes of in-memory locks — defeating the purpose. Striped locks keep memory constant at the cost of rare false contention.

### Reads are lock-free

`get_neighbors` and `get_vector` acquire **no** stripe lock. They rely on bf-tree's internal thread safety and the fact that DiskANN search is approximate by design: a reader may briefly observe a partially-updated neighbor list, which affects recall momentarily rather than correctness. Writers (`set_element`, `delete`, `set_neighbors`/`write_neighbors`, `append_vector`/`write_append`) serialize on the stripe.

## Dual-store write ordering

- Full-precision is now written **first** in `set_element`.
- Failure mode is benign: vector reachable in full-precision but not yet in quantized search (rather than a quantized ghost with no backing data).

## Quant store cleanup on delete

- New `DeleteQuant` trait: real impl for `QuantVectorProvider` (deletes the quant entry), no-op for `NoStore`.
- `delete` now removes both full-precision and quant entries under the stripe lock. Neighbor-topology cleanup remains in the upper `DiskANNIndex` layer (`drop_adj_list`).

# Testing

Three new concurrent stress tests in `diskann-bftree/src/neighbors.rs`:

- `test_concurrent_append_no_lost_edges` — 8 threads × 10 edges to the same vertex; asserts every edge survives.
- `test_concurrent_append_independent_vertices` — 8 threads appending to disjoint vertex ranges; asserts total edge count.
- `test_concurrent_read_write_consistency` — mixed readers + writers on the same vertex; validates no torn reads.

### DIFF
--- a/diskann-bftree/src/neighbors.rs
+++ b/diskann-bftree/src/neighbors.rs
@@ -6,6 +6,7 @@
 //! Bf-Tree neighbor list provider.
 
 use std::marker::PhantomData;
+use std::sync::RwLock;
 
 use bf_tree::{BfTree, Config};
 use bytemuck::{cast_slice, cast_slice_mut};
@@ -19,20 +20,71 @@ use diskann::{
 use super::ConfigError;
 use crate::{bftree_insert, TestCallCount};
 
-pub struct NeighborProvider<I: VectorId> {
+pub struct NeighborProvider<I: VectorId + IntoUsize> {
     adjacency_list_index: BfTree,
     dim: usize, // Max number of neighbors in a neighbor list + 1 for the neighbor count
+    locks: StripedLocks,
     #[allow(dead_code)]
     pub(crate) num_get_calls: TestCallCount,
     _phantom: PhantomData<I>,
 }
 
-impl<I: VectorId> HasId for NeighborProvider<I> {
+/// A fixed-size table of striped locks for per-vertex synchronization.
+///
+/// Vertex IDs are mapped to lock stripes via a multiply-shift hash. This provides
+/// constant memory overhead regardless of dataset size, at the cost of
+/// occasional false contention between vertices that map to the same stripe.
+///
+/// The default stripe count (16384) is chosen to keep false sharing negligible
+/// for typical workloads while using only ~128 KB of memory.
+struct StripedLocks {
+    stripes: Box<[RwLock<()>]>,
+}
+
+impl StripedLocks {
+    const DEFAULT_STRIPE_COUNT: usize = 16384;
+    // Fibonacci hashing constant for 64-bit: floor(2^64 / phi)
+    const HASH_MULTIPLIER: u64 = 0x9E3779B97F4A7C15;
+    // Shift to extract the top log2(16384) = 14 bits
+    const HASH_SHIFT: u32 = 64 - 14;
+
+    fn new() -> Self {
+        Self {
+            stripes: (0..Self::DEFAULT_STRIPE_COUNT)
+                .map(|_| RwLock::new(()))
+                .collect::<Vec<_>>()
+                .into_boxed_slice(),
+        }
+    }
+
+    #[inline]
+    fn stripe_index(&self, id: usize) -> usize {
+        ((id as u64).wrapping_mul(Self::HASH_MULTIPLIER) >> Self::HASH_SHIFT) as usize
+    }
+
+    #[inline]
+    fn read(&self, id: usize) -> std::sync::RwLockReadGuard<'_, ()> {
+        let stripe = self.stripe_index(id);
+        self.stripes[stripe]
+            .read()
+            .unwrap_or_else(|e| e.into_inner())
+    }
+
+    #[inline]
+    fn write(&self, id: usize) -> std::sync::RwLockWriteGuard<'_, ()> {
+        let stripe = self.stripe_index(id);
+        self.stripes[stripe]
+            .write()
+            .unwrap_or_else(|e| e.into_inner())
+    }
+}
+
+impl<I: VectorId + IntoUsize> HasId for NeighborProvider<I> {
     type Id = I;
 }
 
-impl<I: VectorId> NeighborProvider<I> {
-    /// Create a new instance based on bf-tree Config directly
+impl<I: VectorId + IntoUsize> NeighborProvider<I> {
+    /// Create a new instance based on bf-tree Config directly.
     pub fn new_with_config(max_degree: u32, config: Config) -> ANNResult<Self> {
         let key_size = std::mem::size_of::<u32>();
         let value_size = (max_degree as usize + 1) * std::mem::size_of::<u32>();
@@ -49,6 +101,7 @@ impl<I: VectorId> NeighborProvider<I> {
         Ok(Self {
             adjacency_list_index,
             dim,
+            locks: StripedLocks::new(),
             num_get_calls: TestCallCount::default(),
             _phantom: PhantomData,
         })
@@ -79,13 +132,32 @@ impl<I: VectorId> NeighborProvider<I> {
         Self::new(max_degree, adjacency_list_index)
     }
 
-    /// Retrieve the neighbor list of a vector
-    /// `neighbors` is cleared first upon each invocation
-    /// One data copy is involved which copies the data from bf-tree to `neighbors`
+    /// Retrieve the neighbor list of a vector.
+    ///
+    /// Acquires a read lock on the vertex to ensure consistency with concurrent
+    /// mutations (set/append).
     pub fn get_neighbors(&self, vector_id: I, neighbors: &mut AdjacencyList<I>) -> ANNResult<()> {
         #[cfg(test)]
         self.num_get_calls.increment();
 
+        // Clear the output before any early returns so callers always see
+        // a deterministic empty state on error.
+        neighbors.clear();
+
+        let idx = vector_id.into_usize();
+        let _guard = self.locks.read(idx);
+
+        self.get_neighbors_unlocked(vector_id, neighbors)
+    }
+
+    /// Retrieve the neighbor list without acquiring any lock.
+    ///
+    /// Callers must ensure they already hold the appropriate lock.
+    fn get_neighbors_unlocked(
+        &self,
+        vector_id: I,
+        neighbors: &mut AdjacencyList<I>,
+    ) -> ANNResult<()> {
         // Resize 'neighbors' to hold any full-size neighbor list + I-cell for length
         let mut guard = neighbors.resize(self.dim);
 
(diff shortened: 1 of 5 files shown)


harsha-simhadri:
Jordan, what is the perf impact of this change? Is ingestion speed the same as before? thanks

JordanMaples:
> Jordan, what is the perf impact of this change? Is ingestion speed the same as before? thanks

@harsha-simhadri I'm seeing roughly 10% additional index build time on my arm-based laptop in WSL building Wiki100k. 


Here's copilot's summary of my test results for your consideration:
```
--------------------------------------------------------------------------------------------------

Performance A/B results on Wikipedia-100K (768D, float32, inner product):

Static build (multi-insert, batch_size=128, 8 parallel inserters):

┌───────────────────────────────┬────────────┬───────────────┐
│                               │ With locks │ Without locks │
├───────────────────────────────┼────────────┼───────────────┤
│ Build time                    │ 57.82s     │ 52.46s        │
├───────────────────────────────┼────────────┼───────────────┤
│ Insert latency (mean)         │ 73.8ms     │ 66.9ms        │
├───────────────────────────────┼────────────┼───────────────┤
│ Recall@100 (search_l=100)     │ 0.9022     │ 0.9022        │
├───────────────────────────────┼────────────┼───────────────┤
│ Recall@100 (search_l=200)     │ 0.9536     │ 0.9536        │
├───────────────────────────────┼────────────┼───────────────┤
│ Search QPS (8 threads, l=200) │ 519        │ 484           │
└───────────────────────────────┴────────────┴───────────────┘

Streaming (insert/delete/reinsert churn via BigANN runbook, 8 threads):

┌───────────────────────────┬────────────┬───────────────┐
│                           │ With locks │ Without locks │
├───────────────────────────┼────────────┼───────────────┤
│ Insert latency (mean)     │ 3739µs     │ 3965µs        │
├───────────────────────────┼────────────┼───────────────┤
│ Recall@100 (search_l=200) │ 0.9766     │ 0.9765        │
├───────────────────────────┼────────────┼───────────────┤
│ Search QPS (8 threads)    │ 813        │ 812           │
└───────────────────────────┴────────────┴───────────────┘

Summary: 
Locks add ~10% build overhead in the contended parallel-insert path. 
Search performance and recall are unaffected. In the streaming (sequential insert) path, overhead is within noise. 
The locks provide a correctness guarantee against edge loss under high contention — newly added stress unit tests show 11–51% edge loss without them when threads hammer the same vertices concurrently.
```

diskann-bftree/src/neighbors.rs
```diff
@@ -19,20 +20,71 @@ use diskann::{
...
+/// occasional false contention between vertices that map to the same stripe.
+///
+/// The default stripe count (16384) is chosen to keep false sharing negligible
+/// for typical workloads while using only ~128 KB of memory.
+struct StripedLocks {
+    stripes: Box<[RwLock<()>]>,
+}
+
+impl StripedLocks {
+    const DEFAULT_STRIPE_COUNT: usize = 16384;
```
hildebrandmw: Why was 16,384 picked and how does this interact with tests? It seems like a rather hefty allocation .

diskann-bftree/src/provider.rs
```diff
@@ -421,7 +442,11 @@ where
         gid: &Self::ExternalId,
     ) -> impl std::future::Future<Output = Result<(), Self::Error>> + Send {
         let id = *gid;
+        // Only delete vector data here. Neighbor adjacency cleanup (zeroing the
+        // deleted vertex's edge list and patching neighbors-of-neighbors) is
+        // handled by `DiskANNIndex::inplace_delete` → `drop_adj_list`.
         self.full_vectors.delete_vector(id as usize);
+        self.quant_vectors.delete_vector(id as usize);
```
hildebrandmw, line 494: If we're thinking about race conditions: this one rases with a `SetElement` :wink:

JordanMaples: kk this should be sorted now

diskann-bftree/src/neighbors.rs
JordanMaples: removed and switched to something based on current hardware.

JordanMaples:
@copilot give it another review.

JordanMaples:
@harsha-simhadri after chatting a bit with Mark and and changing the approach to locking in the provider. I ran a few more wiki100k streaming benchmarks on my devbox instance it looks like the perf overhead is negligible.

diskann-bftree/src/neighbors.rs
```diff
@@ -236,51 +258,53 @@ impl<I: VectorId> NeighborProvider<I> {
...
     }
 }
 
 pub struct NeighborAccessor<'a, I>
 where
-    I: VectorId,
+    I: VectorId + IntoUsize,
 {
     provider: &'a NeighborProvider<I>,
+    locks: Arc<StripedLocks>,
```
hildebrandmw: One small change is that this can take a `&'a StripedLocks` instead. Since the top level provider  has both a `NeighborProvider` and the `Locks`, references can be used for both.

hildebrandmw (approved):

diskann-bftree/src/provider.rs
```diff
@@ -393,19 +399,40 @@ where
...
-/// committed. Use [`InplaceDeleteMethod::OneHop`] or [`InplaceDeleteMethod::TwoHopAndOneHop`]
-/// instead, as these strategies only require neighbor topology (which remains accessible).
+/// from the underlying bf-tree storage when [`Delete::delete`] is called. When a quantized
+/// store is present, both the full-precision and quantized entries are removed.
+///
+/// This has an important consequence for inplace delete: the
+/// [`InplaceDeleteMethod::VisitedAndTopK`] strategy is **incompatible** with this provider
+/// because it requires reading the deleted vector's data (via
+/// [`InplaceDeleteStrategy::get_delete_element`]) *after* the delete has already been committed.
+/// Use [`InplaceDeleteMethod::OneHop`] or [`InplaceDeleteMethod::TwoHopAndOneHop`] instead,
```
harsha-simhadri, line 468: @metajack, @mhildebrand, would the need to re-access quant vector after delete is committed be alleviated by upcoming algo changes where the quant vector is copied to temp buffer before being deleted?

diskann-bftree/src/provider.rs
```diff
@@ -197,6 +198,10 @@ where
     // Graph configuration parameters for persistence
     //
     pub(crate) graph_params: Option<GraphParams>,
+
+    // Striped locks for per-vertex synchronization across all stores.
+    //
+    pub(crate) locks: StripedLocks,
```
harsha-simhadri, line 224: Could you please document for future reference
- how many locks can be held by a thread (processing insert or delete) at any point in time? 
- if more than one, what is the locking order and arguments on how the order ensures deadlock freedom
- any expectations for threads reading the locked neigbor lists. 
thanks

harsha-simhadri (commented):
Left some comments inline. Ack on perf numbers.

harsha-simhadri (commented):
Left some comments inline. Ack on perf numbers.

diskann-bftree/src/provider.rs
JordanMaples: added

diskann-bftree/src/provider.rs
hildebrandmw: Yes. The algo with either copy out explicitly or avoid accessing the vector entirely.

metajack (approved):

diskann-bftree/src/neighbors.rs
```diff
@@ -547,4 +579,220 @@ mod tests {
...
     }
+
+    /// Stress test: many threads concurrently append to the SAME vertex.
+    ///
+    /// Without per-vertex locking, this test would non-deterministically lose edges
+    /// due to the TOCTOU race in append_vector's read-modify-write cycle. With the
+    /// per-vertex write lock, every appended edge must be present in the final list.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 8)]
+    async fn test_concurrent_append_no_lost_edges() {
+        let locks = Arc::new(StripedLocks::new());
```
harsha-simhadri: COuld we increase the number of threads and iterations over which this test is run. I am concerned that a single run with 8 threads might not be enough to surface non-determinism reliably

diskann-bftree/src/neighbors.rs
```diff
@@ -547,4 +579,220 @@ mod tests {
...
+                while !done_clone.load(std::sync::atomic::Ordering::Acquire) {
+                    provider_clone.get_neighbors(0, &mut result).unwrap();
+                    // The list must never exceed max_degree
+                    assert!(
+                        result.len() <= max_degree as usize,
+                        "Neighbor list exceeded max_degree"
+                    );
+                    // All IDs in the list must be non-zero (our IDs start at 1)
+                    for &id in result.iter() {
+                        assert!(id > 0, "Found invalid zero ID in neighbor list");
```
harsha-simhadri: could we also check that non of the results are out of bounds (id<max_inserted_points or something like that)

diskann-bftree/src/neighbors.rs
JordanMaples: I found that 8 threads did a pretty decent job of uncovering the problem initially, but sure, I can crank up the number of threads in play here. I'll remove the thread count specification so that tokio will pull the max number of threads that the cpu has available. 

I'll also add some internal repetition of the test code to increase our chances of hitting an actual problem

hildebrandmw (approved):

PR #1160, opened 2026-06-12
[flat] Post-processing support and use `DataProvider::InternalId`

arkrishn94:
This PR adds support to do post-processing in the `FlatIndex::knn_search` via the `diskann::glue::SearchPostProcess` trait; which is the same interface used by the graph search API. 

In addition, I had left a hook in the flat search `SearchStrategy` and `DistancesUnordered` to support ID types that were not convertible to and from scalar-like values. With #1145, that constraint is now gone in `VectorId` and we are free to correctly tie the ID type in the `DistancesUnordered` implementation used by the flat index to the `InternalId` of the underlying `DataProvider`. This means: 
- `DistancesUnordered` now has `HasId` as a super-trait. 
- `SearchStrategy` no longer needs `Id` as an associated type.

## Testing 

Added testing in the test module to make sure post-processing works as it should via a custom post-processor that only keeps even ids. Added some extra code in the `test::harness` module just to make the dispatch to post-processing generic enough so I can enumerate easily over cases.

**For reviewers**: Review the `flat\strategy.rs` changes first, then `flat\index.rs`. Optionally, then, review the testing code.


diskann/src/flat/strategy.rs
```diff
@@ -21,18 +21,14 @@ use crate::{
 /// with the supplied computer `C` and invoking `f` with the resulting `(id, distance)`
 /// pair. The associated [`Self::ElementRef`] is the reference shape on which `C` must
 /// be able to compute distances.
-pub trait DistancesUnordered<C>: Send + Sync
+pub trait DistancesUnordered<C>: HasId + Send + Sync
```
doliawu, line 24: I'm not sure how to make sense of this change semantically - "DistancesUnordered should be something that HasId?"

arkrishn94: The closure needs to act on an `Id` so `DistancesUnordered` needs to be implemented for a struct that has underlying IDs.

diskann/src/flat/test/harness.rs
```diff
@@ -5,58 +5,66 @@
...
     ) -> ANNResult<Self> {
-        current_thread_runtime().block_on(Self::run(index, strategy, query, k))
+        current_thread_runtime().block_on(Self::run(index, strategy, oracle, query, k))
     }
 
     /// Async variant of [`KnnOracleRun::run_sync`]. Use this from tests that already
     /// have a Tokio runtime (e.g. `#[tokio::test]`) or that need to drive
     /// `knn_search` concurrently across tasks.
-    pub async fn run(
+    pub async fn run<O: OracleProcessor>(
```
harsha-simhadri, line 64: Is OracleProcessor similar in spirit to QueryLAbelProvider in diskann/graph?

arkrishn94: Not quite, this is just a wrapper used in the testing submodule for the flat index to test different post-processors.

arrayka (approved):

doliawu (approved):

PR #1166, opened 2026-06-15
fix(bftree): validate record sizes at construction 

### LINKED ISSUE #1159: error building index [bftree-full-precision]
## Expected Behavior
Build "graph-index-build-bftree-full-precision" index

## Example Code
using the **./diskann-benchmark/example/graph-index-bftree.json** benchmark.

## Dataset Description
Please tell us about the shape and datatype of your data, (e.g. 128 dimensions, 12.3 billion points, floats)
- Dimensions: 1536
- Number of Points: 1000000
- Data type: float32

**Tried to reduce the number of points to 10k, throws the same error.** 

## Error 
```
Graph Index Bf_Tree Full-Precision Build
               tag: graph-index-build-bftree-full-precision

              file: ada_002_1000000_base_vectors.bin
         data_type: float32
        max degree: 32
           L-build: 50
             alpha: 1.2
start point strategy: Medoid
    backedge ratio: 1
Using Multi Insert: NO
start_point_strategy: Medoid
     build threads: 28
         Save Path: None

  Vector Store:
      cb_size_byte: 67108864
    leaf_page_size: 4096
cb_max_record_size: 1952
cb_min_record_size: 4
read_promotion_rate: 30
scan_promotion_rate: 30
cb_copy_on_access_ratio: 0.1
 read_record_cache: true
        cache_only: false

  Neighbor Store:
      cb_size_byte: 67108864
    leaf_page_size: 4096
cb_max_record_size: 1952
cb_min_record_size: 4
read_promotion_rate: 30
scan_promotion_rate: 30
cb_copy_on_access_ratio: 0.1
 read_record_cache: true
        cache_only: false

Error: ANNError: DiskANN(IndexError)

vector 1000000 not found, escalated: starting point retrieval must succeed -- (./DiskANN/diskann-bftree/src/lib.rs:101)
```

## Additional Details
**graph-index-build-bftree-spherical-quantization** works on the same data
**graph-index-build-bftree-full-precision** works on sift dataset (100k points)

JordanMaples:
Problem

BfTree::insert returns a LeafInsertResult indicating success or failure, but all three sub-providers (vectors, neighbors, quant) were discarding it unconditionally. When a record exceeds cb_max_record_size, the insert silently fails and later surfaces as a misleading "vector not found" error during search.

This is the root cause of #1159 (https://github.com/microsoft/DiskANN/issues/1159): the default cb_max_record_size (1952 bytes) is too small for 1536D f32 vectors (8-byte key + 6144-byte value = 6152 bytes), but the failure was swallowed at insert time.

Solution

Early validation at provider construction:

BfTreeProviderParameters::validate::<T>() checks that cb_max_record_size is large enough for both the vector and neighbor stores before any BfTree is created. For the quant store, validation is performed inside the Poly<dyn Quantizer> impl of CreateQuantProvider::create, where quantizer.bytes() is known — no separate call needed.

Defense in depth at every insert site:

All insert() calls in vectors.rs, neighbors.rs, and quant.rs now match on LeafInsertResult::InvalidKV and propagate a clear ANNError. This catches cases where validation is bypassed (e.g., loading from a snapshot with a changed config).

Error message example:

 vector_provider_config: cb_max_record_size (1952) is too small;
 a record requires 6152 bytes (8-byte key + 6144-byte value);
 increase cb_max_record_size to at least 6152

Fixes #1159. Supersedes #1164 

Testing

 - test_validate_rejects_undersized_vector_config — 1536D f32 + default config
 - test_validate_rejects_undersized_neighbor_config — max_degree=500 exceeding default
 - test_validate_accepts_valid_config — properly sized config passes

### DIFF
--- a/diskann-bftree/src/neighbors.rs
+++ b/diskann-bftree/src/neighbors.rs
@@ -34,6 +34,10 @@ impl<I: VectorId> HasId for NeighborProvider<I> {
 impl<I: VectorId> NeighborProvider<I> {
     /// Create a new instance based on bf-tree Config directly
     pub fn new_with_config(max_degree: u32, config: Config) -> ANNResult<Self> {
+        let key_size = std::mem::size_of::<u32>();
+        let value_size = (max_degree as usize + 1) * std::mem::size_of::<u32>();
+        crate::validate_record_size("neighbor_provider", &config, key_size, value_size)?;
+
         let adj_list_index = BfTree::with_config(config, None).map_err(ConfigError)?;
 
         Self::new(max_degree, adj_list_index)
--- a/diskann-bftree/src/provider.rs
+++ b/diskann-bftree/src/provider.rs
@@ -106,7 +106,6 @@ use diskann_providers::storage::{LoadWith, SaveWith, StorageReadProvider, Storag
 ///     dim: 4,
 ///     metric: Metric::L2,
 ///     max_degree: 32,
-///     graph_slack_factor: 1.3,
 ///     vector_provider_config: Config::default(),
 ///     quant_vector_provider_config: Config::default(),
 ///     neighbor_list_provider_config: Config::default(),
@@ -161,7 +160,6 @@ use diskann_providers::storage::{LoadWith, SaveWith, StorageReadProvider, Storag
 ///     dim: 4,
 ///     metric: Metric::L2,
 ///     max_degree: 32,
-///     graph_slack_factor: 1.3,
 ///     vector_provider_config: Config::default(),
 ///     quant_vector_provider_config: Config::default(),
 ///     neighbor_list_provider_config: Config::default(),
@@ -215,14 +213,10 @@ pub struct BfTreeProviderParameters {
     // The metric to use for distance computations
     pub metric: Metric,
 
-    // The logical maximum degree (target neighbor count per vertex).
-    // The physical neighbor list capacity is computed internally as
-    // `(max_degree * graph_slack_factor) as u32`.
+    // The physical maximum degree (maximum neighbor list capacity per vertex).
+    // Callers are responsible for applying any slack factor externally before
+    // passing this value.
     pub max_degree: u32,
-
-    // Slack factor applied to `max_degree` to determine the physical neighbor
-    // list capacity. Matches the graph construction convention (default: 1.3).
-    pub graph_slack_factor: f32,
 
     // bf-tree config for vector provider.
     // bf-tree requires a minimum circular buffer size relative to leaf_page_size:
@@ -265,8 +259,6 @@ where
         Self: StartPoint<T>,
         TQ: CreateQuantProvider<Target = Q>,
     {
-        let physical_max_degree = (params.max_degree as f32 * params.graph_slack_factor) as u32;
-
         Ok(Self {
             quant_vectors: quant_precursor.create(params.quant_vector_provider_config)?,
             full_vectors: VectorProvider::new_with_config(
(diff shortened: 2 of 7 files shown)


hildebrandmw (commented):
I'm a bit worried that validating like this isn't particularly robust long-term. For example, hooking the validation into `CreateQuantProvider` means someone can still call `QuantVectorProvider::new_with_config` without validation. Would it be better to validate directly in the provider constructors since then there will be one source of truth?

If there's a concern that validating in a constructor is too late (in the case of the top level provider, we may want to validate before creating anything to avoid unnecessary work), there can be custom config states for each sub-provider that does the prevalidation.

Just some thoughts - if you're happy with this direction, I'm happy to approve.

JordanMaples:
> I'm a bit worried that validating like this isn't particularly robust long-term. For example, hooking the validation into `CreateQuantProvider` means someone can still call `QuantVectorProvider::new_with_config` without validation. Would it be better to validate directly in the provider constructors since then there will be one source of truth?
> 
> If there's a concern that validating in a constructor is too late (in the case of the top level provider, we may want to validate before creating anything to avoid unnecessary work), there can be custom config states for each sub-provider that does the prevalidation.
> 
> Just some thoughts - if you're happy with this direction, I'm happy to approve.

I think ultimately validation in each of the constructors will be the most maintainable. I've updated the PR with constructor validations.

diskann-bftree/src/neighbors.rs
```diff
@@ -34,6 +34,10 @@ impl<I: VectorId> HasId for NeighborProvider<I> {
 impl<I: VectorId> NeighborProvider<I> {
     /// Create a new instance based on bf-tree Config directly
     pub fn new_with_config(max_degree: u32, config: Config) -> ANNResult<Self> {
+        let key_size = std::mem::size_of::<u32>();
+        let value_size = (max_degree as usize + 1) * std::mem::size_of::<u32>();
```
harsha-simhadri, line 38: Is graph slack factor already included in the value size?

harsha-simhadri (approved):

diskann-bftree/src/neighbors.rs
JordanMaples: I had it to where it expected a pre-slacked size to be passed in. I changed it to where the provider calculates the slacked size instead. It should prevent misconfiguration.

diskann-bftree/src/provider.rs
```diff
@@ -213,9 +215,15 @@ pub struct BfTreeProviderParameters {
...
 
-    // The maximum number of neighbors to store for each vector
+    // The logical maximum degree (target neighbor count per vertex).
+    // The physical neighbor list capacity is computed internally as
+    // `(max_degree * graph_slack_factor) as u32`.
     pub max_degree: u32,
 
+    // Slack factor applied to `max_degree` to determine the physical neighbor
+    // list capacity. Matches the graph construction convention (default: 1.3).
+    pub graph_slack_factor: f32,
```
hildebrandmw: Suggestion: can we keep the parameters for bftree's provider the physical maximum degree? I know the `SimpleNeighborProviderAsync` takes the slack factor, but I think that was a big mistake since it conflates how the neighbor provider is being used and requires reverse engineering what the slack factor actually is.

In short: let's drop `graph_slack_factor` from the config and ensure that `max_degree` is the physical maximum degree.

hildebrandmw (commented):
Thanks Jordan! This will really help catch early misconfigurations! Just a quick comment on the slack factor. I know the current in-memory passes around the slack factor, but let's try to move away from that approach.

diskann-bftree/src/neighbors.rs
hildebrandmw: Can we change it to accept the true max-degree and not have to worry about pushing knowledge of the slack factor down into the provider?

diskann-bftree/src/provider.rs
JordanMaples: done

hildebrandmw (approved):

PR #1168, opened 2026-06-15
diskann-benchmark refactor pt1 - mechanical move

JordanMaples:
Following up on https://github.com/microsoft/DiskANN/pull/1142#pullrequestreview-4498885825, I'm breaking #1142 into two prs. 
Pt 1 (this) is a purely mechanical module reorganization with no behavioral changes:

- Move backend/index/ contents to top-level index/ module
- Create index/inmem/ submodule for product, scalar, spherical backends
- Update all crate::backend::index:: paths to crate::index::
- Widen pub(super) to pub(crate) where items are now used across submodule boundaries
- Register index benchmarks directly from main.rs instead of through backend/mod.rs


hildebrandmw (commented):
Thanks for splitting this out, it feels partially done: the `index` module gets flattened to the top level, but the rest (`disk_index`, `exhaustive`, etc.) still remain buried one level deeper.

Does it make more sense to flatten everything up into the top level and get rid of `backend` entirely? Having the inputs be separate is kind of annoying in that there is still a "special" module, but that seems like it would be more consistent.

JordanMaples:
> Thanks for splitting this out, it feels partially done: the `index` module gets flattened to the top level, but the rest (`disk_index`, `exhaustive`, etc.) still remain buried one level deeper.
> 
> Does it make more sense to flatten everything up into the top level and get rid of `backend` entirely? Having the inputs be separate is kind of annoying in that there is still a "special" module, but that seems like it would be more consistent.

shouldn't be too hard to flatten out the backend

hildebrandmw (approved):

magdalendobson (approved):

PR #1169, opened 2026-06-15
diskann-benchmark refactor pt2 - shared logic overhaul

JordanMaples:
Follow-up to #1168 (module reorganization). This PR contains the behavioral changes that were previously combined in #1142.

  ## Summary

  Simplifies the streaming benchmark infrastructure, unifies bf-tree input types, and makes `Managed` ID translation opt-in — preparing for its removal with inmem 2.0.

  ## Changes

  ### Managed ID translation is now opt-in

  - `run_streaming` and the streaming stack are now generic over the stream type via a new `StreamingOutput` trait
  - **bftree**: implements `Stream<DataArgs>` directly on `StreamRunner`, using tags as slot IDs with no `Managed` wrapper (no ID translation overhead)
  - **inmem**: continues using `Managed` for tag-to-slot ID translation until inmem 2.0 handles this natively
  - Removed `SlotReclaim::Immediate` and `recycle_tags` (bftree-only dead code)
  - Added `build_direct_streamer` for non-managed paths alongside the existing `build_managed_streamer`

  ### Shared streaming execution layer

  - Introduce `StreamRunner`, `Managed`, `build_managed_streamer`, and `run_streaming` — a reusable stack shared across in-mem and bf-tree streaming benchmarks, replacing duplicated per-backend logic
  - Add `InmemMaintainer` (DropDeleted + Release) extracted from the old inline implementation

  ### Streaming search overhaul (addresses @magdalendobson feedback on #1142)

  - New `StreamingSearchParams` struct with scalar `num_threads` / `search_l` — streaming runs a single search config between runbook stages rather than sweeping a parameter matrix
  - Remove the vestigial groundtruth field that was validated but never read during streaming execution

  ### Input validation

  - Validate `search_l >= search_n` in `StreamingSearchParams` and `GraphSearch`
  - Validate `recall_k <= search_n` in `StreamingSearchParams` and `GraphSearch` to fail fast before long benchmark runs

  ### Input unification

  - Rename Dynamic → Streaming for runbook params and input types (`graph-index-dynamic` → `graph-index-stream-run`)
  - Unify `BfTreeBuild` and `BfTreeStreamingRun` to use a `QuantConfig` enum (`none` | `spherical`) instead of separate input structs per quantization mode
  - Extract `quantizer_util` for shared spherical quantizer construction
  - Restore `InlineFilterSearchPhase` naming (accidentally overwritten)

  ### Housekeeping

  - Simplify `PhantomData<fn() -> T>` to `PhantomData<T>` on `StreamRunner`
  - Delete `streaming_utils.rs` and `streaming/full_precision.rs` (subsumed by the shared layer)
  - Add new example: `graph-index-bftree-stream-spherical.json`
  - Update README with current command syntax

  ## Breaking changes to JSON inputs

  Streaming input files need updating:

  - `"type": "graph-index-dynamic"` → `"graph-index-stream-run"`
  - `"search_phase"` → `"search"` (flat struct with `queries`, `reps`, `num_threads`, `search_l`, `search_n`, `recall_k`)
  - Bf-tree inputs now require `"quantization": { "kind": "none" }` or the spherical config


diskann-benchmark/src/inputs/bftree.rs
```diff
@@ -209,129 +211,65 @@ fn bftree_parameters_from(
...
+    pub(crate) fn validate(&mut self) -> anyhow::Result<()> {
+        match self {
+            Self::None => Ok(()),
+            Self::Spherical {
+                num_bits,
+                quant_store_config,
+                ..
+            } => {
+                match num_bits.get() {
+                    1 | 2 | 4 => {}
```
magdalendobson: From meeting: use enum instead of matching here, use serde::try_from to convert

diskann-benchmark/src/inputs/graph_index.rs
```diff
@@ -218,7 +218,7 @@ impl BetaSearchPhase {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
-pub(crate) struct MultihopFilterSearchPhase {
+pub(crate) struct MultiHopSearchPhase {
```
magdalendobson: Undo overwrite of renaming changes

diskann-benchmark/src/inputs/bftree.rs
```diff
@@ -347,201 +285,81 @@ impl BfTreeDynamicRun {
...
+        writeln!(f, "Graph Index Bf_Tree Build")?;
         if cfg!(not(feature = "bftree")) {
             writeln!(f, "Requires the `bftree` feature")?;
         }
         write_field!(f, "tag", Self::tag())?;
-        write_field!(f, "num_bits", self.num_bits)?;
-        write_field!(f, "seed", self.seed)?;
-        write_field!(f, "transform_kind", self.transform_kind)?;
+
+        if let QuantConfig::Spherical {
```
magdalendobson: Use non-exhaustive match or Display for QuantConfig directly here instead of "if let"

diskann-benchmark/src/inputs/graph_index.rs
JordanMaples: done

diskann-benchmark/src/inputs/graph_index.rs
```diff
@@ -1330,25 +1416,78 @@ impl std::fmt::Display for DynamicRunbookParams {
...
+
+impl StreamingSearchParams {
+    pub(crate) fn max_k(&self) -> usize {
+        self.recall_k
+    }
+
+    pub(crate) fn validate(&mut self, checker: &mut Checker) -> Result<(), anyhow::Error> {
+        self.queries.resolve(checker)?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
magdalendobson: Good comment

diskann-benchmark/src/inputs/graph_index.rs
```diff
@@ -259,7 +259,7 @@ impl AdaptiveL {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
-pub(crate) struct InlineFilterSearchPhase {
```
magdalendobson, line 262: Sorry I missed this earlier, but it looks like you also undid the previous naming changes for `InlineFilterSearchPhase`.

diskann-benchmark/src/inputs/graph_index.rs
```diff
@@ -1398,26 +1484,85 @@ impl std::fmt::Display for DynamicRunbookParams {
...
+///
+/// Streaming runs a single search configuration between each runbook stage rather than
+/// sweeping a parameter matrix, so thread count and search_l are scalars.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct StreamingSearchParams {
+    pub(crate) queries: InputFile,
+    pub(crate) reps: NonZeroUsize,
+    pub(crate) num_threads: NonZeroUsize,
+    pub(crate) search_l: usize,
+    pub(crate) search_n: usize,
```
magdalendobson, line 1492: Do we also need to check that search_k <= search_n?

magdalendobson: Ah, ok, I see that this is done at the recall calculation level. Optional for this PR, but it would be better to check this when params are constructed so that you fail before you potentially wait a long time for your benchmark to run.

magdalendobson (changes requested):
Looks close, just a couple issues in comments.

diskann-benchmark/src/inputs/graph_index.rs
JordanMaples: I'll fix that one

diskann-benchmark/src/inputs/graph_index.rs
JordanMaples: early validation is a good call, i'll add that.

diskann-benchmark/src/index/streaming/runner.rs
```diff
@@ -0,0 +1,215 @@
...
+{
+    index: Arc<DiskANNIndex<DP>>,
+    strategy: S,
+    search: StreamingSearchParams,
+    runtime: Runtime,
+    ntasks: NonZeroUsize,
+    inplace_delete_num_to_replace: usize,
+    inplace_delete_method: InplaceDeleteMethod,
+    maintainer: M,
+    _marker: PhantomData<fn() -> T>,
```
hildebrandmw: Why is this invariant on `T`?

hildebrandmw (commented):
Two big concerns here:
1. Moving from `pub(crate)` members to `pub(crate)` accessors adds a lot of noise to the diff without meaningfully increasing protection. The real danger with sharing inputs is having backends ignore fields entirely, which per-field accessors doesn't really change.
2. This PR centers around `ManagedStream`, but that is a crutch for providers not being able to manage their own ID translation. The upcoming inmem 2.0 (which can do everything except make it to a PR ready state), will not need this management layer. In which case, will a separate streaming implementation be needed anyways? Is there a way to share this infrastructure without requiring `ManagedStream`?

diskann-benchmark/src/index/streaming/runner.rs
JordanMaples: Switched it to `PhantomData<T>`

hildebrandmw (commented):
Something was bothering me about removing the ID translation layer for bf-tree, and I think I managed to figure out what it is. For some reason, bf-tree is pre-configure with a [maximum capacity](https://github.com/microsoft/DiskANN/blob/ab967c1dfdd761eb3b7b77de7fd34777a3d773dd/diskann-bftree/src/vectors.rs#L124-L128). The BAB streaming runs [simulate](https://github.com/microsoft/DiskANN/blob/ab967c1dfdd761eb3b7b77de7fd34777a3d773dd/diskann-benchmark-core/src/streaming/executors/bigann/runbook.rs#L149-L150) the runbook to obtain the maximum number of active points, which could be less than the maximum tag. It's this maximum that is provided to the `BfTreeProvider`'s constructor. So, for runbooks with an active set that is indeed less, then the untranslated IDs will trigger the assert linked earlier. The runbook in `test_data` doesn't hit this because it doesn't have the property of the active points being less than the maximum tag.

Did this latest iteration go through the [wikipedia runbook](https://github.com/microsoft/DiskANN/pull/1106#issuecomment-4605332147).

A small example runbook that triggers this situation:
```yaml
sift-small-256:
  max_pts: 128
  1:
    operation: "insert"
    start: 0
    end: 128
  2:
    operation: "search"
  3:
    operation: "delete"
    start: 0
    end: 64
  4:
    operation: "insert"
    start: 128
    end: 192
  5:
    operation: "search"
  6:
    operation: "replace"
    tags_start: 64
    tags_end: 128
    ids_start: 192
    ids_end: 256
  7:
    operation: "search"
```

magdalendobson (approved):

PR #1170, opened 2026-06-15
Create benchmarks for the flat index (full-precision in-memory implementation)

### LINKED ISSUE #1002: Create benchmarks for the flat index (in-memory implementation)
We need a baseline to start the optimization journey with.

arrayka:
Establishes a baseline benchmark for `FlatIndex::knn_search`, enabling evaluation of new capabilities and innovations in DiskANN's flat-scan implementation.

### Why Add Flat Index Benchmarks if Exhaustive Exists?

**Exhaustive benchmarks** are lightweight benchmarks focused on quantizer speed and accuracy: they measure compress + distance performance in isolation, without involving the `FlatIndex` machinery or any data provider. They operate entirely outside the DiskANN provider/index abstraction.

**Flat benchmarks** are end-to-end benchmarks that exercise `FlatIndex` with different strategy configurations. They exercise `FlatIndex` through the actual provider and search strategy abstractions, so they capture realistic runtime behavior, and integration overhead that exhaustive benchmarks intentionally skip.

This PR enables full-precision in-memory flat search. Future work may add:
  - Quantized flat search
  - Quantized flat search with reranking (disk-based or in-memory)
  - Batched SIMD scan
  - Integration with diversity search

### Changes
 - New `diskann-benchmark/src/flat` benchmark module with:
   - `InMemProvider` wrapping in-memory `Matrix<T>`
   - `ScanStrategy` and `Visitor` implementing `SearchStrategy`/`DistancesUnordered`
   - `Searcher` implementing the `Search` trait from `benchmark_core`
 - Input schema (`diskann-benchmark/src/inputs/flat.rs`) with dataset, distance metric, queries, groundtruth, k, thread counts, and reps
 - Integration test (`flat_search_integration`) and example input (`diskann-benchmark/example/flat-index.json`)
 - Performance test input for wikipedia-100K dataset (`diskann-benchmark/perf_test_inputs/wikipedia-100K-flat-index.json`)
 - Unrelated fix: renamed temp file in `run_integration_test` from `graph-index.json` to `input.json` - the helper is shared by all backends, not just graph-index

Follows established patterns from the graph-index and exhaustive backends.

## Output example:
```
cargo run --package diskann-benchmark --release -- run --input-file diskann-benchmark\perf_test_inputs\wikipedia-100K-flat-index.json --output-file ./target/tmp/flat-index-output.json     
                                                                                                                                                                                
######################
# Running Job 1 of 1 #
######################

              Data: target/tmp\wikipedia_cohere/wikipedia_base.bin.crop_nb_100000
         Data Type: float32
          Distance: inner_product
           Queries: target/tmp\wikipedia_cohere/wikipedia_query.bin
       Groundtruth: target/tmp\wikipedia_cohere/wikipedia-100K
                 K: 100
           Threads: 4, 8
              Reps: 1

Loading dataset...
  Loaded 100000 vectors of dimension 768
  Queries: 5000, Groundtruth: 5000x100


  K,   Avg cmps,   QPS - mean(max),             Avg Latency,           p99 Latency,               Recall,   Threads
===================================================================================================================
100,     100000,     123.0 (123.0),   32430.0us (32430.0us),   35711.0us (35711us),   0.9999899999999999,         4
100,     100000,     250.4 (250.4),   31909.2us (31909.2us),   33323.0us (33323us),   0.9999899999999999,         8
```


diskann-benchmark/src/flat/search.rs
```diff
@@ -0,0 +1,527 @@
...
+    type Error = diskann::ANNError;
+    type Guard = NoopGuard<u32>;
+
+    fn to_internal_id(&self, _ctx: &DefaultContext, gid: &u32) -> Result<u32, Self::Error> {
+        Ok(*gid)
+    }
+
+    fn to_external_id(&self, _ctx: &DefaultContext, id: u32) -> Result<u32, Self::Error> {
+        Ok(id)
+    }
```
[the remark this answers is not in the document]
arrayka: Let's ignore this for now.

diskann-benchmark/src/flat/search.rs
```diff
@@ -0,0 +1,616 @@
...
+            p90_latencies.push(p90);
+            p99_latencies.push(p99);
+        }
+
+        let qps: Vec<f64> = results
+            .iter()
+            .map(|r| recall.num_queries as f64 / r.end_to_end_latency().as_seconds())
+            .collect();
+
+        let mean_cmps = {
```
wuw92: nit: The `mean_cmps` fold here looks identical to  `average_all`  in  `diskann-benchmark-core`. 
Worth considering promoting  `average_all`  to  `pub`  and reusing it

wuw92 (approved):

diskann-benchmark/src/flat/search.rs
```diff
@@ -0,0 +1,616 @@
...
+// FlatSearch  //
+/////////////////
+
+/// A minimal in-memory provider for flat search benchmarks.
+///
+/// Wraps a [`FastMemoryVectorProviderAsync<T>`] and implements [`DataProvider`]
+/// with identity ID mapping.
+struct InMemProvider<T: VectorRepr> {
+    data: FastMemoryVectorProviderAsync<T>,
+}
```
arkrishn94: Can we use the [full precision in-memory provider](https://github.com/microsoft/DiskANN/blob/ccf8d4ce4537c8256ad8ab1c46ede12c10e8248b/diskann-providers/src/model/graph/provider/async_/inmem/full_precision.rs#L41) here instead?

The full precision in memory provider is an instantiation of the `DefaultProvider` and `DataProvider` is [implemented for it](https://github.com/microsoft/DiskANN/blob/ccf8d4ce4537c8256ad8ab1c46ede12c10e8248b/diskann-providers/src/model/graph/provider/async_/inmem/provider.rs#L552).

You'll need to implement the right traits for it to be able to use as a flat index, but I think that's the right move here...

diskann-benchmark/src/flat/search.rs
```diff
@@ -0,0 +1,616 @@
...
+            metric,
+            _phantom: std::marker::PhantomData,
+        }
+    }
+}
+
+/// The visitor that iterates over all vectors in the provider.
+struct FlatVisitor<'a, T: VectorRepr> {
+    data: &'a FastMemoryVectorProviderAsync<T>,
+}
```
arkrishn94, line 235: Same comment as above, we should have a visitor implementation for `FastmemoryVectorProviderAsync` and that should exist along with its definition.

diskann-benchmark/src/flat/search.rs
```diff
@@ -0,0 +1,616 @@
...
+        Ok(T::query_distance(query, self.metric))
+    }
+}
+
+//////////////////////////////////////////
+// benchmark_core::search::Search impl  //
+//////////////////////////////////////////
+
+/// Wraps a [`FlatIndex`] and queries to implement the [`Search`] trait from benchmark_core.
+struct FlatSearcher<T: VectorRepr> {
```
arkrishn94: nit: do we need the `Flat` prefix over all the structs here? Can we use the module path to distinguish?

diskann-benchmark/perf_test_inputs/wikipedia-100K-flat-index.json
arkrishn94, line 1: For my understanding - what is this input provider for? I'm guessing you'll want to create a benchmark for it in CI?

arkrishn94 (commented):
Thanks for working on this Alex. Just have one non-minor comment - we should be using a provider, visitor and `DistancesUnordered` implementation as is in the benchmark. Maybe this PR can include the implementation of these for the full precision provider?

diskann-benchmark/src/flat/search.rs
hildebrandmw: I'd tentatively recommend avoiding using the constructs in `diskann-providers` given the safety issue associated with it. Is there a reason to not use `Matrix` for this simple store initially?

arrayka: @hildebrandmw, this benchmark [originally used](https://github.com/microsoft/DiskANN/pull/1170/changes/b6e742a0516eeab8b40492f4ccb382a698ce70c0) Matrix\<T\>.

Aditya and I agreed that, as an end-to-end benchmark, it should compare FlatIndex against a real data provider rather than an in-memory matrix.

The goal is to predict FlatIndex performance in a production-ready setup, similar to how we benchmark disk-based indexes, so we can predict performance in production.

diskann-benchmark/perf_test_inputs/wikipedia-100K-flat-index.json
arrayka: yes, it's the wikipedia-100K config intended to seed a CI flat-index benchmark later (mirroring the existing disk-index perf inputs). This is an actual benchmark configuration using a real dataset - not a sample configuration.

diskann-benchmark/src/flat/search.rs
arrayka: Done

diskann-benchmark/src/flat/search.rs
arrayka: Done

diskann-benchmark/src/flat/search.rs
arrayka: Let's use FastMemoryVectorProviderAsync, as proposed in this PR.

It looks like the regular MemoryVectorProviderAsync isn't accessible outside the diskann-providers crate.

We should continue the discussion on where the Flat trait implementations for existing data providers should live and come to an agreement there.

arrayka: Switched back to Matrix<T> as per offline discussion with Aditya

diskann-benchmark/src/flat/search.rs
arrayka: As discussed, let's benchmark the flat scan over Matrix<T>.

arkrishn94 (approved):

arkrishn94:
Thanks Alex, looks fine to me. Could you update the PR description to reflect the latest changes?

PR #1173, opened 2026-06-16
Implement filter search for disk path using AdpativeL

### LINKED ISSUE #1181: Filtered search support on disk path

dyhyfu:
This PR is to add filter search to disk path using AdpativeL filter algorithm.

Reference PR: https://github.com/microsoft/DiskANN/pull/1131

The main changes:
1. New filter_search() method on DiskIndexSearcher (parallel to flat_search) that constructs InlineFilterSearch::new(Knn, label_provider, adaptive_l) and hands it to the same self.index.search(...) engine as the plain Knn path.
2. New PredicateLabelProvider that adapts the existing &dyn Fn(&u32) -> bool to QueryLabelProvider<u32>
3. search() and search_internal() gain adaptive_l: Option<AdaptiveL>. When None, behavior is identical to today. When Some(_), the graph path routes through filter_search.
4. Three-way dispatch in search_internal: flat scan → graph + adaptive_l → plain Knn.


magdalendobson:
You may want to take a look at the changes that are coming to filters with https://github.com/microsoft/DiskANN/pull/1141 and make sure your changes are compatible

diskann-benchmark/src/disk_index/search.rs
```diff
@@ -264,20 +268,25 @@ where
         zipped.for_each_in_pool(
             pool.as_ref(),
             |(((((q, vf), id_chunk), dist_chunk), stats), rc)| {
-                let vector_filter = if search_params.vector_filters_file.is_none() {
-                    None
-                } else {
-                    Some(Box::new(move |vid: &u32| vf.contains(vid))
-                        as Box<dyn Fn(&u32) -> bool + Send + Sync>)
+                // Construct the plan from the JSON-driven
```
magdalendobson: Can you add an example yaml file for filtered disk search and wire up an integration test?

diskann-tools/src/utils/search_disk_index.rs
```diff
@@ -8,8 +8,12 @@ use std::{collections::HashSet, sync::atomic::AtomicBool, time::Instant};
 use diskann::utils::IntoUsize;
 use diskann_disk::{
     data_model::{CachingStrategy, GraphDataType},
-    search::provider::{
-        disk_provider::DiskIndexSearcher, disk_vertex_provider_factory::DiskVertexProviderFactory,
+    search::{
+        plan::SearchPlan,
```
magdalendobson: Is there a missing file? I can't see where SearchPlan is defined

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -888,36 +971,52 @@ where
...
+        // unfiltered hot path pays no per-node closure cost.
+        let predicate = plan.predicate();
+
+        let strategy = self.search_strategy(predicate);
         let timer = Instant::now();
         let k = k_value;
         let l = search_list_size as usize;
-        let stats = if is_flat_search {
-            self.runtime.block_on(self.flat_search(
+        let stats = match plan {
```
wuw92: The predicate is applied twice on two of the three paths. `plan.predicate()` is `Some` for `FlatScan { Some }`, `Graph { Some }`, and `InlineFilter`, and it's installed on the strategy before the match — so `RerankAndFilter` re-filters at rerank time. But:

- **FlatScan** already filters the scan iterator at `:840` before inserting into `best`, so the post-filter at `:853-855` can't reject anything.
- **InlineFilter** only pushes `on_visit == Accept` nodes into `matched_results` (`inline_filter_search.rs:189, 199-202`), so candidates reaching `post_process` are already all matches.

Only **Graph** needs the rerank-time filter — greedy traversal doesn't consult the predicate, so the post-filter is its sole enforcement point. 

Consider the following:
```rust
let (strategy, stats) = match plan {
    // Scan filters at :840 — strategy reranks only.
    SearchPlan::FlatScan { filter } => {
        let strategy = self.search_strategy(None);
        let stats = self.runtime.block_on(self.flat_search(
            &strategy, query, filter.as_deref(), l, &mut result_output_buffer,
        ))?;
        (strategy, stats)
    }
    // Greedy traversal doesn't filter — rerank is the enforcement point.
    SearchPlan::Graph { filter } => {
        let strategy = self.search_strategy(filter.as_deref());
        let knn = Knn::new(k, l, beam_width)?;
        let stats = self.runtime.block_on(self.index.search(
            knn, &strategy, &DefaultContext, query, &mut result_output_buffer,
        ))?;
        (strategy, stats)
    }
    // InlineFilterSearch filters at visit time — strategy reranks only.
    SearchPlan::InlineFilter { predicate, adaptive_l } => {
        let strategy = self.search_strategy(None);
        let knn = Knn::new(k, l, beam_width)?;
        let stats = self.runtime.block_on(self.filter_search(
            &strategy, query, knn, predicate.as_ref(), adaptive_l.clone(),
            &mut result_output_buffer,
        ))?;
        (strategy, stats)
    }
};
```
We can also remove `plan.predicate()` after that

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -213,7 +214,11 @@ where
 {
     // This needs to be Arc instead of Rc because DiskSearchStrategy has "Send" trait bound, though this is not expected to be shared across threads.
     io_tracker: IOTracker,
-    vector_filter: &'a (dyn Fn(&u32) -> bool + Send + Sync), // Fn param is u32 as we validate "VectorIdType = u32" everywhere in this provider in trait bounds.
+    /// Borrowed predicate from `SearchPlan::predicate()`. `None` means
+    /// accept-all — downstream consumers short-circuit via `.map_or(true, ...)`
+    /// and skip the dyn-fn call entirely. Fn param is `u32` because
+    /// `VectorIdType = u32` is enforced by trait bounds throughout this provider.
+    vector_filter: Option<&'a (dyn Fn(&u32) -> bool + Send + Sync)>,
```
wuw92: this field is consumed only by `default_post_processor()` → `RerankAndFilter`. The name `vector_filter` reads broader than that — worth renaming to `rerank_filter` / `post_filter` so it states *when* it applies.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -825,16 +864,56 @@ where
...
+            .await
+    }
+
     /// Perform a search on the disk index.
     /// return the list of nearest neighbors and associated data.
+    ///
+    /// The algorithm + filter combination is encoded by `plan` — see
+    /// [`SearchPlan`] for the available variants. The plan replaces the
+    /// previous `(vector_filter, is_flat_search, adaptive_l)` parameter
+    /// triple and makes invalid combinations unrepresentable.
```
wuw92: nit: "replaces the previous  (vector_filter, is_flat_search, adaptive_l)  triple" is migration context, consider removing this as no future reader has the old parameters in mind

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -877,6 +955,11 @@ where
 
     /// Perform a raw search on the disk index.
     /// This is a lower-level API that allows more control over the search parameters and output buffers.
+    ///
+    /// Dispatches on `plan` variants: `FlatScan` → linear scan; `Graph` →
+    /// plain `Knn` with the optional post-filter applied during rerank;
```
wuw92: nit:  The per-variant dispatch list ("`FlatScan` → linear scan; `Graph` → plain `Knn`...") is a prose mirror of the `match` block directly below it, consider simplifying it.

diskann-tools/src/utils/search_disk_index.rs
```diff
@@ -8,8 +8,12 @@ use std::{collections::HashSet, sync::atomic::AtomicBool, time::Instant};
 use diskann::utils::IntoUsize;
```
wuw92, line 8: Worth noting this whole module `diskann-tools/src/utils/search_disk_index.r `looks dead. pub fn search_disk_index / SearchDiskIndexParameters aren't referenced from any binary.
I recommend deleting `search_disk_index.rs` (and its pub mod / pub use in utils/mod.rs) in a follow-up rather than carry the filter search migration.

wuw92 (changes requested):

diskann-tools/src/utils/search_disk_index.rs
dyhyfu: added

diskann-benchmark/src/disk_index/search.rs
dyhyfu: added "diskann-benchmark/example/disk-index-filter.json" for this.

diskann-disk/src/search/provider/disk_provider.rs
dyhyfu: renamed to "postprocess_filter"

diskann-disk/src/search/plan.rs
```diff
@@ -0,0 +1,155 @@
...
+/// Three variants encode the algorithm + filter combination:
+///
+/// * `FlatScan` — brute-force linear scan, with or without an inline filter.
+/// * `Graph` — plain greedy beam search; the optional filter is applied as a
+///   hard post-filter during reranking (no traversal-time effect).
+/// * `InlineFilter` — label-filtered graph search; the predicate is consulted
+///   at visit time (not just during rerank). `adaptive_l = Some(_)` grows the
+///   beam mid-search if the observed match specificity is low.
+
+pub enum SearchPlan<'a> {
```
wuw92: nit:  `Plan`  tends to read as a query/execution plan: something the engine generates and optimize. Consdier naming `SearchMode`  or  `SearchKind`  which would line up more directly with what the variants express

dyhyfu: Renamed to SearchMode

wuw92 (approved):

diskann-benchmark/src/disk_index/search.rs
```diff
@@ -264,20 +268,23 @@ where
         zipped.for_each_in_pool(
             pool.as_ref(),
             |(((((q, vf), id_chunk), dist_chunk), stats), rc)| {
-                let vector_filter = if search_params.vector_filters_file.is_none() {
-                    None
-                } else {
-                    Some(Box::new(move |vid: &u32| vf.contains(vid))
-                        as Box<dyn Fn(&u32) -> bool + Send + Sync>)
+                // Construct the mode from the JSON-driven
```
magdalendobson: Can we expose adaptive L by creating a json deserializable struct here that has a function returning the SearchMode object? It's a bit awkward to just construct it from params here. If we had a dedicated struct it would be much easier to build on what's already here.

magdalendobson:
This looks fine to me pending a couple of things:

- Let's get the `benchmark` integration a little smoother by creating a wrapper struct that's json deserializable and constructs the `SearchMode` directly
- Can you include a few experiments with larger filtered datasets as a sanity check for correctness/performance here?

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -267,40 +280,53 @@ impl IOTracker {
...
+    pub fn new(filter: Option<PostprocessFilter<'a>>) -> Self {
         Self { filter }
     }
 }
 
+/// Adapter exposing the disk-side `&dyn Fn(&u32) -> bool` predicate as a
+/// `QueryLabelProvider<u32>` for `InlineFilterSearch`.
+struct PredicateLabelProvider<'a> {
+    predicate: &'a (dyn Fn(&u32) -> bool + Send + Sync),
+}
```
hildebrandmw: Suggestion: Injecting the predicate this way introduces two levels of dynamic dispatch for each predicate evaluation: one for the `&dyn QueryLabelProvider` used in `diskann::graph::ext::labeled` and one for the inner `predicate` type listed here.

Fortunately, this PR introduces the tools needed to solve this: In `SearchMode`, instead of using `SearchPredicate`, you can do something like this (note this is a sketch):
```rust
enum SearchMode<'a> {
    ...
    InlineFilter {
        predicate: Box<dyn QueryLabelProvider<u32> + 'a>,
        ...
    },
    ...
}

impl<'a> SearchMode<'a> {
    fn inline_filter<F>(predicate: F, adaptive_l: Option<AdaptiveL>) -> Self 
    where
        F: Fn(&u32) -> bool + ...
    {
        struct Wrap<F>(F);
        impl<F> QueryLabelProvider<u32> for Wrap<F> 
        where
           F: Fn(&u32) -> bool + ...
        {
            fn is_match(&self, i: &u32) -> bool {
                (self.0)(i)
            }
        }
    }

    Self::InlineFilter {
        predicate: Box::new(Wrap(f)),
        adaptive_l,
    }
}
```
This means that the `QueryLabelProvider` implementation dispatches directly to the filter and saves a level of indirection.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -337,7 +363,9 @@ where
         let mut uncached_ids = Vec::new();
         let mut reranked = candidates
             .map(|n| n.id)
-            .filter(|id| (self.filter)(id))
+            // `None` short-circuits to `true` — no dyn-fn call when there's
+            // no predicate set, which is the common unfiltered hot path.
+            .filter(|id| self.filter.is_none_or(|f| f(id)))
```
arkrishn94: nit: maybe match on `self.filter`? So something like - 
```rust 
let mut reranked = match self.filter {
    None => ..., 
    Some(p) => ....
};
```

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -1055,55 +1116,97 @@ where
...
+        // * `FlatScan`     — `flat_search` filters the scan iterator at
+        //                    construction; non-matching IDs never enter `best`.
+        // * `Graph`        — plain greedy traversal doesn't consult any predicate;
+        //                    if a predicate is set, `RerankAndFilter` filters out
+        //                    non-matching nodes at rerank time.
+        // * `InlineFilter` — `InlineFilterSearch` only forwards `Accept` nodes
+        //                    into `matched_results`; no filtering in post-process.
+        // * `DiverseGraph` — `index.search_with` runs `DeterminantDiversityAndFilter`
+        //                    as the post-processor over the L candidate pool.
+        let stats = match mode {
```
arkrishn94, line 1130: I'm probably missing something, but remind me - why are we not using the `Search` APIs [here](https://github.com/microsoft/DiskANN/blob/main/diskann/src/graph/search/mod.rs) to dispatch to the different flavors of search here?

arkrishn94 (commented):
Thanks @dyhyfu, just left a couple of comments.

diskann-disk/src/search/provider/disk_provider.rs
dyhyfu: I think we are using the search API. 
The `DiskIndexSearcher::search(query, k, l, bw, mode: SearchMode) `call is the centralized user-facing API for disk search. Behind it, the `match mode` constructs the right setup per variant and then dispatches to corresponding Search API (`index.search(SomeSearchImpl, ...)` or `index.search_with(..., processor)`). The per-variant setup differs in three ways the Search trait can't express on its own:

- **Which Search impl to construct** (`Knn `for Graph/DiverseGraph, `InlineFilterSearch `for InlineFilter).

- **Which strategy type to pass**. `InlineFilterSearch` is bounded on `SearchAccessor: FilteredAccessor`, but `DiskAccessor `only implements `SearchAccessor `— so we need `labeled::Filtered::new(strategy, label_provider)` from #1141 to bridge the gap. Other arms pass the bare strategy.

- **Which entry point to call** (`search `vs `search_with`) depending on whether the post-processor is the strategy default or an explicit one (DeterminantDiversity).

`FlatScan `is the outlier — it's a linear PQ scan, not a graph algorithm, so it bypasses Search entirely.

diskann-disk/src/search/provider/disk_provider.rs
dyhyfu: Thanks for the tip! Updated.

hildebrandmw (approved):

diskann-benchmark/src/disk_index/search.rs
suri-kumkaran: Yes, can we please update the `disk-index` benchmark input to better align with the downstream public API? Specifically, the JSON input could mirror the `SearchMode` structure.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -1055,55 +1094,97 @@ where
...
+                );
+                self.runtime.block_on(self.index.search_with(
+                    knn_search,
+                    &strategy,
+                    processor,
+                    &DefaultContext,
+                    query,
+                    &mut result_output_buffer,
+                ))?
+            }
```
suri-kumkaran, line 1192: The comments explaining the pipeline here are fantastic. Since we currently rely on these comments to clarify what `None` implies for different branches, we could make this obvious through the design itself!

Since this is specifically about the rerank phase, what if we used a domain-specific enum instead of `Option`? It bakes your "accept-all" logic directly into the type system and makes the call sites completely self-documenting:

```rust
enum PostprocessConfig<'a> {
    /// Accept-all at rerank time (used by FlatScan, InlineFilter)
    AcceptAll,
    /// Apply a specific predicate at rerank time (used by Graph)
    Apply(PostprocessFilter<'a>),
}

// SearchMode::FlatScan { ... } =>
let strategy = self.search_strategy(&io_tracker, PostprocessConfig::AcceptAll);

// SearchMode::Graph { filter } =>
let strategy = self.search_strategy(&io_tracker, PostprocessConfig::Apply(filter));

```
Just a thought to let the type system do the heavy lifting for the documentation!
It makes the call sites perfectly self-documenting and scales easily if we ever add a third postprocess config.

diskann-disk/src/search/search_mode.rs
```diff
@@ -0,0 +1,211 @@
...
+    InlineFilter {
+        predicate: Box<dyn QueryLabelProvider<u32> + 'a>,
+        adaptive_l: Option<AdaptiveL>,
+    },
+
+    DiverseGraph {
+        filter: Option<SearchPredicate<'a>>,
+        params: DeterminantDiversityParams,
+    },
+}
```
suri-kumkaran, line 54: nit: Inconsistent field naming between variants

Just a minor readability observation: most of the variants (`FlatScan`, `Graph`, `DiverseGraph`) name this field `filter`, but `InlineFilter` names it `predicate`.

```rust
pub enum SearchMode<'a> {
    FlatScan { filter: Option<SearchPredicate<'a>> },
    // ...
    InlineFilter { predicate: Box<dyn QueryLabelProvider<u32> + 'a>, ... },
}

```

Since they functionally represent the same concept (the yes/no rule applied to the search), it might reduce mental friction to standardize the field names across the board—either all `filter` or all `predicate`. Totally up to you, feel free to ignore if you prefer the distinction here!
If we are fixing, we can update the related comments as well.

suri-kumkaran (approved):
We are good to go after fixing `disk-index` benchmark input and moving towards `PostprocessConfig`.

diskann-disk/src/search/search_mode.rs
dyhyfu: Good catch, aligned the naming!

diskann-disk/src/search/provider/disk_provider.rs
dyhyfu: Added PostprocessStrategy for it.

diskann-tools/src/utils/search_disk_index.rs
dyhyfu: will be removed in https://github.com/microsoft/DiskANN/pull/1185

wuw92 (approved):

diskann-benchmark/src/inputs/disk.rs
```diff
@@ -6,8 +6,15 @@
 use std::{fmt, num::NonZeroUsize, path::Path};
 
 use anyhow::Context;
+#[cfg(feature = "disk-index")]
```
suri-kumkaran, line 9: Any reason we have everything gated by `#[cfg(feature = "disk-index")]`?

PR #1184, opened 2026-06-17
Make diskann-disk runtime-agnostic with async APIs

doliawu:
## Summary

Makes `diskann-disk` runtime-agnostic. Previously the crate created and owned its own tokio runtime (a `create_runtime` helper, a `Runtime` field on `DiskIndexSearcher`, and `block_on` calls inside `build()`/`search()`). This forced a runtime on callers, risked nested-runtime panics, and pulled in tokio's `full` feature set. The crate now exposes `async` APIs and lets callers drive them on their own executor.

## Library changes (`diskann-disk`)

- `DiskIndexBuilder::build()` is now `async`; the redundant `build_internal` is inlined (`BuildInterrupted` handling preserved via an inner `async` block). Deleted `build/builder/tokio.rs` (`create_runtime`).
- `DiskIndexSearcher` no longer holds a `Runtime`; `new()` drops the `Option<Runtime>` parameter; `search()`/`search_internal()` are `async` and await the inner futures directly.
- Trimmed tokio features from `["full"]` to `["rt", "rt-multi-thread", "macros", "sync"]`.
- Tests: single-call search tests use `#[tokio::test]`; the two concurrent rayon tests share one runtime `Handle`; build helpers drive the async build via a local runtime.

## Consumer changes (`diskann-tools`, `diskann-benchmark`)

- Search loops replace rayon `for_each_in_pool` with `futures_util::buffer_unordered(num_threads)` on a caller-owned runtime.
- Build paths wrap `build()` in `block_on` on a multi-thread runtime.

## Validation

- `cargo check --workspace` clean.
- `diskann-disk`: 271 lib tests + 2 doc tests pass.
- `cargo clippy -D warnings` clean on `diskann-disk`, `diskann-tools`, and `diskann-benchmark` (`disk-index` feature).
- `cargo fmt --all` clean.

## Note on performance

The consumer search loops now use single-task `buffer_unordered` concurrency rather than the previous `num_threads`-wide rayon pool. This yields concurrency (interleaving at await points) but not CPU parallelism; for the CPU-bound search path this may reduce multi-core QPS in the benchmark. Restoring true parallelism would require `Arc<searcher>` + `JoinSet`/`tokio::spawn` with `Send + Sync + 'static` bounds on the generic search functions.

🤖 Generated with [GitHub Copilot CLI](https://github.com/github/copilot-cli)


diskann-tools/src/utils/build_disk_index.rs
```diff
@@ -169,7 +169,13 @@ where
     }
 
     let timer = Timer::new();
-    disk_index.build()?;
+    let runtime = tokio::runtime::Builder::new_multi_thread()
```
wuw92: For awareness — `build_disk_index.rs` and `search_disk_index.rs` are both dead code. 
I've opened #1185 to remove them.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -1291,6 +1280,8 @@ mod disk_provider_tests {
             load_query_result(params.storage_provider, params.truth_result_file_path);
 
         let pool = create_thread_pool(params.thread_num.into_usize()).unwrap();
+        let runtime = tokio::runtime::Builder::new_multi_thread().build().unwrap();
```
wuw92, line 1566: The parallelism here comes from the rayon `for_each_in_pool`, and the search path never spawn or suspend, so the tokio workers stay idle. 
`new_current_thread()` would be enough for search, since the runtime is only here to `block_on`.

diskann-disk/src/build/builder/build.rs
```diff
@@ -205,22 +204,7 @@ where
         )
     }
 
-    pub fn build(&mut self) -> ANNResult<()> {
-        let runtime = create_runtime(self.index_configuration.num_threads)?;
```
wuw92, line 209: Worth noting: build parallelism is governed by two numbers — `num_tasks` (the tokio tasks spawned via `JoinSet`) and the runtime's worker-thread count. The old `create_runtime(num_threads)` derived both from `num_threads`, so they stayed in lockstep. 
Now the runtime is caller-owned, so the two are independent: if the caller sizes worker threads below `num_tasks`, the build still runs but with reduced parallelism. Might be worth a note on `build()` about the expected worker-thread count.

diskann-benchmark/src/disk_index/search.rs
```diff
@@ -227,12 +224,11 @@ where
         &index_reader,
         vertex_provider_factory,
         search_params.distance.into(),
-        None,
     )?;
 
     logger.log_checkpoint("index_loaded");
 
-    let pool = create_thread_pool(search_params.num_threads)?;
+    let runtime = tokio::runtime::Builder::new_current_thread().build()?;
```
wuw92: The CI benchmark shows QPS dropping ~70% on this path ([run](https://github.com/microsoft/DiskANN/actions/runs/27720261469/job/82003596523?pr=1184)).
The cause looks like the `current_thread` runtime + `buffer_unordered`: the search futures never suspend (`get_element` is `future::ready`), so on a single worker thread they run one after another — `buffer_unordered(num_threads)` gives no real parallelism here.

hildebrandmw: We may want to consider going [benchmark core](https://github.com/microsoft/DiskANN/blob/3aa44ace775d2d32e7f9d37d5a99b5f4f26d8e48/diskann-benchmark-core/src/search/api.rs#L42-L81). It will manage the `tokio` parallelism.

diskann-disk/src/search/provider/disk_provider.rs
doliawu: switched to multi-threaded runtime, the benchmark no longer shows regression

PR #1187, opened 2026-06-18
[WIP] IVF index interface

arkrishn94:
This is a major work in progress! 

To see what I'm going for, see `diskann/src/ivf/glue.rs` and `index.rs`.


diskann/src/ivf/index.rs
```diff
@@ -0,0 +1,154 @@
...
+
+            let result_count =
+                output.extend(queue.iter().take(k).map(|n| (n.id, n.distance))) as u32;
+
+            Ok(SearchStats { cmps, result_count })
+        }
+    }
+
+    /// Insert a vector under external id `id`.
+    pub fn insert<'a, S, T>(
```
magdalendobson, line 100: Do we also want `delete` in this interface?

diskann/src/ivf/index.rs
```diff
@@ -0,0 +1,154 @@
...
+        &'a self,
+        strategy: &'a S,
+        context: &'a P::Context,
+        id: &P::ExternalId,
+        vector: T,
+    ) -> impl SendFuture<ANNResult<()>>
```
magdalendobson, line 106: Should we have the concept of an index that is read-only? I am thinking of designs where the index is loaded into memory for maintenance and then queries are served from a slower provider. In that case, we don't necessarily want to support insert/delete in the slower provider.

diskann/src/ivf/index.rs
```diff
@@ -0,0 +1,154 @@
...
+    /// Borrow the underlying provider.
+    pub fn provider(&self) -> &P {
+        &self.provider
+    }
+
+    /// Run IVF k-nearest-neighbor search.
+    pub fn knn_search<'a, S, T, OB>(
+        &'a self,
+        k: NonZeroUsize,
+        nprobe: usize,
```
magdalendobson, line 52: Should `nprobe` be a `NonZeroUsize` that we enforce is greater than or equal to `k`?

Also, how will this contract deal with parameters used for searching the centroid index (i.e. L_search)? Maybe it should take in a type `CentroidSearchParams` or something similar instead

diskann/src/ivf/index.rs
```diff
@@ -0,0 +1,154 @@
...
+                .await
+                .escalate("IVF coarse list selection must complete")?;
+
+            let mut search_accessor = strategy
+                .search_accessor(&self.provider, context, query)
+                .into_ann_result()?;
+
+            let k = k.get();
+            let mut queue = NeighborPriorityQueue::new(k);
+            let mut cmps: u32 = 0;
```
magdalendobson, line 82: We should also count distance comparisons done to select centroids, right?

diskann/src/ivf/glue.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn append(
+        &mut self,
+        list: Self::ListId,
+        id: Self::Id,
+        vector: T,
+    ) -> impl SendFuture<Result<(), Self::Error>>;
+}
+
+/// Per-call factory for IVF insert.
+pub trait InsertStrategy<'a, Provider, T>: Send + Sync
```
magdalendobson, line 108: Trying to understand the vision for the dynamic case here--is the idea that we will later extend this to deletes, at which point we call it something like `DynamicStrategy`, or that there will be a separate `DeleteStrategy` object?

rfcs/01187-incremental-ivf.md
```diff
@@ -0,0 +1,820 @@
...
+- **Scan payload**: the representation stored with a list for query scoring. It
+  may be full precision, quantized, or otherwise encoded.
+- **Partition update**: the complete logical centroid and point-membership
+  change calculated by the algorithm.
+
+### Required Logical Invariants
+
+After a maintenance accessor returns `Ok(())` from `apply`:
+
+1. Every visible point is assigned to exactly one live list.
```
magdalendobson, line 147: This seems like something we do not want to deeply commit to in our architecture. Replication of points is a pretty standard technique in IVF.

rfcs/01187-incremental-ivf.md
```diff
@@ -0,0 +1,820 @@
...
+2. `assignment[point] == list` if and only if `point` is a member of `list`.
+3. Every live list has exactly one live centroid with the same logical id.
+4. Every list member has an available canonical vector and scan payload.
+5. Retired list ids are never reused for a different centroid.
```
magdalendobson, line 151: Why can't we reuse list ids? This seems like something we might want to do.

rfcs/01187-incremental-ivf.md
```diff
@@ -0,0 +1,820 @@
...
+6. Point identity mappings agree with point visibility in the partition.
+7. Inserts may trigger splits but not dissolves.
+8. Deletes may trigger dissolves but not splits.
```
magdalendobson, line 154: Do we want to commit that deletes can't trigger splits? I think this is a bit contrary to some of our discussions on a batching API, and iirc we did at least one experiment that shows that this policy would degrade recall.

diskann/src/ivf/dynamic.rs
```diff
@@ -0,0 +1,316 @@
...
+/// One point membership change in a partition mutation.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct PointMove<Id, ListId> {
+    /// Internal point id being inserted, deleted, or reassigned.
+    pub id: Id,
+    /// Previous list, or `None` for a newly inserted point.
+    pub from: Option<ListId>,
+    /// New list, or `None` for a deleted point.
+    pub to: Option<ListId>,
+}
```
magdalendobson, line 189: It might make debugging/reading code easier if we made this an enum over delete, insert, and reassign, but just a thought

rfcs/01187-incremental-ivf.md
```diff
@@ -0,0 +1,820 @@
...
+13. Consume the accessor with `apply(update)`.
+
+With one overflow parent, joint `2l`-means reduces to local 2-means. Routes are
+computed before structural change; regional reassignment deliberately revisits
+points whose routes become stale due to the split.
+
+### Split Semantics
+
+A split retires one parent and installs two children, for a net increase of one
+live cluster. Parent ids are never reused. Candidate centroids for a region are
```
magdalendobson, line 539: Why can't parent ids be reused?

rfcs/01187-incremental-ivf.md
```diff
@@ -0,0 +1,820 @@
...
+The provider may rewrite complete lists, append delta segments, or mutate
+in-memory lists. Those choices do not change `PartitionUpdate` semantics.
+
+### Batch Delete And Dissolve
+
+1. Construct one maintenance accessor and resolve every requested point's
+   current assignment.
+2. Stage deletion of each point id.
+3. Group deletes by list and compute projected post-delete sizes.
+4. Admit underfull victims without crossing `min_clusters`.
```
magdalendobson, line 552: Is a victim a centroid that is being deleted, or the points in the deleted centroid's posting list? What does it mean to "admit" them?

rfcs/01187-incremental-ivf.md
```diff
@@ -0,0 +1,820 @@
...
+7. Assign those members exactly among the saved survivor candidates.
+8. Build one update containing point deletes, retired victim centroids, and
+   reassigned victim members.
+9. Consume the accessor with `apply(update)`.
+
+A dissolve fits no centroid and does not read survivor members. Removing a
+centroid cannot make a point already assigned to another surviving centroid
+prefer the removed centroid. Deletes never trigger a split in the same
+operation. A survivor above the split threshold remains eligible for the next
+insert-driven split.
```
magdalendobson, line 565: We should add a section for a combined insert/delete batch API since that will likely be necessary. Also, I don't think we should allow deletes to over-fill a posting list. In fact, we should maintain an invariant that posting list updates passed via the `PartitionUpdate` can never exceed the maximum number of points. We can be flexible on the minimum number of points, but by being inflexible on the max we maintain openness to providers that do a fixed allocation for each posting list.

rfcs/01187-incremental-ivf.md
```diff
@@ -0,0 +1,820 @@
...
+- Add hysteresis, minimum-cluster, delete/reinsert, and overlapping-batch tests.
+
+### Phase 6: Durable Providers
+
+- Design local-disk consistency and recovery inside its concrete accessor.
+- Add crash/reopen tests around each irreversible write.
+- Design blob coordination separately rather than forcing disk and blob through
+  one generic publication protocol.
+
+## Testing Strategy
```
magdalendobson, line 685: We should add some integration tests with exact baselines using the same framework as is used for DiskANN.

diskann/src/lib.rs
```diff
@@ -15,6 +15,7 @@ pub(crate) mod internal;
 // Index Implementations
 pub mod flat;
 pub mod graph;
+pub mod ivf;
```
magdalendobson, line 18: Meta-question, do we want a separate `diskann-ivf` crate or do we want inclusion within `diskann`?

diskann/src/ivf/dynamic.rs
```diff
@@ -0,0 +1,316 @@
...
+/// batch reads, coalesce blob requests, prefetch, decode quantized payloads, or
+/// fan work out across tasks.
+pub trait SearchAccessor: HasId + Send + Sync {
+    /// Stable logical centroid/list id.
+    type ListId: VectorId;
+
+    /// Errors from list selection or scanning.
+    type Error: ToRanked + Debug + Send + Sync + 'static;
+
+    /// Select exactly the available requested number of lists for the bound query.
```
magdalendobson, line 117: This should probably also return some kind of stats object

diskann/src/ivf/index.rs
magdalendobson: Well, `>= k` maybe doesn't make sense, but maybe `n_probe * min_size >= k`? That would, however, mean we needed to enforce the minimum size strictly.

diskann/src/ivf/dynamic.rs
```diff
@@ -0,0 +1,316 @@
...
+    pub comparisons: u64,
+    /// Number of distinct lists read.
+    pub lists_scanned: u32,
+}
+
+/// An in-memory index over the live centroids.
+///
+/// The centroid catalog is authoritative. Approximate implementations, such as
+/// a graph navigator, must retain enough catalog information to reject retired
+/// ids and recover with exact selection when navigation cannot produce a usable
```
magdalendobson, line 73: We expect a graph index to handle fallback to flat scan? Why is that? That's not typical for DiskANN in general, is there a reason it would be needed for an IVF centroid index?

PR #1188, opened 2026-06-18
Introduce `diskann-record` crate to support serialization + deserialization of DiskANN indexes

suhasjs:
This PR is part 1/2 of  #1079  and only introduces the `diskann-record` crate. See #1079 for sample output when the `save::Save` and `load::Load` traits from this crate are implemented for a simple in-memory index.

#### Reference Issues/PRs
Part 1/2 of #1079. Also see #737. 

Introduces a new create `diskann-record` to support serialization + deserialization for DiskANN indexes. It provides a small, backend-agnostic framework for persisting structured Rust values as a **versioned** manifest plus side-car binary artifacts, and reloading them later. `diskann_record` is intended to be a new __foundational__ crate, so it only depends on `anyhow` (`serde` is feature-gated and only used for on-disk formats).

Summary of changes:
- Introduces two new traits (`diskann_record::save::Save` and `diskann_record::load::Load`), along with two macros `load_fields` and `save_fields` for simple, plain `struct`s.
- Records carry `Version` that allow for durable indexes: `load_legacy` explicitly supports loading serialized representations of older versions of the same struct.
- Manifest is JSON-compatible and carries a list of artifact filenames (if serialized using disk-backend)
- Implement two custom backends -- `Disk` backend that writes artifacts to individual files, and a `Memory` backend that simply serializes to a `Vec<u8>`. `Disk` backend depends on `serde`, whereas `Memory` does not.
- `diskann-record` is generic enough to allow saving with one backend and loading with another provider. It boils down to how `SaveContext`/`LoadContext` are implemented for each provider, except for `Memory` provider -- it pipes the output of its `SaveContext` directly into its `LoadContext`, so it never leaves in-process memory.

### DIFF
--- a/diskann-record/README.md
+++ b/diskann-record/README.md
@@ -1,0 +1,14 @@
+# DiskANN Record
+
+This crate provides a small framework for persisting structured Rust values as a 
+manifest (can be serialized to JSON) plus a set of side-car binary artifacts, and 
+reloading them later. It is can be used by `diskann` providers and indexes to 
+implement durable, consistent and backward-compatible checkpoints.
+
+Types describe how they map to a versioned record by implementing the `save::Save` and
+`load::Load` traits; the `save_fields!` and `load_fields!` macros handle the
+field-by-field plumbing for plain structs. Every record carries a `Version` so loaders
+can detect schema changes and either upgrade or fall back through a probing chain.
+
+The goal is to allow crates like `diskann` to checkpoint their state without depending on
+a particular serialization backend. This crate has minimal dependencies by design.
--- a/diskann-record/README.md
+++ b/diskann-record/README.md
@@ -1,9 +1,9 @@
 # DiskANN Record
 
-This crate provides a small framework for persisting structured Rust values as a 
-manifest (can be serialized to JSON) plus a set of side-car binary artifacts, and 
-reloading them later. It is can be used by `diskann` providers and indexes to 
-implement durable, consistent and backward-compatible checkpoints.
+This crate provides a small framework for persisting structured Rust values (`struct`, `enum`, 
+`Option<T>`, `Vec<T>`, and primitives) as a  manifest (can be serialized to JSON) plus a set 
+of side-car binary artifacts, and reloading them later. It is can be used by `diskann` 
+providers and indexes to implement durable, consistent and backward-compatible checkpoints.
 
 Types describe how they map to a versioned record by implementing the `save::Save` and
 `load::Load` traits; the `save_fields!` and `load_fields!` macros handle the
--- a/diskann-record/README.md
+++ b/diskann-record/README.md
@@ -11,4 +11,9 @@ field-by-field plumbing for plain structs. Every record carries a `Version` so l
 can detect schema changes and either upgrade or fall back through a probing chain.
 
 The goal is to allow crates like `diskann` to checkpoint their state without depending on
-a particular serialization backend. This crate has minimal dependencies by design.
+a particular serialization backend. Currently, this crate implements `Disk` and `Memory` backends
+-- `Disk` for persistent storage and `Memory` for in-memory operations. 
+`Memory` pipes the output of its `SaveContext` directly into the input of its `LoadContext`, so it is not compatible with other backends.
+A hypothetical `ObjectStore` backend could be implemented to be compatible with `Disk` backends to support caching and other advanced features.
+
+This crate has minimal dependencies by design.
--- a/diskann-record/src/backend/disk.rs
+++ b/diskann-record/src/backend/disk.rs
@@ -380,6 +380,76 @@ mod tests {
     }
 
     #[test]
+    fn write_rejects_preexisting_file_on_disk() {
+        let dir = tempfile::tempdir().unwrap();
+        let ctx = DiskSaveContext::new(dir.path().into(), dir.path().join("meta.json")).unwrap();
+        // The first artifact is named with a `000` count prefix; pre-create that exact file
+        // on disk so the `full.exists()` guard rejects the allocation.
+        std::fs::write(dir.path().join("000-artifact.bin"), b"stale").unwrap();
+        let err = SaveContext::write(&ctx, Some("artifact.bin"))
+            .expect_err("an artifact whose file already exists on disk must be rejected");
+        assert!(format!("{err}").contains("already exists"));
+    }
+
+    #[test]
+    fn finish_rejects_unfinished_artifact() {
+        let dir = tempfile::tempdir().unwrap();
+        let ctx = DiskSaveContext::new(dir.path().into(), dir.path().join("meta.json")).unwrap();
+        // Reserve an artifact slot but drop the writer without calling `finish`, leaving the
+        // slot marked as not-yet-finished.
+        let writer = SaveContext::write(&ctx, Some("artifact.bin")).unwrap();
+        drop(writer);
+        let err = ctx
+            .finish(Value::Null)
+            .expect_err("finish must fail when an artifact was reserved but never finished");
+        assert!(format!("{err}").contains("was reserved but never finished"));
+    }
+
+    #[test]
+    fn write_rejects_name_collision() {
+        let dir = tempfile::tempdir().unwrap();
+        let ctx = DiskSaveContext::new(dir.path().into(), dir.path().join("meta.json")).unwrap();
+        // The count prefix normally makes a collision impossible, so seed the bookkeeping map
+        // directly with the exact name the next `write` will generate (one entry => count 1 =>
+        // `001-artifact.bin`).
+        ctx.files
+            .lock()
+            .unwrap()
+            .insert("001-artifact.bin".to_string(), true);
+        let err = SaveContext::write(&ctx, Some("artifact.bin"))
+            .expect_err("a generated name that is already registered must be rejected");
+        assert!(format!("{err}").contains("collides with an existing artifact"));
+    }
+
+    #[test]
+    fn write_reports_file_creation_failure() {
+        let dir = tempfile::tempdir().unwrap();
+        let artifacts = dir.path().join("artifacts");
+        std::fs::create_dir(&artifacts).unwrap();
+        let ctx = DiskSaveContext::new(artifacts.clone(), dir.path().join("meta.json")).unwrap();
+        // Remove the validated directory so `create_new` fails with a non-"exists" IO error
+        // (the `full.exists()` guard passes because the path is gone).
+        std::fs::remove_dir(&artifacts).unwrap();
+        let err = SaveContext::write(&ctx, Some("artifact.bin"))
+            .expect_err("creating an artifact in a missing directory must fail");
+        assert!(format!("{err}").contains("while creating new file"));
+    }
+
+    #[test]
+    fn finish_reports_rename_failure() {
+        let dir = tempfile::tempdir().unwrap();
+        // Make the final manifest path an existing directory: the `<metadata>.temp` file is
+        // created and serialized fine, but renaming a file onto a directory fails.
+        let metadata = dir.path().join("meta.json");
+        std::fs::create_dir(&metadata).unwrap();
+        let ctx = DiskSaveContext::new(dir.path().into(), metadata.clone()).unwrap();
+        let err = ctx
+            .finish(Value::Null)
+            .expect_err("renaming the temp manifest onto a directory must fail");
+        assert!(format!("{err}").contains("while renaming temp manifest"));
+    }
+
+    #[test]
     fn drop_without_finish_cleans_up_artifacts() {
         let dir = tempfile::tempdir().unwrap();
         let metadata = dir.path().join("meta.json");
@@ -497,5 +567,37 @@ mod tests {
             panic!("a handle escaping the manifest directory must be rejected");
         };
         assert!(format!("{err}").contains("escapes the manifest directory"));
+    }
+
+    #[test]
+    fn read_reports_missing_artifact_file() {
+        let dir = tempfile::tempdir().unwrap();
+        // Register the artifact in the manifest but never create the file on disk, so the
+        // path/registration checks pass and only the file `open` fails.
+        let metadata = write_manifest(dir.path(), &["artifact.bin"]);
+        let ctx = DiskLoadContext::new(&metadata, dir.path()).unwrap();
+        let Err(err) = ctx.read("artifact.bin") else {
+            panic!("a registered artifact missing from disk must be reported");
+        };
+        assert!(format!("{err}").contains("while opening artifact file"));
+    }
+
+    #[test]
+    fn new_load_reports_missing_manifest() {
+        let dir = tempfile::tempdir().unwrap();
+        let metadata = dir.path().join("does-not-exist.json");
+        let err = DiskLoadContext::new(&metadata, dir.path())
+            .expect_err("a missing manifest file must be reported");
+        assert!(format!("{err}").contains("while trying to open"));
+    }
+
+    #[test]
+    fn new_load_rejects_malformed_manifest() {
+        let dir = tempfile::tempdir().unwrap();
+        let metadata = dir.path().join("metadata.json");
+        std::fs::write(&metadata, b"this is not json").unwrap();
+        let err = DiskLoadContext::new(&metadata, dir.path())
+            .expect_err("a malformed manifest must be rejected");
+        assert!(format!("{err}").contains("could not deserialize manifest"));
     }
 }
--- a/diskann-record/src/backend/disk.rs
+++ b/diskann-record/src/backend/disk.rs
@@ -249,7 +249,7 @@ impl LoadContext for DiskLoadContext {
             load::Error::new(err).context(format!("while opening artifact file {}", full.display()))
         })?;
 
-        Ok(Reader::new(Box::new(file)))
+        Ok(Reader::new(file))
     }
 }
 
--- a/diskann-record/src/backend/disk.rs
+++ b/diskann-record/src/backend/disk.rs
@@ -1,0 +1,339 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+//! Disk-backed save/load contexts.
+//!
+//! [`DiskSaveContext`] writes the manifest as JSON plus side-car artifact files into a
+//! directory; [`DiskLoadContext`] reads them back. The two halves are independent and
+//! communicate only through the filesystem. Both are available under the `disk` feature.
+
+use std::{
+    collections::HashSet,
+    io::BufReader,
+    path::{Path, PathBuf},
+    sync::Mutex,
+};
+
+use crate::{
+    load::{self, LoadContext, Reader},
+    save::{self, SaveContext, Value, Writer},
+};
+
+/// The disk-backed [`SaveContext`].
+///
+/// Holds the manifest directory, the manifest path, and the set of artifact file names
+/// registered so far. Lookup and insertion go through a [`Mutex`] so that concurrent
+/// [`Save`](crate::save::Save) impls cannot accidentally hand out the same artifact name
+/// twice.
+#[derive(Debug)]
+pub(crate) struct DiskSaveContext {
+    dir: PathBuf,
+    metadata: PathBuf,
+    files: Mutex<HashSet<String>>,
+}
+
+#[derive(serde::Serialize)]
+struct Final<'a> {
+    files: Vec<&'a str>,
+    value: &'a Value<'a>,
+}
+
+impl DiskSaveContext {
+    /// Create a disk-backed save context targeting `dir` for side-car artifacts and
+    /// `metadata` for the manifest. Validates that `dir` is an actual directory.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`save::Error`] if `dir` does not exist, cannot be inspected, or exists but
+    /// is not a directory.
+    pub(crate) fn new(dir: PathBuf, metadata: PathBuf) -> save::Result<Self> {
+        match std::fs::metadata(&dir) {
+            Ok(meta) if meta.is_dir() => {}
+            Ok(_) => {
+                return Err(save::Error::message(format!(
+                    "path {} exists but is not a directory",
+                    dir.display()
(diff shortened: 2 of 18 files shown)


diskann-record/src/save/context.rs
```diff
@@ -0,0 +1,228 @@
...
+///
+/// Holds the manifest directory, the manifest path, and the set of artifact file names
+/// registered so far. Lookup and insertion go through a [`Mutex`] so that concurrent
+/// [`Save`](super::Save) impls cannot accidentally hand out the same artifact name twice.
+#[derive(Debug)]
+pub(super) struct ContextInner {
+    dir: PathBuf,
+    metadata: PathBuf,
+    files: Mutex<HashSet<String>>,
+}
```
hildebrandmw: I think one of the biggest things we can do in this PR is put the `ContextInner` types behind a trait object:
```rust
// Save path
pub trait SaveContext {
    type Output;
    fn write(&self, key: &str) -> Result<Write<'_>>;
    fn finish(self, value: Value<'_>) -> Result<Self::Output>;
}

trait GetWrite {
    fn write(&self, key: &str) -> Result<Write<'_>>;
}

impl<T> GetWrite for T
where
    T: SaveContext,
{
    fn write(&self, key: &str) -> Result<Write<'_>> {
        <T as SaveContext>::write(key)
    }
}

pub struct Context<'a> {
    inner: &'a dyn GetWrite,
}

impl<'a> Context<'a> {
    pub(super) fn new(inner: &'a dyn GetWrite) -> Self {
        Self { inner }
    }
}

pub fn save<T, C>(x: &T, context: C) -> Result<C::Output>
where
    T: Saveable,
    C: SaveContext,
{
    let value = x.save(Context::new(&context))?;
    context.finish(value)
)
```
and the load path is
```rust
pub trait LoadContext {
    fn value(&self) -> Result<&Value<'_>>>;
    fn read(&self, key: &str) -> Result<Reader<'_>>;
}

pub struct Context<'a> {
    inner: &'a dyn LoadContext,
    value: &'a Value<'a>,
}

impl<'a> Context<'a> {
    pub(super) fn new(inner: &'a dyn LoadContext, value: &'a Value<'a>) -> Self {
         Self {
              inner,
              value
         }
    }
}

pub fn load<'a, T, C>(context: &'a C) -> Result<T>
where
    T: Loadable<'a>,
    C: LoadContext,
{
    let value = context.value()?;
    T::load(Context::new(context, &value))
}
```

This gets us a few things:
1. The crate with no features can have no actual implementations of the `{Save/Load}Context` traits. This means no serde/serde_json as required dependencies be default, making `diskann-record` a lighter-weight dependency for lower-level crates (will need to make the serialize implementation for Value feature-gated).
2. It provides a great hook for plugging in a virtual-filesystem for testing.
3. Additional testing can use a purely in-memory implementation completely adjacent to a VFS backed system for even less overhead (will need to add a `Value::into_owned` implementation).

diskann-record/src/value.rs
```diff
@@ -0,0 +1,362 @@
...
+
+use crate::{Number, Version, save::Error};
+
+/// The wire-level union of every saveable kind.
+///
+/// See the module-level docs for an overview of when each variant is produced. The
+/// borrowing parameter `'a` lets [`Value::String`], [`Value::Bytes`], and nested
+/// records reuse memory owned by the caller without copying.
+#[derive(Debug)]
+pub enum Value<'a> {
```
hildebrandmw, line 45: Probably worth moving `value` up one level since it's shared between saving and loading.

hildebrandmw (commented):
Thanks Suhas - I have one big architectural comment about allowing pluggable backend contexts that I think will address many of the concerns about how heavy this is as a dependency and support for VFS that I think is probably worth doing. Happy to help out if needed.

diskann-record/src/save/context.rs
suhasjs: Thanks for the sketch, Mark. I've created `SaveContext` and `LoadContext` traits as you suggested. Please take a look.

suhasjs:
@hildebrandmw Requesting review for the following changes: 
- Moved up `value.rs` to reflect the shared usage between load/save paths
- Trait objects to enable VFS pluggability  (with a `DiskContext` impl for disk-based serialization, feature gated)
- Change `SaveContext::write` to take `Option<&str>` to treat the input `key` as a _hint_. Filename is now `INTEGER-{key}`, where INTEGER is just the number of artifacts written so far. Creating a random value would pull in `rand`, and I didn't want to add that dependency.
- Improved test coverage

diskann-record/src/load/context.rs
```diff
@@ -0,0 +1,475 @@
...
+
+        Ok(reader)
+    }
+}
+
+/// A borrowed reader over a side-car artifact.
+///
+/// Produced by [`Object::read`]. Implements [`std::io::Read`] and [`std::io::Seek`].
+pub struct Reader<'a> {
+    io: BufReader<File>,
```
hildebrandmw: This needs to be something more like `BufRead<Box<dyn std::io::Read>>` to be compatible with non-file backed providers. Otherwise we're still forcing files on everyone.

diskann-record/src/load/context.rs
```diff
@@ -0,0 +1,475 @@
...
+    _lifetime: std::marker::PhantomData<&'a ()>,
```
hildebrandmw: There's also a bit of tension: if `LoadContext` is public and supposed to return a `Reader`, then there needs to be some publicly available way to build a `Reader` aound an `io::Read + io::Seek` instance.

diskann-record/src/version.rs
```diff
@@ -0,0 +1,111 @@
...
+            minor,
+            patch,
+        }
+    }
+}
+
+#[cfg(feature = "serde")]
+impl Serialize for Version {
+    fn serialize<S: Serializer>(&self, ser: S) -> Result<S::Ok, S::Error> {
+        ser.collect_str(&format_args!(
```
hildebrandmw: This functionality can live behind `Display` and maybe a `from_str` method? Making them inherent methods allows them to be used even when `serde` isn't active.

diskann-record/src/save/context.rs
```diff
@@ -0,0 +1,391 @@
...
+    }
+}
+
+/// A borrowed side-car artifact writer produced by [`Context::write`].
+///
+/// Implements [`std::io::Write`] and [`std::io::Seek`]. Writes are buffered; calling
+/// [`Writer::finish`] flushes the buffer, closes the file, and returns a [`Handle`].
+#[derive(Debug)]
+pub struct Writer<'a> {
+    io: BufWriter<File>,
```
hildebrandmw: One thing you might find useful when working through the backend design is developing an in-memory only implementation of `SaveContext` and `LoadContext` ;)

diskann-record/src/save/context.rs
```diff
@@ -0,0 +1,391 @@
...
+        // When a human-readable hint is supplied it must be a simple relative file name:
+        // reject absolute paths, parent traversal, and multi-component paths so the prefix
+        // below produces a single, well-formed file name in the manifest directory.
+        if let Some(key) = key {
+            let mut components = std::path::Path::new(key).components();
+            match components.next() {
+                Some(std::path::Component::Normal(_)) if components.next().is_none() => {}
+                _ => {
+                    return Err(Error::message(format!(
+                        "artifact file name hint {:?} must be a relative file name with no path \
```
hildebrandmw: It seems like a shame to bail here, especially since the trait level documentation doesn't say "don't use anything that looks like a file-path". Would it make sense instead to fall back to using `None` instead?

diskann-record/src/save/context.rs
```diff
@@ -0,0 +1,391 @@
...
+        let mut files = self
+            .files
+            .lock()
+            .unwrap_or_else(|poison| poison.into_inner());
+
+        // Prefix each artifact with the count of artifacts written so far, so that reusing
+        // the same `key` (or omitting it) still yields a unique file name.
+        let name = match key {
+            Some(key) => format!("{:03}-{}", files.len(), key),
+            None => format!("{:03}", files.len()),
```
hildebrandmw: Maybe include a little more information in the file name? Like `001-record.bin`?

diskann-record/src/save/context.rs
```diff
@@ -0,0 +1,391 @@
...
+        };
+
+        if !files.insert(name.clone()) {
+            return Err(Error::message(format!(
+                "generated artifact name {:?} collides with an existing artifact",
+                name,
+            )));
+        }
+        let full = self.dir.join(&name);
```
hildebrandmw: We can consider using [`cap_std`](https://docs.rs/cap-std/latest/cap_std/fs/struct.Dir.html) to systematically force files to be generated in a subdirectory instead of relying on path name manipulation. It also adds a bit of safety to the saving process.

hildebrandmw (commented):
Thanks Suhas - one more round on the `SaveContext/LoadContext` traits. I think it would be helpful to implement a purely in-memory backend to see how it interactws with the reader/writers. Such a backend can work directly with `Value` and avoid pulling in `serde` entirely.

diskann-record/src/save/context.rs
suhasjs: I thought about adding the `.bin` suffix, but then decided against it because we already have a `.bin` format that is pervasive throughout the repo, and there's no guarantee that the the `save` impl for a struct will look anything like the `.bin` file we know.

diskann-record/src/save/context.rs
suhasjs: Doesn't that also pull in a whole new crate? Right now, we only depend on `anyhow`. Do you think it's worth adding?

diskann-record/src/load/context.rs
suhasjs: Added a `Reader::new()` to construct a `Reader` with a `io::Read + io::Seek`.

diskann-record/src/load/context.rs
suhasjs: Done.

diskann-record/src/save/context.rs
suhasjs: Adding the in-memory only impls actually improved the organization of the code now.  Previously, the disk-variants of SaveContext and LoadContext were part of the `{load|save}/context.rs`. After adding the in-memory, I realized I could abstract that into a `Backend` enum and moved the two backends to `backend/{memory|disk}.rs`.

This keeps the Context traits in `load|save` directories and puts the impls under `backend`.

diskann-record/src/save/context.rs
suhasjs: Added fallback to `None` as suggested.

diskann-record/src/save/context.rs
```diff
@@ -0,0 +1,202 @@
...
+///
+/// A `SaveContext` decides where side-car artifacts are written ([`SaveContext::write`])
+/// and how the final manifest is committed ([`SaveContext::finish`]). The concrete
+/// implementations live under `crate::backend`: a disk-backed context (under the `disk`
+/// feature) and an in-memory context. Alternative implementations (e.g. a virtual
+/// filesystem) can be supplied for testing or to avoid touching the filesystem.
+///
+/// The generic [`save`](super::save) entry point is parameterized over this trait so
+/// that the base crate carries no hard dependency on any particular implementation.
+pub trait SaveContext {
```
hildebrandmw: What are your thoughts on intending this to be a public extension point from the get-go versus keeping it `pub(crate)` for now? The issue with it currently is that it's public, but `write` returns a `Writer` and `Writer::new` is just `pub(super)`. So external users cannot implement it. Even if they could, they would still need to create a `Handle`, which I firmly believe should not have a public constructor. It might be a better idea in the short-term to keep this (and `LoadContext`) `pub(crate)` at least initially. It's better to go from private to public than the other way.

diskann-record/src/value.rs
```diff
@@ -0,0 +1,453 @@
...
+    Null,
+    Bool(bool),
+    Number(Number),
+    String(Cow<'a, str>),
+    Bytes(Cow<'a, [u8]>),
```
hildebrandmw: I do think `Bytes` has value, but maybe not in its current form (where it will be store in JSON as integers rather than as a byte-string). We could either remove it, or figure out how to get it represented in binary form in JSON. Thoughts?

diskann-record/src/number.rs
```diff
@@ -0,0 +1,208 @@
...
+///
+/// The wire format distinguishes unsigned, signed, and floating-point numbers; the
+/// deserializer preserves that distinction by selecting the matching variant. Use the
+/// narrowing accessors (e.g. [`Number::as_u32`], [`Number::as_f64`]) to extract a Rust
+/// value of the desired type.
+#[derive(Debug, Clone, Copy)]
+pub enum Number {
+    U64(u64),
+    I64(i64),
+    F64(f64),
```
hildebrandmw, line 28: Something subtle here is that this silently loses data when the float64 value is infinity or NaN, as [`serde_json`](https://docs.rs/serde_json/latest/serde_json/struct.Number.html#method.from_f64) does not support these representations. We can support this by checking at for these values and using strings "inf", "neg_ing", and "nan" (and then teaching the deserializer to process these values).

diskann-record/src/backend/memory.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        if files.contains_key(&name) {
+            return Err(save::Error::message(format!(
+                "generated artifact name {:?} collides with an existing artifact",
+                name,
+            )));
+        }
+        // Reserve the name so the count advances and concurrent writers cannot collide;
+        // the placeholder is overwritten with the real bytes by `Writer::finish`.
+        files.insert(name.clone(), Vec::new());
```
hildebrandmw: Both backends do this to some extent, but it's maybe a little worse here. Eagerly registering `Vec::new()` means we cannot differentiate between a writer that was created but dropped before properly finishing, and a valid writer that just didn't write everything.

It's better to make files `Mutex<HashMap<String, Option<Vec<u8>>>>`. When we first pass out a file, set the corresponding value to `None` and only over-write it to `Vec` when finishing. This lets us differentiate properly between successfully completed files. As a bonus, we can also check in `SaveContext::finish` that all created files finished correctly and error out if this is not the case.

For the disk backend, we can do the same thing.

diskann-record/src/backend/memory.rs
```diff
@@ -0,0 +1,221 @@
...
+        if let Some(key) = key {
+            let mut components = std::path::Path::new(key).components();
+            match components.next() {
+                Some(std::path::Component::Normal(_)) if components.next().is_none() => {}
+                _ => {
+                    return Err(save::Error::message(format!(
+                        "artifact file name hint {:?} must be a relative file name with no path \
+                         separators",
+                        key,
+                    )));
```
hildebrandmw: This behavior diverges from the disk backend. Can we make name mangling infallible like the disk path?

diskann-record/src/backend/disk.rs
```diff
@@ -0,0 +1,360 @@
...
+                "generated artifact name {:?} collides with an existing artifact",
+                name,
+            )));
+        }
+        let full = self.dir.join(&name);
+        if full.exists() {
+            return Err(save::Error::message(format!(
+                "file {} already exists",
+                full.display()
+            )));
```
hildebrandmw, line 144: If a previous save fails for some reason, it will leave around junk which will then make a second round fail. I think defensively not over-writing is the call, but do we want to support some amount of cleanup on failure?

diskann-record/src/load/mod.rs
```diff
@@ -0,0 +1,265 @@
...
+where
+    T: Load<'a>,
+{
+    fn load(context: Context<'a>) -> Result<Self> {
+        let object = context.as_object().ok_or(error::Kind::TypeMismatch)?;
+        let version = object.version();
+        if version == T::VERSION {
+            T::load(object)
+        } else {
+            T::load_legacy(object)
```
hildebrandmw, line 135: Many of the built-in impls are not tested for round trippability, and this `load_legacy` path is completely uncovered. Please add tests for the unhappy paths as well and verify that `load_legacy` works as expected.

hildebrandmw (commented):
Thanks Suhas, this is coming together. I love how light-weight it is getting when the disk backend and serde are excluded. I have a few higher-level comments. Mostly about testing and fortifying the unhappy paths in addition to the happy paths.

diskann-record/README.md
```diff
@@ -0,0 +1,14 @@
+# DiskANN Record
+
+This crate provides a small framework for persisting structured Rust values as a
```
harsha-simhadri: what are "structured rust values"

diskann-record/README.md
```diff
@@ -0,0 +1,14 @@
...
+manifest (can be serialized to JSON) plus a set of side-car binary artifacts, and 
+reloading them later. It is can be used by `diskann` providers and indexes to 
+implement durable, consistent and backward-compatible checkpoints.
+
+Types describe how they map to a versioned record by implementing the `save::Save` and
+`load::Load` traits; the `save_fields!` and `load_fields!` macros handle the
+field-by-field plumbing for plain structs. Every record carries a `Version` so loaders
+can detect schema changes and either upgrade or fall back through a probing chain.
+
+The goal is to allow crates like `diskann` to checkpoint their state without depending on
```
harsha-simhadri, line 13: Are manifest and binary artifacts compatible across providers or defined per provider?

diskann-record/src/value.rs
```diff
@@ -0,0 +1,453 @@
+/*
+ * Copyright (c) Microsoft Corporation.
```
harsha-simhadri, line 2: could the primitives in number.rs, value.rs, and version.rs be used from some existing external crate. Is there a strong reason to define this here

harsha-simhadri (commented):
couple of comments inline. will read backed code once you have final updates

harsha-simhadri (commented):
couple of comments inline. will read backed code once you have final updates

diskann-record/src/save/context.rs
suhasjs: I want to keep it public and not `pub(crate)`. I see the issue though. We'll need to also make `WriteInner` `pub` as well. For now, I'll make this `pub(crate)` and we can open it up later if there's a need for it.

suhasjs: I want to keep it public and not `pub(crate)`. I see the issue though. We'll need to also make `WriteInner` `pub` as well. For now, I'll make this `pub(crate)` and we can open it up later if there's a need for it.

diskann-record/src/number.rs
suhasjs: Thanks for pointing out this subtle issue, Mark. I've added explicit handling of these special values.

diskann-record/src/value.rs
suhasjs: We could represent as hex strings, but that would require us adding code to do this. Default `serde` impl for `Bytes` is an array of `u8`, and it currently doesn't round-trip back to `Bytes` natively, so I've removed `Bytes` entirely to reduce unused code.

diskann-record/src/backend/memory.rs
suhasjs: `MemorySaveContext::write()` now treats `hint` as `None`, mirroring the `disk` backend.

diskann-record/src/backend/memory.rs
suhasjs: Added this check to both memory and disk contexts.

diskann-record/src/backend/disk.rs
suhasjs: Added cleanup in `DiskSaveContext`. `MemorySaveContext` did not need this check since the default `Drop` impl for `Vec<u8>` already takes care of the buffer cleanup.

diskann-record/src/load/mod.rs
suhasjs: Added two tests for `load_legacy` -- one that must succeed and another that must fail. Also added tests to cover save/load for primitives at various bitwidths.

diskann-record/src/save/context.rs
suhasjs: Decided against adding `cap_std`. Current way of generating paths looks safe enough.

diskann-record/src/value.rs
suhasjs: `serde_json` provides similar looking `Number` and `Value`, but have a couple of issues:
- `serde_json::Number` does not go down to <64 bit values. It only provides `i64/u64/f64`. We need more than just those 3. It also does not explicitly handle `NaN, +inf, -inf`, which our impl does.
- `serde_json::Value` does not have support for `$handle, $version` (reserved keys). In addition, our `Value` uses `Cow<'a, str>` for zero-copy save where `serde_json::Value` is always an owned `String`. 

Using `serde_json::Number` instead of our `Number` does not meaningfully cut down LoC here

diskann-record/README.md
suhasjs: Right now, artifacts saved by one provider are readable only by another provider (i.e., saved to Disk ==> must use DiskLoadContext). `diskann-record` is generic enough to allow saving with one provider and loading with another provider (boils down to how `SaveContext/LoadContext` are implemented for each provider). 

For example: say we want to save to object store (using PUT requests for each write), then download the artifacts (using a side-channel) to local disk, and load the disk-local artifacts. 
To do this, we write a `ObjectStoreSaveContext` that emulates the same format as `DiskSaveContext`, except uses `PUT` requests to write data. `DiskSaveContext` cannot know (and will not care) who wrote the artifacts, just that it is in a format it can read and understand.

harsha-simhadri: thanks, could you add this note to readme.

harsha-simhadri (approved):
lgtm, but please see which of the comments in #1079 still apply here. There are quite a few unresolved comments from Jordan and Alex there that seem relevant.

JordanMaples (approved):
lgtm

JordanMaples (approved):

JordanMaples (approved):

JordanMaples (approved):

diskann-record/README.md
suhasjs: Structs + Option<> + Enum + primitives (nothing dynamic, with known schema). Updated readme with this info.

JordanMaples (approved):

PR #1193, opened 2026-06-20
Improve diskann-disk test coverage

arrayka:
Adds 18 targeted unit tests to the `diskann-disk` crate covering previously
untested error paths, control flow branches, and pure functions.

| **Files with missing lines**<br> | **Coverage Δ**<br> |
| --- | --- |
| .../checkpoint/checkpoint_record_manager_with_file.rs<br> | 96.20% <100.00%> (+1.81%)<br> |
| diskann-disk/src/search/pq/pq_scratch.rs<br> | 100.00% <100.00%> (+10.00%)<br> |
| ...kann-disk/src/search/provider/disk_sector_graph.rs<br> | 97.70% <100.00%> (+0.75%)<br> |
| diskann-disk/src/storage/quant/generator.rs<br> | 94.31% <100.00%> (+1.64%)<br> |
| diskann-disk/src/utils/kmeans.rs<br> | 96.61% <100.00%> (+5.62%)<br> |
| diskann-disk/src/utils/partition.rs<br> | 92.85% <100.00%> (+0.34%)<br> |
| ...kann-disk/src/build/chunking/continuation/utils.rs<br> | 93.90% <97.87%> (+3.61%)<br> |


diskann-disk/src/search/pq/pq_scratch.rs
```diff
@@ -128,4 +128,30 @@ mod tests {
...
+        let mut pq_scratch = PQScratch::new(64, dim, 4, 256).unwrap();
+
+        // Query longer than dim should succeed (only first `dim` elements used)
+        let long_query: Vec<f32> = (1..=dim + 10).map(|i| i as f32).collect();
+        pq_scratch.set(&long_query).unwrap();
+
+        for (i, &val) in long_query.iter().enumerate().take(dim) {
+            assert_eq!(pq_scratch.query_scratch[i], val);
+        }
+    }
```
arkrishn94, line 152: Is this expected behavior? Shouldn't the scratch fail for an incorrectly sized query?

arrayka: Yes, it is the expected behavior, according to the documentation of `pq_scratch.set()`:
```
    /// Copy the first `dim` elements of `query` into `query_scratch`.
    ///
    /// `query` must already be in full-precision `f32` representation; quantized
    /// inputs (e.g. `MinMaxElement`) should be decoded via `VectorRepr::as_f32`
    /// at the caller boundary before invoking this method.
    ///
    /// Accepts oversized `query` (only the first `dim` elements are used) for
    /// backwards compatibility with callers that hold alignment-padded buffers.
    /// Returns `DimensionMismatchError` if `query.len() < query_scratch.len()`.
    pub fn set(&mut self, query: &[f32]) -> ANNResult<()>
```

arkrishn94: I followed the chain of calls to `set()` and `new()` and I don't think supporting larger input query dimension is actually needed here (the scratch length is derived from the dimension of the fp vectors in the PQ table, which is the same as the query dimension).

Can we fix this since we've found it? The reason I'm asking is cause this is actually incorrect behavior; I'm not actually sure how this ended up getting supported (It's probably my fault!). The f32 dimension is not larger than the dimension of minmax vectors, it's actually smaller :)

arrayka: Done

arkrishn94 (approved):
Thanks Alex! Looks good to me.

diskann-disk/src/build/chunking/checkpoint/checkpoint_record_manager_with_file.rs
```diff
@@ -104,6 +104,77 @@ mod tests {
...
+            .to_string();
+        // Two managers with the same prefix+identifier should see the same checkpoint state
+        let manager_a = CheckpointRecordManagerWithFileStorage::new(&index_prefix, 42);
+        let manager_b = CheckpointRecordManagerWithFileStorage::new(&index_prefix, 42);
+        assert_eq!(
+            manager_a.get_resumption_point(WorkStage::Start)?,
+            manager_b.get_resumption_point(WorkStage::Start)?
+        );
+        // A different identifier should be independent
+        let manager_c = CheckpointRecordManagerWithFileStorage::new(&index_prefix, 99);
```
metajack: It's unclear to me how this asserts independence from a and b like the comment states.

diskann-disk/src/build/chunking/checkpoint/checkpoint_record_manager_with_file.rs
```diff
@@ -104,6 +104,77 @@ mod tests {
...
+    #[test]
+    fn test_has_completed_false_when_no_file() -> ANNResult<()> {
+        let temp_dir = tempdir()?;
+        let index_prefix = temp_dir
+            .path()
+            .join("nonexistent_index")
+            .to_str()
+            .unwrap()
+            .to_string();
+        let manager = CheckpointRecordManagerWithFileStorage::new(&index_prefix, 999);
```
metajack, line 106: This appears to be the identical test to the last thing above?

diskann-disk/src/build/chunking/checkpoint/checkpoint_record_manager_with_file.rs
```diff
@@ -104,6 +104,77 @@ mod tests {
...
+
+        // Mark as invalid - progress resets to 0 (is_valid=false => progress read as 0)
+        let mut manager3 = CheckpointRecordManagerWithFileStorage::new(&index_prefix, identifier);
+        manager3.mark_as_invalid()?;
+        assert_eq!(
+            manager3.get_resumption_point(WorkStage::QuantizeFPV)?,
+            Some(0)
+        );
+
+        clean_checkpoint_file(&index_prefix, identifier);
```
metajack: It's not a big deal, but part of point of `tempdir()` is that the object returned cleans up after itself on drop. This kind of clean is only really needed if you were doing multiple things in the same tempdir and wanted to clean in between.

diskann-disk/src/build/chunking/continuation/utils.rs
```diff
@@ -150,6 +150,150 @@ mod tests {
...
+        let result = process_while_resource_is_available(
+            |item| {
+                processed.push(item);
+                Ok::<(), TestError>(())
+            },
+            items.into_iter(),
+            Box::new(tracker),
+        );
+
+        assert!(result.is_ok());
```
metajack: This line is redundant with the next one. If you match on result.unwrap() it will already panic if it's not ok.

diskann-disk/src/build/chunking/continuation/utils.rs
```diff
@@ -150,6 +150,150 @@ mod tests {
...
+        match result.unwrap() {
+            Progress::Processed(idx) => {
+                assert_eq!(idx, 3); // stopped before processing item at index 3
```
metajack: Is this comment correct? Processed(3) means processed to 3, or  processed until 3? I really hope it is the former.

diskann-disk/src/build/chunking/continuation/utils.rs
```diff
@@ -150,6 +150,150 @@ mod tests {
...
+            |item| {
+                processed.push(item);
+                Ok::<(), TestError>(())
+            },
+            items.into_iter(),
+            Box::new(tracker),
+        );
+
+        assert!(result.is_ok());
+        match result.unwrap() {
```
metajack: This is probably better written as an `if let Progress::processed(idx) = result.unwrap() { } else { panic!(); }`. I'm surprised clippy didn't warn on this; it is usually not happy with single arm match statements.

diskann-disk/src/build/chunking/continuation/utils.rs
```diff
@@ -150,6 +150,150 @@ mod tests {
...
+        let result = process_while_resource_is_available(
+            |item| {
+                processed.push(item);
+                Ok::<(), TestError>(())
+            },
+            items.into_iter(),
+            Box::new(tracker),
+        );
+
+        assert!(result.is_ok());
```
metajack: Same comments from above apply here.

diskann-disk/src/build/chunking/continuation/utils.rs
```diff
@@ -150,6 +150,150 @@ mod tests {
...
+                    processed.lock().await.push(item);
+                    Ok::<(), TestError>(())
+                }
+            },
+            items.into_iter(),
+            Box::new(tracker),
+        )
+        .await;
+
+        assert!(result.is_ok());
```
metajack: And same here.

metajack (approved):

diskann-disk/src/build/chunking/checkpoint/checkpoint_record_manager_with_file.rs
arrayka: Done

diskann-disk/src/build/chunking/continuation/utils.rs
arrayka: Yes. it is the former. fixed.

diskann-disk/src/build/chunking/continuation/utils.rs
arrayka: Fixed

diskann-disk/src/build/chunking/continuation/utils.rs
arrayka: Fixed. Even removed panic!() to increase code coverage number.

diskann-disk/src/build/chunking/continuation/utils.rs
arrayka: Fixed

diskann-disk/src/build/chunking/continuation/utils.rs
arrayka: Fixed

diskann-disk/src/build/chunking/checkpoint/checkpoint_record_manager_with_file.rs
arrayka: True. Deleted `test_new_checkpoint_creates_fresh_record()`

diskann-disk/src/build/chunking/checkpoint/checkpoint_record_manager_with_file.rs
arrayka: Deleted this test.

metajack (approved):

PR #1205, opened 2026-06-26
Remove `diskann-platform` & decouple `diskann-providers`; tidy `diskann-disk` modules

suri-kumkaran:
Folds `diskann-platform` into `diskann-disk` and moves perf timing (`Timer`) out of `diskann-providers` into `diskann-disk`'s instrumentation, then reorganizes the affected `diskann-disk` modules. Refactor-only except one thing: `generate_pq`'s completion log now reports elapsed seconds only, dropping cycles/CPU/peak-memory (already zero on Linux/macOS).

## Changes
- **Delete `diskann-platform`**: perf counters become a private detail of `Timer`; io primitives (io_uring / Windows IOCP / handles) move next to their only consumer, `aligned_file_reader`. Workspace, CI filters, and docs updated.
- **Move `Timer` out of `diskann-providers`**: `Timer` + perf counters relocate to `diskann-disk`'s `utils/instrumentation/` beside `PerfLogger` (canonical path `utils::instrumentation::Timer`). Its two callers — providers' `pq_construction` and tools' `build_pq` — switch to `std::time::Instant` (`build_pq` drops the perf fields). Providers also loses its `opentelemetry` / `diskann-platform` deps and `perf_test` feature.
- **Relocate `aligned_file_reader`**: `utils/` → `search/provider/`, split into `reader/`, `factory/`, `traits/`, `platform/`.
- **Simplify disk-search**: add `DiskVertexProviderFactory::from_disk_index_path`, dropping the benchmark's coupling to the low-level reader factory.
- **Config cleanup**: remove tools' now-orphan `perf_test` feature and the `diskann-providers/perf_test` + `diskann-tools/perf_test` entries in the benchmark's `disk-index` feature.
- **Trim dead code**: remove `diskann-disk`'s unused `virtual_storage` feature and `Timer::{elapsed_seconds, elapsed_seconds_for_step}` + their OpenTelemetry import.


Cargo.lock
```diff
@@ -1847,6 +1835,17 @@ dependencies = [
...
+[[package]]
+name = "io-uring"
+version = "0.7.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4d09b98f7eace8982db770e4408e7470b028ce513ac28fecdc6bf4c30fe92b62"
+dependencies = [
+ "bitflags 2.10.0",
+ "cfg-if",
+ "libc",
+]
```
hildebrandmw: This introduces another `io-uring` dependency (0.7.12 in addition to 0.6.4). Is there a good reason for this? Version bumps are good, but I'd prefer to keep our dependencies more consistent.

Cargo.lock
```diff
@@ -1847,6 +1835,17 @@ dependencies = [
...
+
```
hildebrandmw: On a related note - apparently, we have three different `windowsw-sys` dependency versions.

diskann-disk/src/search/provider/aligned_file_reader/factory/file.rs
```diff
@@ -6,12 +6,12 @@
 use diskann::ANNResult;
 
 #[cfg(all(not(miri), target_os = "linux"))]
-use super::LinuxAlignedFileReader;
+use super::super::reader::LinuxAlignedFileReader;
```
hildebrandmw: Non-blocking nit: Often, `super::super` is a sign that something structured quite right. Consider if another style of imports might be better.

diskann-disk/src/search/provider/aligned_file_reader/platform/windows/file_handle.rs
```diff
@@ -48,6 +49,7 @@ pub enum AccessMode {
 /// - `Write`: Allows subsequent open operations on the same file file to request write access.
 ///
 /// - `Delete`: Allows subsequent open operations on the same file file to request delete access.
+#[allow(dead_code)] // Win32 wrapper: all modes are valid; only `Read` is exercised today.
```
hildebrandmw: Nit: Now that these are no longer part of a public API, they are truly dead-code. Maybe it can be removed?

diskann-disk/src/search/provider/aligned_file_reader/platform/windows/file_handle.rs
```diff
@@ -29,6 +29,7 @@ pub const FILE_ATTRIBUTE_READONLY: DWORD = 0x00000001;
 /// - `Write`: The file is opened in write-only mode.
```
hildebrandmw: non-blocking nit: After this PR, this st5ruct is buried 6-levels deep in the module hierarchy. While not necessarily a bad thing, it is a little unpleasant digging so deep. Somewhere higher in the module hierarchy might be a bit easier to work with.

suri-kumkaran: I feel it's fine to be here as it's a private, Windows-only detail with a single consumer in the sibling  reader/.
Co-location is more valuable than shallowness here.

Cargo.lock
suri-kumkaran: This is really sad as all of them are coming from different dependencies :)

diskann-disk/src/search/provider/aligned_file_reader/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+#[cfg(all(not(miri), target_os = "linux"))]
+pub use reader::LinuxAlignedFileReader;
+pub use reader::StorageProviderAlignedFileReader;
+#[cfg(all(not(miri), target_os = "windows"))]
+pub use reader::WindowsAlignedFileReader;
+
+mod factory;
+pub use factory::AlignedFileReaderFactory;
+#[cfg(test)]
+pub(crate) use factory::VirtualAlignedReaderFactory;
```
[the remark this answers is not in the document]
suri-kumkaran: No users other than test.

diskann-disk/src/search/provider/aligned_file_reader/factory/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+
+//! Concrete [`AlignedReaderFactory`](super::traits::AlignedReaderFactory) implementations.
+
+mod file;
+pub use file::AlignedFileReaderFactory;
+
+#[cfg(test)]
+mod virtual_storage;
+#[cfg(test)]
+pub(crate) use virtual_storage::VirtualAlignedReaderFactory;
```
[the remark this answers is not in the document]
suri-kumkaran: No users other than test.

Cargo.lock
suri-kumkaran: I see, I reverted the version bump to keep our dependencies more consistent.

diskann-disk/src/search/provider/aligned_file_reader/platform/windows/file_handle.rs
suri-kumkaran: Make sense, removed.

diskann-disk/src/search/provider/aligned_file_reader/factory/file.rs
suri-kumkaran: Make sense, changed the import style.

harsha-simhadri:
could we break this PR into components? It sounds like the 4 themes you called can be separated.

suri-kumkaran:
> could we break this PR into components? It sounds like the 4 themes you called can be separated.

@harsha-simhadri, the only goal was to remove `diskann-platform`. All other changes were necessary to make that removal cleanly.

harsha-simhadri (approved):

diskann-disk/src/search/provider/aligned_file_reader/factory/mod.rs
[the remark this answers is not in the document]
wuw92: Now `VirtualAlignedReaderFactory`  is  `#[cfg(test)]` only, this feature has no consumer — no  #[cfg(feature = "virtual_storage")]  left in  diskann-disk , and nothing enables  `diskann-disk/virtual_storage` (disk tests reach  `VirtualStorageProvider`  via the  `diskann-providers`  dev-dep directly). 
Worth removing `virtual_storage` feature from `diskann-disk` crate.

wuw92 (approved):

diskann-disk/src/search/provider/aligned_file_reader/factory/mod.rs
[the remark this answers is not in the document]
suri-kumkaran: Make sense, removed.

hildebrandmw (approved):

PR #1206, opened 2026-06-27
In-Mem 2.0

### LINKED ISSUE #712: Move in-mem provider from inside diskann-providers to its own crate
In-memory providers are in `diskann-providers/src/model/graph/provider/async_/inmem`. Move them up to `diskann-inmem` crate.

Move traits and utilities in `diskann-providers/src/model/graph/provider/async_/common.rs` that are used solely by the in-mem providers to the inmem crate.

If you encounter further Rust orphan rule violations, list all the violations.

hildebrandmw:
Introduce a second in-memory provider with the intention of replacing the current provider.

# Why
The [RFC](https://github.com/microsoft/DiskANN/blob/9d5f1435fd3a327a8fc2cdf7f96d0ffeb164fa1b/rfcs/01206-inmem2.md) outlines much of the motivation. In short, the goal here is to:

* Make the provider safe under concurrent inserts/search/deletes.
* Support proper external/internal ID translation.
* Improve test coverage
* Do so with minimal performance overhead.

The concurrency argument comes from the epoch-based-reclamation (EBR) protection scheme for internal slots. See the RFC for more details.

# Known Follow-up Items
- Perf Parity: This generally has pretty good performance, but could use a little more tuning to bring it fully on-par with our current inmem index.
- Quantization: This initial prototype is lacking quantization support. Adding quantization is relatively straightforward in the primary `Store`, but will need some thought on how to add the reranking layer. This shouldn't be an architectural blocker, though.
- Hybrid PQ: A larger open question is how to support the `max_fp_vecs` feature of our current PQ implementation, which reads some full-precision and some quantized vectors during prune. Like with quantization in general, I think this is not a fundamental issue.
- Support for non-uniform sized items in slots: For this, I'm mainly thinking of multi-vectors where the number of vectors within each multi-vector can vary. In the context of multi-vectors, fast element access is less important than for traditional vectors as distance computations take considerably longer.

# Suggested Reviewing Order

The majority of this PR is in a new `diskann-inmem` crate. To facilitate testing, this crate has an "integration-test" feature, which enables the code in `diskann-inmem/src/integration`. This code is an unstable public reexport of internal types meant only for consumption in the `diskann-inmem/integration` integration test binary.

The integration test binary is powered by `diskann-benchmark-runner`.

### Independent Low-Level Utilities
* `num.rs`: Strong type utilities for byte and alignment representations.
* `buffer.rs`: A miri-compliant version of [`AlignedMemoryVectorStore`](https://github.com/microsoft/DiskANN/blob/b17240a610ba51aeb2b8bbfebab98d0c8240e12d/diskann-providers/src/model/graph/provider/async_/common.rs#L89-L95). This type allows vectors/neighbors to be stored in a single larger allocation. The use of `RawSlice` allows slots within the `Buffer` to be inspected and manipulated without forcing reference materialization (which is important to prevent aliasing).
* `neighbors.rs`: The new version of [`SimpleNeighborVectorProviderAsync`](https://github.com/microsoft/DiskANN/blob/main/diskann-providers/src/model/graph/provider/async_/simple_neighbor_provider.rs#L18-L26). This reuses the sharded-lock idea, but provides additional utility, including the ability to perform read-modify-write operations on adjacency lists.
* `counters.rs`: Event counters. When the "integration-test" feature is not enabled, counters become a no-op. These are enabled for testing to monitor changes.
* `sharded.rs`: An external-to-internal ID translation utility. The main trick with this struct is to provide utilities like `Sharded::occupied_entry`, which locks and returns an `external/internal` mapping. The proxy `Entry` struct is important as it verifies that such a mapping exists and provides an infallible way of deleting the mapping. This is chained with higher level operations (e.g. `Provider::delete`) to delete both the ID-mapping and the internal data-slot in lock-step.

### Concurrency Protocol
The concurrency protocol is built upon three main layers:
* `tag.rs`: An atomic slot tag for controlling access to data.
* `epoch.rs`: The central registry where readers register and deregister. This is the crux of this PR and probably the most important file.
* `store.rs`: A binary blob store built on top of `epoch.rs` to provide the safe concurrent store for data. This provides the following operations:
  - Storage of binary data in "slots".
  - Reading of data in slots (provided by `Reader`).
  - Tracking on the valid/invalid state slots.
  - Finding available slots into which new data can be inserted.
  - Safe retirement of slots and eventual reclamation.

The `Store` in `store.rs` has some help from `freelist.rs` to accelerate locating available slots internally.

**Testing**: `epoch.rs` has unit tests with injectable delays to set up known pathological orderings. The sequencing is helped by `test/sequencer.rs`. In addition, `test/epoch.rs` includes a direct stress test for the `Registry`. This is particularly helpful when run under Miri, which has the ability to detect some race conditions.

A larger concurrency stress test lives in the integration-test binary. This directly tests `store.rs` by spinning up readers, writers, and retirers and hammers a single `Store`. Data is read and written into the store in a knowable pattern, allowing readers to detect torn reads, implying a race condition. For this PR, I ran the following stress test file
```json
{
  "search_directories": [
  ],
  "output_directory": null,
  "jobs": [
    {
      "type": "store-stress",
      "content": {
        "capacity": 8192,
        "duration_secs": 600,
        "entry_bytes": 256,
        "epoch_guard_slots": 256,
        "freelist_recycle_capacity": 1024,
        "low_watermark": 4096,
        "max_ops": 50000000000,
        "readers": 32,
        "retirers": 16,
        "seed": 11935966405698895599,
        "writers": 32
      }
    }
  ]
}
```
with the command
```
cargo run --package diskann-inmem \
  --bin integration-test \
  --features integration-test \
  --release -- \
   run --input-file stress.json --output-file temp.json
```
The generated output was
```
readers:       32
writers:       32
retirers:      16
capacity:      8192
entry_bytes:   256
low_watermark: 4096
duration_secs: 600
max_ops:       50000000000
seed:          11935966405698895599
elapsed_secs:  600.001182962
reads:         110206469888
acquires_ok:   248086767
acquires_fail: 619145
retires_ok:    248082573
retires_fail:  243312644
reclaims:      108847566
transitions:   42006520
peak_live:     8131
```
While not a proof of correctness, this is a pretty decent stress test.

### Providers
The implementation of the data provider is split into two logical pieces. The first lives in `layers/` and is focused on computing distances. With this approach, I am trying to avoid the need to replicate `Accessor`s and `Strategy`s for each future quantization type. `layers/full.rs` is the full-precision implementation. One thing to note is the use of the `FullPrecision` marker trait from which the implementations of the whole `layers` API is derived for `layers::Full`. This allows users to include just a `T: FullPrecision` trait bound and **really** simplifies the generics upstream.

Within `provider.rs` - my goal here is to minimize the use of generics as much as possible. In particular for search, I use a trait object for `ExpandBeam`. When coupled with the `layers::QueryDistance` API, we can create implementations where the distance function is inlined *and* the number of prefetch instructions can be tailored to the data length. Fully optimizing this is still a work-in-progress.

Another thing to call out in `provider.rs` is the care needed for data insertion and deletion. Since external/internal ID translation is supported, we need to ensure that the translation table stays in-sync with the internal store. On insert, if we allocate an internal slot only to find the external ID already exists, we need to abort the operation rather than publish the internal slot. Similarly on delete, we first need to establish if the external/internal ID mapping exists. If so, then we can try to retire the slot. If slot retiring fails (it shouldn't, but bugs can happen) - we need to not commit the ID mapping deletion and instead return an error.

#### Testing
Like the concurrency stress tests, testing uses the `integration-test` binary. Here, the 10k YFCC dataset is used for non-trivial runs. Using the A/B functionality in `diskann-benchmark-runner`, we can compare against checked-in baselines. This allows us to capture rich metrics for recall, number of operations, etc. and update easily. The main logic for checking and reporting baseline mismatches is in `integration/support/check.rs`. The goal is to summarize all such mismatches for presentation to provide the highest signal possible.

## `diskann-benchmark`

Integration into the benchmarks is straightforward. I elected to put everything in a single file to minimize disruption. I'm trying an approach of using `diskann_benchmark_runner::Input::from_raw` to separate out the deserialization types from the actual inputs, allowing richer types (e.g., a full `diskann_benchmark_core::streaming::bigann::RunBook`) to be loaded.

Also note how relatively simple the streaming benchmark integration is. Since the new inmem provider supports ID translation and internal slot allocation, it does not need the same level of hand-holding that the current inmem provider needs.


Cargo.toml
```diff
@@ -24,7 +24,7 @@ members = [
     "diskann-benchmark",
     "diskann-tools",
     "vectorset",
-    "diskann-bftree",
+    "diskann-bftree", "diskann-inmem",
```
JordanMaples:
```suggestion
    "diskann-bftree", 
    "diskann-inmem",
```

hildebrandmw: Thanks!

diskann-inmem/src/tag.rs
```diff
@@ -0,0 +1,328 @@
...
+#[repr(transparent)]
+pub(crate) struct Tag(u8);
+
+impl Tag {
+    //-------------//
+    // High Values //
+    //-------------//
+
+    /// The slot is permanently readable and never mutated again. See [`Tag`].
+    pub(crate) const FROZEN: Self = Self::new(u8::MAX);
```
magdalendobson, line 89: What is `FROZEN` for? Why would we want to use it?

diskann-inmem/src/store.rs
```diff
@@ -0,0 +1,963 @@
...
+
+    // The unpadded size of each row in `buffer`. This includes both the data **and** the
+    // 1-byte tag. Tags are located at byte `unpadded - 1`.
+    unpadded: Bytes,
+
+    // The number of unfrozen points. This is guaranteed to be less than `buffer`.
+    unfrozen: usize,
+
+    // The authoritative source of truth for the state of each slot.
+    tags: Vec<AtomicTag>,
```
magdalendobson, line 150: What would go wrong if I only ever used/updated the inline tags?

diskann-inmem/src/epoch.rs
```diff
@@ -0,0 +1,847 @@
...
+    queue(epoch.wrapping_sub(2))
+}
+
+impl Registry {
+    /// Construct a new [`Registry`] with the default number of guard slots (256).
+    pub(crate) fn new() -> Self {
+        Self::with_capacity(CAPACITY)
+    }
+
+    /// Construct a new [`Registry`] with `capacity` guard slots.
```
magdalendobson, line 129: Does 256 guard slots mean that with default capacity we're effectively limited to 256 threads? If so this is a pretty subtle limitation and might trip people up if they are trying out DiskANN on a large machine. Would there be a downside to making the default capacity slightly larger, maybe 512?

diskann-inmem/src/store.rs
```diff
@@ -0,0 +1,963 @@
...
+        )]
+        let padded_bytes = unpadded
+            .checked_next_multiple_of(Bytes::CACHELINE.div(TWO))
+            .expect("unreachabel because `init` cannot exceed `isize::MAX` bytes");
+
+        let too_many_entries = || StoreError::too_many_entries(entries, init.nrows());
+
+        // We have a hard upper-bound of `u32::MAX` total slots.
+        //
+        // Thiis enforces that bound.
```
magdalendobson: Typo :)

diskann-inmem/src/store.rs
```diff
@@ -0,0 +1,963 @@
...
+            clippy::expect_used,
+            reason = "we expect `init` to have at least one row, so this should never happen"
+        )]
+        let padded_bytes = unpadded
+            .checked_next_multiple_of(Bytes::CACHELINE.div(TWO))
+            .expect("unreachabel because `init` cannot exceed `isize::MAX` bytes");
+
+        let too_many_entries = || StoreError::too_many_entries(entries, init.nrows());
+
+        // We have a hard upper-bound of `u32::MAX` total slots.
```
magdalendobson, line 215: How do we feel about this given that 10 billion size indices may be a possibility? Is this something we want to change in the future?

diskann-inmem/src/tag.rs
hildebrandmw: I used `FROZEN` for the start points. Basically, it's a mechanism to mark slots a being available for reading, but prevents code from unintentionally changing them as slot retirement needs to transition from `AVAILABLE` to `RETIRED`.

There's a bigger discussion to be had about how to safely change start points and all that for a truly dynamic index that I kind of punted on for this PR.

diskann-inmem/src/store.rs
hildebrandmw: Used: nothing. These tags are meant to be authoritative, and so should always be correct.

Updated: lots (unless the invasive tags for the primary buffer were removed). Currently, the invasive tags are the only ones checked on the read path, so failure to update them properly would be pretty disastrous.

diskann-inmem/src/epoch.rs
hildebrandmw: No problem bumping up to 512. It does make epoch scans a little more expensive, but those should be relatively rare. I meant to expose this as part of the top level provider configuration and ... kind of forgot.

diskann-inmem/src/store.rs
hildebrandmw: Wouldn't be a Mark PR without rampant typos in the comments.

diskann-inmem/src/store.rs
hildebrandmw: Do we really see someone using an in-memory index for a 10-billion sized dataset? This would require TBs of memory. I'm not necessarily opposed to making this change, but I feel like at that scale, you would want a different data structure that could swap out to disk more efficiently.

diskann-inmem/src/store.rs
magdalendobson: I think my broader question is what do we gain from having the invasive tags? Do they make the read path faster or something?

diskann-inmem/src/epoch.rs
magdalendobson: I feel mildly uncomfortable with having an effective limit on the number of threads buried deep inside the code, maybe we should bring it up to the top level configuration after all

diskann-inmem/src/store.rs
magdalendobson: For something SIFT-ish or a little bit bigger, for a 4B sized index you need 1.5-2 TB main memory. I took a quick look at the largest VMs out there you can rent and you can definitely get up to 3 TB. I don't think it's fully unreasonable to say we just don't think this is something you should be doing with this inmem provider, though.

diskann-inmem/src/epoch.rs
```diff
@@ -0,0 +1,847 @@
...
+        // It can only fail if someone else is holding the drain lock, which means we can't
+        // proceed anyways.
+        //
+        // This can help save an expensive slot scan.
+        let drain = self.drain.try_lock()?;
+
+        let (can_advance, current) = self.can_advance(&mut delay);
+
+        // Don't wrap around!
+        if current == u64::MAX {
```
magdalendobson, line 308: We mean epochs here, not IDs, right?

Also, what is so bad about wrapping around? Couldn't we just have a special case in try_advance for handling u64::MAX?

diskann-inmem/src/epoch.rs
hildebrandmw: I agree. Fortunately, this is just a matter of plumbing that I hadn't bothered to do yet. Thanks for bringing it up!

diskann-inmem/src/store.rs
hildebrandmw: The thought was basically this: for some quantization types (e.g. spherical), the overall vector length does not end up being a multiple of the cache line size anyways. Since we pad to cache line boundaries, we can basically get the invasive tag "for free", which by collocating it with the data means we don't necessarily need to generate extra bandwidth when checking the tags.

Unfortunately, this doesn't always work out for full-precision data, which often ends up being a multiple of the cache line size. However, moving to only using the tags out of the authoritative source does mean that we need to introduce an extra prefetch loop somewhere, either prefetching the tags during filtering higher in the `expand_beam` implementation, or emitting prefetches to the tags inline with the data access like I am currently doing.

I think the structure of `Store` needs to be a bit more open-ended anyways to support quantization/filtering metadata, so we might be able to make invasive tags more optional anyways.

diskann-inmem/src/epoch.rs
hildebrandmw: I'd have to think though this - this might be a remnant from before when the actual epochs mattered. I do think there is a nasty corner case in the current implementation with the rotating queue strategy: wrapping from `u64::MAX` to `1` (needed to differentiate between unclaimed and claimed slots) would skip one queue in the rotation and break things upstream.

If we determine that we *can* support wrapping, we could use a 32-bit epoch with still 64-bit slots in the registry, which would allow us to differentiate between "taken" vs "not-taken" and epoch 0.

That said, I'm not entirely convinced this is really needed since this would need to be running for a long-long time before overflowing.

diskann-inmem/src/epoch.rs
```diff
@@ -0,0 +1,847 @@
...
+
+    // We can only retire a single generation at a time.
+    //
+    // This guard avoids situations where two threads concurrently advance the epoch and
+    // hand out overlapping `Drain`s referring to the same retiring queue.
+    drain: Mutex<()>,
+
+    // We use four queues for storing retiring items. The rationale is documented below.
+    //
+    // ```text
```
magdalendobson: Should probably remove this bash header

diskann-inmem/src/epoch.rs
```diff
@@ -0,0 +1,847 @@
...
+    {
+        // GUARD CHECK
+        let mut epoch = self.epoch();
+        let hint = self.hint.fetch_add(1, Ordering::Relaxed);
+        delay.post_guard_check();
+        let nguards = self.guards.len();
+        for i in 0..nguards {
+            let slot = hint.wrapping_add(i) % nguards;
+
+            let m = &self.guards[slot];
```
magdalendobson: Can we give `m` a more descriptive name? Maybe something like `current_slot_value` or `my_guard_value`?

diskann-inmem/src/neighbors.rs
```diff
@@ -0,0 +1,718 @@
...
+    pub(crate) fn max_length(&self) -> usize {
+        // We reserve 4 bytes at the beginning for the length of the adjacency list.
+        (self.neighbors.stride().value() - std::mem::size_of::<Id>()) / std::mem::size_of::<Id>()
+    }
+
+    /// Return the maximum length for any adjacency list as a 32-bit integer.
+    pub(crate) fn max_length_u32(&self) -> u32 {
+        // Lossless by the invariants on `Self::new`.
+        self.max_length() as u32
+    }
```
magdalendobson, line 113: Is it better to have this be a method rather than having the user do the conversion?

diskann-inmem/src/neighbors.rs
```diff
@@ -0,0 +1,718 @@
...
+use thiserror::Error;
+
+use crate::{
+    buffer::{Buffer, BufferError},
+    num::{Align, Bytes},
+};
+
+type Id = u32;
+
+/// Locks are shared among groups of adjacency lists.
```
magdalendobson, line 39: Why do we want to do lock grouping here? It seems like if we are inserting a dataset in an ordering with strong spatial locality without any slot retiring (so points get ids in insertion order), this could become an issue.

diskann-inmem/src/provider.rs
```diff
@@ -0,0 +1,1237 @@
...
+    pub fn prefetch_lookahead(&mut self, prefetch_lookahead: Option<NonZeroUsize>) {
+        self.prefetch_lookahead = prefetch_lookahead;
+    }
+}
+
+///////////////////
+// Data Provider //
+///////////////////
+
+/// A zero-sied [`diskann::provider::ExecutionContext`] for [`Provider`].
```
magdalendobson: zero-sized?

diskann-inmem/src/store.rs
```diff
@@ -0,0 +1,963 @@
...
+            neighbors: &self.neighbors,
+            _guard: self.registry.guard()?,
+        })
+    }
+
+    /// Attempt to acquire a new [`Slot`] for writing.
+    ///
+    /// This method first consults the freelist and falls back to scanning the tags list
+    /// if no ID is available from the fast path.
+    pub(crate) fn acquire(&self) -> Option<Slot<'_>> {
```
magdalendobson, line 337: Maybe this would be a strange thing to do but I am wondering whether we could/should use `try_drain` before doing `scan_acquire`. The thinking is that once the freelist thins out, we would do multiple more expensive scans before finding that you can't acquire anything from the freelist. Instead, we could repopulate the happy path before needing to scan. Would this be reasonable?

diskann-inmem/src/neighbors.rs
```diff
@@ -0,0 +1,718 @@
...
+/// share a single `RwLock`. This means that holding a [`Lock`] on slot `i` will also block
+/// operations on any slot `j` in the same group.
+///
+/// **Deadlock hazard**: attempting to acquire two [`Lock`]s simultaneously can deadlock if
+/// they fall in the same lock group — or even across groups, depending on acquisition order.
+/// Callers must not hold more than one [`Lock`] at a time.
+const LOCK_GRANULARITY: usize = 16;
+
+fn lock_index(i: u32) -> usize {
+    i.into_usize() / LOCK_GRANULARITY
```
harsha-simhadri, line 51: minor nitpick: we may want to stride lock id rather than have contiguous ids map to the same lock because:
- ids 0...15 might be close to the start and present higher chance of contention
-  when adjacent ids have correlated data, there might lock contention during insertion.
Both are speculative. If you have observed neither in your runs, please ignore

diskann-inmem/src/neighbors.rs
```diff
@@ -0,0 +1,718 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// * `i` exceeds [`Self::entries`].
+    /// * `neighbors.len()` exceeds [`Self::max_length_u32`].
+    ///
+    /// If an error is returned, the graph is left unmodified.
+    pub(crate) fn set(&self, i: u32, neighbors: &[u32]) -> Result<(), SetError> {
```
harsha-simhadri, line 207: I suppose append is not supported since it would be a read and an unchecked set, and we dont need to check if previous read is overwritten since there is a epoch based guards at a higher level?

diskann-inmem/src/sharded.rs
```diff
@@ -0,0 +1,364 @@
...
+};
+use diskann::utils::IntoUsize;
+use parking_lot::{RwLock, RwLockWriteGuard};
+use thiserror::Error;
+
+const SHARD_SIZE: usize = 1024;
+
+/// Bidirectional mapping between an external id `I` and a dense internal `u32` id.
+#[derive(Debug)]
+pub(crate) struct Sharded<I>
```
harsha-simhadri: why is this mapping called sharded?

diskann-inmem/src/ids.rs
```diff
@@ -0,0 +1,364 @@
...
+const SHARD_SIZE: usize = 1024;
+
+/// Bidirectional mapping between an external id `I` and a dense internal `u32` id.
+#[derive(Debug)]
+pub(crate) struct Sharded<I>
+where
+    I: Hash + Eq,
+{
+    forward: DashMap<I, u32>,
+    backward: Vec<RwLock<Box<[Option<I>]>>>,
```
harsha-simhadri, line 29: could we add a note about why we choose DashMap for the fwd map and Vec<RwLock<>> for reverse map

diskann-inmem/src/tag.rs
```diff
@@ -0,0 +1,328 @@
...
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(transparent)]
+pub(crate) struct Tag(u8);
+
+impl Tag {
+    //-------------//
+    // High Values //
+    //-------------//
+
+    /// The slot is permanently readable and never mutated again. See [`Tag`].
```
harsha-simhadri, line 88: "permanently" or for the epoch?

diskann-inmem/src/store.rs
```diff
@@ -0,0 +1,963 @@
...
+            buffer: Buffer::new(total.into_usize(), padded_bytes, Align::_128)?,
+            unpadded,
+            unfrozen: entries.into_usize(),
+            tags: repeat_n(Tag::AVAILABLE, total.into_usize())
+                .map(AtomicTag::new)
+                .collect(),
+
+            // NOTE: The `Freelist` is initialized to `entries` and not `total` because
+            // we do not want it to release frozen IDs.
+            freelist: Freelist::new(entries, FREELIST_SIZE),
```
harsha-simhadri: with freelist and id mapping, what is the overall space overhead of this design?

harsha-simhadri (commented):
more to come

diskann-inmem/src/tag.rs
harsha-simhadri: based on discussion, readers have to yield for epoch to transition. Request updating comments for copilot/reviewers to catch long running readers.

harsha-simhadri (approved):

diskann-inmem/src/neighbors.rs
hildebrandmw: Append is not supported directly as an API, but if implemented via `Lock`. My thought here was that if you are doing an `append`, then you usually want the append to be part of a read+modify+write operation. Trying to do that by a `get` + `append` API opens up the possibility for race conditions where another thread appends/sets in the interim.

On the other hand, `Lock` prevents this.

diskann-inmem/src/neighbors.rs
hildebrandmw: Yes. Internally, we can guarantee that `self.max_length()` is less than `u32`. A caller doing that conversion looses that information.

diskann-inmem/src/store.rs
hildebrandmw: I'm a little hesitant to always reach directly for `try_drain` as it's potentially fairly expensive. That said, what's currently happening (a full scan before draining) is probably sub-optimal. What about this: what if `scan_acquire` only scans `self.unfrozen.div_ceil(RETRY_LIMIT)`. So it bails a little earlier before doing a reclamation  attempt?

To be honest, all of this needs dedicated benchmarking, though, to figure out what the right thing is. I'm aiming for something reasonable-ish for the initial version.

diskann-inmem/src/neighbors.rs
hildebrandmw: This might require a bit more work to investigate. Initially, it looks like introducing a light-weight hash for lock makes performance considerably worse. Will revisit.

diskann-inmem/src/tag.rs
hildebrandmw: For frozen points, it is "permanent". Though we could use opt-in to the normal epoch handling if we ever needed to replace start points. The `frozen` tag does add an extra level of safety because the usual logic for deleting points will refuse to retire a `FROZEN` tag.

diskann-inmem/src/store.rs
hildebrandmw: The `Freelist` has a overhead that doesn't scale with the size of the dataset - its queue has a defined upper bound. Increasing the queue size may improve performance (by decreasing the number of scans).

ID translation does introduce overhead - on the order of `(2 * std::mem::size_of::<I>() + std::mem::size_of::<u32>()) * N` where `I` is the size of the external ID type and `N` is the number of elements in the dataset.

diskann-inmem/src/ids.rs
hildebrandmw: Done. TLDR: `DashMap` handles a sparse key-space well. But for the backward mapping, we know our IDs are in the range `[0..capacity)`, so we can store the keys implicitly as indices and save some space.

diskann-inmem/src/store.rs
```diff
@@ -0,0 +1,963 @@
...
+//!
+//! ## Reading
+//!
+//! Read access requires a [`Reader`] produced by [`Store::reader`]. [`Reader::read`]
+//! provides read-only access to data at slot `i` if the data is valid for reads.
+//!
+//! ## Writing
+//!
+//! [`Store::acquire`] is used to find and claim an unused internal [`Slot`]. A [`Slot`]
+//! provides write access to its coresponding data which is published when the [`Slot`] is
```
JordanMaples: typo


```suggestion
//! provides write access to its corresponding data which is published when the [`Slot`] is
```

diskann-inmem/src/store.rs
```diff
@@ -0,0 +1,963 @@
...
+//! dropped.
+//!
+//! The index of the slot chosen may be obtained via [`Slot::slot`].k
```
JordanMaples:
```suggestion
//! The index of the slot chosen may be obtained via [`Slot::slot`].
```

diskann-inmem/src/store.rs
```diff
@@ -0,0 +1,963 @@
...
+        // Pad to half a cache line. When data occupies just part of a cache line, this
+        // results in the same total number of cache lines being fetched while potentially
+        // enabling more compact memory.
+        #[expect(
+            clippy::expect_used,
+            reason = "we expect `init` to have at least one row, so this should never happen"
+        )]
+        let padded_bytes = unpadded
+            .checked_next_multiple_of(Bytes::CACHELINE.div(TWO))
+            .expect("unreachabel because `init` cannot exceed `isize::MAX` bytes");
```
JordanMaples:
```suggestion
            .expect("unreachable because `init` cannot exceed `isize::MAX` bytes");
```

diskann-inmem/src/store.rs
```diff
@@ -0,0 +1,963 @@
...
+        // SAFETY: We've guaranteed that `tag` belongs to `slot`.
+        unsafe { self.try_acquire(tag, i) }
+    }
+
+    /// Try to acquire `slot` with the associated `tag`.
+    ///
+    /// # Safety
+    ///
+    /// Caller asserts that `tag` was obtained from `self.tags[slot]`. This is meant as
+    /// a perfomance optimization where `tag` is first queried for potential availability.
```
JordanMaples:
```suggestion
    /// a performance optimization where `tag` is first queried for potential availability.
```

diskann-inmem/src/epoch.rs
```diff
@@ -0,0 +1,847 @@
...
+                drain.is_empty(),
+                "100 must not drain on 1st advance after A"
+            );
+        }
+
+        // Retire 200 at generation B (= A - 1).
+        let gen_b = retire_at(200);
+        assert_eq!(gen_b, gen_a + 1);
+
+        // 2st advance after A: must NOT drain item 100.
```
JordanMaples:
```suggestion
        // 2nd advance after A: must NOT drain item 100.
```

diskann-inmem/src/neighbors.rs
```diff
@@ -0,0 +1,718 @@
...
+        }
+
+        // SAFETY: `self.ptr` is guaranteed to be valid for at least 4-bytes, and we own the
+        // underlying data until `drop`.
+        unsafe { self.ptr.write(newlen as u32) };
+
+        Ok(())
+    }
+
+    /// Write the contents of `neighbors` into `self` without validating lenghts.
```
JordanMaples:
```suggestion
    /// Write the contents of `neighbors` into `self` without validating lengths.
```

diskann-inmem/src/provider.rs
```diff
@@ -0,0 +1,1237 @@
...
+//!   support ergonomic insert or delete guards to protect slots during insert or delete
+//!   operations. This leaves open a situation where an item can be inserted and during
+//!   the insertion algorithm, it is deleted, and then re-inserted.
+//!
+//!   This can cause some issue within the main indexing algorithms which assume the inserted
+//!   ID is present but requires upstream changes to properly fix.
+//!
+//! * Failed insert rollback: again, this needs some upstream changes to full support.
+//!
+//! * Quantization + reranking: Ths current version of this index targets just a single
```
JordanMaples:
```suggestion
//! * Quantization + reranking: The current version of this index targets just a single
```

diskann-inmem/src/layers/full.rs
```diff
@@ -0,0 +1,703 @@
...
+    arch::{Current, FTarget2},
+};
+use half::f16;
+use thiserror::Error;
+
+use crate::{Hidden, layers, num::Bytes};
+
+/// A useful trait bound for types compatible with [`Full`].
+///
+/// This encompases *everything* required for `Full: layers::Insert` and can be used as
```
JordanMaples:
```suggestion
/// This encompasses *everything* required for `Full: layers::Insert` and can be used as
```

JordanMaples (commented):
just some typos

JordanMaples (approved):
I've spent the last handful of days trying to get a good understanding of the underlying EBR mechanisms and how they play into this PR. It all seems sound to me. 
I'm sure there are some things that I might have missed while going over this PR given its size. However, the tests seem to be sound, and it appears to work, so I'll sign off pending the remaining comments from everyone gets addressed.

diskann-inmem/src/epoch.rs
hildebrandmw: Okay, the number of epoch guard slots (and the capacity of the freelist queue) have been propagated up to a top level configuration.

diskann-inmem/src/epoch.rs
hildebrandmw: I'm going to defer this for now. It's possible we can make wrap-around safe, but I also don't think the current use with a 64-bit counter is entirely unreasonable. Please re-open if you disagree.

diskann-inmem/src/epoch.rs
hildebrandmw: Fair - forgot this wasn't a `///` block.

diskann-inmem/src/neighbors.rs
hildebrandmw: I did a few offline experiments, and it really doesn't seem to affect things much. I think at least part of the intuition is that locks are held for pretty short durations. Trying to do something fancier like using a hash for the lock IDs just made things slower, and having one lock per adjacency list wasn't any faster than grouping.

In any case, it's a fairly easy knob to tweak.

diskann-inmem/src/store.rs
hildebrandmw: @magdalendobson - just pinging you on this :smile:

diskann-inmem/src/neighbors.rs
hildebrandmw: After some benchmarking, hashing appears to be strictly worse than the simpler indexing, and a coarser granularity does not appear to be slower than a finer granularity. Part of the intuition, perhaps, is the short duration for which these locks are held, and decreasing the memory required for the locks (i.e. coarser granularity) perhaps compensates for potential concurrency benefits of finer locks? Not sure. In any case, this is a relatively easy assumption to revisit in the future.

diskann-inmem/src/ids.rs
hildebrandmw: A note has been added.

diskann-inmem/src/epoch.rs
hildebrandmw: LMAO - that's a good one.

diskann-inmem/src/store.rs
hildebrandmw: Fixed - thanks!

diskann-inmem/src/store.rs
hildebrandmw: Fixed - thanks!

diskann-inmem/src/store.rs
hildebrandmw: Fixed - thanks!

diskann-inmem/src/store.rs
hildebrandmw: Fixed - thanks!

diskann-inmem/src/epoch.rs
hildebrandmw: Fixed - thanks!

diskann-inmem/src/neighbors.rs
hildebrandmw: Fixed - thanks!

diskann-inmem/src/layers/full.rs
hildebrandmw: Fixed - thanks!

diskann-inmem/src/tag.rs
```diff
@@ -0,0 +1,328 @@
...
+///   the tag before retirement may still exist.
+///
+///   Only transition away from this value when the corresponding slot is returned from a
+///   [`Drain`](super::epoch::Drain).
+///
+/// # Allowed Transitions
+///
+/// The following protocol must be used when working with [`AtomicTag`]ged data and a
+/// [`Registry`](super::Registry).
+///
```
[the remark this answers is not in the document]
hildebrandmw: Fixed.

diskann-inmem/src/buffer.rs
```diff
@@ -0,0 +1,539 @@
...
+
+impl Buffer {
+    /// Construct a new [`Buffer`] capable of holding `entries` with each entry occupying
+    /// exactly `bytes_per_entry`. Subsequent entries are separated by exactly
+    /// `bytes_per_entry` bytes. The base point will be aligned to at least `align`.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the number of bytes `bytes_per_entry * entries` rounded up to
+    /// the next multiple of `align` exceeds `isize::MAX`.
```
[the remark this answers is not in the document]
hildebrandmw: This is what would cause the allocation layout to not be created.

diskann-inmem/src/ids.rs
```diff
@@ -0,0 +1,367 @@
...
+    I: Eq + Hash,
+{
+    pub(crate) fn internal(&self) -> u32 {
+        *self.forward.get()
+    }
+
+    pub(crate) fn delete(mut self) {
+        self.forward.remove();
+        self.backward[self.entry] = None;
+    }
```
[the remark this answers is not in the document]
hildebrandmw: No, this is incorrect. Both `forward` and `backward` acquire write locks to their respective slots, so no reader can observe the contents while `Entry` is live.

diskann-inmem/integration/index/runner.rs
```diff
@@ -0,0 +1,598 @@
...
+            let mut io = std::fs::File::open(&*self.queries)
+                .with_context(|| format!("could not open {}", self.queries.display()))?;
+
+            load_and_convert(&mut io, self.data_type, data_type, &self.preprocess)?
+        };
+
+        let groundtruth = {
+            let mut io = std::fs::File::open(&*self.groundtruth)
+                .with_context(|| format!("could not open {}", self.queries.display()))?;
+
```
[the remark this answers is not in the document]
hildebrandmw: Thanks.

diskann-inmem/src/epoch.rs
hildebrandmw: Updated to `guard_slot`.

diskann-inmem/src/provider.rs
hildebrandmw: Fixed- thanks.

diskann-inmem/src/provider.rs
hildebrandmw: Fixed - thanks!

rfcs/01206-inmem2.md
```diff
@@ -0,0 +1,270 @@
...
+* Faster to compile.
+* Support external/internal ID translation.
+* Have much better tests.
+* Without completely compromising performance.
+
+## Proposal
+
+This RFC accompanies [1206](https://github.com/microsoft/DiskANN/pull/1206), which is a MVP of the proposed design.
+
+The design assumes that traffic to the data store is primarily **read** heavy (to facilitated searches).
```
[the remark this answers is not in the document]
hildebrandmw: Fixed.

diskann-inmem/DEV.md
```diff
@@ -0,0 +1,7 @@
+# Dev Docs
+
+Fully testing this crate requires enabling the `integration-test` feature.
+The suggested command is
+```
+cargo test --package diskann-inmem --all-features --profile ci
+```
```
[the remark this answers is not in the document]
hildebrandmw: Fixed.

diskann-benchmark/src/index/mod.rs
```diff
@@ -15,9 +15,32 @@ mod result;
...
+            diskann_benchmark_runner::Features::new("inmem2"),
+            "Index build-and-search using version 2 of the inmem index",
+        )?;
+
+        registry.register_gated(
+            "inmem2-streaming",
+            "inmem2-stream",
+            diskann_benchmark_runner::Features::new("inmem2"),
+            "Streaming runs using version 2 of the inmem index",
+        )?;
```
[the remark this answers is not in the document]
hildebrandmw: This is intentional - the actual stream benchmarks can encompass many different types, so the gated entry just serves to note that more benchmarks exist.

diskann-benchmark/src/index/mod.rs
```diff
@@ -15,9 +15,32 @@ mod result;
...
+    inmem2::register_benchmarks(registry)?;
+
+    #[cfg(not(feature = "inmem2"))]
+    {
+        registry.register_gated(
+            "inmem2",
+            "inmem2",
+            diskann_benchmark_runner::Features::new("inmem2"),
+            "Index build-and-search using version 2 of the inmem index",
+        )?;
```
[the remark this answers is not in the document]
hildebrandmw: This is intentional.

diskann-inmem/src/buffer.rs
```diff
@@ -0,0 +1,539 @@
...
+/// The memory in the range `[RawSlice::as_ptr(), RawSlice::as_ptr().add(slice.len()))` is
+/// guaranteed to be within a single alive allocation.
+///
+/// This has borrowing semantics of a raw pointer.
+#[derive(Debug)]
+pub(crate) struct RawSlice<'a> {
+    ptr: NonNull<u8>,
+    len: Bytes,
+    _lifetime: PhantomData<&'a ()>,
+}
```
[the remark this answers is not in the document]
hildebrandmw: It contains non-null, that makes it `!Send` and `!Sync`. Besides - your suggestion doesn't even compile!

diskann-inmem/src/store.rs
```diff
@@ -0,0 +1,1045 @@
...
+        #[expect(clippy::panic, reason = "we cannot proceed if we observe this")]
+        fn release(tag: &AtomicTag, kind: &'static str) {
+            // Relaxed ordering is sufficient as all readers/writers are synchronized on
+            // the central generation.
+            if let Err(got) = tag.compare_exchange(
+                Tag::RETIRING,
+                Tag::AVAILABLE,
+                Ordering::Relaxed,
+                Ordering::Relaxed,
+            ) {
```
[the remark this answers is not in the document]
hildebrandmw: Decent find - switched to `Release`.

diskann-inmem/src/store.rs
magdalendobson: Sorry I didn't see this. Yeah, I think your suggestion to not always reach for `try_drain` and instead avoid a full scan by scanning fewer buckets before reclaiming is reasonable. I'm fine with going with that for this PR.

diskann-inmem/src/sharded.rs
hildebrandmw: I had originally envisioned using multiple different types for Id translation depending on the use case. But we can cross that bridge when we get to it. For now, I just renamed this to `IdMap`.

diskann-inmem/src/store.rs
hildebrandmw: Per offline discussion, we're keeping IDs as 32-bits for now.

diskann-inmem/src/store.rs
hildebrandmw: Okay, I actually fixed it this time. Sorry about that.

diskann-inmem/src/provider.rs
```diff
@@ -0,0 +1,1265 @@
...
+
+    fn append_vector(
+        &mut self,
+        id: Self::Id,
+        neighbors: &[Self::Id],
+    ) -> impl std::future::Future<Output = ANNResult<()>> + Send {
+        let work = move || -> ANNResult<()> {
+            self.counters.append_vector(1);
+            let lock = self.reader.neighbors().lock(id)?;
+
```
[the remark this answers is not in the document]
hildebrandmw: Better luck next time ...

harsha-simhadri (approved):

PR #1207, opened 2026-06-29
Fix CRLF line endings caused by broken `.gitattributes`

suri-kumkaran:
**Problem:** The first rule `text eol=crlf` was parsed as the literal filename `text`, not a glob — so no EOL normalization ever applied. Line endings fell back to each committer's local `core.autocrlf`, leaving **16 files committed with CRLF** against the repo's LF convention.

**Fix:**
- `text eol=crlf` → `* text=auto eol=lf` (normalize all text files to LF)
- Drop the dead `*.cer` / `*.qencr` / `*.dat` binary rules (boilerplate from initial import; none exist in the repo, and `text=auto` auto-detects binaries anyway)
- Renormalize the 16 CRLF files to LF (line endings only — no content change)

**Verification:** `git ls-files --eol` now reports **zero CRLF/mixed** across all tracked files; `git check-attr` resolves `text: auto, eol: lf` for source files (previously empty).


harsha-simhadri (approved):

.gitattributes
```diff
@@ -1,7 +1,7 @@
-text eol=crlf
-.cer binary
-.qencr binary
-.dat binary
+* text=auto eol=lf
+*.cer binary
+*.qencr binary
+*.dat binary
```
hildebrandmw: I think we can drop `.cer`/`.qencr`/`.dat`. None of these show up in the repo.

hildebrandmw (approved):
Thanks!

.gitattributes
suri-kumkaran: Make sense, done.

hildebrandmw (approved):

PR #1216, opened 2026-06-30
Support u64 vertex ids in the bftree provider

JordanMaples:
## What

Make the bf-tree provider generic over its vertex-id type so it can index past
the `u32::MAX` (~4.3B) ceiling, while keeping `u32` as the default (no memory or
storage cost for existing datasets). Also add defensive hardening, wide-id test
coverage, and a benchmark-layer path to validate the `u64` id space against
real-world datasets.

- **New `id.rs`** — `BfTreeId` trait with `u32` / `u64` impls, `from_index` /
  `try_from_index`, and `validate_id_capacity`.
- **`provider.rs`** — `BfTreeProvider<T, Q = QuantVectorProvider, I = u32>`.
  The vertex id type threads through construction, save, and load. The persisted
  `SavedParams` records `id_width`; load validates it against the requested type
  so a `u32` index can't be silently reopened as `u64` (or vice versa).
- **`neighbors.rs` / `vectors.rs`** — neighbor cells and starting-point seeding
  are generic over `I`, sized via `size_of::<I>()`.

## Why

Larger-than-memory, billion-scale datasets need vertex ids wider than `u32`.
Defaulting `I = u32` keeps the common case unchanged; opting into `u64` is a
single type parameter at the call site.

## Hardening

Defensive coverage for the wide-id path:
- Capacity guard rejecting id counts that overflow the chosen width.
- Save/load rejects an `id_width` mismatch instead of misreading data.
- Tests for high-bit (`> u32::MAX`) ids and their low-bit collisions, plus a
  `u64` save/load round-trip.

## Wide-id tests + 32-bit compile guard

The earlier `u64`-typed tests only used values in `0..N`, which still fit in
`u32` and would pass even under silent 32-bit truncation. Added tests that
actually exceed `u32::MAX`:
- `test_quantized_index_search_u64_high_ids` — end-to-end quantized insert +
  search with ids above `u32::MAX`, using a sparse `max_points` so no
  billion-scale allocation is needed.
- `test_u64_id_recall_parity_on_transposed_dataset` — builds a random dataset
  twice (dense `u32` identity vs an order-preserving `u64` offset above
  `u32::MAX`) and asserts identical per-query neighbor sets plus a recall floor
  against brute-force ground truth.

Added `BfTreeId::INDEX_CONVERSION_LOSSLESS`, overridden for `u64` to assert a
64-bit `usize` and referenced from the monomorphized `validate_id_capacity`.
This turns `as_index` truncation into a **compile error only for `u64` providers
on 32-bit targets**, leaving the default `u32` path (even on 32-bit) unaffected.

## Real-world `u64` validation (benchmark layer)

Real datasets ship as a dense `0..N` block whose ids always fit in `u32`, so
they never exercise the `u64` paths, and reaching an id `> u32::MAX`
*positionally* would need billions of rows. Added an **opt-in, validation-only**
benchmark path that transposes a dataset's dense `u32` row ids into the
`u64`-only range (`base > u32::MAX`) **without touching production code**:

- `diskann-benchmark-core`: an `Offset` `ToId<u64>` mapper (`base + i`, checked
  add), documented as validation-only.
- `diskann-benchmark`: a `single_or_multi_insert_offset` helper, a
  `U64IdValidation { offset }` JSON config on `BfTreeFullPrecisionBuild`
  (`#[serde(default)]`, non-breaking), a widened `bftree_parameters_offset`, and
  a dedicated `run_u64_id_validation` orchestration that builds a
  `BfTreeProvider<T, NoStore, u64>`, inserts with the offset, and widens
  ground-truth ids via `Matrix::map` to keep recall aligned. This reuses the
  already-generic provider/search/recall stacks rather than relaxing the shared
  `u32` build pipeline.
- Example config + an end-to-end integration test asserting the transposed
  `u64` run reproduces the identity-`u32` recall baseline.

Validated on the 100K Wikipedia/Cohere dataset (offset `2^32`, so every vertex
id lives above `u32::MAX`): recall@100 **0.9536 (u32) vs 0.9537 (u64)** — parity
confirmed (the sub-`1e-4` delta is nondeterministic 8-thread multi-insert batch
ordering, not an id-path difference).

**Deferred:** the streaming benchmark is not yet transposable — its shared
executor is `u32`-pinned (the `TagSlotManager` slot allocator, the
`ManagedStream` trait, the `Managed` translation layer, and `BfTreeStream` all
hardcode `u32`), so a transposed stream run needs a broader genericization of
that shared infrastructure. Tracked as follow-up.

## Notes

- Each commit builds and lints clean in isolation (clean bisect).
- **Open design question:** bf-tree's `I` is an *internal* id. The `inmem2`
  direction uses `InternalId = u32` (hardcoded) + generic *external* id. Worth
  aligning on whether bftree needs wide *internal* ids or should follow the same
  internal-u32 / external-generic split before merge.

### DIFF
--- a/diskann-bftree/src/id.rs
+++ b/diskann-bftree/src/id.rs
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+//! Vertex-id abstraction for the bf-tree provider.
+
+use diskann::utils::{IntoUsize, VectorId};
+
+/// Identifier type usable as a `BfTreeProvider` vertex id.
+///
+/// This bundles the bounds the core algorithm requires of an id ([`VectorId`])
+/// with the ability to convert *to* `usize` ([`IntoUsize`], used to key the
+/// per-vector stores) and *from* a zero-based index. The provider mints ids
+/// densely from `0..total`, so it needs a way to build an `I` from a counter.
+///
+/// Implemented for `u32` (the default, capping at ~4.29B vertices) and `u64`
+/// (for billion-scale-and-beyond, larger-than-memory datasets). On a 64-bit
+/// target `u64` covers every representable `usize`, so its conversions never
+/// fail.
+pub trait BfTreeId: VectorId + IntoUsize {
+    /// Build an id from a zero-based index, truncating on overflow.
+    ///
+    /// Only call this for indices already known to fit (e.g. ids drawn from
+    /// `0..total`, which the provider guarantees fit by construction).
+    fn from_index(index: usize) -> Self;
+
+    /// Build an id from a zero-based index, returning `None` if it does not fit.
+    fn try_from_index(index: usize) -> Option<Self>;
+}
+
+impl BfTreeId for u32 {
+    #[inline(always)]
+    fn from_index(index: usize) -> Self {
+        index as u32
+    }
+
+    #[inline(always)]
+    fn try_from_index(index: usize) -> Option<Self> {
+        u32::try_from(index).ok()
+    }
+}
+
+impl BfTreeId for u64 {
+    #[inline(always)]
+    fn from_index(index: usize) -> Self {
+        index as u64
+    }
+
+    #[inline(always)]
+    fn try_from_index(index: usize) -> Option<Self> {
+        u64::try_from(index).ok()
+    }
+}
--- a/diskann-bftree/src/id.rs
+++ b/diskann-bftree/src/id.rs
@@ -5,21 +5,26 @@
 
 //! Vertex-id abstraction for the bf-tree provider.
 
-use diskann::utils::{IntoUsize, VectorId};
+use std::marker::PhantomData;
+
+use diskann::utils::VectorId;
 use diskann::{ANNError, ANNResult};
 
 /// Identifier type usable as a `BfTreeProvider` vertex id.
 ///
-/// This bundles the bounds the core algorithm requires of an id ([`VectorId`])
-/// with the ability to convert *to* `usize` ([`IntoUsize`], used to key the
-/// per-vector stores) and *from* a zero-based index. The provider mints ids
-/// densely from `0..total`, so it needs a way to build an `I` from a counter.
+/// This bundles the bounds the core algorithm requires of an id ([`VectorId`],
+/// mandated by `HasId::Id` and `DataProvider::InternalId`) with the index
+/// arithmetic the provider needs: converting an id *to* a `usize` index
+/// ([`as_index`](BfTreeId::as_index), used to key the per-vector stores) and
+/// *from* a zero-based index ([`from_index`](BfTreeId::from_index)). The
+/// provider mints ids densely from `0..total`, so it needs a way to build an
+/// `I` from a counter.
 ///
 /// Implemented for `u32` (the default, capping at ~4.29B vertices) and `u64`
 /// (for billion-scale-and-beyond, larger-than-memory datasets). On a 64-bit
 /// target `u64` covers every representable `usize`, so its conversions never
 /// fail.
-pub trait BfTreeId: VectorId + IntoUsize {
+pub trait BfTreeId: VectorId {
     /// Build an id from a zero-based index, truncating on overflow.
     ///
     /// Only call this for indices already known to fit (e.g. ids drawn from
@@ -28,6 +33,22 @@ pub trait BfTreeId: VectorId + IntoUsize {
 
     /// Build an id from a zero-based index, returning `None` if it does not fit.
     fn try_from_index(index: usize) -> Option<Self>;
+
+    /// Convert this id to its zero-based `usize` index.
+    ///
+    /// The provider uses the identity map, so an id *is* its own index. This is
+    /// lossless on the 64-bit targets diskann supports.
+    fn as_index(&self) -> usize;
+
+    /// An iterator over the dense id range `0..total`.
+    ///
+    /// The conversion is monomorphized and inlined per id type (no stored
+    /// function pointer), and the iterator preserves the exact-size and
+    /// double-ended properties of the underlying index range.
+    #[inline]
+    fn id_range(total: usize) -> IdRange<Self> {
+        IdRange::new(total)
+    }
 }
 
 impl BfTreeId for u32 {
@@ -40,6 +61,11 @@ impl BfTreeId for u32 {
     fn try_from_index(index: usize) -> Option<Self> {
         u32::try_from(index).ok()
     }
+
+    #[inline(always)]
+    fn as_index(&self) -> usize {
+        *self as usize
+    }
 }
 
 impl BfTreeId for u64 {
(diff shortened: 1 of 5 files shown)


harsha-simhadri:
Does u64 bloat adjacency list, and if so, is there a chance we can bring in a delta compression ?

JordanMaples:
> Does u64 bloat adjacency list, and if so, is there a chance we can bring in a delta compression ?

@harsha-simhadri yes, in u64 mode it doubles the adjacency-list storage (neighbor ids are stored inline), but that's explicit opt-in and there are no u64 consumers yet. I can tackle delta compression as a follow-up. Neighbor order doesn't affect search correctness, so we can sort ids and delta+varint them, which recovers essentially all of the u64 overhead.

harsha-simhadri:
> > Does u64 bloat adjacency list, and if so, is there a chance we can bring in a delta compression ?
> 
> @harsha-simhadri yes, in u64 mode it doubles the adjacency-list storage (neighbor ids are stored inline), but that's explicit opt-in and there are no u64 consumers yet. I can tackle delta compression as a follow-up. Neighbor order doesn't affect search correctness, so we can sort ids and delta+varint them, which recovers essentially all of the u64 overhead.

thanks, could you create an issue for future follow up. thanks

diskann-bftree/src/provider.rs
```diff
@@ -378,13 +382,13 @@ where
...
-    pub fn starting_points(&self) -> ANNResult<Vec<u32>> {
+    pub fn starting_points(&self) -> ANNResult<Vec<I>> {
         self.full_vectors.starting_points()
     }
 
     /// An iterator over all ids including start points (even if they are deleted).
-    pub fn iter(&self) -> std::ops::Range<u32> {
-        0..(self.full_vectors.total() as u32)
+    pub fn iter(&self) -> std::iter::Map<std::ops::Range<usize>, fn(usize) -> I> {
+        (0..self.full_vectors.total()).map(I::from_index as fn(usize) -> I)
```
hildebrandmw: This is probably not the move, nor is it a great precedent to set. The problem is that `std::iter::Map<.., fn(usize) -> I>` folds the function pointer into the map. In particular, a function pointer is (usually) not inlineable, so the resulting iterator has a large number of undesirable properties, including an indirect call for each evaluation. [Godbolt Link](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,selection:(endColumn:14,endLineNumber:17,positionColumn:14,positionLineNumber:17,selectionStartColumn:14,selectionStartLineNumber:17,startColumn:14,startLineNumber:17),source:'%23%5Binline(never)%5D%0Afn+iter_slow(x:+std::ops::Range%3Cusize%3E)+-%3E+std::iter::Map%3Cstd::ops::Range%3Cusize%3E,+fn(usize)+-%3E+u32%3E+%7B%0A++++let+f+%3D+%7Cx:+usize%7C+x+as+u32%3B%0A++++x.map(f)%0A%7D%0A%0A%23%5Binline(never)%5D%0Afn+iter_fast(x:+std::ops::Range%3Cusize%3E)+-%3E+impl+ExactSizeIterator%3CItem+%3D+u32%3E+%7B%0A++++x.start+as+u32..x.end+as+u32%0A%7D%0A%0Apub+fn+sum_slow(x:+std::ops::Range%3Cusize%3E)+-%3E+u32+%7B%0A++++iter_slow(x).sum()%0A%7D%0A%0Apub+fn+sum_fast(x:+std::ops::Range%3Cusize%3E)+-%3E+u32+%7B%0A++++iter_fast(x).sum()%0A%7D'),l:'5',n:'0',o:'Rust+source+%231',t:'0')),k:34.04004329564551,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:r1940,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,libs:!(),options:'-Copt-level%3D3+-Ctarget-cpu%3Dx86-64-v4',overrides:!((name:edition,value:'2024')),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+rustc+1.94.0+(Editor+%231)',t:'0')),header:(),k:40.94066278201313,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'rustc+1.94.0',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+rustc+1.94.0+(Compiler+%231)',t:'0')),k:25.019293922341355,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

It's probably worth baking more logic into the `BfTreeId` trait to avoid needing these kinds of inefficient fallbacks.

diskann-bftree/src/id.rs
```diff
@@ -0,0 +1,74 @@
...
+/// This bundles the bounds the core algorithm requires of an id ([`VectorId`])
+/// with the ability to convert *to* `usize` ([`IntoUsize`], used to key the
+/// per-vector stores) and *from* a zero-based index. The provider mints ids
+/// densely from `0..total`, so it needs a way to build an `I` from a counter.
+///
+/// Implemented for `u32` (the default, capping at ~4.29B vertices) and `u64`
+/// (for billion-scale-and-beyond, larger-than-memory datasets). On a 64-bit
+/// target `u64` covers every representable `usize`, so its conversions never
+/// fail.
+pub trait BfTreeId: VectorId + IntoUsize {
```
hildebrandmw: Suggestion: Try to avoid `IntoUsize` as a trait bound and just make this an `as_index` method of `BfTreeId`. It's a little more defensive and self contained. It may also be a good idea to drop the `VectorId` bound as well and instead include the required traits directly, but that's perhaps less important since `VectorId` still pulls in way too much.

harsha-simhadri:
- Coud you please link the follow up issues on delta encoding for u64 in bf-tree
- Share the testing done for u64 id type, and the results.

JordanMaples:
@harsha-simhadri 
> * Coud you please link the follow up issues on delta encoding for u64 in bf-tree

The follow up issue is already linked above your comment. 
https://github.com/microsoft/DiskANN/issues/1234

JordanMaples:
@harsha-simhadri
As far as tests for u64 go, I added 3 unit tests in this PR but haven't run against a real-word dataset.

diskann-bftree/src/provider.rs
```diff
@@ -2268,13 +2531,217 @@ mod tests {
...
+
+            let index = Arc::new(DiskANNIndex::new(index_config, provider, None));
+            let ctx = &DefaultContext;
+
+            for (row, vector) in data.iter().enumerate() {
+                index
+                    .insert(&FullPrecision, ctx, &id_of(row), vector.as_slice())
+                    .await
+                    .unwrap();
+            }
```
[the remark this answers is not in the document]
JordanMaples: it compiles though. Marking as false.

harsha-simhadri (approved):

hildebrandmw (approved):

PR #1228, opened 2026-07-06
Add algorithm and benchmark for filtered range search

### LINKED ISSUE #1179: Support filtered range search

magdalendobson:
This PR adds an algorithm for filtered range search to the DiskANN repository. Details of the algorithm, along with the experiments supporting it, are at [this Wiki page](https://github.com/microsoft/DiskANN/wiki/Filtered-Range-Search-in-DiskANN).

This PR:
1. Adds the new algorithm to `diskann`, along with integration tests.
2. Adds the new algorithm as an option in `diskann-benchmark`, along with an integration test.
3. Adds filtered range groundtruth files in `test_data`.  

Some enhancements/fixes to range search along the way:
1. Added range groundtruth in `test_data` for yfcc.
2. Fixed a bug where range search was not always honoring the maximum number of returned results.
3. Added integration tests for range search making sure a set maximum number of results is actually respected.


diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,490 @@
...
+            });
+
+            let result_count = processor
+                .post_process(
+                    &mut accessor,
+                    query,
+                    all_matched.iter().copied(),
+                    &mut filtered,
+                )
+                .await
```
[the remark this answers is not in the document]
magdalendobson: Good suggestion

diskann/src/graph/test/cases/range_search.rs
```diff
@@ -28,11 +28,41 @@ use crate::{
     },
 };
 
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub(super) struct RangeSearchBaseline {
+    pub(super) grid_size: usize,
```
hildebrandmw, line 39: Request: can you add a "description" field to describe what's being tests and what the expected results are? Something like [this](https://github.com/microsoft/DiskANN/blob/053830b75234866d63b95f154791f6f100f58293/diskann/src/graph/test/cases/grid_search.rs#L110-L116).

We want to help future developers a little.

diskann/src/graph/search/range_search.rs
```diff
@@ -212,6 +212,7 @@ where
 
             let stats = if scratch.in_range.len()
                 >= ((self.starting_l() as f32) * self.initial_slack()) as usize
+                && scratch.in_range.len() < self.max_returned().unwrap_or(usize::MAX)
```
hildebrandmw: Can this bug fix be split out into its own little PR?

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,491 @@
...
+                .best
+                .iter()
+                .take(self.starting_l())
+                .chain(matched_results.iter().copied())
+            {
+                if neighbor.distance <= self.radius() * self.range_slack() {
+                    merged
+                        .entry(neighbor.id)
+                        .and_modify(|best: &mut Neighbor<_>| {
+                            if neighbor.distance < best.distance {
```
hildebrandmw: What's happening here? Why are we expecting repeats for IDs and why is picking the best the right choice?

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,491 @@
...
+                        })
+                        .or_insert(neighbor);
+                }
+            }
+
+            let mut in_range: Vec<_> = merged.into_values().collect();
+            in_range.sort_unstable_by(|left, right| {
+                left.distance
+                    .total_cmp(&right.distance)
+                    .then_with(|| left.id.cmp(&right.id))
```
hildebrandmw: Don't `Neighbor`s already sort themselves?

diskann-benchmark-core/src/search/graph/filtered_range.rs
```diff
@@ -0,0 +1,216 @@
...
+            setup: run.setup().clone(),
+            parameters: *run.parameters(),
+            end_to_end_latencies: results.iter().map(|r| r.end_to_end_latency()).collect(),
+            mean_latencies,
+            p90_latencies,
+            p99_latencies,
+            average_precision,
+        })
+    }
+}
```
hildebrandmw, line 215: Please add tests like all the other entries in `search/graph/`

diskann-benchmark/src/index/benchmarks.rs
```diff
@@ -571,7 +575,84 @@ where
...
+
+        let filtered_range = benchmark_core::search::graph::filtered_range::FilteredRange::new(
+            index,
+            queries,
+            benchmark_core::search::graph::Strategy::broadcast(strategy.inner()),
+            labels,
+        )?;
+
+        let result = search::range::run(&filtered_range, &groundtruth, steps, |range_search| {
+            diskann::graph::search::FilteredRange::with_options(
```
hildebrandmw: This giant wall is of constructors is repeated multiple times. Please rework it into a better abstraction.

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,491 @@
...
+        }
+    }
+}
+
+/// A [`SearchOutputBuffer`] wrapper that filters results by distance before
+/// forwarding them to an inner buffer.
+struct DistanceFiltered<'a, F, B: ?Sized> {
+    predicate: F,
+    inner: &'a mut B,
+}
```
hildebrandmw: Is this copied verbatim from `range_search.rs`? Why is that?

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,491 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+//! Range-based search within a distance radius.
```
hildebrandmw: Copy+paste error?

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,498 @@
...
+                    matched_from_first_round.push(neighbor);
+                }
+            }
+
+            // clear the visited set and repopulate it with all in-range points found so far, filtered and unfiltered
+            scratch.visited.clear();
+            for neighbor in in_range.iter() {
+                scratch.visited.insert(neighbor.id);
+            }
+            scratch.in_range = in_range;
```
hildebrandmw: I feel like we really need to stop messing with the internals of `scratch` like this and actually develop a proper API.

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,498 @@
...
+
+impl FilteredRange {
+    /// Create range search with default slack values.
+    pub fn new(starting_l: usize, radius: f32) -> Result<Self, RangeSearchError> {
+        let range_params = Range::with_options(None, starting_l, None, radius, None, 1.0, 1.0)?;
+        Ok(Self { range_params })
+    }
+
+    /// Create range search with full options.
+    #[allow(clippy::too_many_arguments)]
```
hildebrandmw: We should probably listen to clippy here. This is a *lot* of arguments, and while it mirrors `Range` - I don't think that's sufficient justification. There are several problems:

1. There's no real documentation about *how* these are expected to affect the algorithm.
2. Callers of `FilteredRange::with_options` just have a sea of values, which can be difficult to remember.
3. Any changes to the parameters requires a breaking change for all callers.

It would be way better to use a [builder](https://github.com/microsoft/DiskANN/blob/a37302639e9828c25878d8e1752296543df8c8e6/diskann/src/graph/config/mod.rs#L576-L592) interface for these. As a side-benefit, the same builder could be used for both `Range` and `FilteredRange`.

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,498 @@
...
+        context: &'a DP::Context,
+        query: T,
+        output: &mut OB,
+    ) -> impl SendFuture<ANNResult<Self::Output>>
+    where
+        O: Send,
+        PP: glue::SearchPostProcess<S::SearchAccessor, T, O> + Send + Sync,
+        OB: SearchOutputBuffer<O> + Send + ?Sized,
+    {
+        async move {
```
hildebrandmw, line 133: This and `filtered_range_search_internal` need at least one clean up pass, if not more. The intermediate storage states are messy and often redundant.

Several things I notices:
* The computation `self.range() * self.range_slack()` happens often, especially within loops. Should this not be a pre-computed value?
* We may be able to skip `merged` entirely with a  vec + sort + dedup
* I don't know what's happening with `scratch.in_range`. It's overwritten in `scratch` simply to pass it as an argument to `filtered_range_search_internal` (this could be completely avoided and likely save an allocation). Further, it is **mutated** in `filtered_range_search_internal` but that mutation is never observed. So why is it being pushed to?
* The initialization of `scratch.range_frontier` is sketchy - it's really relying on the pooled variable being cleared properly to prevent garbage.
* Many repetitions of the pattern `self.max_returned().unwrap_or(usize::MAX)`, which presumably is related to the bug fix in `range_search`. Perhaps this should be a dedicated helper or done just once?

Basically, there are a lot of little things and implicit coupling that could do with a solid scrutinization. Much of this is inherited from range search, but many of my complaints also hold for range search; newer code should be held to a higher standard.

diskann/src/graph/glue.rs
```diff
@@ -314,7 +314,7 @@ impl<T> Reject<T> {
 /// Filter decision for [`FilteredAccessor`].
 ///
 /// Both variants carry the item `T` since rejected items are useful for graph navigation.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
hildebrandmw: Why are `PartialEq` and `Eq` needed? The code compiles fine without them.

diskann/src/graph/test/cases/filtered_range_search.rs
```diff
@@ -0,0 +1,505 @@
...
+    let filter = AlwaysTrueFilter;
+
+    let filtered_range =
+        FilteredRange::with_options(None, starting_l, None, radius, Some(inner_radius), 1.0, 1.0)
+            .unwrap();
+
+    let (filtered_stats, filtered_results) =
+        run_filtered_range_search(&index, query.as_slice(), filtered_range, &filter);
+
+    let baseline = RangeSearchBaseline {
```
hildebrandmw: There's probably a better way to construct these baselines. A constructor taking the stats object and the `Range` could automatically fill in a lot of the fields (and we could capture *all* the fields of `Range~ rather then just like 3).

diskann-benchmark/src/index/search/range.rs
```diff
@@ -30,21 +31,26 @@ impl<'a> RangeSearchSteps<'a> {
...
-        parameters: Vec<Run>,
+        parameters: Vec<core_search::Run<Self::Parameters>>,
         groundtruth: &dyn benchmark_core::recall::Rows<I>,
     ) -> anyhow::Result<Vec<RangeSearchResults>>;
 }
 
-pub(crate) fn run<I>(
-    runner: &dyn Range<I>,
+pub(crate) fn run<I, R, F>(
+    runner: &R,
```
hildebrandmw: Why did this change from a `&dyn Range` to a true generic? The former is quite important to keep compile times down.

Presumably it's because there's a different in the parameters being passed. Please find an alternate solution that isn't as regressive on compile times.

hildebrandmw (changes requested):
Thanks Magdalen. I've left a number of comments throughout the PR, but they all fall into a few general themes.

* This PR duplicates a lot of logic when that duplication should be abstracted instead. The worst offender is probably `DistanceFiltered`.
* We need to move away from messing around with the internal of scratch space (`in_range`, `range_frontier`, `visited` etc) - that can quickly become very brittle.
* The logic in the filtered search implementation can be tightened.
* We should start tagging baselines with more verbose descriptions of *what* the test is covering and what the corner cases being exercised are.
* Please keep to high level trait objects in `diskann-benchmark` instead of generics. The former is significantly better for compile times.

In addition, is the intention of this algorithm for research, or is there a production ask for it? I ask because maintaining such algorithms/code is not free. If this is meant for research, please put it behind an `experimental` feature flag to avoid committing us to it long term.

diskann/src/graph/search/range_search.rs
magdalendobson: Done, I split this and the new tests out to #1253

diskann/src/graph/test/cases/range_search.rs
magdalendobson: Done

diskann/src/graph/search/filtered_range_search.rs
magdalendobson: Overwriting the distance was an error and I've gotten rid of it now. We are expecting repeats because we are merging the best elements found so far with all the filter-satisfying elements found so far, which might have overlap.

diskann/src/graph/search/filtered_range_search.rs
magdalendobson: No, because `scratch.best` is merged with the predicate-satisfying points, which are not in sorted order.

diskann/src/graph/search/filtered_range_search.rs
magdalendobson: Thanks, I believe I have fixed most of these now. I have:
- Removed computation `self.range() * self.range_slack()` in several places. In some cases it was being used mistakenly, because we don't test for `range_slack > 1.0` sufficiently (at all). I will follow up with another PR with tests for this value.
- We now use a vec + sort + dedup to gather all points within the radius
- I got rid of use of `scratch.in_range` and now directly modify only `scratch.visited` and `scratch.range_frontier`, which I now properly clear and then initialize
- pattern `self.max_returned().unwrap_or(usize::MAX)` is now called just twice, once in each search function

diskann/src/graph/glue.rs
magdalendobson: I think this was leftover from an intermediate state. Removed now.

diskann-benchmark-core/src/search/graph/filtered_range.rs
magdalendobson: Yes, done

diskann/src/graph/search/filtered_range_search.rs
magdalendobson: ty

diskann/src/graph/search/filtered_range_search.rs
magdalendobson: Because I'm silly

diskann/src/graph/search/filtered_range_search.rs
magdalendobson: When you're back let's discuss overhauling the use of scratch members in all kinds of range search

diskann/src/graph/search/filtered_range_search.rs
magdalendobson: Thanks, I set up a builder

diskann/src/graph/test/cases/filtered_range_search.rs
magdalendobson: Thanks for the suggestion, I added a constructor

diskann-benchmark/src/index/search/range.rs
magdalendobson: I found an alternative way to do this, so the generic isn't used anymore

diskann-benchmark/src/index/benchmarks.rs
magdalendobson: This was cleaned up when I made other changes

diskann/src/graph/test/cases/range_search.rs
```diff
@@ -113,6 +151,10 @@ fn assert_range_invariants(results: &[Neighbor<u32>], radius: f32, inner_radius:
 
 #[test]
 fn basic_range_search() {
+    let description = "Basic range search test to validate that the range /
+     search returns results within the specified radius and that there are /
+     no duplicate results.";
```
hildebrandmw, line 160: Wrong slash type. I think you want a backslash `\`.

diskann/src/graph/test/cases/range_search.rs
```diff
@@ -205,6 +254,10 @@ fn inner_radius_filtering() {
 
 #[test]
 fn two_round_search() {
+    let description = "Two round search test to validate that a /
+    low starting L with a large radius triggers a second round /
+    of range search.";
```
hildebrandmw, line 265: Use backslash `\`?

diskann/src/graph/search/range_search.rs
```diff
@@ -276,13 +347,13 @@ where
 
 /// A [`SearchOutputBuffer`] wrapper that filters results by distance before
 /// forwarding them to an inner buffer.
-struct DistanceFiltered<'a, F, B: ?Sized> {
+pub(crate) struct DistanceFiltered<'a, F, B: ?Sized> {
```
hildebrandmw: `pub(super)` is probably sufficient.

diskann/src/graph/search/range_search.rs
```diff
@@ -407,14 +521,23 @@ mod tests {
         assert!(Range::new(0, 0.5).is_err());
 
         // Invalid slack values
-        assert!(Range::with_options(None, 100, None, 0.5, None, 1.5, 1.0).is_err());
-        assert!(Range::with_options(None, 100, None, 0.5, None, 1.0, 0.5).is_err());
+        assert!(Range::builder(100, 0.5).initial_slack(1.5).build().is_err());
```
hildebrandmw, line 533: No action needed, just wanted to say I love this so much more.

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -159,16 +159,16 @@ where
 }
 
 #[derive(Debug)]
-struct Ret<I>
+pub(crate) struct Ret<I>
```
hildebrandmw, line 157: As mentioned elsewhere, `pub(super)` is sufficient.

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,358 @@
...
+        index::{DiskANNIndex, InternalSearchStats, SearchStats},
+        search::inline_filter_search::{Ret, inline_filter_search_internal},
+        search::{Range, RangeSearchError, range_search::RangeBuilder, record::NoopSearchRecord},
+        search_output_buffer::SearchOutputBuffer,
+    },
+    neighbor::Neighbor,
+    provider::DataProvider,
+};
+
+use super::range_search::DistanceFiltered;
```
hildebrandmw: Nit: the imports here are kind of all over the place: two `use super`, and the `use crate` has imports at the same hierarchy as `super`. Recommend just consolidating to a single `use crate`.

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+}
+
+impl RangeBuilder {
+    /// Build validated [`FilteredRange`] parameters.
+    pub fn build_filtered(self) -> Result<FilteredRange, RangeSearchError> {
+        let range_params = self.build()?;
+        Ok(FilteredRange { range_params })
+    }
+}
```
hildebrandmw, line 109: Please move this to where `RangeBuilder` is defined. Scattered `impl` blocks should usually be avoided.

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,358 @@
...
+                .take(self.starting_l())
+                .chain(matched_results.iter().copied())
+                .filter(|neighbor| neighbor.distance <= self.radius())
+                .collect();
+
+            in_range.sort_unstable_by(|left, right| {
+                left.id
+                    .cmp(&right.id)
+                    .then_with(|| left.distance.total_cmp(&right.distance))
+            });
```
hildebrandmw: I think it's probably fine to assume that if there are duplicates that they will retain the same `(id, distance)` pair (we can throw in a debug assert) as we no longer allow tweaking of distances that the old `QueryLabelProvider` allowed and the user returning an ID multiple times in `expand_beam_filtered` would be a contract violation. With that in mind a total order by distance then ID should be sufficient, requiring only a single sort then dedup.

In any case, the code below is already relying on this assumption since `matched_within_radius` is only constructed from `matched_results`. If `best` ever diverged somehow, this could cause issues.

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,358 @@
...
+                    matched_within_radius.push(neighbor);
+                }
+            }
+
+            // clear the visited set and repopulate it with all in-range points found so far, filtered and unfiltered
+            scratch.visited.clear();
+            scratch.range_frontier.clear();
+            for neighbor in in_range.iter() {
+                scratch.visited.insert(neighbor.id);
+                scratch.range_frontier.push_back(neighbor.id);
```
hildebrandmw: Small optimization: using [`Extend`](https://doc.rust-lang.org/std/iter/trait.Extend.html) for the re-population is probably a little more efficient.

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,358 @@
...
+            let inner_radius = self.inner_radius();
+
+            let mut filtered = DistanceFiltered::new(output, |dist| {
+                if let Some(ir) = inner_radius
+                    && dist <= ir
+                {
+                    return false;
+                }
+                dist <= radius
+            });
```
hildebrandmw, line 231: Question: should we be early filtering by inner radius in addition to the outer radius - especially when computing the number of matched results at various intermediate steps? I ask because it's possible if there are a lot of results less than the inner radius, `max_returned` can trigger but this final filtering step would remove everything.

Another benefit of doing early inner-radius filtering would be that we could potentially avoid `DistanceFiltered` entirely since `MatchedResults` would already be correctly constrained.

If we're going down that route, I'd suggest using a small helper struct to make it happen automatically:
```rust
struct InRange<I> {
    candidates: Neighbor<I>,
    lower: Option<f32>,
    upper: f32,
}

impl<I> InRange<I> {
    fn push(&mut self, n: Neighbor<I>) {
        if n.distance <= self.distance 
            && let Some(lower) = self.lower
            && n.distance >= self.lower {
            self.candidates.push(n);
        }
    }
}
```

diskann-benchmark-core/src/search/graph/filtered_range.rs
```diff
@@ -0,0 +1,379 @@
...
+        parameters: &Self::Parameters,
+        buffer: &mut O,
+        index: usize,
+    ) -> ANNResult<Self::Output>
+    where
+        O: graph::SearchOutputBuffer<DP::ExternalId> + Send,
+    {
+        let context = DP::Context::default();
+        let filtered_range_search =
+            graph::search::FilteredRange::builder(parameters.starting_l(), parameters.radius())
```
hildebrandmw: I see what's happening here: we want to keep the parameters the same to manage monomorphization in `diskann-benchmark`. But, it might be better to define a conversion from `Range` to `FilteredRange` rather than going back through the builder (which runs the risk of going stale). I'm not sure what the implications are, though, if we ever decide that `FilteredRange` needs to have additional parameters, but I guess we'll cross that bridge when we come to it.

diskann-benchmark/src/index/inmem/spherical.rs
```diff
@@ -428,6 +431,63 @@ mod imp {
...
+
+            let bit_maps = generate_bitmaps(
+                &filtered_range.query_predicates,
+                &filtered_range.data_labels,
+            )?;
+
+            let labels: Arc<[_]> = bit_maps
+                .into_iter()
+                .map(utils::filters::as_query_label_provider)
+                .collect();
```
hildebrandmw, line 471: We should really be thinking of factoring these patterns out. I don't think there's anything actionable for *this* PR, but it's starting to become a problem.

hildebrandmw (commented):
Thanks Magdalen, appreciate the work on this iteration!

diskann-benchmark/src/index/inmem/spherical.rs
magdalendobson: Marking resolved

diskann-benchmark-core/src/search/graph/filtered_range.rs
magdalendobson: Defined a conversion

diskann/src/graph/search/filtered_range_search.rs
magdalendobson: Yeah, this is a good point. This is kind of the result of me just putting all these parameters in when requested externally without deeply considering them in the design. Let's do a follow-up that addresses this plus the extra members of `scratch` in both filtered range and regular range to minimize the footprint of this PR.

diskann/src/graph/search/filtered_range_search.rs
magdalendobson: Done

diskann/src/graph/search/filtered_range_search.rs
magdalendobson: Sounds good, I refactored to reflect the assumption.

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Done

diskann/src/graph/search/filtered_range_search.rs
magdalendobson: You're right, fixed now

diskann/src/graph/search/filtered_range_search.rs
magdalendobson: Done

diskann/src/graph/test/cases/range_search.rs
magdalendobson: Done

diskann/src/graph/test/cases/range_search.rs
magdalendobson: Done

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,356 @@
...
+        async move {
+            let mut accessor = strategy
+                .search_accessor(&index.data_provider, context, query)
+                .into_ann_result()?;
+            let num_start_ids = accessor.num_starting_points().await?;
+            let mut scratch = index.search_scratch(self.starting_l(), num_start_ids);
+
+            // Perform an initial inline filtered search, store both filtered and unfiltered results
+
+            let search_knn = Knn::new(self.starting_l(), self.beam_width())?;
```
magdalendobson: Add conversion from range_params to Knn?

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,490 @@
...
+                let mut all_matched = matched_from_first_round;
+                let mut matched_from_second_round = matched_from_second_round;
+                all_matched.append(&mut matched_from_second_round);
+
+                (
+                    InternalSearchStats {
+                        cmps,
+                        hops: hops + range_stats.hops,
+                        range_search_second_round: true,
+                    },
```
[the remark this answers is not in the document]
magdalendobson: Good suggestion

diskann/src/graph/search/range_search.rs
```diff
@@ -50,9 +53,9 @@ pub struct Range {
     /// Maximum results to return (None = unlimited).
     max_returned: Option<usize>,
     /// Initial search list size.
-    starting_l: usize,
-    /// Optional beam width.
-    beam_width: Option<usize>,
+    starting_l: NonZeroUsize,
+    /// Beam width.
+    beam_width: NonZeroUsize,
```
hildebrandmw, line 58: One thing for the larger range search cleanup: it would be nice to have a central summary of what all these parameters are expected to do algorithmically.

diskann/src/graph/test/cases/filtered_range_search.rs
```diff
@@ -0,0 +1,439 @@
...
+///////////////////////////////////////////////////////
+// Tests with an always-true filter that validate    //
+// that filtered range search gives correct answers  //
+// and respects parameters such as inner_radius and  //
+// max_results. Note that identical behavior to the  //
+// unfiltered range search is NOT expected even when //
+// using an always-true filter, because the filtered //
+// initial search returns every predicate-satisfying //
+// point it finds as opposed to only those in        //
+// `scratch.best`.                                   //
```
hildebrandmw: Please avoid referring to specific variable names when describing behavior. If the variable name changes, this will go out of date.

diskann/src/graph/test/cases/filtered_range_search.rs
```diff
@@ -0,0 +1,439 @@
...
+    assert_range_invariants(&filtered_results, radius, None);
+    assert_no_duplicates(&filtered_results);
+}
+
+#[test]
+fn max_results_respected_and_second_round_triggered() {
+    let description = "Test of `max_results` that sets `initial_l_search` \
+     below `max_results` so that a second round of search is triggered. Note that \
+     the result set size is expected to be below `max_results` due to filtering \
+     out start points during `post_process`.";
```
hildebrandmw, line 306: Should we increase the internal `max_results` in case start points are filtered?

diskann/src/graph/test/cases/range_search.rs
```diff
@@ -28,11 +29,71 @@ use crate::{
     },
 };
 
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub(super) struct RangeSearchBaseline {
+    /// A description of what to expect, what trends to observe, and anything else
+    /// a reviewer may need to either understand why this test is checked in or to validate
+    /// any changes that occur in the checked-in file.
+    pub(super) description: String,
+    pub(super) grid_size: usize,
```
hildebrandmw, line 39: Can we include `grid_dim` as well?

diskann/src/graph/search/filtered_range_search.rs
```diff
@@ -0,0 +1,362 @@
...
+                    &self,
+                    &mut accessor,
+                    &mut scratch,
+                    &mut matched_within_radius,
+                )
+                .await?;
+
+                InternalSearchStats {
+                    cmps: cmps + range_stats.cmps,
+                    hops: hops + range_stats.hops,
```
hildebrandmw: Is there some double counting? As far as I can tell, there is no reset of the variables in scratch, so they contain history from the first call to `inline_filter_search_internal`.

hildebrandmw (commented):
Thanks Magdalen, just a few small comments. After this PR lands, need a serious garbage collection pass on algorithm implementations. The sharing of scratch space and general code organization has reached a point where we need to reevaluate the design. This reevaluation is not actionable for this PR.

diskann/src/graph/search/knn_search.rs
```diff
@@ -110,6 +110,13 @@ impl Knn {
     pub fn beam_width(&self) -> NonZeroUsize {
         self.beam_width
     }
+
+    pub(crate) fn new_unchecked(l_value: NonZeroUsize, beam_width: NonZeroUsize) -> Self {
```
metajack: This name smells wrong. `unchecked` usually implies that it is also unsafe, in that some assumption is unchecked, usually for performance reasons. I cannot find a method with `unchecked` in the name that isn't unsafe in the standard library.

Two suggestions:
1. Use `new_infallible()` instead (or something similar)
2. Rename `new()` to `try_new()` and make this constructor `new()`. `try_new()` appears common in the standard library, although the corresponding `new()` is not always infallible.

diskann-benchmark-core/src/search/graph/filtered_range.rs
```diff
@@ -0,0 +1,371 @@
...
+    type Id = DP::ExternalId;
+    type Parameters = graph::search::FilteredRange;
+    type Output = Metrics;
+
+    fn num_queries(&self) -> usize {
+        self.queries.nrows()
+    }
+
+    fn id_count(&self, parameters: &Self::Parameters) -> search::IdCount {
+        search::IdCount::Dynamic(NonZeroUsize::new(parameters.starting_l().into()))
```
metajack, line 109: Should this not be the k value instead of Lsearch?

metajack (approved):

diskann/src/graph/test/cases/filtered_range_search.rs
magdalendobson: Yeah, I suppose we could increase it by the number of start points during search, and then truncate to the "real" value at the end of the calculation. Since that would need to be done for range and filtered range, I'll defer to the follow-up if that's okay with you.

diskann-benchmark-core/src/search/graph/filtered_range.rs
magdalendobson: There is no `k_value` for range search since there's no requested k.

diskann/src/graph/search/range_search.rs
magdalendobson: Noted, marking resolved

diskann/src/graph/test/cases/range_search.rs
magdalendobson: Yep, done

diskann/src/graph/search/filtered_range_search.rs
magdalendobson: You're right, fixed

diskann/src/graph/test/cases/filtered_range_search.rs
magdalendobson: Done

diskann/src/graph/search/knn_search.rs
magdalendobson: Changed to `new_infallible`

hildebrandmw (approved):

PR #1230, opened 2026-07-06
[diskann-garnet] Fix cache coherence on set_neighbors

metajack:
When diagnosing some issues last week I realized that there was a cache coherency problem with `set_neighbors()`. `append_vector()` updates the start point neighbor cache atomically alongside the real neighbor list, but `set_neighbors()` did two writes sequentially which could result in coherency issues.

As far as I'm aware, this did not cause any actual problems, but it was easy enough to make it coherent by using RMW instead in `set_neighbors()`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -789,7 +789,7 @@ dependencies = [
 
 [[package]]
 name = "diskann-garnet"
-version = "4.0.0"
+version = "4.0.1"
 dependencies = [
  "bytemuck",
  "crossbeam",
--- a/diskann-garnet/Cargo.toml
+++ b/diskann-garnet/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "diskann-garnet"
-version = "4.0.0"
+version = "4.0.1"
 edition = "2024"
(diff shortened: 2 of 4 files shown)


harsha-simhadri:
Could you share before and after perf numbers? Perhaps we should make it easier to run some of these going forward

hildebrandmw (approved):

metajack:
> Could you share before and after perf numbers? Perhaps we should make it easier to run some of these going forward

It appears negligible at least on 1M scale dataset.

build (with fix):

RESULTS (1000000 vectors ingested in 406.11s):
    vps: 2462.41

search (with fix):

RESULTS (6144 queries finished in 2.22s):
        qps: 2768.76
     recall: 93.89%
    latency: 10633.92us

build (before fix):

RESULTS (1000000 vectors ingested in 403.97s):
    vps: 2475.41

search (before fix):
RESULTS (6144 queries finished in 2.31s):
        qps: 2659.20
     recall: 93.92%
    latency: 10719.20us

harsha-simhadri (approved):

PR #1232, opened 2026-07-07
Refactor(diskann-benchmark): consolidate disk search config under DiskSearchMode

dyhyfu:
## Summary

Consolidates disk-index search configuration under `DiskSearchMode` and decouples the benchmark's JSON input schema from the optional `diskann-disk` backend.

Previously, the input schema both defined the config and constructed `diskann_disk::SearchMode`, forcing several `#[cfg(feature = "disk-index")]` gates and scattering related fields (`vector_filters_file`, `post_processor`) across `DiskSearchPhase`. This PR moves the backend-specific construction into the search execution path and groups the search-mode fields together.

## Changes

- **Move `SearchMode` construction to the backend.** The match logic that builds `diskann_disk::SearchMode` now lives in a `build_search_mode` helper in the search execution module, so the input schema is pure config data with no dependency on the disk backend's `SearchMode` type.
- **Consolidate fields into `DiskSearchMode`.** `vector_filters_file` and `post_processor` are nested inside `DiskSearchMode` alongside `is_flat_search` and `adaptive_l`. Validation and `Display` moved accordingly.
- **Remove unnecessary feature gates.** `DiskSearchMode` and its fields no longer need `#[cfg(feature = "disk-index")]`. The only remaining gates are for `QuantizationType`, which is a `diskann-disk` type (intentionally left as-is — see below).
- **Migrate JSON fixtures** (examples and perf test inputs) to the nested `search_mode` format.


diskann-benchmark/src/inputs/disk.rs
```diff
@@ -6,27 +6,18 @@
 use std::{fmt, num::NonZeroUsize, path::Path};
 
 use anyhow::Context;
```
hildebrandmw, line 8: One request I have (either in this PR or in a follow-up) is that running 
```
cargo test --package diskann-benchmark --features disk-index --profile ci
```
generates the following files
```
diskann-benchmark/siftsmall_index_filter_flat_disk.index
diskann-benchmark/siftsmall_index_filter_flat_pq_compressed.bin
diskann-benchmark/siftsmall_index_filter_flat_pq_pivots.bin
diskann-benchmark/siftsmall_index_filter_graph_disk.index
diskann-benchmark/siftsmall_index_filter_graph_pq_compressed.bin
diskann-benchmark/siftsmall_index_filter_graph_pq_pivots.bin
```

Can this be fixed to properly use a tempdir or a virtual file system to avoid polluting the workspace? Thanks!

dyhyfu: Fixed, thanks for discovering this problem.

diskann-benchmark/src/disk_index/search.rs
```diff
@@ -158,6 +160,52 @@ impl DiskSearchResult {
     }
 }
 
+/// Construct the disk [`SearchMode`] from the JSON-driven [`DiskSearchMode`]
+/// config plus the per-query filter and post-processor supplied at search time.
+fn build_search_mode<'a>(
+    mode: &'a DiskSearchMode,
+    has_vector_filters: bool,
+    vector_filter: &'a HashSet<u32>,
```
wuw92: `vector_filter` and `has_vector_filters` both are derived from `mode.vector_filters_file`. Recommend dropping `has_vector_filters` here to avoid redundancy

diskann-benchmark/src/disk_index/search.rs
```diff
@@ -158,6 +160,52 @@ impl DiskSearchResult {
...
+            .expect("validated adaptive L must construct")
+    });
+
+    match (
+        mode.is_flat_search,
+        has_vector_filters,
+        post_processor,
+        adaptive_l,
+    ) {
+        (true, false, _, _) => SearchMode::flat(),
```
wuw92: nit: Flat search silently discards `post_processor` and `adaptive_l` via the `_` arms, so a config with `is_flat_search: true` plus a `post_processor` (or `adaptive_l`) parses and drops them rather than erroring. 
Worth having `DiskSearchMode::validate` reject those mutually-exclusive combos up front or modeling the mode as an enum ( like `SearchMode`) so the illegal combination is unrepresentable in the first place?

diskann-benchmark/src/inputs/disk.rs
```diff
@@ -155,21 +110,11 @@ pub(crate) struct DiskSearchPhase {
...
-    // Backward compatibility for older benchmark inputs that used
-    // `is_flat_search` directly at the search-phase level.
-    #[cfg(feature = "disk-index")]
-    #[serde(default, skip_serializing)]
-    pub(crate) is_flat_search: Option<bool>,
-    #[cfg(not(feature = "disk-index"))]
-    pub(crate) is_flat_search: bool,
     pub(crate) distance: SimilarityMeasure,
-    pub(crate) vector_filters_file: Option<InputFile>,
     pub(crate) num_nodes_to_cache: Option<usize>,
```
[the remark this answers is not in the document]
dyhyfu: The old benchmark JSON files are also updated.

diskann-benchmark/src/disk_index/search.rs
dyhyfu: Fixed.

diskann-benchmark/src/disk_index/search.rs
dyhyfu: Updated.

diskann-benchmark/src/inputs/disk.rs
```diff
@@ -270,16 +215,7 @@ impl DiskSearchPhase {
...
 
-        #[cfg(feature = "disk-index")]
-        if let Some(is_flat_search) = self.is_flat_search {
-            self.search_mode.is_flat_search = is_flat_search;
-        }
-
-        #[cfg(feature = "disk-index")]
         self.search_mode
             .validate(checker)
             .context("invalid disk search mode")?;
```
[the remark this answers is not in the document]
dyhyfu: Added  "#[serde(deny_unknown_fields)]" to enforce failure of the old "is_falt_search" parameters. Users using this config will get an error.

diskann-benchmark/src/inputs/disk.rs
```diff
@@ -72,69 +63,50 @@ pub(crate) struct DiskIndexBuild {
...
 #[derive(Debug, Serialize, Deserialize, Default)]
 pub(crate) struct DiskSearchMode {
     pub(crate) is_flat_search: bool,
     #[serde(default)]
     pub(crate) adaptive_l: Option<AdaptiveL>,
+    #[serde(default)]
+    pub(crate) vector_filters_file: Option<InputFile>,
+    #[serde(default)]
+    pub(crate) post_processor: Option<TopkPostProcessor>,
 }
```
suri-kumkaran: Can we use a enum here to avoid impossible combinations? Something similar and in spirit of:
```rust
pub enum SearchMode<'a> {
    FlatScan {
        filter: Option<SearchPredicate<'a>>,
    },

    Graph {
        filter: Option<SearchPredicate<'a>>,
    },

    InlineFilter {
        filter: Box<dyn QueryLabelProvider<u32> + 'a>,
        adaptive_l: Option<AdaptiveL>,
    },

    DiverseGraph {
        filter: Option<SearchPredicate<'a>>,
        params: DeterminantDiversityParams,
    },
}
```

I understand `SearchMode` is per query and has relevant types, but can we have something similiar on the benchmark input level eliminating the impossible scenarios?

diskann-benchmark/src/inputs/disk.rs
```diff
@@ -72,104 +63,168 @@ pub(crate) struct DiskIndexBuild {
...
+    /// vector filter, and/or a top-k post-processor.
+    Graph {
+        #[serde(default)]
+        adaptive_l: Option<AdaptiveL>,
+        #[serde(default)]
+        vector_filters_file: Option<InputFile>,
+        #[serde(default)]
+        post_processor: Option<TopkPostProcessor>,
+    },
+}
```
suri-kumkaran, line 88: This is my main comment, and it's a "go one step further", not a "this is wrong".

Turning flat/graph into an enum was the right move. But the backend has four variants keyed on strategy, and this has two keyed on traversal with strategy in the optionals. `build_search_mode` still produces 8 leaf constructions for those 4 strategies, because every strategy is written twice, once for `None` filter and once for `Some`.

Mirroring the backend collapses that. Note `vector_filters_file` is **required** on the inline-filter variant, because the backend's `InlineFilter.filter` is a plain `Box<dyn QueryLabelProvider<u32>>`, not an `Option`:

```rust
#[serde(tag = "mode", rename_all = "kebab-case")]
pub(crate) enum DiskSearchMode {
    Flat              { vector_filters_file: Option<InputFile> },
    Graph             { vector_filters_file: Option<InputFile> },
    GraphInlineFilter { vector_filters_file: InputFile, adaptive_l: Option<AdaptiveL> },
    GraphDiverse      { vector_filters_file: Option<InputFile>, post_processor: DeterminantDiversityParams },
}
```

The mapping then normalizes the predicate once and matches four ways. I compiled this against `diskann-disk` to be sure:
```rust
let filter: Option<SearchPredicate<'a>> =
    vector_filter.map(|f| Box::new(move |vid: &u32| f.contains(vid)) as SearchPredicate<'a>);
match mode {
    Flat  { .. } => SearchMode::FlatScan { filter },
    Graph { .. } => SearchMode::Graph { filter },
    GraphInlineFilter { .. } => {
        let f = vector_filter.expect("inline filter requires a vector filter");
        SearchMode::inline_filter(move |vid: &u32| f.contains(vid), adaptive_l)
    }
    GraphDiverse { post_processor, .. } =>
        SearchMode::DiverseGraph { filter, params: *post_processor },
}
```

What this buys: `adaptive_l` can no longer coexist with a post-processor. This also makes some comments becomes unrepresentable.

diskann-benchmark/src/inputs/disk.rs
```diff
@@ -72,104 +63,168 @@ pub(crate) struct DiskIndexBuild {
...
+        )
+        .expect_err("legacy phase-level search settings must be rejected");
+
+        assert!(error.to_string().contains("is_flat_search"));
+    }
+}
+
 /// Search phase configuration
 #[derive(Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
```
suri-kumkaran, line 215: This is the one thing I'd want settled before merge, because it's a user-facing decision.

`#[serde(deny_unknown_fields)]` is the **only** occurrence in the repository, and it's inconsistent with the rest of the config parsing.

That said, I don't think it should just be deleted here. `main` deliberately carried a back-compat `is_flat_search` field, and this PR removes it. Without `deny_unknown_fields`, an old config could silently ignore `is_flat_search` and flip from flat scan to graph search.

I'd suggest keeping `#[serde(deny_unknown_fields)]` locally for this migration to verify that all existing configs are migrated correctly and no old config changes behavior silently. Once we've verified the migration, we can remove it in this PR.

If we want unknown-field strictness permanently, let's propose it repo-wide in a separate PR.

diskann-benchmark/src/disk_index/search.rs
```diff
@@ -158,6 +160,58 @@ impl DiskSearchResult {
...
+            match (post_processor, adaptive_l, vector_filter) {
+                (Some(TopkPostProcessor::DeterminantDiversity(params)), _, None) => {
+                    SearchMode::diverse_graph(*params)
+                }
+                (Some(TopkPostProcessor::DeterminantDiversity(params)), _, Some(vector_filter)) => {
+                    SearchMode::diverse_graph_filtered(
+                        move |vid: &u32| vector_filter.contains(vid),
+                        *params,
+                    )
+                }
```
suri-kumkaran: Not introduced here, `main` has these arms verbatim. Flagging it because this PR is reshaping exactly this code, so it's the cheapest moment to fix it.
```rust
(Some(TopkPostProcessor::DeterminantDiversity(params)), _, None) => ...
(Some(TopkPostProcessor::DeterminantDiversity(params)), _, Some(vector_filter)) => ...
```
The `_` swallows a user-specified `adaptive_l`. So `{"mode":"graph", "adaptive_l": {...}, "post_processor": {...}}` parses, passes `validate`, and runs with adaptive L off, while `Display` at `disk.rs:154` keys off `adaptive_l.is_some()` and prints `"graph + adaptive-l"`. Results labelled as adaptive L that aren't.

The root cause is that `SearchMode::Graph` has no `adaptive_l` field, so the value has nowhere to go. The config is offering a combination the backend can't express.

Best fix is my main comment, which makes it unrepresentable.

diskann-benchmark/src/main.rs
```diff
@@ -709,6 +709,18 @@ mod tests {
...
+        // Redirect each build job's `save_path` into the tempdir so the disk index
+        // artifacts are not written relative to the process cwd (the repo tree).
+        let jobs = raw["jobs"]
+            .as_array_mut()
+            .expect("\"jobs\" should be an array");
+        for (i, job) in jobs.iter_mut().enumerate() {
+            let save_path = tempdir.path().join(format!("disk_index_filter_job_{i}"));
+            job["content"]["source"]["save_path"] =
+                serde_json::Value::String(save_path.to_str().unwrap().to_string());
+        }
```
suri-kumkaran, line 722: Bug: the loop is unconditional, and `serde_json`'s `IndexMut` creates missing keys, so a `Load` job gets a `save_path` invented for it. Not hypothetical: `disk-index.json` job 1 is a `Load`. It only works here because this fixture happens to be two `Build` jobs and `DiskIndexLoad` ignores unknown fields.

`run_bftree_save_roundtrip` (L405-448) already sets the precedent for extracting this. I'd go further and call it from `run_integration_test` (L267-302), so every example gets it:

```diff
 let tempdir = tempfile::tempdir().unwrap();
+redirect_save_paths(&mut raw, tempdir.path());

 let input_path = tempdir.path().join("input.json");
```

```rust
/// Redirect each build job's `save_path` into `dir` so artifacts don't land in the repo.
/// Only rewrites keys that already exist: `Load` sources have no `save_path`.
fn redirect_save_paths(raw: &mut serde_json::Value, dir: &Path) {
    let Some(jobs) = raw.get_mut("jobs").and_then(Value::as_array_mut) else { return };
    for (i, job) in jobs.iter_mut().enumerate() {
        let Some(source) = job.get_mut("content")
            .and_then(|c| c.get_mut("source"))
            .and_then(Value::as_object_mut) else { continue };
        let Some(slot) = source.get_mut("save_path") else { continue };
        *slot = Value::String(dir.join(format!("index_job_{i}")).to_string_lossy().into_owned());
    }
}
```

Checking for the key rather than a `"Build"` tag keeps it benchmark-agnostic and means it can't invent anything. It's also panic-free, unlike the current `raw["jobs"]` indexing.

The test then collapses to the same shape as the others, since `run_integration_test` already does `prefix_search_directories`:

```rust
let raw = value_from_file(&example_directory().join("disk-index-filter.json"));
run_integration_test(raw);
```

Leave `run_bftree_save_roundtrip` as-is; it leaks its tempdir on purpose for the load-back step.

diskann-benchmark/src/disk_index/search.rs
```diff
@@ -158,6 +160,58 @@ impl DiskSearchResult {
...
+                (None, Some(adaptive_l), None) => {
+                    SearchMode::inline_filter(|_| true, Some(adaptive_l))
+                }
+                (None, Some(adaptive_l), Some(vector_filter)) => SearchMode::inline_filter(
+                    move |vid: &u32| vector_filter.contains(vid),
+                    Some(adaptive_l),
+                ),
```
suri-kumkaran, line 205: `adaptive_l` is mishandled on both paths

Neither is a regression; both cases already exist on `main`. I'm calling them out because this PR is already reshaping this code, so it's a good opportunity to fix them.

**With a post-processor, `adaptive_l` gets dropped.** The `_` in L190 and L193 ignores it, so a config like `{"mode":"graph", "adaptive_l": {...}, "post_processor": {...}}` runs with adaptive L disabled. But `Display` still sees `adaptive_l.is_some()` and prints `"graph + adaptive-l"`, so the output is labeled as adaptive L when it isn't.

**Without a filter, `adaptive_l` does nothing.** L200 uses `inline_filter(|_| true, Some(adaptive_l))`. Since the predicate is always true, the filter is effectively 100% specific, so the adaptive-L scaling never kicks in.

So today, of the four possible `adaptive_l` paths, only one actually has an effect: graph search with a real filter. We also don't have any disk fixture covering `adaptive_l`.

I think both issues come from the same mismatch: the config allows combinations that `SearchMode` can't actually represent. The main comment fixes this cleanly by making `adaptive_l` and `post_processor` mutually exclusive variants and requiring a filter for inline filtering.

diskann-benchmark/src/inputs/disk.rs
```diff
@@ -72,104 +63,168 @@ pub(crate) struct DiskIndexBuild {
...
-#[cfg(feature = "disk-index")]
-#[derive(Debug, Serialize, Deserialize, Default)]
-pub(crate) struct DiskSearchMode {
-    pub(crate) is_flat_search: bool,
-    #[serde(default)]
-    pub(crate) adaptive_l: Option<AdaptiveL>,
+/// Disk search mode. The `flat` / `graph` split is encoded as an enum so that
+/// combinations invalid for flat scan (adaptive L, post-processor) are
+/// unrepresentable rather than rejected at validation time. Mirrors the spirit
+/// of `diskann_disk::search::search_mode::SearchMode` at the config level.
```
suri-kumkaran, line 69: Both comments say combinations invalid for flat scan are "unrepresentable by construction" and that "there is no runtime compatibility check left to test."

They're representable. `{"mode":"flat", "post_processor": {...}}` and `{"mode":"flat", "adaptive_l": {...}}` both deserialize fine and the extra fields are dropped. The enum removed the check, not the state. 

The main comment would make the claim true.

diskann-benchmark/src/disk_index/search.rs
```diff
@@ -158,6 +160,58 @@ impl DiskSearchResult {
...
+            post_processor,
+            ..
+        } => {
+            let adaptive_l = adaptive_l.as_ref().map(|adaptive_l| {
+                graph::search::AdaptiveL::new(
+                    adaptive_l.sample_count.into(),
+                    adaptive_l.scale_factor,
+                )
+                .expect("validated adaptive L must construct")
+            });
```
suri-kumkaran, line 187: nit: `.expect("validated adaptive L must construct")`

`graph_index.rs:291-298` handles the same conversion with `?`. Here, `build_search_mode` returns a bare `SearchMode`, so there’s no way to propagate the error and we end up panicking instead.

The bigger issue is that this happens inside `for_each_in_pool`, so a failure would panic from the per-query path, and the safety of the `expect` depends on `validate` having run earlier - which isn't obvious from the function signature.

Since `adaptive_l` is fixed config, we can build it once before the `for &l in search_list` loop and use `?` there. This also avoids doing the same conversion for every query. That's what the approach the main comment assumes.

diskann-benchmark/src/disk_index/search.rs
suri-kumkaran, line 56:
```rust
pub(crate) is_flat_search: bool,
pub(crate) uses_vector_filters: bool,
```
Two bools encode four states, and neither records whether adaptive L or a post-processor was active. Two runs with materially different search behaviour now serialize to identical stats, which makes the JSON ambiguous for regression and comparison tooling.

`DiskSearchMode` already implements `Display` and `Serialize`, so emitting the mode itself would be strictly more useful.

suri-kumkaran (changes requested):
I like the overall direction here. Decoupling the input schema from `diskann-disk`, keeping backend construction in `build_search_mode`, and moving `vector_filters_file` / `post_processor` out of `DiskSearchPhase` all make the code cleaner. The `Option<&HashSet>` runtime filter change is a nice cleanup too.

My main concern is that the enum refactor stops a bit short. The backend already models the four actual search strategies (`FlatScan`, `Graph`, `InlineFilter`, `DiverseGraph`), but the new config still models traversal separately and keeps strategy in a few optional fields. That makes `build_search_mode` more complicated than it needs to be and leaves invalid combinations representable.

I also called out two pre-existing `adaptive_l` issues in the inline comments. They're not regressions, but this PR is already touching the relevant code, and the enum shape gives us a clean way to make both cases impossible.

The one thing I'd want settled before merge is the `is_flat_search` migration. This PR removes the old back-compat field, and `deny_unknown_fields` is currently what prevents an old config from silently switching from flat to graph. I suggest temporarily keeping it locally to verify the migration, then removing it once we've confirmed all configs are clean. See the inline comment for the details.

Nothing else blocking. Happy to talk through the enum shape live if that's easier.

PR #1238, opened 2026-07-07
Fix recall computation for beta-filtered search in benchmark

magdalendobson:
It turns out that when we implemented the fix to recall computation for filtered search in #1069, the fix did not propagate to beta search. This happened because beta search used the same runner as regular knn search in `search/knn.rs`, so the wrong `GroundTruthMode` was passed in. This PR fixes that problem by including `GroundTruthMode` in `SearchSteps` so it can be passed in earlier in the stack.


hildebrandmw (approved):

harsha-simhadri:
Does the change the results reported in the wiki comparing different filtered search algos?

magdalendobson:
> Does the change the results reported in the wiki comparing different filtered search algos?

No, this error doesn't lead to incorrect recall, but rather to throwing an error if any query point has fewer than `k` groundtruth results. This wasn't the case for the datasets I tested in the Wiki post. I think I found the error originally when working with smaller datasets where small selectivity led to fewer than `k` valid results for some points.

harsha-simhadri (approved):

metajack (approved):

Issue #150, opened 2022-11-15
Warn of failures in aligned_alloc 


harsha-simhadri:
https://github.com/microsoft/DiskANN/blob/8635dadb63391c09da0e450278136de5d88ad7cc/include/utils.h#L177

Replace assert with if and a meaningful message.

PR #1250, opened 2026-07-10
(diversity) attribute diversity implementation with post process and adaptive L

narendatha:
### NOTE: THIS WORK IS BLOCKED
 The block is: how to get attribute information for vectors during operations in diversity queue? The current code has experimental implementation that reads attribute information from a file and gives it in a side channel. that will not work in production because 1) attribute information and vector information are retrieved in different APIs causing performance issue of reading twice or from different places, 2) there is consistency issue if vector gets deleted between vector read and attribute read.

The right solution is to read both vector and attribute together. The current discussed design is to keep both vector and attribute info in InternalId and retrieve it together. This needs design and implementation change in core library. This PR cannot be merged until that work is done.

### Summary

Adds two independent, opt-in strategies for **attribute-bucket diverse** nearest-neighbor search on the on-disk index, so that a top-`k` result set contains at most `diverse_results_k` vectors sharing the same attribute value (e.g. author, `doc_id`, camera model). Neither strategy modifies the shared neighbor queue or the default search path — both are selected only when a diversity attribute is configured.

This also removes the old `experimental_diversity_search` feature gate (the queue-mutating prototype) and replaces it with these two cleaner designs.

### Motivation

A plain KNN search over a corpus where many vectors share an attribute value returns highly redundant results (e.g. 30%+ of neighbors from a single bucket). Enforcing "at most `k` per bucket" against a **diverse ground truth** requires either filtering after search or fetching a larger candidate pool. This PR implements and benchmarks both approaches.

### Algorithms

**Design A — post-processing over a fixed pool** (`AttributeBucketDiversity`, `diskann-disk/src/search/provider/disk_provider.rs`)

Runs an ordinary greedy beam search with the caller's `L`, then applies a post-processor over the top-`L` pool:

1. **Rerank with exact distances.** Full-precision distances computed during traversal are reused from the accessor's `distance_cache`; only cache-missed candidates trigger a vertex load + `evaluate_similarity` recompute.
2. **Sort** candidates ascending by exact distance (true nearest-first order).
3. **Greedy bucket selection.** Walk the sorted candidates keeping a `HashMap<bucket, count>`; emit a candidate only while its bucket count is `< diverse_results_k`. The output buffer truncates the stream to `k`.

Cost is essentially free over a normal search (no extra IO), but recall depends on the caller choosing an `L` large enough that the top-`L` pool holds enough distinct buckets.

**Design B — adaptive-L over-fetch** (`DiverseAdaptiveSearch`, `diskann/src/graph/search/diverse_adaptive_search.rs`)

Layers a traversal-time sampler on top of Design A to size `L` per query:

1. During the greedy walk, sample the attribute bucket of each newly visited node into a `HashMap<bucket, count>` until `sample_count` nodes are seen.
2. Compute the diversity **yield** as `(Σ_b min(count_b, k)) / visited` — i.e. the fraction of visited nodes that would survive bucket selection. For `k = 1` this is `distinct_buckets / visited`.
3. Feed that yield into the existing `compute_adaptive_l` piecewise scaler (shared with inline-filter search):
   - yield ≥ 50% → 1× (no growth)
   - 10–50% → 2×
   - < 10% → log-scale, `2^(-log10(yield))` (1% → 4×, 0.1% → 8×), clamped to `scale_factor`
4. Resize the search-list scratch **once** to the new `L`, continue the walk, then hand the enlarged pool to the same Design A post-processor.

When buckets are diffuse the yield is high and `L` is left unchanged, so Design B degrades exactly to Design A with no over-fetch. When buckets are concentrated it over-fetches proportionally to the observed skew — no per-dataset `L` tuning required.

### Results

Benchmarked against diverse ground truth (k=10, `diverse_results_k=1`) on three datasets spanning both regimes; full tables and charts in `diverse-search-results.md`.

- **Concentrated attributes (Enron, YFCC `camera`):** Design B wins recall at every `L` (+13 over the queue prototype on Enron @ L=20; on YFCC it climbs to 96% while the queue plateaus near 81%), at a throughput cost from the extra IO.
- **Diffuse attribute (Caselaw `doc_id`):** Design B performs no over-fetch and matches Design A on recall and IO exactly; both slightly beat the queue prototype via exact-distance reranking.

**Recommendation: Design B** — matches Design A when diversity is free and outperforms both Design A and the queue prototype when the attribute is concentrated, without per-dataset tuning.

### Changes

- **Core:** new `DiverseAdaptiveSearch` strategy; `compute_adaptive_l` / `AdaptiveL` accessors made `pub(crate)`; removed `experimental_diversity_search` feature gate across `diskann`, `diskann-disk`, `diskann-providers`, `diskann-bftree`.
- **Disk:** `AttributeBucketDiversity` post-processor; `SearchMode::DiverseAttribute { provider, diverse_attribute_id, diverse_results_k, adaptive_l }` dispatching to plain KNN (A) or adaptive search (B).
- **Benchmark:** `FileAttributeProvider`, disk-search config fields for attributes / diversity / adaptive-L.
- **Docs:** benchmark results write-up with recall/QPS tables and Mermaid charts.

### Testing

- Unit tests for `diverse_matched_slots` and adaptive-L growth behavior.
- Existing diversity search + queue retain/truncate tests migrated off the removed feature gate.

> Note: QPS figures in the doc are best-of-N single-threaded on a contended dev machine — treat them as indicative; recall and IO counts are the authoritative comparison.


diverse-search-results.md
```diff
@@ -0,0 +1,225 @@
+# Attribute-Diversity Search — Benchmark Results
```
magdalendobson, line 1: This doc and the attached plots should be moved to the Wiki.

diskann/src/graph/search/diverse_adaptive_search.rs
```diff
@@ -0,0 +1,318 @@
...
+
+//! Attribute-diversity search with adaptive-L over-fetch (Design B).
+//!
+//! This is a plain greedy graph search whose only addition over [`Knn`] is a
+//! traversal-time sampler that estimates *bucket concentration* — how many
+//! distinct attribute buckets appear among the nodes visited so far — and grows
+//! the search list `L` accordingly. The actual bucket-diverse top-`k` selection
+//! is still performed by a downstream post-processor over the enlarged pool.
+//!
+//! # Motivation
```
magdalendobson, line 14: Maybe I'm misunderstanding something about this calculation, but it seems like the space of possible values might be constrained in a bad way when the number of attributes is small. If `diverse_results_k`=1, and the number of buckets is low (say, 60-70 like in some of your experiments), it seems like if the number of points visited is set to the default 1000, the yield could only ever be as high as .06 even with a maximally diverse set, and L might be increased even if the yield is already perfect. It seems like there needs to be some coupling of `diverse_results_k` with `num_buckets` and the number of samples in `AdaptiveL` to make sure that yield is actually allowed to be anywhere from 0 to 1.0.

magdalendobson: Thinking about this more, maybe what we should be doing is waiting for a certain number of samples to accumulate, and then computing `yield = (Σ_b min(count_b, k)) / num_buckets * k`. Then, we would probably want to scale the adaptive L a bit differently from how filtered search does it. For filtered search, a specificity of .5 over 1000 samples is really good and an indication that we probably don't need to search more. On the other hand, with my proposed definition, a yield of .5 would be pretty bad and would indicate that we have only reached <= 50% recall so far and probably need to search quite a lot more.

narendatha: Thanks — both points were spot on. Fixed in the latest push:

**1. Yield is now coupled to `num_buckets`, `k`, and the sample size.** The metric is:

`yield = (Σ_b min(count_b, k)) / min(sample_visited, num_buckets·k)`

The denominator is the tight upper bound on the numerator, so yield always spans the full `[0, 1]`:
- **few buckets / large sample** → denominator is `num_buckets·k`, so covering every bucket reaches 1.0 (your 60–70 bucket case);
- **many buckets / small sample** → denominator is `sample_visited`, so a fully distinct sample reaches 1.0 and diffuse data leaves `L` unchanged.

`num_buckets` comes from a new `AttributeValueProvider::num_buckets()` (defaults to `None`, which disables growth); the file-backed provider reports its distinct-value count.

**2. Diversity yield is scaled differently from filter specificity.** Design B no longer reuses the inline-filter piecewise curve. `L` now scales linearly from the *deficit* `1 − yield`, so a yield of 0.5 grows `L` substantially (≈ halfway to the max multiplier) rather than treating 0.5 as "good enough" — matching your point that 0.5 yield ≈ ≤50% recall and needs much more search.

narendatha: Followed up by rerunning **Design B across all four datasets** after the yield-normalization change, and regenerated the charts. Recall is deterministic (authoritative); QPS is kept as best-of-several single-threaded runs since it varies 10–30× from disk contention.

**Design B recall @ L = 20 / 40 / 80 / 100:**

| Dataset (attribute) | Regime | Before | After |
|---|---|---|---|
| Enron | concentrated, uncorrelated | 73.4 / 83.6 / 89.8 / 91.0 | **79.9 / 87.4 / 92.1 / 93.0** |
| Caselaw `doc_id` | diffuse | 92.0 / 97.5 / 99.1 / 99.3 | 92.0 / 97.5 / 99.1 / 99.3 (unchanged) |
| Caselaw `court_jurisdiction` | concentrated + distance-correlated | 59.2 / 69.6 / 77.6 / 79.9 | 57.3 / 68.2 / 76.6 / 79.0 |
| YFCC `camera` | concentrated, ~uncorrelated | 83.6 / 92.5 / 95.4 / 96.3 | **88.4 / 94.2 / 96.5 / 96.9** |

Takeaways matching the intent of the review:

- **Concentrated + uncorrelated (Enron, YFCC):** clear recall gains — normalizing the yield by *achievable* diversity drives more appropriate over-fetch. +6.5 pts on Enron and +4.9 pts on YFCC at L=20.
- **Diffuse (Caselaw `doc_id`):** unchanged — yield ≈ 1.0, so no over-fetch; still matches Design A on recall and IO exactly. Correct behavior preserved.
- **Concentrated + distance-correlated (Caselaw `court_jurisdiction`):** a small, honest regression (~1–2 pts). With only 60 buckets the denominator `num_buckets·k` caps at 60, so seeing many courts in the sample reads as good coverage and over-fetches less (IO 193 vs old 210). Design B now sits **just below** the queue at every L (79.0 vs 79.1 at L=100) rather than edging past it — which actually reinforces the doc's conclusion that **in-traversal diversity (the queue) is the right tool for distance-correlated attributes**, and post-processing shouldn't be expected to win there.

I also checked in the chart tooling so this is reproducible going forward: `assets/diverse-search/chart_data.json` (the plotted recall/QPS values) + `assets/diverse-search/plot_qps_vs_recall.py`. All four PNGs and the results tables/prose are updated. 

I will remove these chart tooling scripts once we are close to shipping level with right way to get attributes becomes available. We will also need to discuss how to get the num_buckets attribute information.

narendatha: **Follow-up: `num_buckets`-unavailable fallback**

Rather than disabling adaptive growth when a provider can't report `num_buckets`, `diverse_yield` now falls back to normalizing by `sample_visited` alone — the original pre-normalization heuristic. So on that path `L` still grows on concentrated samples (never worse than fixed-`L` Design A) instead of collapsing to Design A.

Verified by temporarily forcing `num_buckets() -> None` and re-running Design B:

| Dataset (buckets) | With `num_buckets` (L=20/40/80) | `None` fallback (L=20/40/80) |
|---|---|---|
| Enron (~12,849) | 79.9 / 87.4 / 92.1 | 79.9 / 87.4 / 92.1 (identical) |
| YFCC (67) | 88.4 / 94.2 / 96.5 | 91.5 / 95.4 / 96.9 (over-fetches more) |

- **Many buckets (Enron):** fallback is a no-op — `num_buckets·k` already exceeds the sample, so both paths cap the denominator at `sample_visited`.
- **Few buckets (YFCC):** fallback grows `L` more than necessary (the sample looks concentrated without the bucket total), yielding higher recall at higher IO. Still bounded below by Design A.

Recall stays deterministic; QPS omitted as usual due to run-to-run disk-contention noise. Added a `yield_unknown_buckets_uses_sample_size` unit test covering the `None` path.

magdalendobson: Thanks, it's good to see that recall mostly improved after better normalization. I was wondering what `sample_visited` is set to in the experiments? In particular with `L=20`, if it's set to the default 1000, are we always seeing 1000 samples and giving the algorithm a chance to adapt?

It's interesting that normalizing didn't help for Caselaw with the `court_jurisdiction` attribute. I guess one interpretation of what happened there is that since Caselaw has much more correlation between distance and attribute, the buckets were saturated earlier on in the search by points that were far away from the query, but all of the points that are close to the query were from a smaller set of attributes. Then it would look to the Adaptive L calculation as if diversity had been achieved, but actually the attribute-satisfying points closest to the query hadn't been found. Previously with `num_samples` normalization, `L` would have just been coincidentally increased because the cap on the ratio was artificially small. Normalizing correctly is still the right thing to do, but we could consider playing with increasing `k` a bit for a small number of buckets?

diskann/src/graph/search/diverse_adaptive_search.rs
```diff
@@ -0,0 +1,481 @@
...
+        accessor
+            .expand_beam(
+                scratch.beam_nodes.iter().copied(),
+                glue::NotInMut::new(&mut scratch.visited),
+                |id, distance| neighbors.push(Neighbor::new(id, distance)),
+            )
+            .await?;
+
+        for neighbor in neighbors.iter() {
+            scratch.best.insert(*neighbor);
```
magdalendobson, line 241: Is `scratch.best` a `DiversePriorityQueue` here?

diskann-benchmark/src/disk_index/search.rs
```diff
@@ -37,7 +37,7 @@ use serde::{Deserialize, Serialize};
 use crate::{
```
magdalendobson, line 37: When this goes into `benchmark`, can you add generated diverse groundtruth for the two filter datasets in `test_data` and enable an example benchmark json with them?

diverse-search-results.md
narendatha: Sure. will do this once we integrate attribute getting code done.

diskann/src/graph/search/diverse_adaptive_search.rs
narendatha: Great questions.

**On `sample_count`:** it's a required config field (no default), and in every experiment here it's set to **100**, not 1000 — so the sampler estimates yield after the first 100 visited nodes, then resizes `L` and continues. Even at `L=20` the search visits far more than 100 nodes before it would naturally converge (fixed-`L`=20 Design A already does ~1,300 comparisons / ~40 IOs per query on Enron), so the 100-sample threshold is always reached and adaptation always fires. You can see it in the L=20 Enron numbers: Design B does ~1,880 comps / ~81 IOs vs Design A's ~1,300 / ~40 — i.e. `L` really was grown mid-search, and recall went 60.0 → 79.9 as a result.

**On the `court_jurisdiction` case:** your interpretation matches the data exactly. With 60 buckets and `sample_count=100`, the correct denominator is `min(100, 60·k) = 60`, so once the sample has touched all 60 courts (which happens early, dominated by far-away points) the yield saturates at ~1.0 and `L` stops growing. The old `sample_visited` normalization divided by 100 while `matched` could never exceed 60 — so the ratio was **capped at 0.6**, guaranteeing a ≥40% "deficit" and thus coincidental over-fetch every time. That's precisely why the old numbers were a bit higher here (57.3 vs 59.2 at L=20, with IO 193 vs 210): the previous version was over-fetching for the wrong reason. Correct normalization is still the right call.

**On bumping `k` for few-bucket datasets:** I think this is a promising direction. The clean version is to **decouple the yield-`k` from the output `diverse_results_k`** — i.e. require the sample to see, say, 3–4 points *per bucket* before the yield saturates, even though the final result still caps at 1 per bucket. On a correlated attribute that would keep `L` growing past the point where the far-away points have merely "covered" the buckets, giving the search a chance to reach the closer attribute-satisfying points. It's a small change (an extra `yield_k` knob feeding `diverse_matched_slots`/`diverse_yield`), and it degrades gracefully on diffuse data where each bucket only has one point anyway. Does it match with your thinking? Will prototype this now.

diskann/src/graph/search/diverse_adaptive_search.rs
narendatha: No. `DiversePriorityQueue` comes in for Queue based diversity search. This diverse adaptive search does not have dependency or connection to queue based approach. If we settle on using adaptive search only and no customer scenario needs diverse priority queue, we can get rid of queue version safely.

diskann-benchmark/src/disk_index/search.rs
narendatha: Yes, of course, that is the plan. Will keep this comment open until that happens.

diskann/src/graph/search/diverse_adaptive_search.rs
narendatha: ### Follow-up on the `yield_k` idea → moved to a separate experiment draft PR https://github.com/microsoft/DiskANN/pull/1262

Following the discussion about the jurisdiction regression on distance-correlated attributes, I prototyped the decoupled `yield_k` knob (sampler per-bucket cap independent of the output `diverse_results_k`). It works — on jurisdiction, `yield_k ≥ 2` recovers the regression and beats the plain-queue baseline, and `yield_k = 4` reaches 81.55 recall@10 @ L=100 (vs. 78.97 today) for ~28% more IOs.

However, it adds a knob whose optimal value depends on the (unobservable) attribute/distance correlation, and it interacts with the existing `scale_factor` / `sample_count` over-fetch controls. Rather than expand this PR's surface area with a parameter that's hard to specify a default for, I've split it onto its own branch/draft PR so we can:

- sweep `yield_k` across more datasets (Enron, YFCC, Caselaw doc_id) to map where  it helps vs. is a no-op, and
- explore auto-deriving `yield_k` from observed bucket saturation so it needn't be  a knob at all.

**This PR is unaffected** — it stays as the clean Design B (`yield_k = 1` behavior, no new config surface). Experiment branch: `u/narendatha/diverse-yield-k-experiment` (draft PR here https://github.com/microsoft/DiskANN/pull/1262)

What do you feel about adding another knob that may be hard to tune? Perhaps it depends on scenarios. Let me continue experimenting in that other branch with other datasets meanwhile.

diskann/src/graph/search/diverse_adaptive_search.rs
magdalendobson: Oh, I see. Then is there a possibility that we are throwing away diverse results that could be valid final answers by just using `scratch.best` instead of the entire set of visited neighbors?

narendatha: Good thought. Yes — that's a real and intended tradeoff, and it's the crux of the design.

scratch.best is a NeighborPriorityQueue bounded to size L and ordered purely by distance, so it only keeps the L closest nodes seen. A node that satisfies a rare attribute bucket (valuable for diversity) but ranks poorly by distance — say #150 with L=100 — gets evicted before the post-processor ever sees it. So we can indeed drop candidates that would have made valid diverse final answers.

This is exactly the failure mode on distance-correlated attributes (e.g. jurisdiction): the diverse candidates live far out in distance rank, right where the fixed-L queue truncates. Design B's answer is adaptive-L — detect bucket concentration during traversal and grow L so more of those far candidates survive in scratch.best. It's a heuristic that widens the window in proportion to the concentration signal, not a guarantee that nothing is lost.

I'd argue against the "post-process over all visited neighbors" alternative, for two reasons:

1. scratch.visited is a HashSet IDs only, no distances or buckets. It exists purely for dedup. Using it as a candidate pool would require accumulating a separate, much larger distance-bearing buffer across the whole traversal (visited » L), with the corresponding memory/post-processing cost.

2. "All visited" isn't actually "all valid diverse answers." The greedy traversal already declines to expand beyond what the frontier deems promising, so the visited set is itself an arbitrary, traversal-order-dependent subset. Post-processing over it would be more expensive and still incomplete.

So the principled question is "how wide should the candidate window be?" rather than "bounded queue vs. unbounded visited set." I've been quantifying exactly this on the sibling experiment branch ( [PR](https://github.com/microsoft/DiskANN/pull/1262)): deferring when the yield estimate saturates keeps L growing longer, and jurisdiction recall@10 climbs from 78.97 ? 81.55 @ L=100 as a result. That's the direct measurement of the candidates we're currently dropping and how many we can recover by widening the window — which feels like the right place to iterate, rather than switching to an unbounded visited pool.

magdalendobson: You don't have to use `scratch.visited`, you could instead keep the distances around in a separate struct to hand to post-process in the same way that [inline filtered search](https://github.com/microsoft/DiskANN/blob/1552e51c6eb072fa905fb8f04eccabb721c75de5/diskann/src/graph/search/inline_filter_search.rs#L279) does. This made a HUGE difference in the performance of filtered search, which is why I am a bit concerned with potentially leaving it on the table here.

narendatha: Great call — you were right, and this was a big win left on the table.

I've decoupled the diversity post-processor's candidate pool from the `L`-bounded `scratch.best` frontier, exactly mirroring `inline_filter_search`: the adaptive-L traversal now accumulates every visited node **with its distance** into a side `Vec<Neighbor<_>>`, and `search` hands the full distance-sorted set to `post_process` instead of `scratch.best.iter()`. (I don't `.take(L)` the way filtered search does, since for diversity the whole point is to let a diverse-but-distant node that the frontier evicted survive as a final answer.)

It improves recall@10 at every `L` with **identical IO** across all four regimes:

| Dataset | L=100 before → after |
|---|---|
| Caselaw jurisdiction (dist-correlated) | 78.97 → **90.19** |
| YFCC camera | 96.92 → **97.49** |
| Enron | 93.01 → **93.74** |
| Caselaw doc_id (diffuse) | 99.25 → **99.26** |

The jurisdiction case is the striking one: Design B previously sat *just below* the in-traversal queue (79.0 vs 79.1) and now beats it decisively (90.2 vs 79.1), which inverts the conclusion in the results write-up. It also subsumes most of what the experimental `yield_k` knob was doing — and helps Enron, where `yield_k` actually hurt — so I'm inclined to keep `yield_k` on its separate experiment branch rather than in this PR. I've updated `diverse-search-results.md` and the charts accordingly.

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -300,7 +312,12 @@ where
 ///   0 matches in sample → `max_multiplier`× L (maximum expansion)
 ///
 /// Clamped to [1×, max_multiplier] range.
-fn compute_adaptive_l(base_l: usize, visited: usize, matched: usize, max_multiplier: f64) -> usize {
+pub(crate) fn compute_adaptive_l(
```
magdalendobson, line 315: I think this doesn't need to be pub(crate) anymore

diskann/src/graph/search/diverse_adaptive_search.rs
```diff
@@ -0,0 +1,502 @@
...
+/// to grow `L` when bucket concentration is low. The final bucket-diverse
+/// selection is delegated to the post-processor supplied at search time.
+pub struct DiverseAdaptiveSearch<P>
+where
+    P: AttributeValueProvider + ?Sized,
+{
+    inner: Knn,
+    provider: Arc<P>,
+    diverse_results_k: usize,
+    adaptive_l: AdaptiveL,
```
magdalendobson, line 87: It might be worth making AdaptiveL and Option the way it is in inline filtered search. This can be helpful for benchmarking how well regular search can perform without adaptations, scenarios where the user really wants to be able to control how much compute is being used, and perhaps also might be the right thing to do when `num_buckets` isn't known.

magdalendobson:
Thanks Naren! It was good to see that normalization and using the entire visited set made a big impact. I don't have any more concrete thoughts at the moment on how the algorithm could be improved, although I think it would be interesting to look at estimating distance<->attribute correlation at some point. I can return for another pass once we're ready to move this out of draft.

PR #1251, opened 2026-07-10
[diskann-garnet] Add persistence for quantized indexes

### LINKED ISSUE #1108: diskann-garnet should recover from failed quantization backfill
Currently we make a best effort and operate in full precision mode indefinitely if it doesn't succeed. There should be some way to signal failure or report errors.

metajack:
Quantizer state was not persisted to the store so the index would operate in full precision mode after a restart. This PR adds support for reloading and resuming quantized indexes.

For Q8 quantizers, the quantized start point cache needs to be filled on restart. For BIN quantizers, the quantizer code book must be saved and loaded as well as a flag indicating whether backfill is complete or not.

Note, because we can't signal errors during backfill yet, we cannot fully recover if the index is restarted during backfill. This will be addressed in the future as it will require some FFI changes.

### DIFF
--- a/diskann-garnet/src/provider.rs
+++ b/diskann-garnet/src/provider.rs
@@ -208,7 +208,15 @@ impl<T: VectorRepr> GarnetProvider<T> {
                 let quantizer = Box::new(quantization::MinMax8Bit::new(dim, metric_type)?)
                     as Box<dyn GarnetQuantizer>;
                 let canonical_bytes = quantizer.bytes();
-                // NOTE: Q8 needs no training, so it always starts with backfill complete.
+                // NOTE: Q8 needs no training, so it always starts with backfill
+                // complete. However, we still need to load the start point if
+                // it exists.
+
+                let mut qsv = Poly::broadcast(0u8, canonical_bytes, AlignToEight)?;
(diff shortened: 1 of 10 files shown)


diskann-garnet/src/provider.rs
```diff
@@ -467,6 +505,23 @@ impl<T: VectorRepr> GarnetProvider<T> {
...
         match quantizer.train(self.metric_type, view) {
             Ok(()) => {
+                let quant_state = if let Ok(s) = quantizer.serialize() {
+                    s
+                } else {
+                    return false;
+                };
+
+                let mut total_quant_state = vec![0u8; quant_state.len() + 1];
+                total_quant_state[1..].copy_from_slice(&quant_state);
```
hildebrandmw, line 515: It's a bit of a bummer that we have to do this - but I don't have a good alternative :(

hildebrandmw (approved):

diskann-garnet/src/provider.rs
metajack: It's probably possible to use a custom allocator so I get the space, but since this happens basically never it didn't seem worth worrying about much.

harsha-simhadri (approved):

harsha-simhadri (approved):

PR #1252, opened 2026-07-13
perf: ANN optimizations - SwissHashSet, AVX2 distance, insert_fast

cave-g-f:
Port hot-path optimizations from parent repo PR 6610806 (commit 844664001):

- swiss_hash_set.h: New SSE2/SIMD-optimized Swiss table hash set for uint32_t keys with O(occupied_slots) clear and ~87.5% load factor. Replaces tsl::robin_set for inserted_into_pool_rs.
- neighbor.h: Add NeighborMinHeap::insert_fast() with AVX2-accelerated linear scan over a contiguous float distance array (_dist_arr) for faster insertion-point finding.
- index.cpp: Refactor iterate_to_fixed_point() to interleave prefetch, distance computation, and immediate insert; remove separate dist_scratch allocation; mark nodes visited and process neighbors under the same shared lock (eliminates redundant lock/unlock).
- pq_l2_distance.cpp: AVX2 inner loop for chunk-wise PQ centroid distance computation (8-wide _mm256_fmadd_ps).
- scratch.h/.cpp: Switch _inserted_into_pool_rs type to SwissHashSet and skip clear() on empty containers.

### DIFF
--- a/include/neighbor.h
+++ b/include/neighbor.h
@@ -7,6 +7,7 @@
 #include <mutex>
 #include <vector>
 #include <assert.h>
+#include <immintrin.h>
 #include <tsl/robin_map.h>
 //#include "utils.h"
 #include "color_info.h"
@@ -90,6 +91,7 @@ public:
     NeighborVector(size_t capacity)
         : _capacity(capacity)
         , _data(capacity + 1, Neighbor(std::numeric_limits<uint32_t>::max(), std::numeric_limits<float>::max()))
+        , _dist_arr(capacity + 1, std::numeric_limits<float>::max())
     {
     }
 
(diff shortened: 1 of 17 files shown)


include/neighbor.h
```diff
@@ -108,12 +110,46 @@ class NeighborVector : public NeighborVectorBase
...
         std::memmove(&_data[lo + 1], &_data[lo], (size - lo) * sizeof(Neighbor));
+        std::memmove(&_dist_arr[lo + 1], &_dist_arr[lo], (size - lo) * sizeof(float));
         _data[lo] = { nbr.id, nbr.distance };
+        _dist_arr[lo] = nbr.distance;
+    }
+
+    // SIMD lower-bound insertion (Rust-style): finds first position where
+    // distance >= target, inserts there. Caller's visited set prevents duplicates.
+    // Returns insertion index, or capacity+1 if rejected.
+    size_t insert_fast(const Neighbor& nbr, size_t size, size_t capacity)
```
Sanhaoji2, line 121: add UT for this function

src/pq_l2_distance.cpp
```diff
@@ -260,18 +261,20 @@ template <typename data_t>
 void PQL2Distance<data_t>::prepopulate_chunkwise_distances(const float *query_vec, float *dist_vec)
```
Sanhaoji2, line 260: this function isn't called in current scenarios, the change should not take effect

include/neighbor.h
cave-g-f: done

PR #1254, opened 2026-07-14
Remove checkpoint manager from disk index builds

### LINKED ISSUE #1119: Deprecate checkpoint manager from disk index

wuw92:
## Why

The checkpoint path existed so a host could stop a long-running disk index build at chunk boundaries and resume it in a later invocation. That caller requirement has been removed: disk index builds are now allowed to run to completion once started. We still need bounded compression batches to control memory usage, but that batching does not require persisted build stages, continuation grants, or restart support.

## What

Remove checkpoint and continuation support from disk index construction and simplify the build around single-pass execution.

1. **Build lifecycle:** Disk index construction no longer persists or resumes partially completed stages; the build path writes only final outputs.
2. **In-memory build:** Remove checkpoint-oriented insertion waves and process vectors through long-lived workers over a shared iterator.
3. **Configuration API:** Remove `ChunkingConfig`. Its continuation provider and in-memory chunk count are removed, while the compression chunk count moves to `DiskIndexBuildParameters` with the existing 25,000-vector default and an optional override.
4. **Compression:** Retain bounded PQ compression blocks, validate empty inputs and zero-sized blocks, and flush the output after generation completes.


diskann-disk/src/storage/quant/generator.rs
```diff
@@ -209,54 +158,15 @@ where
...
 
-        let progress = process_while_resource_is_available(
-            action,
-            0..num_blocks,
-            chunking_config.continuation_checker.clone_box(),
-        )?
-        .map(|processed| processed * block_size + offset);
+        for block_index in 0..num_blocks {
+            action(block_index)?;
+        }
```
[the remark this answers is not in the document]
wuw92: Fixed in 3f77fa49. Compression blocks are written without per-block flushes, and the writer is flushed once after the loop completes.

diskann-disk/README.md
```diff
@@ -12,7 +12,7 @@ The `disk-index` crate contains all the components specifically needed for build
 src/
 ├── build/             # Disk index building pipeline
 │   ├── builder/       # Core disk index builder and logic
-│   ├── chunking/      # Checkpointing and continuation handling
+│   ├── chunking/      # Build batch size configuration
```
harsha-simhadri: how is build batch size configuration different than the configuration in folder below? Could they be consolidated into one configuration?

wuw92: Good point. I consolidated the compression chunk size into  `DiskIndexBuildParameters` and removed the separate chunking configuration module.

diskann-disk/src/storage/quant/generator.rs
```diff
@@ -113,52 +78,36 @@ where
...
+            .write(&mut compressed_data_writer)?;
 
         let compressed_size = self.quantizer.compressed_bytes();
         let max_block_size = chunking_config.data_compression_chunk_vector_count;
-        let num_remaining = num_points - offset;
-
         let block_size = std::cmp::min(num_points, max_block_size);
-        let num_blocks =
-            num_remaining / block_size + !num_remaining.is_multiple_of(block_size) as usize;
+        let num_blocks = num_points / block_size + !num_points.is_multiple_of(block_size) as usize;
```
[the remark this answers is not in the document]
wuw92: fix it.

diskann-disk/src/build/builder/build.rs
```diff
@@ -166,58 +125,30 @@ where
...
-                )
-            },
+        BuildQuantizer::train::<Data, _>(
+            build_quantization_type,
+            index_path_prefix,
+            index_configuration,
+            pq_storage,
+            storage_provider,
         )
     }
```
suri-kumkaran: After removing `checkpoint_record_manager` logic, this function does nothing other than param passing.

diskann-disk/src/build/builder/build.rs
```diff
@@ -348,159 +249,67 @@ where
...
-                let message = format!(
-                    "[Stage:{:?}] Build interrupt at progress {}",
-                    stage, processed
-                );
-                Err(ANNError::log_build_interrupted(message))
-            }
-            Progress::Completed => Ok(()),
-        }
+        .await
     }
```
suri-kumkaran: After removing `checkpoint_context` logic, this function does nothing other than param passing.

diskann-disk/src/build/builder/core.rs
```diff
@@ -508,63 +490,31 @@ impl<'a> MergedVamanaIndexWorkflow<'a> {
...
+        let num_points = builder.index_configuration.max_points;
+        let sampling_rate = MAX_PQ_TRAINING_SET_SIZE / num_points as f64;
+        let ram_budget_in_bytes = builder.disk_build_param.build_memory_limit().in_bytes() as f64;
+
+        partition_with_ram_budget::<Data::VectorDataType, _, _>(
+            &self.dataset_file,
+            builder.index_configuration.dim,
+            sampling_rate,
+            ram_budget_in_bytes,
+            2,
```
suri-kumkaran: Optional: This isn't introduced by this change, but since we're here, would it be worth extracting this into a constant or default config for better discoverability?

diskann-disk/src/build/builder/core.rs
```diff
@@ -577,47 +527,11 @@ impl<'a> MergedVamanaIndexWorkflow<'a> {
...
-        CheckpointContext::new(
-            builder.checkpoint_record_manager.as_ref(),
-            current_stage,
-            next_stage,
+        builder.merge_shards_and_cleanup(
+            &self.merged_index_prefix,
+            num_parts,
+            self.max_degree,
+            &mut self.rng,
         )
```
suri-kumkaran: Can we check whether we actually need this public `merge_and_cleanup` method, since it mostly just forwards parameters to a private method?

suri-kumkaran (commented):
Can we simplify the build flow overall? With this change making the flow simpler, I'm not sure we still need all of these helper functions for clarity and modularity.

diskann-disk/src/build/builder/core.rs
wuw92: Add  `PARTITION_ASSIGNMENTS_PER_VECTOR=2`

diskann-disk/src/build/builder/core.rs
wuw92: Remove unnecessary helper class.

suri-kumkaran (approved):

hildebrandmw (commented):
Thanks! I'm all for simplification where possible. Can you please add some info to the PR description describing why this feasible to do now when it wasn't in the past to provide some breadcrumbs in case this is ever needed again?

arkrishn94 (approved):
Thanks Wei, this will help us clean up a bunch of extra code in the disk index build path. As Mark mentioned, can you write a short explanation on what change allowed us to delete the checkpointing?

wuw92:
Added the background to the PR description. Checkpointing was required when the host needed to stop a build at chunk boundaries and resume it in a later invocation. That caller requirement has been removed, so builds can now run to completion.

PR #1259, opened 2026-07-15
Upgrade Rust toolchain to 1.95 and fix clippy lints

harsha-simhadri:
## Summary

Bumps the pinned Rust toolchain from **1.92 → 1.95** and fixes the clippy warnings that surface under 1.95. All CI clippy jobs run with `-Dwarnings`, so each warning is a hard failure.

## Changes

- **`rust-toolchain.toml`**: `channel = "1.92"` → `"1.95"`. CI reads the channel via `rustup show`, so no workflow edits are needed.
- **Clippy fixes** (all three CI configs — default, all-features, and per-crate no-default-features — pass clean):
  - `useless_conversion`: removed redundant `.into_iter()` calls in `diskann-benchmark-runner`, `diskann-benchmark`, `diskann-disk`, and `diskann-wide` test helpers (plus the follow-up `unused_parens` / rustfmt reflow they triggered).
  - `collapsible_match`: folded the nested `is_trained()` check into a match guard in `GarnetProvider::train_quantizer`. Written as `Some(q) if !q.is_trained() => q, _ => return false` to stay exhaustive (clippy's own `--fix` produced a non-exhaustive match).
  - `manual_checked_ops`: added a targeted `#[allow]` + comment on `DiskSectorGraph::node_sector_index`. The `else` branch is genuine multi-sector-per-node layout logic, not a divide-by-zero guard, so `checked_div` would be semantically wrong.
- **`agents.md`**: refreshed the toolchain reference to 1.95.

## Validation (local, under 1.95)

- `cargo clippy` with `-Dwarnings` — default features, all features, and per-crate no-default-features: **0 warnings**.
- `cargo fmt --all --check`: clean.
- `--locked` (`Cargo.lock`): unchanged.
- Tests for all changed crates pass (e.g. `diskann-garnet` 40/40).

## Notes

The SDE baseline/AVX-512, codecov, and CodeQL gates aren't reproducible on Windows locally and are left to CI; the diff is lint-only, so risk is low.

### DIFF
--- a/agents.md
+++ b/agents.md
@@ -1,6 +1,6 @@
 # DiskANN Repository - Agent Onboarding Guide
 
-**Last Updated**: 2026-02-11 (based on v0.45.0, Rust 1.92)
+**Last Updated**: 2026-07-14 (based on v0.45.0, Rust 1.95)
 
 This guide helps coding agents understand how to work efficiently with the DiskANN repository.
 
--- a/diskann-benchmark-runner/src/internal/regression.rs
+++ b/diskann-benchmark-runner/src/internal/regression.rs
@@ -229,7 +229,7 @@ impl<'a> Checks<'a> {
         // We can now package everything together!
         debug_assert_eq!(input_to_parsed.len(), inputs.jobs().len());
 
-        let checks = std::iter::zip(inputs.into_inner(), input_to_parsed.into_iter())
+        let checks = std::iter::zip(inputs.into_inner(), input_to_parsed)
             .map(|(input, index)| {
                 // This index should always be inbounds.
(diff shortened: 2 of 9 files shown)


hildebrandmw:
CI failure is because of the compile tests in diskann-quantization. I've audited the results and it's fine. Running
```
TRYBUILD=overwrite cargo test --package diskann-quantization compile_tests
```
should fix the issue.

Regarding the update - we should eventually have a policy regarding minimum Rust version. Current stable - 2 (so in this case, 1.95) is a decent policy. It allows our users some slack while still allowing us to (eventually) get shiny new features.

PR #1261, opened 2026-07-17
Fix benchmark ground truth for disk_index_siftsmall 10-query workload

### LINKED ISSUE #1172: diskann-benchmark\example\flat-index.json uses graph ground truth which is not equal
diskann-benchmark\example\flat-index.json uses graph ground truth which is not equal to knn bruteforce ground truth, hence the example shows 0.90 recall:

```
cargo run --package diskann-benchmark --release -- run --input-file ./diskann-benchmark/example/flat-index.json --output-file ./target/tmp/flat-index-output.json                           
    Finished `release` profile [optimized + debuginfo] target(s) in 8.83s
     Running `target\release\diskann-benchmark.exe run --input-file ./diskann-benchmark/example/flat-index.json --output-file ./target/tmp/flat-index-output.json`
######################
# Running Job 1 of 1 #
######################

              Data: test_data/disk_index_search\disk_index_siftsmall_learn_256pts_data.fbin
         Data Type: float32
          Distance: squared_l2
           Queries: test_data/disk_index_search\disk_index_sample_query_10pts.fbin
       Groundtruth: test_data/disk_index_search\disk_index_10pts_idx_uint32_truth_search_res.bin
                 K: 10
           Threads: 1
              Reps: 1

Loading dataset...
  Loaded 256 vectors of dimension 128
  Queries: 10, Groundtruth: 10x10


 K,   Avg cmps,   QPS - mean(max),       Avg Latency,     p99 Latency,               Recall,   Threads
======================================================================================================
10,        256,   7745.9 (7745.9),   15.0us (15.0us),   79.0us (79us),   0.9099999999999999,         1
```

partychen:

#### Reference Issues/PRs

Fixes #1172.

This PR adds an exact brute-force ground-truth file for benchmark examples using
the `disk_index_siftsmall_learn_256pts_data.fbin` and
`disk_index_sample_query_10pts.fbin` workload.

- Added
  `test_data/disk_index_search/disk_index_10pts_idx_uint32_exact_ground_truth.bin`,
  generated using `compute_groundtruth`.
- Updated the applicable benchmark example configurations to use the new exact
  ground-truth file.
- Added no dependencies and made no API changes.

This fixes the recall mismatch where `flat-search` could report recall below
1.0 because the benchmark configurations previously referenced results that
were not exact brute-force ground truth.

#### Any other comments?

Validated with:

- `cargo test --locked --workspace`
- All benchmark example JSON files parse successfully.

### DIFF
--- a/test_data/disk_index_search/disk_index_10pts_idx_uint32_exact_ground_truth.bin
+++ b/test_data/disk_index_search/disk_index_10pts_idx_uint32_exact_ground_truth.bin
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:de012fa2aba9dc6e67031d90ca0422bd8189871293322ad179cd12a6ea22ebbb
+size 808
--- a/diskann-benchmark/example/async-determinant-diversity.json
+++ b/diskann-benchmark/example/async-determinant-diversity.json
@@ -23,7 +23,7 @@
         "search_phase": {
           "search-type": "topk-determinant-diversity",
           "queries": "disk_index_sample_query_10pts.fbin",
-          "groundtruth": "disk_index_10pts_idx_uint32_truth_search_res.bin",
+          "groundtruth": "disk_index_10pts_idx_uint32_exact_ground_truth.bin",
           "reps": 5,
           "num_threads": [
             1
--- a/diskann-benchmark/example/disk-index-determinant-diversity.json
+++ b/diskann-benchmark/example/disk-index-determinant-diversity.json
@@ -22,7 +22,7 @@
         },
         "search_phase": {
           "queries": "disk_index_sample_query_10pts.fbin",
-          "groundtruth": "disk_index_10pts_idx_uint32_truth_search_res.bin",
+          "groundtruth": "disk_index_10pts_idx_uint32_exact_ground_truth.bin",
           "search_list": [10, 20, 40],
           "beam_width": 4,
           "recall_at": 10,
--- a/diskann-benchmark/example/disk-index.json
+++ b/diskann-benchmark/example/disk-index.json
@@ -22,7 +22,7 @@
         },
         "search_phase": {
           "queries": "disk_index_sample_query_10pts.fbin",
-          "groundtruth": "disk_index_10pts_idx_uint32_truth_search_res.bin",
+          "groundtruth": "disk_index_10pts_idx_uint32_exact_ground_truth.bin",
           "search_list": [10, 20, 40],
           "beam_width": 4,
           "recall_at": 10,
@@ -43,7 +43,7 @@
         },
         "search_phase": {
           "queries": "disk_index_sample_query_10pts.fbin",
-          "groundtruth": "disk_index_10pts_idx_uint32_truth_search_res.bin",
+          "groundtruth": "disk_index_10pts_idx_uint32_exact_ground_truth.bin",
           "search_list": [10, 20, 40],
           "beam_width": 4,
           "recall_at": 10,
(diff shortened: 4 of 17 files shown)


harsha-simhadri:
Thanks! Are there any other workflows using the older file? Asking to check if it  might break any workflow.

partychen:
> Thanks! Are there any other workflows using the older file? Asking to check if it might break any workflow.

Thanks for calling this out. I checked all references to the existing file.

The disk-provider regression tests still use the older
`disk_index_10pts_idx_uint32_truth_search_res.bin` file and expect its contents
to remain unchanged.

The benchmark examples were the other consumers. They now reference the new
`disk_index_10pts_idx_uint32_exact_ground_truth.bin` file, which contains the
brute-force KNN ground truth.

So the existing file remains unchanged for workflows that depend on its current
results, while benchmark recall calculations use the new exact ground-truth
file. I also ran `cargo test --locked --workspace`, and all tests passed.

test_data/disk_index_search/disk_index_10pts_idx_uint32_exact_ground_truth.bin
```diff
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:de012fa2aba9dc6e67031d90ca0422bd8189871293322ad179cd12a6ea22ebbb
+size 808
```
hildebrandmw, line 3: Thanks for finding this - the old groundtruth is just straight up wrong. I don't know how it was computed, but it's not right at all. I vote that we replace the current groundtruth file with your new one (which I have verified using a different procedure) and completely ditch the old one which appears to be quite misleading.

hildebrandmw (commented):
Thanks! See my comment below - I think we should get rid of the incorrect groundtruth entirely.

test_data/disk_index_search/disk_index_10pts_idx_uint32_exact_ground_truth.bin
hildebrandmw: Hmmm, I see. The checked in results appear to be the groundtruth as computed directly by the disk index. Which leads me wonder why the disk index isn't getting 100% recall ... @suri-kumkaran, any ideas?

Okay, I take back the suggestion to remove the ground-truth.

hildebrandmw (approved):

arkrishn94 (approved):
Thanks for fixing this!

test_data/disk_index_search/disk_index_10pts_idx_uint32_exact_ground_truth.bin
arkrishn94: Yeah this is definitely weird. There are [tests](https://github.com/microsoft/DiskANN/blob/1552e51c6eb072fa905fb8f04eccabb721c75de5/diskann-disk/src/build/builder/tests.rs#L151) in the disk index crate that check the graph index for this dataset against groundtruth computed using bruteforce and get 100% recall.

Link to recall validation - [link](https://github.com/microsoft/DiskANN/blob/1552e51c6eb072fa905fb8f04eccabb721c75de5/diskann-disk/src/build/builder/core.rs#L1058).

suri-kumkaran: This is very weird, let me try to find the reasoning behind it.

suri-kumkaran (approved):

test_data/disk_index_search/disk_index_10pts_idx_uint32_exact_ground_truth.bin
partychen: Thanks for looking into this. The new brute-force ground-truth data appears to be correct and has been independently verified, so I’ll merge this PR to fix the current benchmark recall issue. Thanks @suri-kumkaran for continuing to investigate why the disk index does not achieve the expected recall here.

PR #1263, opened 2026-07-17
Add a tool for computing groundtruth for a runbook

magdalendobson:
The groundtruth computation tool in the deprecated C++ repository seems to have some bugs, so I wrote a new one in Rust using our Runbook abstractions. This requires `diskann-tools` to take a dependency on `diskann-benchmark-core`, but this is acceptable.

Once this PR lands I will update the [streaming groundtruth tool](https://github.com/harsha-simhadri/big-ann-benchmarks/blob/main/benchmark/streaming/compute_gt.py) in big-ann-benchmarks to use the Rust code instead.


diskann-tools/src/bin/compute_streaming_groundtruth.rs
```diff
@@ -0,0 +1,326 @@
...
+    let mut active: Vec<bool> = vec![false; n_base];
+    // For each active dataset offset, the external ID that should appear in the groundtruth.
+    // For plain inserts external_id == dataset_offset; for Replace they diverge.
+    let mut ext_id: Vec<u32> = vec![0u32; n_base];
+    // Reverse map: external_id -> dataset_offset, needed to resolve Delete/Replace removals.
+    let mut ext_to_offset: HashMap<u32, usize> = HashMap::new();
+
+    let distance_fn = V::distance(args.distance_function, Some(dataset.ncols()));
+
+    for (stage_idx, stage) in runbook.stages().iter().enumerate() {
```
hildebrandmw: Can this be re-written as a `Stream` instead of iterating over `Stages`?

magdalendobson: Yes, done

diskann-tools/src/bin/compute_streaming_groundtruth.rs
```diff
@@ -0,0 +1,342 @@
...
+
+    // FindGroundtruth impl that always returns the expected output path whether
+    // or not it exists yet -- we are about to generate the files.
+    struct AllowMissing {
+        dir: PathBuf,
+        suffix: String,
+    }
+    impl FindGroundtruth for AllowMissing {
+        fn find_groundtruth(&mut self, stage: usize) -> anyhow::Result<PathBuf> {
+            Ok(self.dir.join(format!("step{}.{}", stage, self.suffix)))
```
hildebrandmw: This is a small detail, but we can expose the name mangling for ground-truth as part of the `ScanDirectory` struct in `diskann-benchmark-core` - just as a single source of truth for the expected file name.

diskann-tools/src/bin/compute_streaming_groundtruth.rs
```diff
@@ -0,0 +1,342 @@
...
+    Ok(())
+}
+
+struct GroundtruthStream<'a, V: VectorRepr + Send + Sync> {
+    storage: &'a FileStorageProvider,
+    dataset: &'a Matrix<V>,
+    queries: &'a Matrix<V>,
+    distance_fn: V::Distance,
+    n_base: usize,
+    n_queries: usize,
```
hildebrandmw: Why are `n_queries` and `n_base` tracked here? Can't they be fetched from `queries` and `dataset` respectively?

diskann-tools/src/bin/compute_streaming_groundtruth.rs
```diff
@@ -0,0 +1,342 @@
...
+    recall_at: usize,
+    active: Vec<bool>,
+    ext_id: Vec<u32>,
+    ext_to_offset: HashMap<u32, usize>,
```
hildebrandmw: Unless I'm missing something, isn't `ext_to_offset` sufficient? It can track whether or not an ID is active by its presence or absence. And you can iterate over `(k, v)` pairs to get all the necessary distances.

diskann-tools/src/bin/compute_streaming_groundtruth.rs
```diff
@@ -0,0 +1,342 @@
...
+                    pq.insert(Neighbor {
+                        id: self.ext_id[offset],
+                        distance: dist,
+                    });
+                }
+                pq
+            })
+            .collect();
+
+        let under_k: Vec<usize> = results
```
hildebrandmw: Wouldn't it be a lot easier to check if the number of active IDs was ever less than `K`? I think that's the only reason this could happen - and the error message would be much clearer. This is also something that can be easily added to the validation step for `RunBook` and could be queried through an API. That way, we'd get the error almost immediately instead of potentially deep into a computation.

diskann-benchmark-core/src/streaming/executors/bigann/runbook.rs
```diff
@@ -176,8 +176,7 @@ impl RunBook {
     }
 
     /// Returns a reference to the stages in this runbook.
-    #[cfg(test)]
-    pub(super) fn stages(&self) -> &[Stage] {
+    pub fn stages(&self) -> &[Stage] {
```
hildebrandmw: I think for now this can be left as `#[cfg(test)]`.

diskann-tools/src/bin/compute_streaming_groundtruth.rs
```diff
@@ -0,0 +1,342 @@
...
+fn main() -> CMDResult<()> {
+    init_subscriber();
+    let args = Args::parse();
+    match args.data_type {
+        DataType::Float => run::<f32>(&args),
+        DataType::Fp16 => run::<diskann_vector::Half>(&args),
+        DataType::Uint8 => run::<u8>(&args),
+        DataType::Int8 => run::<i8>(&args),
+    }
+}
```
hildebrandmw, line 113: IF you want a small compile-time win, we can do something like this:
```
trait Distance {
    fn distance(&self, query: usize, data: usize) -> anyhow::Result<f32>;
}

struct Foo<T: VectorRepr> {
    data: Matrix<T>,
    queries: Matrix<T>,
    distance: T::Distance,
}

impl<T: VectorRepr> Distance for Foo<T> {
    fn distance(&self, query: usize, data: usize) -> anyhow::Result<f32> {
        // Should return a better error type for out-of-bounds
        distance.evaluate_similarity(self.queries.row[query], self.data.row[data])
    }
}
```

The rest of the code can take `&dyn Distance` and that way all the streaming/computation loops are only compiled once. Alternaticely, the `distance` could take a `&HashMap<u32, usize>` and a `&mut NeighborPriorityQueue` and do all the distance computations behind the trait object boundary.

Just a thought to move us towards faster compiling patterns.

diskann-tools/src/bin/compute_streaming_groundtruth.rs
magdalendobson: Done

diskann-tools/src/bin/compute_streaming_groundtruth.rs
magdalendobson: Changed to fetch them

diskann-tools/src/bin/compute_streaming_groundtruth.rs
magdalendobson: Yep, changed.

diskann-tools/src/bin/compute_streaming_groundtruth.rs
magdalendobson: Good idea. I moved that check up to the beginning of the search implementation.

diskann-benchmark-core/src/streaming/executors/bigann/runbook.rs
magdalendobson: Yep, you're right, this was intermediate state

diskann-tools/src/bin/compute_streaming_groundtruth.rs
magdalendobson: I did something similar to this--let me know if it looks good!

diskann-tools/src/bin/compute_streaming_groundtruth.rs
```diff
@@ -0,0 +1,342 @@
...
+        );
+
+        let active_ids: Vec<usize> = self
+            .active
+            .iter()
+            .enumerate()
+            .filter_map(|(i, &on)| if on { Some(i) } else { None })
+            .collect();
+
+        // it's generally find to use the global threadpool in diskann-tools
```
metajack: s/find/fine/

metajack (approved):

hildebrandmw (approved):

PR #1265, opened 2026-07-20
Relax `Copy` to `Clone` on `Standard<T>` matrix elements

suri-kumkaran:
**What**

`Mat<Standard<T>>` required `T: Copy`. Now it works for any `Clone` element (and any `T` for views) — matching what `diskann-utils::Matrix<T>` already allows.

**Why**

The `Copy` bound wasn't load-bearing; it came from a naive derive and a few over-broad `where` clauses. Relaxing it is a prerequisite for unifying `Matrix<T>` with `Mat<Standard<T>>`.

**How**

- `Standard<T>`: hand-write unconditional `Copy`/`Clone` (it's just dims + `PhantomData`), drop the unused `Eq`, keep `PartialEq`.
- Drop `T: Copy` from `Repr`/`ReprMut`/`ReprOwned`/`NewRef`/`NewMut` and the `Mat`/`MatRef`/`MatMut` impl blocks.
- `NewOwned`/`NewCloned`: `Copy` → `Clone` (`value.clone()`, `.cloned()`).
- Add a `String`-element test covering fill / `from_fn` / mutation / clone / views.

Repr metadata stays `Copy` (required by `Repr: Copy`, free for `PhantomData`) — only the *element* bound relaxes.

### DIFF
--- a/diskann-quantization/src/multi_vector/matrix.rs
+++ b/diskann-quantization/src/multi_vector/matrix.rs
@@ -256,7 +256,7 @@ pub trait NewCloned: ReprOwned {
 // Standard //
 //////////////
 
-/// Metadata for dense row-major matrices of `Copy` types.
+/// Metadata for dense row-major matrices.
 ///
 /// Rows are stored contiguously as `&[T]` slices. This is the default representation
 /// type for standard floating-point multi-vectors.
@@ -265,14 +265,25 @@ pub trait NewCloned: ReprOwned {
 ///
 /// - `Row<'a>`: `&'a [T]`
 /// - `RowMut<'a>`: `&'a mut [T]`
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, PartialEq)]
 pub struct Standard<T> {
     nrows: usize,
     ncols: usize,
     _elem: PhantomData<T>,
 }
 
-impl<T: Copy> Standard<T> {
+// Hand-written so `Standard<T>` is `Copy`/`Clone` for every `T`: it only stores two `usize`
+// and a `PhantomData<T>`, so a derive would spuriously require `T: Copy`/`T: Clone` (and the
+// `Repr: Copy` supertrait must hold regardless of the element type).
+impl<T> Copy for Standard<T> {}
+
+impl<T> Clone for Standard<T> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<T> Standard<T> {
     /// Create a new `Standard` for data of type `T`.
     ///
     /// Successful construction requires:
@@ -417,7 +428,7 @@ pub enum SliceError {
 // SAFETY: The implementation correctly computes row offsets as `i * ncols` and
 // constructs valid slices of the appropriate length. The `layout` method correctly
 // reports the memory layout requirements.
-unsafe impl<T: Copy> Repr for Standard<T> {
+unsafe impl<T> Repr for Standard<T> {
     type Row<'a>
         = &'a [T]
     where
@@ -447,7 +458,7 @@ unsafe impl<T: Copy> Repr for Standard<T> {
 
 // SAFETY: The implementation correctly computes row offsets and constructs valid mutable
 // slices.
-unsafe impl<T: Copy> ReprMut for Standard<T> {
+unsafe impl<T> ReprMut for Standard<T> {
     type RowMut<'a>
         = &'a mut [T]
     where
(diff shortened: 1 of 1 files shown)


diskann-quantization/src/multi_vector/matrix.rs
```diff
@@ -265,14 +265,25 @@ pub trait NewCloned: ReprOwned {
...
+// Hand-written so `Standard<T>` is `Copy`/`Clone` for every `T`: it only stores two `usize`
+// and a `PhantomData<T>`, so a derive would spuriously require `T: Copy`/`T: Clone` (and the
+// `Repr: Copy` supertrait must hold regardless of the element type).
+impl<T> Copy for Standard<T> {}
+
+impl<T> Clone for Standard<T> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
```
hildebrandmw, line 284: `PartialEq` and `Eq` could probably get the same treatment to be independent of `T`.

diskann-quantization/src/multi_vector/matrix.rs
```diff
@@ -492,11 +503,11 @@ unsafe impl<T: Copy> ReprOwned for Standard<T> {
 // initialized by it is non-null and properly aligned to the underlying type.
 unsafe impl<T> NewOwned<T> for Standard<T>
 where
-    T: Copy,
+    T: Clone,
 {
     type Error = crate::error::Infallible;
     fn new_owned(self, value: T) -> Result<Mat<Self>, Self::Error> {
-        let b: Box<[T]> = (0..self.num_elements()).map(|_| value).collect();
+        let b: Box<[T]> = (0..self.num_elements()).map(|_| value.clone()).collect();
```
hildebrandmw: We can use [`repeat_n`](https://doc.rust-lang.org/beta/std/iter/fn.repeat_n.html) to avoid a clone for the last element.

diskann-quantization/src/multi_vector/matrix.rs
```diff
@@ -506,7 +517,7 @@ where
 // SAFETY: This safely reuses `<Self as NewOwned<T>>`.
 unsafe impl<T> NewOwned<Defaulted> for Standard<T>
 where
-    T: Copy + Default,
+    T: Clone + Default,
```
hildebrandmw: Is it worth dropping the `Clone` requirement here? Requiring just `T: Default` seems desirable.

diskann-quantization/src/multi_vector/matrix.rs
```diff
@@ -552,10 +557,10 @@ where
 
 impl<T> NewCloned for Standard<T>
 where
-    T: Copy,
+    T: Clone,
 {
     fn new_cloned(v: MatRef<'_, Self>) -> Mat<Self> {
-        let b: Box<[T]> = v.rows().flatten().copied().collect();
+        let b: Box<[T]> = v.rows().flatten().cloned().collect();
```
hildebrandmw: Small nit: We can use `as_slice` now to avoid going through rows + flatten. The advantage is that `as_slice` will be a single `ExactSizeIterator` and so may optimize a bit better.

hildebrandmw (commented):
Thanks! Just a few minor comments.

diskann-quantization/src/multi_vector/matrix.rs
suri-kumkaran: Make sense, done.

diskann-quantization/src/multi_vector/matrix.rs
suri-kumkaran: Make sense, done.

diskann-quantization/src/multi_vector/matrix.rs
suri-kumkaran: Done.

diskann-quantization/src/multi_vector/matrix.rs
suri-kumkaran: Done.

diskann-quantization/src/multi_vector/matrix.rs
```diff
@@ -265,14 +265,25 @@ pub trait NewCloned: ReprOwned {
 ///
 /// - `Row<'a>`: `&'a [T]`
 /// - `RowMut<'a>`: `&'a mut [T]`
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, PartialEq)]
 pub struct Standard<T> {
     nrows: usize,
     ncols: usize,
     _elem: PhantomData<T>,
```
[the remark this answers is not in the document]
suri-kumkaran: Added it back.

hildebrandmw (approved):

wuw92 (approved):

PR #1269, opened 2026-07-21
Unify the two matrix types

suri-kumkaran:
We carry two matrix types for overlapping jobs: the dense `MatrixBase` in `diskann-utils` (the `Matrix`/`MatrixView` everything reaches for), and `diskann-quantization`'s `Repr`-generic `Mat<Standard<T>>`, which also covers the exotic layouts (block-transposed, quantized). This PR folds them into one.

The `Repr` framework is the more capable of the two, so it stays — moving into `diskann-utils`, where the dense types already live. Dense becomes a thin alias layer (`Matrix<T> = Mat<RowMajor<T>>`) that keeps every existing name, so callers barely notice, and the old `MatrixBase` is deleted.

A few things worth flagging:

- `Standard` is renamed to `RowMajor` — the codebase already says "row-major" everywhere; `Standard` was the odd one out.
- The generic constructor `Mat::new(repr, init)` becomes `Mat::from_repr`, freeing plain `new` for the dense `Matrix::new(gen, rows, cols)` most callers use.
- Quantization keeps its block-transposed and quantized layouts; they now `impl diskann_utils::matrix::Repr` instead of owning the framework.
- No crate-root re-export in `diskann-utils` — it's already a grab-bag, so the types stay under `diskann_utils::matrix`.

## How to review

Most of the diff (~110 files) is mechanical and safe to skim: `Standard`→`RowMajor`, `MatrixBase`→`Matrix`, the constructor renames, `views::*`→`matrix::*`. The parts that want real attention:

1. **`diskann-utils/src/matrix.rs`** — the heart of it. Git tracks it as a rename of `diskann-quantization/src/multi_vector/matrix.rs`, so read it as "that file, `Standard`→`RowMajor`, plus the dense aliases/API and `DenseData` folded in."
2. **`lib.rs`, the deleted `views.rs`, `strided.rs`** — the module wiring, and where `MatrixBase` used to live.
3. **`multi_vector/mod.rs`, the deleted `multi_vector/matrix.rs`, `block_transposed.rs`** — the quantization side now building on the shared framework.
4. **`product/tables/basic.rs` and `minmax/multi/meta.rs`** — the only spots with real logic changes, not mechanical ones.


diskann-quantization/tests/compile-fail/multi/mat_as_view.rs
```diff
@@ -3,12 +3,12 @@
  * Licensed under the MIT license.
  */
 
-use diskann_quantization::multi_vector::{Mat, Standard};
+use diskann_quantization::multi_vector::{Mat, RowMajor};
```
hildebrandmw: A bit strange to keep the compile tests in `diskann-quantization` when the things they are testing against have been moved to `diskann-utils`. Please move these to `diskann-utils` as well.

Since these don't rely on post-monomorphization errors, the implementation in `diskann-utils` doesn't need to use the "force compilation" trick that `diskann-quantization` does.

diskann-utils/src/views.rs
hildebrandmw, line 1: This may slightly delay when we actually do the cross over, but is it possible to keep the new matrix in `views` for one PR and then rename the module in a follow-up. Will help with auditing across all the files.

diskann-utils/src/views.rs
```diff
@@ -635,7 +648,10 @@ impl<T: ReprOwned> Mat<T> {
         }
     }
 
-    pub(crate) unsafe fn get_row_unchecked(&self, i: usize) -> T::Row<'_> {
+    /// # Safety
```
hildebrandmw, line 674: Please add a tiny docstring - though the behavior of this function is pretty self-explanatory.

diskann-utils/src/views.rs
```diff
@@ -879,19 +899,15 @@ impl<'a, T> MatRef<'a, Standard<T>> {
...
-        // SAFETY: `Standard<T>` guarantees `nrows * ncols` contiguous `T` elements
+        // SAFETY: `RowMajor<T>` guarantees `nrows * ncols` contiguous `T` elements
         // starting at `self.ptr`. The lifetime `'a` is tied to the original data.
         unsafe { std::slice::from_raw_parts(self.ptr.as_ptr().cast::<T>(), len) }
     }
 
     /// Return a [`MatrixView`] over the backing data.
-    #[allow(clippy::expect_used)]
     #[inline]
     pub fn as_matrix_view(&self) -> MatrixView<'a, T> {
```
hildebrandmw, line 943: Should we deprecate this method in a follow-up?

diskann-utils/src/matrix.rs
```diff
@@ -827,7 +844,10 @@ impl<'a, T: Repr> MatRef<'a, T> {
     ///
     /// `i` must be less than `self.num_vectors()`.
     #[inline]
-    pub(crate) unsafe fn get_row_unchecked(&self, i: usize) -> T::Row<'_> {
+    /// # Safety
```
hildebrandmw: What's going on here?

diskann-utils/src/views.rs
```diff
@@ -1335,7 +2006,7 @@ mod tests {
         }
     }
```
hildebrandmw, line 2135: One thing that stands out is that there is a large increase in duplicated code for the constructors/accessors for `Mat<RowMajor>`/`MatRef<'_, RowMajor>`/`MatMut<'_, RowMajor>`, but not so much in an increase of test coverage. Are we confident that all the non-trivial duplicated methods are being hit? I'm not exactly a big fan of all the duplication to be honest.

diskann-utils/src/matrix.rs
```diff
@@ -1119,6 +1135,596 @@ impl<'a, T> MatMut<'a, Standard<T>> {
...
+    /// `row < self.nrows()` and `col < self.ncols()`.
+    pub unsafe fn get_unchecked_mut(&mut self, row: usize, col: usize) -> &mut T {
+        let ncols = self.ncols();
+        // SAFETY: guaranteed in-bounds by the caller.
+        unsafe { self.as_mut_slice().get_unchecked_mut(row * ncols + col) }
+    }
+
+    /// Consume the matrix, returning its backing storage as a `Box<[T]>`.
+    pub fn into_inner(self) -> Box<[T]> {
+        let len = self.nrows() * self.ncols();
```
hildebrandmw: `self.repr.num_elements()`? Or even better: `self.as_slice()`?

diskann-utils/src/matrix.rs
```diff
@@ -1119,6 +1135,596 @@ impl<'a, T> MatMut<'a, Standard<T>> {
...
+            _ => return Err(err),
+        };
+        // SAFETY: `data.len()` was checked to equal `repr.num_elements()`.
+        Ok(unsafe { repr.box_to_mat(data) })
+    }
+
+    /// Take ownership of `data` as a single-row matrix.
+    pub fn row_vector(data: Box<[T]>) -> Self {
+        let ncols = data.len();
+        Self::try_from(data, 1, ncols).expect("row vector length")
```
hildebrandmw: The advantage of the `row_vector`/`column_vector` constructors are to provide light-weight, infallible ways of building matrics. Adding a panicking path here seems like a step backwards in library design.

diskann-utils/src/matrix.rs
```diff
@@ -1119,6 +1135,596 @@ impl<'a, T> MatMut<'a, Standard<T>> {
...
+}
+
+impl<T> Mat<RowMajor<T>> {
+    /// Fill a new `nrows x ncols` matrix with clones of `value`.
+    pub fn filled(value: T, nrows: usize, ncols: usize) -> Self
+    where
+        T: Clone,
+    {
+        Self::new(value, nrows, ncols)
+    }
```
hildebrandmw: Why do we need `filled`? Does the blanket implementation of `Generator<T> for T where T: Clone` not suffice?

diskann-utils/src/matrix.rs
```diff
@@ -1119,6 +1135,596 @@ impl<'a, T> MatMut<'a, Standard<T>> {
...
+    /// Number of columns.
+    #[inline]
+    pub fn ncols(&self) -> usize {
+        self.vector_dim()
+    }
+
+    /// Row `i`. Panics if `i >= self.nrows()`.
+    pub fn row(&self, i: usize) -> &'a [T] {
+        let ncols = self.ncols();
+        &self.as_slice()[i * ncols..i * ncols + ncols]
```
hildebrandmw: Why can't this delegate to `get_row`?

diskann-utils/src/strided.rs
```diff
@@ -497,17 +497,29 @@ where
...
+}
+
+impl<'a, T> From<crate::matrix::MatMut<'a, crate::matrix::RowMajor<T>>>
+    for StridedBase<&'a mut [T]>
 {
-    fn from(matrix: views::MatrixBase<T>) -> Self {
-        let nrows = matrix.nrows();
-        let ncols = matrix.ncols();
+    fn from(mut m: crate::matrix::MatMut<'a, crate::matrix::RowMajor<T>>) -> Self {
+        let (nrows, ncols) = (m.nrows(), m.ncols());
```
hildebrandmw, line 514: Is there not already an `as_mut_slice` method?

diskann-utils/src/matrix.rs
```diff
@@ -1119,6 +1135,596 @@ impl<'a, T> MatMut<'a, Standard<T>> {
...
+/// Shared dense row-major view.
+pub type MatrixView<'a, T> = MatRef<'a, RowMajor<T>>;
+
+/// Mutable dense row-major view.
+pub type MatrixViewMut<'a, T> = MatMut<'a, RowMajor<T>>;
+
+/// Error returned when a slice length is incompatible with the requested dimensions.
+#[derive(Debug, Error)]
+#[error("cannot view a slice of length {len} as a {nrows}x{ncols} matrix")]
+pub struct TryFromError {
```
hildebrandmw: This breaks what the original `TryFromError` was trying to achieve. If you pass a large `Box<[T]>` to something and happen to get the sizes wrong, `TryFromError` will allow you to retrieve the original `Box<[T]>` without deallocating. Dropping this seems like a step back in library design.

diskann-utils/src/strided.rs
```diff
@@ -497,17 +497,29 @@ where
     }
 }
 
-impl<T, U> From<views::MatrixBase<T>> for StridedBase<U>
-where
-    T: DenseData,
-    U: DenseData,
-    T: Into<U>,
+impl<'a, T> From<crate::matrix::MatRef<'a, crate::matrix::RowMajor<T>>> for StridedBase<&'a [T]> {
```
hildebrandmw: Please keep the style consistent and use top level imports.

diskann-quantization/src/product/tables/basic.rs
```diff
@@ -22,23 +22,40 @@ use thiserror::Error;
...
 /// A `BasicTableBase` that references its contents. Construction of such a table will
 /// not result in a memory allocation.
-pub type BasicTableView<'a> = BasicTableBase<&'a [f32], &'a [usize]>;
+pub type BasicTableView<'a> = BasicTableBase<MatrixView<'a, f32>, &'a [usize]>;
+
+/// Bridges owned ([`Matrix`]) and borrowed ([`MatrixView`]) pivot storage, exposing a
+/// [`MatrixView`] so a [`BasicTableBase`] reads its dimensions and rows from one place.
+pub trait PivotTable {
+    fn view(&self) -> MatrixView<'_, f32>;
+}
```
hildebrandmw, line 46: I'm not sure how I feel about this. I'll need to think about it more ...

diskann-quantization/src/multi_vector/block_transposed.rs
```diff
@@ -578,30 +578,22 @@ unsafe impl<T: Copy, const GROUP: usize, const PACK: usize> ReprOwned
 // ════════════════════════════════════════════════════════════════════
 
 // SAFETY: The returned `Mat` contains a `Box` with exactly `self.storage_len()` elements.
-unsafe impl<T: Copy, const GROUP: usize, const PACK: usize> NewOwned<T>
+//
+// No generic `NewOwned<T>` value-fill: `Defaulted` is foreign (diskann-utils), so a value-fill
+// impl would overlap with this under intercrate coherence. These matrices are only default-init'd.
+unsafe impl<T: Copy + Default, const GROUP: usize, const PACK: usize> NewOwned<Defaulted>
```
hildebrandmw: This seems pretty bad that users can't implement a broadcasted constructor.

diskann-utils/src/views.rs
suri-kumkaran: Fair point. Reads are now de-duplicated: the non-trivial accessors (`row`, `row_iter`, `window_iter`, `par_*`, `subview`, `try_get`, `map`, `get_unchecked`) have a single body on `MatrixView`, and `Matrix`/`MatrixViewMut` forward to it via a `delegate_read!` macro (like the block-transposed layout's `delegate_to_ref!`).

For the mut methods (`as_mut_slice`, `row_mut`, `*_iter_mut`) I stopped short: centralising them needs the consuming `_inner(self) -> &'a mut …` helpers that block-transposed uses, which trades a couple of trivial duplicated lines for `unsafe` + indirection. I leaned toward leaving them direct, but I don't feel strongly - do you want the muts folded into the same pattern for full consistency, or kept simple?

Coverage: reads funnel through the one `MatrixView` impl, so I tested it directly - `subview` (incl. OOB), `try_get`, `map`, `window_iter` batches/remainder + `batchsize == 0` panic, `row` OOB panic, `get_unchecked` - plus a test that `Matrix`/`MatrixView`/`MatrixViewMut` return identical rows, so each forwarding path is hit.

diskann-utils/src/views.rs
suri-kumkaran: Yes, I also identified this and I plan to deprecate this in a follow-up.

diskann-utils/src/views.rs
```diff
@@ -1119,6 +1135,596 @@ impl<'a, T> MatMut<'a, Standard<T>> {
...
+    /// Iterator over rows.
+    pub fn row_iter(&self) -> impl ExactSizeIterator<Item = &'a [T]> {
+        self.as_slice().chunks_exact(self.ncols())
+    }
+
+    /// Iterator over sub-matrices of up to `batchsize` rows.
+    ///
+    /// # Panics
+    /// Panics if `batchsize == 0`.
+    pub fn window_iter(&self, batchsize: usize) -> impl Iterator<Item = MatrixView<'a, T>> {
```
arkrishn94, line 1322: Instead of panicking, maybe throw an error? Or better yet, use `NonZeroUsize`?

diskann-utils/src/views.rs
```diff
@@ -1119,6 +1135,596 @@ impl<'a, T> MatMut<'a, Standard<T>> {
...
+    #[cfg(feature = "rayon")]
+    pub fn par_window_iter(
+        &self,
+        batchsize: usize,
+    ) -> impl rayon::iter::IndexedParallelIterator<Item = MatrixView<'a, T>>
+    where
+        T: Send + Sync,
+    {
+        use rayon::slice::ParallelSlice;
+        assert!(batchsize != 0, "par_window_iter batchsize cannot be zero");
```
arkrishn94, line 1353: Same here re the error :)

diskann-utils/src/matrix.rs
```diff
@@ -1119,6 +1135,596 @@ impl<'a, T> MatMut<'a, Standard<T>> {
...
+        self.as_mut_slice().chunks_exact_mut(ncols)
+    }
+
+    /// Iterator over sub-matrices of up to `batchsize` rows.
+    pub fn window_iter(&self, batchsize: usize) -> impl Iterator<Item = MatrixView<'_, T>> {
+        assert!(batchsize != 0, "window_iter batchsize cannot be zero");
+        let ncols = self.ncols();
+        self.as_slice()
+            .chunks(ncols * batchsize)
+            .map(move |d| MatrixView::try_from(d, d.len() / ncols, ncols).expect("exact chunk"))
```
arkrishn94: Is panicking within these iterators the right move here? If the params check out, shouldn't the user be able to construct this iterator in an error free way? Same goes with the other iterator-like types below

arkrishn94 (commented):
Thanks for taking this on Suryansh. I left some minor comments re errors. 

Apart from that, I notice there is quite a large increase in the number of constructors and accessors for `Mat` and its friends. I recognize that most of this is needed because of the switch from the old `view` type but can you - 

1. Do an audit and confirm that each of these accessors/constructors in `matrix.rs` is needed and can't be consolidated. 
2. That these are also tested correctly? I don't notice any extra testing there.

diskann-utils/src/views.rs
suri-kumkaran: Panics here predate this PR. On `main` these iterators already did `assert!(batchsize != 0)` plus a per-chunk `assert_eq!(blobsize % ncols == 0)` — I just carried them over (the `assert_eq!` became `try_from(..).expect(..)`, same behavior). So the unification didn't introduce the panicking, and it's the same for the parallel ones below.

That said, I agree with the direction — once the params check out, these should be constructible error-free. `NonZeroUsize` for `batchsize` removes the first panic at compile time, and the per-chunk `expect` is unreachable given valid params, so it can be infallible construction instead. Since it's pre-existing and spans all the window/par-window iterators plus their call sites, I'll do it as a small dedicated follow-up rather than widen this PR.

diskann-utils/src/views.rs
suri-kumkaran: Replied in the earlier comment.

diskann-utils/src/matrix.rs
suri-kumkaran: Replied in the earlier comment.

diskann-quantization/tests/compile-fail/multi/mat_as_view.rs
suri-kumkaran: Done.

diskann-utils/src/views.rs
suri-kumkaran: That makes sense. I'll handle the module reordering and streamlining in a follow up PR.

diskann-utils/src/views.rs
suri-kumkaran: Done.

diskann-utils/src/matrix.rs
suri-kumkaran: Sorry for the glitch, fixed it.

diskann-utils/src/matrix.rs
suri-kumkaran: Done.

diskann-utils/src/matrix.rs
suri-kumkaran: Fixed.

diskann-utils/src/matrix.rs
suri-kumkaran: Removed.

diskann-utils/src/matrix.rs
suri-kumkaran: Used `BlockTransposed` pattern of delegation.

diskann-utils/src/strided.rs
suri-kumkaran: Removed.

diskann-utils/src/matrix.rs
suri-kumkaran: Sorry for going backwards, now fixed it.

diskann-utils/src/strided.rs
suri-kumkaran: Fixed.

diskann-quantization/src/product/tables/basic.rs
suri-kumkaran: Do you have any better suggestion?

diskann-quantization/src/multi_vector/block_transposed.rs
suri-kumkaran: Fixed. That broadcast constructor got dropped by accident: on `main`, `BlockTransposed` had both `NewOwned<T>` (value-fill) and `NewOwned<Defaulted>` (default), which overlap at `Repr<Defaulted>` — the compiler only tolerated it while `Defaulted` was a local type. Moving the framework here made `Defaulted` foreign, the impls collided, and round-1 resolved it by cutting the value-fill one.

I've split default-init into its own `NewDefault` trait (+ `Mat::from_default`) and deleted the `Defaulted` marker, so there's no overlap. `BlockTransposed` gets its `NewOwned<T>` broadcast constructor back, now alongside a separate `NewDefault` impl - so a custom layout can implement either or both. `test_standard_non_copy_element` covers both paths.

diskann-utils/src/views.rs
```diff
@@ -3,139 +3,1226 @@
...
+        }
+    }
+
+    /// Return the base pointer for the [`Mat`].
+    pub fn as_raw_ptr(&self) -> *const u8 {
+        self.ptr.as_ptr()
+    }
+
+    /// Return a mutable base pointer for the [`Mat`].
+    pub fn as_raw_mut_ptr(&mut self) -> *mut u8 {
```
arkrishn94, line 753: Should this be public?

suri-kumkaran: It was `pub(crate)` on main. It's public now only because the framework moved into `diskann-utils` while the layouts that build on it (`block_transposed` and friends) stayed in `diskann-quantization` — their `Repr` impls need `as_raw_mut_ptr` to construct views over their own storage, and `pub(crate)` doesn't reach across the crate boundary. Same reason `from_raw_parts` and `get_row_unchecked` are public now.

If you'd still rather it not be public, do you have a preferred approach? Happy to explore alternatives.

diskann-quantization/src/minmax/multi/mod.rs
```diff
@@ -54,15 +54,15 @@
...
 //!     0.0, 0.0, 1.0, 0.0,  // doc vector 2
 //! ];
-//! let doc_input = MatRef::new(
-//!     Standard::new(num_doc_vectors, dim).unwrap(), &doc_data
+//! let doc_input = MatRef::from_repr(
+//!     RowMajor::new(num_doc_vectors, dim).unwrap(), &doc_data
 //! ).unwrap();
 //!
-//! // Create owned matrices for quantized output using Mat::new
+//! // Create owned matrices for quantized output using Mat::from_repr
```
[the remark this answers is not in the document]
suri-kumkaran: Fixed.

hildebrandmw (commented):
I'm not entirely convinced that the replacement is ready yet. The `Repr` abstraction basically forces duplication of inherent methods (e.g. `BlockTranspose` and now `Mat<RowMajor<T>>`) which will be a nightmare to keep in sync and explodes the test surface area. Plus, `Rows` as the first-class aspect of a generic matrix representation is a somewhat arbitrary choice. It makes sense from a multi-vector perspective, but less from a core-abstraction perspective.

I think we need to rework `Repr` and friends so they work more like the existing `DenseData` with an ability to reborrow. That would take care of the API surface duplication. The `Rows` machinery could be split into its own `Rows`/`RowsMut` traits to keep `Repr` simple. Then we might be able to keep the `MatrixBase` style class like the current class as a new-type wrapper over e.g. `Mat<RowMajor<T>>` to avoid polluting `Mat` with a bunch of inherent methods.

Also, the variance properties of `Mat` And friends needs revisiting: I don't think it's quite right (my bad). For example, `Mat` does not need to be invariant. `MatRef`'s `Send` bound is also incorrect as it should require `T: Sync` plus a proper bound on its `Repr`.

Basically, I think `Mat` + `Repr` is still too complicated to make a core abstraction.

diskann-utils/src/views.rs
```diff
@@ -3,139 +3,1226 @@
...
+        $vis unsafe fn $name(&self $(, $a: $t)*) $(-> $r)? {
+            // SAFETY: the caller upholds the delegated method's safety contract.
+            unsafe { self.as_view().$name($($a),*) }
+        }
+    };
+}
+
+impl<T> Mat<RowMajor<T>> {
+    /// Construct a [`Mat`] by filling each element in row-major order from `gen`.
+    pub fn new<U: Generator<T>>(mut gen: U, nrows: usize, ncols: usize) -> Self {
```
partychen, line 799: This public constructor now routes caller-supplied dimensions through `RowMajor::new(...).unwrap_or_else(...)`, so overflow or an `isize::MAX` byte-budget violation panics, but the rustdoc does not state that contract. Please add a `# Panics` section describing those conditions, or expose a fallible constructor if callers are expected to handle them.

diskann-utils/src/views.rs
```diff
@@ -161,2023 +1247,1626 @@ impl<T: DenseData> TryFromError<T> {
...
-        T: MutDenseData,
-        T::Elem: Send,
-    {
-        let ncols = self.ncols();
-        self.as_mut_slice().par_chunks_exact_mut(ncols)
+// Indexing by `(row, col)`.
+impl<T> std::ops::Index<(usize, usize)> for Mat<RowMajor<T>> {
+    type Output = T;
+    fn index(&self, (row, col): (usize, usize)) -> &T {
+        self.try_get(row, col).expect("index out of bounds")
```
partychen, line 1678: The unified indexing path regresses the old dense matrix diagnostics: `MatrixBase` distinguished an invalid row from an invalid column and included the corresponding bound, while all four new `Index`/`IndexMut` implementations collapse failures to `index out of bounds`. Please preserve the row or column and its bound in these panic messages so failures remain actionable.

PR #1270, opened 2026-07-22
perf: software prefetch bitmask filter K=8 steps ahead in iterate_to_…

cave-g-f:
…fixed_point

Reduces contain_filtered_label self-cost from 49.2% to 14.3% by hiding DRAM latency on the random-access node bitmask load. Measured on DANNMemoryTable with 21M-node index, 189 filter labels, 12 threads: QPS 7,705 -> 12,435 (+61.4%), P99 3012us -> 1656us (-45%).

Changes:
- filter_match_proxy: add prefetch_bitmask(id) virtual method that issues _mm_prefetch on the node's bitmask address. No-op for the integer_label path.
- iterate_to_fixed_point: pre-prefetch first K=8 neighbors' bitmasks before the loop, then sliding-window prefetch K steps ahead inside the loop. Guarded by use_filter to avoid overhead on unfiltered queries. Applied to both dynamic and static graph paths.
- Use NeighborList::data() for indexed access (no operator[]).
- Also brings AVX2 buffer padding from downstream d9583a32:
  * bitmask_filter_match ctor: query_bitmask_buf resized to at least 4 words for safe 256-bit loads.
  * build_bitmask_index: node bitmask buffer padded with 4 extra uint64 words at end for safe AVX2 reads on the last node.

The prefetch instruction itself accounts for 11% self-cost in the post- optimization profile, which is the expected steady-state cost of hiding the DRAM stall on 44 billion filter checks (4.6M queries x 200 search list x ~48 out-edges).

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -1186,14 +1186,36 @@ std::pair<uint32_t, uint32_t> Index<T, TagT, LabelT>::iterate_to_fixed_point(
         {
             LockGuard guard(_locks[n]);
             auto neighbour_list = _graph_store->get_neighbours(n);
-            for (auto id : neighbour_list)
+            const location_t* neighbour_data = neighbour_list.data();
+            const size_t nbrs_count = neighbour_list.size();
+            constexpr size_t BITMASK_PREFETCH_K = 8;
+
+            // Pre-prefetch bitmasks for first K neighbors (only if filtering)
+            if (use_filter)
             {
+                const size_t prefetch_init = std::min(BITMASK_PREFETCH_K, nbrs_count);
+                for (size_t p = 0; p < prefetch_init; ++p)
+                {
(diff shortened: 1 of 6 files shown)


src/index.cpp
```diff
@@ -1057,10 +1057,31 @@ std::pair<uint32_t, uint32_t> Index<T, TagT, LabelT>::iterate_to_fixed_point(
...
+                }
+            }
+
+            for (size_t i = 0; i < nbrs_count; ++i)
+            {
+                auto id = neighbour_data[i];
                 assert(id < _max_points);
 
+                // Prefetch bitmask K steps ahead (sliding window)
+                if (use_filter && i + BITMASK_PREFETCH_K < nbrs_count)
```
Sanhaoji2: we can move the prefetch part after visited check?

PR #1271, opened 2026-07-22
Refactor disk index build structure

wuw92:
## Why

This follows #1254. Removing checkpoint and continuation support left `core.rs` grouping several independent parts of disk-index construction, making the remaining build path harder to review and maintain.

## What

This PR is primarily a **mechanical** file split of the existing disk-index build implementation. Existing code blocks move to files that match their current responsibilities. No build algorithm or storage-format change is intended.

1. **Split `core.rs`:** move the partition, shard-build, merge, and cleanup workflow to `merged_index.rs`; move shared fixtures and existing end-to-end tests to `tests.rs`; then remove the empty source module.
2. **Split `build.rs`:** move the existing single-graph construction pipeline to `inmem_builder.rs`, leaving top-level PQ construction, one-shot versus merged selection, and disk-layout creation in `build.rs`.
3. **Reconnect the moved blocks:** update private module wiring, imports, and test fixture re-exports without introducing another build path.

```text
build.rs
├── one-shot ────────────────> inmem_builder.rs
└── merged ──> merged_index.rs ──> inmem_builder.rs
```

This PR does not introduce a new Vamana abstraction or change runtime dispatch. Beyond the file moves, it intentionally:

- narrows internal builder modules and helpers that are not supported extension points;
- removes direct stdout progress output from the merge path;
- uses each graph header's block size when comparing test indices; 
- removes a RAM-estimation test that duplicated the production formula.

## Existing test coverage moved by this PR

- One-shot builds cover L2, inner-product, cosine, associated data, and nodes spanning multiple sectors.
- The low-memory path covers merged construction, search results, and maximum-degree persistence.
- Error cases cover unsupported build quantization settings.
- The build matrix covers one-shot and merged FP, SQ, PQ, and MinMax construction and search.


diskann-disk/src/build/builder/tests.rs
```diff
@@ -4,17 +4,636 @@
  */
 
 //! Disk index builder tests.
+use crate::{data_model::GraphDataType, storage::DiskIndexWriter};
+use diskann_providers::{
+    model::IndexConfiguration,
+    storage::{StorageReadProvider, StorageWriteProvider},
+    utils::load_metadata_from_file,
+};
+use diskann_utils::io::read_bin;
```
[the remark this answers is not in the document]
wuw92: Good point—the imports were consumed through `use super::*`, so CI did not report them as unused, but that made the test module dependencies implicit. I moved the imports into `disk_index_builder_tests` and removed the wildcard import in 6c76dbeb.

diskann-disk/src/build/builder/mod.rs
```diff
@@ -5,11 +5,14 @@
...
 pub mod build;
-pub mod core;
-pub mod quantizer;
 
-pub mod inmem_builder;
-pub mod tokio;
+mod inmem_builder;
+mod merged_index;
+mod quantizer;
+mod tokio;
```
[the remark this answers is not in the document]
wuw92: Updated `BuildQuantizer`, `BuildQuantizer::train`, and `create_runtime` to `pub(super)`, matching their sibling-module-only usage.

diskann-disk/src/build/builder/tests.rs
```diff
@@ -5,16 +5,629 @@
...
+                test_node_per_block as usize,
+                DEFAULT_DISK_SECTOR_LEN,
+            );
+
+            let truth_node_id_offset = node_data_offset(
+                idx,
+                truth_max_node_length as usize,
+                truth_node_per_block as usize,
+                DEFAULT_DISK_SECTOR_LEN,
+            );
```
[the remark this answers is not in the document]
wuw92: Updated the graph comparison to calculate each node offset with the block size stored in that graph's own header.

diskann-disk/src/build/builder/merged_index.rs
```diff
@@ -0,0 +1,538 @@
...
+
+                let nnbrs: u32 = std::cmp::min(final_nbrs.len() as u32, max_degree);
+                merged_vamana_cached_writer.write(&nnbrs.to_le_bytes())?;
+
+                let bytes = final_nbrs
+                    .iter()
+                    .take(nnbrs as usize)
+                    .flat_map(|x| x.to_le_bytes())
+                    .collect::<Vec<u8>>();
+                merged_vamana_cached_writer.write(&bytes)?;
```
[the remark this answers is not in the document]
wuw92: Updated this path to write the contiguous neighbor prefix through `bytemuck::must_cast_slice`, avoiding the per-node `Vec<u8>` allocation.

diskann-disk/src/build/builder/merged_index.rs
```diff
@@ -0,0 +1,538 @@
...
+
+                merged_index_size += (size_of::<u32>() + nnbrs as usize * size_of::<u32>()) as u64;
+                if cur_id % 499999 == 1 {
+                    print!(".");
+                }
```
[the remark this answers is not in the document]
wuw92: Removed the direct stdout progress output from the library merge path.

diskann-disk/src/build/builder/merged_index.rs
```diff
@@ -0,0 +1,538 @@
...
+
+        let nnbrs: u32 = std::cmp::min(final_nbrs.len() as u32, max_degree);
+        merged_vamana_cached_writer.write(&nnbrs.to_le_bytes())?;
+
+        let bytes = final_nbrs
+            .iter()
+            .take(nnbrs as usize)
+            .flat_map(|x| x.to_le_bytes())
+            .collect::<Vec<u8>>();
+        merged_vamana_cached_writer.write(&bytes)?;
```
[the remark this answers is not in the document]
wuw92: Updated the final-node path to write the contiguous neighbor prefix directly as bytes as well.

diskann-disk/src/build/builder/merged_index.rs
```diff
@@ -0,0 +1,527 @@
...
+    graph_degree: u64,
+    build_quantization_type: &QuantizationType,
+) -> f64 {
+    let graph_size =
+        (num_points * graph_degree * mem::size_of::<u32>() as u64) as f64 * GRAPH_SLACK_FACTOR;
+
+    let single_vec_size = match *build_quantization_type {
+        QuantizationType::FP => dim.next_multiple_of(8u64) * datasize,
+        // We can skip PQ pivots data as it is very small(~3MB) for even large datasets like OAI-3072.
+        QuantizationType::PQ { num_chunks } => num_chunks as u64,
```
harsha-simhadri, line 53: I feel this pattern is too fragile. when the data store change, these can out of sync.

Could all these size functions be implemented by the store that holds the data? So, if the builder is instantiating some data stores + internal data structures, it just calls the data_store->bytes() data structure and adds them up.

diskann-disk/src/build/builder/merged_index.rs
```diff
@@ -0,0 +1,527 @@
...
+    use rstest::rstest;
+
+    use super::*;
+    use crate::QuantizationType;
+
+    #[rstest]
+    #[case(QuantizationType::FP)]
+    #[case(QuantizationType::PQ { num_chunks: 15 })]
+    #[case(QuantizationType::SQ { nbits: 1, standard_deviation: None })]
+    fn test_estimate_build_index_ram_usage(#[case] build_quantization_type: QuantizationType) {
```
harsha-simhadri: What are we testing against here? how are we computing the actual usage?

harsha-simhadri (commented):
could you please briefly list the tests in this PR  and their intent in the PR description? thanks

diskann-disk/src/build/builder/merged_index.rs
wuw92: Agreed that store-owned byte estimates would prevent this from drifting as the backing stores change. This estimator predates this PR and was moved unchanged from `core.rs`; implementing that design would add APIs across `diskann-providers` and change the build-strategy estimate. I am keeping the estimator unchanged in this structural PR so that change can be reviewed independently.

diskann-disk/src/build/builder/merged_index.rs
wuw92: You are right: this recomputed the same formula and did not measure actual memory usage. I removed the test.

diskann-disk/src/build/builder/merged_index.rs
```diff
@@ -0,0 +1,484 @@
...
+                DiskIndexWriter::get_merged_index_subshard_mem_index_file(merged_index_prefix, p);
+
+            self.storage_provider.delete(&shard_base_file)?;
+            self.storage_provider.delete(&shard_ids_file)?;
+            self.storage_provider.delete(&shard_index_file)?;
+        }
+
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
wuw92: I would prefer to address this in a follow-up PR together with the store-owned RAM estimator. Reintroducing the old unit test here would duplicate the estimation formula without validating the allocations created by the stores.

wuw92:
> could you please briefly list the tests in this PR and their intent in the PR description? thanks

 I’ve updated the PR description with a brief list of the covered build scenarios and the intent of each test.

diskann-disk/src/build/builder/merged_index.rs
wuw92: done in #1277, stacked on this PR. 
Each store now exposes a sizing helper its own constructor consumes, so the estimate just sums what the stores report.

partychen:
Before this change, `builder::quantizer` and `builder::tokio` were public modules, and downstream crates could use at least:

- `diskann_disk::build::builder::quantizer::BuildQuantizer`
- `BuildQuantizer::train`
- `diskann_disk::build::builder::tokio::create_runtime`

Making both modules private, together with reducing those items to `pub(super)`, causes existing downstream code using those paths to stop compiling. This is more than an internal module reorganization, although the PR is otherwise presented as a behavior-preserving structural refactor.

If these APIs are intentionally being removed, could the PR description call out the breaking API change? Otherwise, should the old public paths be preserved or deprecated for a compatibility window?

wuw92:
> Before this change, `builder::quantizer` and `builder::tokio` were public modules, and downstream crates could use at least:
> 
> * `diskann_disk::build::builder::quantizer::BuildQuantizer`
> * `BuildQuantizer::train`
> * `diskann_disk::build::builder::tokio::create_runtime`
> 
> Making both modules private, together with reducing those items to `pub(super)`, causes existing downstream code using those paths to stop compiling. This is more than an internal module reorganization, although the PR is otherwise presented as a behavior-preserving structural refactor.
> 
> If these APIs are intentionally being removed, could the PR description call out the breaking API change? Otherwise, should the old public paths be preserved or deprecated for a compatibility window?

Thanks for calling this out. The visibility reduction is intentional. `BuildQuantizer`, `BuildQuantizer::train`, and the builder-specific `create_runtime` helper are implementation details used to orchestrate `diskann-disk` builds; they are not intended as downstream extension points.
I checked the current internal and M365 sources and found no remaining uses of these APIs. I will update the PR description to explicitly call out the breaking visibility changes.

diskann-disk/src/build/builder/build.rs
```diff
@@ -3,47 +3,33 @@
...
 };
-use tokio::task::JoinSet;
-use tracing::{debug, info};
+use tracing::info;
 
 use crate::{
     build::builder::{
-        core::{determine_build_strategy, IndexBuildStrategy, MergedVamanaIndexBuilder},
-        inmem_builder::{new_inmem_index_builder, InmemIndexBuilder},
+        inmem_builder::build_inmem_index,
```
partychen, line 21: The name `build_inmem_index` is now used at two different abstraction levels. `DiskIndexBuilder::build_inmem_index` selects either the one-shot or merged strategy, while this imported helper builds and persists one complete Vamana graph (for either the full dataset or one shard). The outer operation is not necessarily in-memory, and the inner helper is specifically the one-shot leaf operation.

Could we distinguish these responsibilities explicitly - for example, `build_vamana_index` for the outer phase and `build_one_shot_vamana` or `build_vamana_graph` for the leaf helper? That would make the call graph understandable without relying on the historical `_mem.index` terminology.

diskann-disk/src/build/builder/mod.rs
```diff
@@ -5,11 +5,14 @@
 
 //! Disk index builders and related functionality.
 pub mod build;
-pub mod core;
-pub mod quantizer;
 
-pub mod inmem_builder;
-pub mod tokio;
+mod inmem_builder;
+mod merged_index;
```
partychen, line 10: With `core` removed, I expected the new module boundaries to expose the build model more directly, but these sibling names still use different concepts: `inmem_builder` describes memory residency/an implementation detail, while `merged_index` describes a build strategy or output. Their contents are not symmetric either: the former owns quantizer dispatch plus the complete graph-construction pipeline, while the latter owns the merged-build workflow.

Could we align the ownership and naming around the actual domain - Vamana construction with one-shot and merged strategies - for example through private `vamana::{one_shot, merged}` modules or an equivalent layout? The exact number of files is not important; the goal is for the module names to state which algorithm and responsibility they own.

diskann-disk/src/build/builder/build.rs
```diff
@@ -212,186 +198,42 @@ where
...
+    Merged,
 }
 
-#[cfg(debug_assertions)]
-/// Log statistics about the build process
-async fn log_build_stats<T: VectorRepr>(index: &Arc<dyn InmemIndexBuilder<T>>) -> ANNResult<()> {
-    debug!(
-        "Number of points reachable in the graph: {}",
-        index.count_reachable_nodes().await?
+pub(crate) fn determine_build_strategy<Data: GraphDataType>(
```
partychen, line 206: `IndexBuildStrategy` sounds like it controls the complete disk-index build, but it only selects how the Vamana graph is constructed. Its current placement also leaves the decision split across modules: `build.rs` owns the strategy, while the RAM estimate it requires remains in `merged_index.rs`, even though that estimate is not specific to the merged path.

Could the strategy and its estimate be named and owned together with the Vamana builders (for example, `VamanaBuildStrategy`)? That would leave `build.rs` focused on the top-level PQ, Vamana, and disk-layout orchestration.

partychen (commented):
I will close #1306 to avoid maintaining two competing refactors. My feedback here is not a request to adopt its exact layout. My main concern is that a structural refactor should make the domain model visible from the module boundaries: the disk build orchestrates PQ construction, Vamana construction, and disk layout, while Vamana construction has one-shot and merged strategies.

I would like the naming and ownership in this PR to reflect that model more directly. The enum-dispatch suggestion can be handled separately; the module boundaries, strategy ownership, and overloaded `build_inmem_index` terminology are the main changes I am requesting.

diskann-disk/src/build/builder/inmem_builder.rs
```diff
@@ -23,17 +30,23 @@ use diskann_providers::{
...
+use crate::error::{diskann_error, ErrorKind};
 
 use super::quantizer::BuildQuantizer;
 
 /// Builder facade for in memory index construction and persistence.
 ///
 /// Thread safety:
 /// Implementors must be `Send` and `Sync`. Methods can be called from many tasks.
-pub(super) trait InmemIndexBuilder<T: Sized>: Send + Sync {
+trait InmemIndexBuilder<T: Sized>: Send + Sync {
```
partychen, line 49: Now that this facade is private, it only erases the closed set of implementations selected by `BuildQuantizer` (FP, SQ, and PQ). That requires boxed futures and dynamic dispatch for operations such as every vector insertion, while also hiding the supported variants from the type structure.

Would an explicit enum over those variants make this pipeline easier to follow and avoid the per-operation boxing? I see this as a useful simplification, but it need not block the module and naming cleanup if it is better handled as a follow-up.

PR #1273, opened 2026-07-24
Make `Neighbor` less incorrect

hildebrandmw:
Our current `Neighbor` type has several inter-related issues:

1. It structurally requires `Eq` for the ID type, even though this is only needed when comparing equality. This makes it impossible to construct a `Neighbor` containing arbitrary external ID types and is the reason [`SearchOutputBuffer`](https://github.com/microsoft/DiskANN/blob/6caca35ddcfbacb2de38e57b7b53b7a7a95e1cf0/diskann/src/graph/search_output_buffer.rs#L20-L36) has to take tuples instead of `Neighbors`. This leads to [duplicated sorting logic](https://github.com/microsoft/DiskANN/blob/6caca35ddcfbacb2de38e57b7b53b7a7a95e1cf0/diskann-providers/src/model/graph/provider/async_/inmem/full_precision.rs#L397-L399) all over the code base.

    This structural requirement is preventing some aspects of ID handling cleanup.
3. Its members are public.
4. Its implementation of [`Ord`](https://github.com/microsoft/DiskANN/blob/6caca35ddcfbacb2de38e57b7b53b7a7a95e1cf0/diskann/src/neighbor/mod.rs#L80-L95) is [not consistent](https://doc.rust-lang.org/std/cmp/trait.Ord.html) with its [`PartialEq`](https://github.com/microsoft/DiskANN/blob/6caca35ddcfbacb2de38e57b7b53b7a7a95e1cf0/diskann/src/neighbor/mod.rs#L66-L74) method.
5. Its implementation of `Ord` is [explicitly called out](https://doc.rust-lang.org/std/cmp/trait.Ord.html#examples-of-incorrect-ord-implementations) as invalid.

This PR attempts to address many of these issues. Mainly, it turns `Neighbor` into a pretty simple tuple-like type with the various ordering methods implemented as free functions in `neighbors::ord`. Algorithms that need to sort `Neighbors` can pick and choose the correct `ord::*` methods needed, which in turn are quite `greppable`. These `ord::*` methods do not quite patch the issue of NaNs showing up in as distances, but at least we have a better way of documenting our requirements.

### DIFF
--- a/diskann/src/neighbor/mod.rs
+++ b/diskann/src/neighbor/mod.rs
@@ -4,7 +4,7 @@
  */
 
 // Imports
-use std::{cmp::Ordering, fmt::Debug};
+use std::fmt::Debug;
 
 use crate::graph::{SearchOutputBuffer, search_output_buffer};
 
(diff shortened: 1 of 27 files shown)


diskann/src/neighbor/mod.rs
```diff
@@ -23,114 +23,197 @@ pub use diverse_priority_queue::{
...
-    pub distance: f32,
+/// A pairing of opaque ID with a distance.
+///
+/// For all intents and purposes, this is a simple aggregate with minimal additional semantics.
+///
+/// # Sorting
+///
+/// An exceedingly common algorithmic operation is to sort [`Neighbor`]s according to some
+/// protocol. Usually, this is by distance (from smallest to largest) ignoring the ID entirely,
+/// but certain situations call for different orderings. This raises two problems:
```
harsha-simhadri, line 32: could we capture these other orderings currently used here. Would nice to have a reference

harsha-simhadri (approved):

diskann/src/neighbor/mod.rs
hildebrandmw: The idea behind this PR is to move the orderings into the `ord` submodule, which then serves as kind of an "automatic" collection of the various orderings. As a bonus, the functions names like `fast_distance` are pretty easy to `grep`, making it easier to locate where these are being used. Is the ask here to also record *when* other recordings are used?

JordanMaples (approved):

arkrishn94 (approved):
Thanks Mark, lgtm

JordanMaples (approved):

diskann/src/neighbor/mod.rs
```diff
@@ -23,114 +23,197 @@ pub use diverse_priority_queue::{
...
+    /// let z = Neighbor::new(12, f32::NAN);
+    ///
+    /// // The following line can return any `Ordering`.
+    /// // neighbor::ord::fast_distance(&z, &z);
+    /// ```
+    pub fn fast_distance<I>(x: &Neighbor<I>, y: &Neighbor<I>) -> std::cmp::Ordering {
+        x.distance()
+            .partial_cmp(&y.distance())
             .unwrap_or(std::cmp::Ordering::Equal)
     }
```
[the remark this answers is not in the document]
hildebrandmw: No - it's explicitly documented a not working on NaNs.

arkrishn94 (approved):

PR #1276, opened 2026-07-27
Converge PQ pivot loading

### LINKED ISSUE #1015: Converge load_existing_pivot_data() and load_pq_pivots_bin()
It seems that we have two functions doing the same thing - loading pq pivots:
1. load_existing_pivot_data
2. load_pq_pivots_bin

We should converge them (in a separate PR). Let me create a work item for that.

_Originally posted by @arrayka in https://github.com/microsoft/DiskANN/pull/1010#discussion_r3185433718_

xinyuwen2:

#### Reference Issues/PRs
Fixes https://github.com/microsoft/DiskANN/issues/1015.

This PR consolidates PQ pivot loading around `PQStorage::load_pivots`, which uses `PQStorage::pivot_data_path` as the single source of truth and returns a fully validated `BasicTable`.

- Route existing PQ pivot loading through a shared parser for offsets, pivots, centroid data, and chunk offsets.
- Fold legacy centroid data into the pivot matrix before constructing `BasicTable`, so the loaded table is fully formed at construction time.
- Use `BasicTable` / `ChunkOffsetsBase` to validate structural table invariants in the quantization layer.
- Remove the raw pivot-data loading API and the `load_pq_pivots_bin` compatibility wrapper.
- Remove the explicit pivot path parameter from pivot loading APIs and remove `get_pivot_data_path`.
- Add `TryFrom<BasicTable> for FixedChunkPQTable` so callers validate fixed-table center-count bounds when converting loaded tables.
- Preserve the previous fixed PQ table loading behavior that accepts up to `NUM_PQ_CENTROIDS` centers while rejecting tables with more centers.
- Move configuration-specific compatibility checks, such as expected chunk count and dimensionality, to the call sites that own those expectations.
- Update PQ construction, PQ generation, disk index loading, and memory quant vector providers to use `load_pivots`.

#### Any other comments?
Validated with:
- `cargo test -p diskann-providers pq_storage`
- `cargo test -p diskann-providers pq_construction`
- `cargo test -p diskann-providers fixed_chunk_pq_table_test`
- `cargo test -p diskann-providers fast_memory_quant_vector_provider::tests`
- `cargo test -p diskann-providers memory_quant_vector_provider::tests`


partychen (approved):

diskann-providers/src/storage/pq_storage.rs
```diff
@@ -301,51 +318,85 @@ impl PQStorage {
...
-            accum_row_inplace(pivots.as_mut_view(), centroids.as_slice())
+        let parts = PivotFileParts {
+            pivots,
+            centroid,
+            chunk_offsets,
+        };
+        if chunk_offset_validation == ChunkOffsetValidation::ShapeAndBounds
+            && (parts.chunk_offsets.nrows() < 2
+                || parts.chunk_offsets[(0, 0)] != 0
+                || parts.chunk_offsets[(parts.chunk_offsets.nrows() - 1, 0)] != parts.dim())
```
arkrishn94: It looks like this case can only be reached when `num_pq_chunks == 0`. Does that ever happen? Should it be allowed? @hildebrandmw, @wuw92,

diskann-providers/src/storage/pq_storage.rs
```diff
@@ -301,51 +318,85 @@ impl PQStorage {
...
+
+        let pivots = read_bin_from::<f32>(&mut reader, file_offset_data[(0, 0)])?;
+        if let Some(num_centers) = expected_num_centers {
+            if pivots.nrows() != num_centers {
+                return Err(ANNError::log_pq_error(format_args!(
+                    "Error reading pq_pivots file {}. file_num_centers = {}, but expecting {} centers.",
+                    pq_pivots,
+                    pivots.nrows(),
+                    num_centers
+                )));
```
arkrishn94: We're moving away from this pattern - https://github.com/microsoft/DiskANN/pull/1282. How can we get ahead of this here? Or maybe we don't need to if 1282 is checked in first :)

diskann-providers/src/storage/pq_storage.rs
```diff
@@ -289,7 +264,49 @@ impl PQStorage {
...
+
+    fn load_pivot_file_parts<Storage: StorageReadProvider>(
+        &self,
+        pq_pivots: &str,
+        expected_num_pq_chunks: Option<usize>,
+        expected_num_centers: Option<usize>,
+        expected_dim: Option<usize>,
+        chunk_offset_validation: ChunkOffsetValidation,
+        storage_provider: &Storage,
+    ) -> ANNResult<PivotFileParts> {
```
arkrishn94: Can we add some documentation to this function to make sure we understand how the arguments here are used?

nit: It feels a bit hacky to me to case on these values when performing the checks in this function. It looks like the last two are used to validate pivots and the first to validate the chunks_offset. Maybe you can make this check more explicit?

diskann-providers/src/storage/pq_storage.rs
```diff
@@ -289,7 +264,49 @@ impl PQStorage {
...
+        if centroid.as_slice().iter().any(|c| *c != 0.0) {
+            accum_row_inplace(pivots.as_mut_view(), centroid.as_slice())
+        }
+
+        FixedChunkPQTable::new(
+            pivots.ncols(),
+            pivots.into_inner(),
+            chunk_offsets.into_inner(),
+        )
+    }
```
hildebrandmw: Thank you for looking at this! I have a higher-level comment that I think we may want to think about instead.

The current divergent loading paths are a symptom of larger architectural issues and while unifying the load paths is a step in the right direction, it is treating a symptom rather than the cause. Much of the complexity here comes from the (optional) inline invariant checks. However, these checks are not exhaustive. For example, the offsets also need to be strictly monotonically increasing, which is not covered by the checks here. However, duplicating these checks is not strictly necessary as [`BasicTable`](https://github.com/microsoft/DiskANN/blob/3e7c1f37a500d35e4cc2bd8576170640118b407c/diskann-quantization/src/product/tables/basic.rs#L65-L82) and [`ChunkOffsets`](https://github.com/microsoft/DiskANN/blob/3e7c1f37a500d35e4cc2bd8576170640118b407c/diskann-quantization/src/views.rs#L28-L39) already assert the exact invariants that are needed. Duplication runs the risk of things getting out-of-sync.

This loader has two basic return paths:

* `load_pq_pivots_bin`: This returns a `FixedChunkPQTable`, which is really a [`BasicTable` in disguise](https://github.com/microsoft/DiskANN/blob/3e7c1f37a500d35e4cc2bd8576170640118b407c/diskann-providers/src/model/pq/fixed_chunk_pq_table.rs#L24-L27) with a [terrible constructor](https://github.com/microsoft/DiskANN/blob/3e7c1f37a500d35e4cc2bd8576170640118b407c/diskann-providers/src/model/pq/fixed_chunk_pq_table.rs#L133-L142). Anyways, since `load_pq_pivots_bin` already returns a `FixedChunkPQTable`, it's already performing the necessary invariant checks. Errors returned on `FixedChunkPQTable` construction can be chained with a [`context`](https://github.com/microsoft/DiskANN/blob/3e7c1f37a500d35e4cc2bd8576170640118b407c/diskann/src/error/ann_error.rs#L239-L247) to provide a nice message that construction failed during file loading.
* `load_existing_pivot_data`: This is where a lot of the complexity from this localized change comes from as this is the path the requests more pedantic checking due to the looser return type. However, there are just two call-sites that actually use this method:
  - [This one](https://github.com/microsoft/DiskANN/blob/3e7c1f37a500d35e4cc2bd8576170640118b407c/diskann-providers/src/model/pq/pq_construction.rs#L354-L359), which also goes to the trouble of re-accumulating the centroid before calling [`TransposedTable::new`](https://github.com/microsoft/DiskANN/blob/3e7c1f37a500d35e4cc2bd8576170640118b407c/diskann-quantization/src/product/tables/transposed/table.rs#L73-L76). Note that the arguments to the transposed table can come directly from the [`BasicTable`](https://github.com/microsoft/DiskANN/blob/3e7c1f37a500d35e4cc2bd8576170640118b407c/diskann-quantization/src/product/tables/basic.rs#L84-L92). So this call site can be updated to take method returning `BasicTable`.
  -  [This test function](https://github.com/microsoft/DiskANN/blob/3e7c1f37a500d35e4cc2bd8576170640118b407c/diskann-providers/src/model/pq/pq_construction.rs#L940-L947). Again, though, this can instead use a `BasicTable` instead if we update [`generate_pq_data_from_pivots_from_membuf`](https://github.com/microsoft/DiskANN/blob/3e7c1f37a500d35e4cc2bd8576170640118b407c/diskann-providers/src/model/pq/pq_construction.rs#L489-L518) to take a `&BasicTable` instead. Though if we do that - there's no need for this helper method at all because we can simply use `table.compress_into`! All the callers of the `_membuf` APIs already have a `FixedChunkPQTable`, so getting the inner `BasicTable` should be a breeze.

To recap, if this loading function only has a single pivot loading method returning a `BasicTable`, we get most of the invariant checking for free (outside of the file format verification checks) and a bunch of PQ related code can be simplified because it basically already operates on a `BasicTable`. We can add a new trivial constructor for `FixedChunkPQTable` taking a `BasicTable` and the net effect will probably be a healthy reduction in code.

diskann-providers/src/storage/pq_storage.rs
xinyuwen2: Happy to see it has been merged. I have replaced it with the new pattern.

diskann-providers/src/storage/pq_storage.rs
xinyuwen2: Updated the structure per the later review feedback. The helper now keeps only file-format checks local to the file parser, while structural PQ table invariants are delegated to `ChunkOffsetsBase` and `BasicTable`. This removed the extra `ChunkOffsetValidation` branching and makes the expected checks more direct.

diskann-providers/src/storage/pq_storage.rs
xinyuwen2: Thank you so much for pointing out this direction!. I have updated to make it cleaner.

diskann-providers/src/storage/pq_storage.rs
xinyuwen2: I see the code [here](https://github.com/microsoft/DiskANN/blob/c1f8c7d9cb2c660cbaee7269de3cd70c6a6c36a4/diskann-disk/src/storage/disk_index_reader.rs#L34), it pass 0 then inter num_pq_chunks from pivot file chunk_offsets. It's not allowed when building/training, but it's allowed when loading existing pivot file. So I think we can still preserve the old behavior and I added `load_pq_pivots_bin_infer_chunks` as the explicit version for new callers.

diskann-providers/src/storage/pq_storage.rs
hildebrandmw: Thanks. We're part of the way there. A good rule of thumb for suggestions like these is that ideally we're looking for a net reduction in complexity, not an increase. The most recent change gets part of the way there, but now there are four public methods for loading a pivot table. Here is a concrete suggestion:

* There should only be **one** method for loading pivots (call it `load_pivots`) and it should return a `BasicTable`. If we do that, almost all of the extra scaffolding in `pq_storage.rs` introduced in this PR (including `PivotFileParts`) goes away. The code withing `load_pivots` can verify the layout of the file, but doesn't need to worry about the consistency of the final `BasicTable` because that's what `BasicTable::new` does.
* The call sites for `load_existing_pivot_data` get simpler (there's one more in [`diskann-disk`](https://github.com/microsoft/DiskANN/blob/74c060bb0a55a5578c780a435f3ed0ee37a0aeab/diskann-disk/src/storage/quant/pq/pq_generation.rs#L118-L141) that I missed) because they *all* do the centroid accumulation work + validation that `load_pivots` would already do.
* Completely get rid of `generate_pq_data_from_pivots_from_membuf` and do not introduce another method that just calls `{BasicTable/FixedChunkPQTable}::compress_into`. Instead, prefer to call `compress_into` directly. (Note that either can be used for this purpose).

  The users of this `_membuf` all already have a `FixedChunkPQTable`, so again that will be a net cleanup: instead of decomposing and recomposing a `BasicTable`/`FixedChunkPQTable` ... we can reuse the one we already have :smile:.

Appreciate your help here!

xinyuwen2: Thank you! I have updated with below changes.
- For point 1, I removed all extra PR-added methods, now only have `load_pivots`. And I keep `load_existing_pivot_data` and `load_pq_pivots_bin` for backwards compatibility.
- For point 2, the call sites that were using `load_existing_pivot_data` now use `load_pivots` directly to avoid duplicate 'centroid accumulation work + validation'.
- For point 3, I removed the extra helper that only wrapped compress_into, and the compatibility wrappers now call compress_into directly.

diskann-providers/src/storage/pq_storage.rs
```diff
@@ -179,66 +180,51 @@ impl PQStorage {
...
         ))
     }
 
+    /// Load the effective PQ pivot table from a pivot file.
+    ///
+    /// If `expected_num_pq_chunks` is `None`, the chunk count is inferred from the
+    /// file. The loader verifies the pivot file layout and folds any stored legacy
+    /// centroid into the pivots. `BasicTable::new` validates the resulting table
+    /// invariants, including chunk-offset bounds and monotonicity.
+    pub fn load_pivots<Storage: StorageReadProvider>(
```
wuw92, line 164: Could we use `PQStorage::pivot_data_path` as the single source of truth here? Every in-tree caller passes the same path already stored in `PQStorage`, while accepting another path permits contradictory object state.

Then we can remove the path parameter from both `load_pivots` and `load_pq_pivots_bin`,  and remove the now-unnecessary `get_pivot_data_path()` method.

diskann-providers/src/storage/pq_storage.rs
wuw92: Consider removing `load_existing_pivot_data`? It no longer has any in-tree production callers; the remaining test usages could be migrated to `load_pivots`.

diskann-providers/src/storage/pq_storage.rs
```diff
@@ -179,66 +180,51 @@ impl PQStorage {
...
+        &self,
+        pq_pivots: &str,
+        expected_num_pq_chunks: Option<usize>,
```
wuw92: Could we make the validation boundary explicit here?

- The loader validates internal consistency: file layout, centroid/pivot dimensions, offset monotonicity and bounds, and `BasicTable` invariants.
- The caller validates configuration compatibility: whether `table.nchunks()` matches the expected build or search configuration.

This would let `load_pivots` infer the chunk count and leave expected chunk-count validation to its callers.

diskann-providers/src/storage/pq_storage.rs
wuw92: Could we keep `load_pivots` as the single loading API and model this as a type conversion instead?

Since wrapping a validated `BasicTable` in `FixedChunkPQTable` is infallible, an `impl From<BasicTable> for FixedChunkPQTable` would let callers use:

```rust
let table: FixedChunkPQTable = pq_storage.load_pivots(storage_provider)?.into();
```

Then `load_pq_pivots_bin` could either be removed or retained only as a deprecated compatibility wrapper.

wuw92 (commented):
Thanks for working through this convergence. 
One point I am unsure about is the compatibility boundary. This PR adds or retains several `pub` APIs primarily to support existing in-workspace patterns, Although these are technically available to downstream users, they appear intended mainly for other DiskANN crates rather than as a stable external API.

Should we preserve these conservatively as public compatibility surfaces, or can we treat them as workspace-internal implementation details and simplify them directly? If compatibility is required, deprecating the legacy APIs explicitly may be clearer. Otherwise, this PR could leave `load_pivots` as the single loader, remove the raw-data compatibility path.  @hildebrandmw @arkrishn94

hildebrandmw:
> Thanks for working through this convergence. One point I am unsure about is the compatibility boundary. This PR adds or retains several `pub` APIs primarily to support existing in-workspace patterns, Although these are technically available to downstream users, they appear intended mainly for other DiskANN crates rather than as a stable external API.
> 
> Should we preserve these conservatively as public compatibility surfaces, or can we treat them as workspace-internal implementation details and simplify them directly? If compatibility is required, deprecating the legacy APIs explicitly may be clearer. Otherwise, this PR could leave `load_pivots` as the single loader, remove the raw-data compatibility path. @hildebrandmw @arkrishn94

Thanks @wuw92 - while we're getting closer, DiskANN is not quite at a place (in most parts of the code) where I think trying to provide exact API backwards compatibility is realistic or desirable. In situations like this where implementations and API surface get worse as a result, I'd rather take the "L" on compatibility and do the full simplification. Especially since there is a clear path to go from before this change to after this change.

In other words, I fully support removing the older methods.

diskann-quantization/src/product/tables/basic.rs
```diff
@@ -86,6 +86,14 @@ where
         self.pivots.as_view()
     }
 
+    /// Return a mutable view over the pivot table.
+    pub fn view_pivots_mut(&mut self) -> MutMatrixView<'_, f32>
+    where
+        T: MutDenseData,
+    {
+        self.pivots.as_mut_view()
+    }
```
hildebrandmw: Please remove this method. Types like `BasicTable` should be fully formed upon construction and not require "patching up" after the fact.

diskann-providers/src/storage/pq_storage.rs
xinyuwen2: Removed the explicit pivot path parameter from `load_pivots`, also removed `get_pivot_data_path()`.

diskann-providers/src/storage/pq_storage.rs
xinyuwen2: Removed `load_existing_pivot_data` and migrated the remaining tests to use `load_pivots` directly.

diskann-providers/src/storage/pq_storage.rs
xinyuwen2: Thank you! `load_pivots` now only validates the pivot file's internal consistency and constructs a valid `BasicTable`, and expected chunk-count validation are now performed by the callers.

diskann-providers/src/storage/pq_storage.rs
xinyuwen2: Removed `load_pq_pivots_bin`. Callers now use `load_pivots(...)?.into()` when they need a `FixedChunkPQTable`, with any expected chunk-count validation handled at the call site.

diskann-quantization/src/product/tables/basic.rs
xinyuwen2: Thank you! Removed.

diskann-providers/src/model/graph/provider/async_/fast_memory_quant_vector_provider.rs
```diff
@@ -245,7 +245,15 @@ impl FastMemoryQuantVectorProviderAsync {
...
-            let table = pq_storage.load_pq_pivots_bin(pivots, pq_bytes, provider)?;
+            let table = pq_storage.load_pivots(provider)?;
+            if table.nchunks() != pq_bytes {
+                return Err(ANNError::message(format!(
+                    "PQ pivot table mismatch: file has {} chunks but expected {} chunks.",
+                    table.nchunks(),
+                    pq_bytes
+                )));
+            }
+            let table: FixedChunkPQTable = table.into();
```
wuw92: nit: it can be simplified to `Ok(Self::new(metric, num_points, table.into()))`

diskann-quantization/src/product/tables/basic.rs
```diff
@@ -105,6 +105,11 @@ where
     pub fn dim(&self) -> usize {
         self.pivots.ncols()
     }
+
+    /// Consume this table and return the underlying pivots and offsets.
+    pub fn into_parts(self) -> (MatrixBase<T>, ChunkOffsetsBase<U>) {
```
wuw92: Seems not necessary, remove it.

diskann-quantization/src/views.rs
```diff
@@ -211,6 +211,11 @@ where
     pub fn as_slice(&self) -> &[usize] {
         self.offsets.as_slice()
     }
+
+    /// Consume the offsets, returning the inner representation.
+    pub fn into_inner(self) -> T {
```
wuw92: Seems not necessary, remove it.

diskann-providers/src/storage/pq_storage.rs
```diff
@@ -274,76 +210,117 @@ impl PQStorage {
         Ok(data)
     }
 
-    /// Load pre-trained pivot table
-    pub fn load_pq_pivots_bin<Storage: StorageReadProvider>(
+    fn read_pivot_file<Storage: StorageReadProvider>(
         &self,
         pq_pivots: &str,
-        num_pq_chunks: usize,
+        expected_num_centers: Option<usize>,
```
wuw92: Could we remove `expected_num_centers` and `expected_dim` from `read_pivot_file`? The only caller now always passes `None` for both

wuw92 (approved):
Thanks for the update! left a few small comments

diskann-providers/src/model/pq/pq_construction.rs
```diff
@@ -546,17 +542,20 @@ pub fn generate_pq_data_from_pivots_from_membuf_batch<T: Copy + Sync + Into<f32>
...
-        .try_for_each_in_pool(pool, |(pq_slice, vector_slice)| {
-            generate_pq_data_from_pivots_from_membuf(
-                vector_slice,
-                pivot_data,
-                parameters.num_centers(),
-                offsets,
-                pq_slice,
-            )
+        .try_for_each_in_pool(pool, |(pq_slice, vector)| {
+            let data = vector.iter().map(|x| (*x).into()).collect::<Vec<f32>>();
```
hildebrandmw: Use [`VectorRepr::as_f32`](https://github.com/microsoft/DiskANN/blob/bea135b96d1d2fe689bc2877b06bbff941c6c906/diskann/src/utils/vector_repr.rs#L94-L103) to avoid allocating in the common case.

diskann-providers/src/model/pq/pq_construction.rs
```diff
@@ -952,9 +950,9 @@ mod pq_test {
             .for_each_in_pool(pool.as_ref(), |(i, membuf_slice)| {
                 generate_pq_data_from_pivots_from_membuf(
```
hildebrandmw: We can fully get rid of `generate_pq_data_from_pivots_from_membuf`, right? The only callers are now in tests?

diskann-providers/src/storage/pq_storage.rs
```diff
@@ -488,36 +461,212 @@ mod pq_storage_tests {
...
-            "centroid should be all zeros when written with None"
+            PQStorage::new(wrong_centroid_path, PQ_COMPRESSED_PATH, None)
+                .load_pivots(&storage_provider)
+                .is_err()
         );
-        assert_eq!(loaded_offsets, chunk_offsets);
+    }
+
+    #[test]
+    fn load_pivot_data_rejects_invalid_chunk_bounds() {
```
hildebrandmw: Can you please audit the newly added tests and just retain those that test logic within the loader? Tests that reaffirm the behavior of BasicTable are

diskann-providers/src/model/pq/pq_construction.rs
xinyuwen2: I updated this to use `VectorRepr::as_f32`, which avoids the per-vector `Vec<f32>` allocation in the common `f32` case. This is fine for the in-tree callers, but it does tighten the public generic bound from `T: Copy + Sync + Into<f32>` to `T: VectorRepr + Sync`. Is it acceptable?

diskann-providers/src/model/pq/pq_construction.rs
xinyuwen2: yes removed.

diskann-providers/src/storage/pq_storage.rs
xinyuwen2: Removed new tests for BasicTable.

diskann-providers/src/model/pq/pq_construction.rs
xinyuwen2: Kindly ping. Just wanted to follow up on this PR when you have a chance. Thanks! @hildebrandmw

hildebrandmw: Apologies for the delays, I have been dealing with an injury this week that has made it difficult to work. I am slowly getting back up to speed.

Yes, strengthening the bound to `VectorRepr` is acceptable. The `Into<f32> bounds predate the `VectorRepr` trait and while not perfect, `VectorRepr` does allow for some things to be more efficient like bulk conversion.

hildebrandmw (approved):
Awesome thanks for your patience throughout this process. The PR looks great now!

diskann-providers/src/model/graph/provider/async_/memory_quant_vector_provider.rs
```diff
@@ -178,8 +178,15 @@ impl MemoryQuantVectorProviderAsync {
             // We can use that information to load the pivots, then finish the rest
             // of initialization.
             let pq_storage = storage::PQStorage::new(pivots, data, None);
-            let table = pq_storage.load_pq_pivots_bin(pivots, pq_bytes, provider)?;
-            Ok(Self::new(metric, num_points, table))
+            let table = pq_storage.load_pivots(provider)?;
+            if table.nchunks() != pq_bytes {
+                return Err(ANNError::message(format!(
+                    "PQ pivot table mismatch: file has {} chunks but expected {} chunks.",
```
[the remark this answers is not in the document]
xinyuwen2: Resolved.

diskann-providers/src/model/graph/provider/async_/fast_memory_quant_vector_provider.rs
```diff
@@ -245,8 +245,15 @@ impl FastMemoryQuantVectorProviderAsync {
             // We can use that information to load the pivots, then finish the rest
             // of initialization.
             let pq_storage = storage::PQStorage::new(pivots, data, None);
-            let table = pq_storage.load_pq_pivots_bin(pivots, pq_bytes, provider)?;
-            Ok(Self::new(metric, num_points, table))
+            let table = pq_storage.load_pivots(provider)?;
+            if table.nchunks() != pq_bytes {
+                return Err(ANNError::message(format!(
+                    "PQ pivot table mismatch: file has {} chunks but expected {} chunks.",
```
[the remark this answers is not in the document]
xinyuwen2: Resolved.

diskann-disk/src/storage/disk_index_reader.rs
```diff
@@ -29,11 +31,7 @@ impl DiskIndexReader {
         storage_provider: &Storage,
     ) -> ANNResult<Self> {
         let pq_storage = PQStorage::new(&pq_pivot_path, &pq_compressed_data_path, None);
-        let pq_pivot_table = pq_storage.load_pq_pivots_bin::<Storage>(
-            &pq_pivot_path,
-            0, // Use 0 to infer num_pq_chunks from the file
-            storage_provider,
-        )?;
+        let pq_pivot_table: FixedChunkPQTable = pq_storage.load_pivots(storage_provider)?.into();
```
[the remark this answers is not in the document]
xinyuwen2: Resolved.

partychen (approved):

PR #1277, opened 2026-07-27
Make build RAM estimation store-owned

wuw92:
# Make build RAM estimation store-owned

Stacked on #1271 — review against `wuw92/refactor-disk-build-structure-main`.

## Why

Addresses https://github.com/microsoft/DiskANN/pull/1271#discussion_r3653942378.

The build RAM estimate was a hand-written byte formula with no link to the stores it described, so layout changes silently invalidated it. 

## What

Make each store answer for its own size, so an allocation and its estimate cannot diverge. Every estimate mirrors one constructor at its own level and knows only what that constructor directly builds:

```
DefaultProviderParameters::estimated_bytes        mirrors  new_empty
  ├─ CU/CV/CD: EstimateBytes                      mirrors  CreateVectorStore::create
  │    ├─ CreateFullPrecision<T> ─┐
  │    ├─ FixedChunkPQTable       ├─ <Store>::estimate_bytes   mirrors  <Store>::new
  │    ├─ WithBits<NBITS>       ──┘
  │    └─ NoStore / NoDeletes → 0
  └─ SimpleNeighborProviderAsync::estimate_bytes   mirrors  SimpleNeighborProviderAsync::new
       └─ AlignedMemoryVectorStore::estimate_bytes mirrors  with_capacity
```

The estimator moves beside the in-memory builder.
Scoped to the stores disk build uses; small or fixed overheads such as PQ pivots stay uncounted.

Also removes the `graph_slack_factor` parameter from `SimpleNeighborProviderAsync`: slack is already folded into `max_degree`, so every call site passed `1.0`.

### DIFF
--- a/diskann-disk/src/build/builder/build.rs
+++ b/diskann-disk/src/build/builder/build.rs
@@ -4,7 +4,7 @@
  */
 
 //! Disk index builder implementation.
-use std::{marker::PhantomData, mem};
+use std::marker::PhantomData;
 
 use crate::data_model::GraphDataType;
 use diskann::{utils::VectorRepr, ANNResult};
@@ -18,8 +18,8 @@ use tracing::info;
 
 use crate::{
     build::builder::{
-        inmem_builder::build_inmem_index,
-        merged_index::{estimate_build_index_ram_usage, MergedVamanaIndexBuilder},
+        inmem_builder::{build_inmem_index, estimate_build_index_ram_usage},
+        merged_index::MergedVamanaIndexBuilder,
         quantizer::BuildQuantizer,
         tokio::create_runtime,
     },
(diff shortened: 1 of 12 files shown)


hildebrandmw (commented):
I don't like the direction taken in this PR. My main issue is that it worsens the already haggard initialization story of `diskann-providers`. The whole precursor stuff really needs to go away, and mimicking it with size estimation feels like doubling down instead of heading in the right direction.

While I agree in spirit, I don't think this is something that should be done while we're trying to deprecate `diskann-providers`. It is a net addition of code, increases complexity, does not cover the interface exposed by the inmem provider (e.g., `TableBasedDeletes`), and requires somewhat dubious definitions like `EstimateBytes for FixedChunkPQTable`.

harsha-simhadri:
Lets pause on this until providers redesign is complete. @wuw92 sorry for the randomization

harsha-simhadri:
@wuw92, lets file using in-mem2's size estimators as a future work item and proceed with reviewing your previous PR.

wuw92:
Thanks for the review. I'm going to close this PR, it makes sense to reuse in-mem2's size estimator in disk path.

PR #1278, opened 2026-07-28
Remove the old log* style constructors

xinyuwen2:
- [x] Does this PR have a descriptive title that could go in our release notes?
- [ ] Does this PR add any new dependencies?
- [ ] Does this PR modify any existing APIs?
- [ ] Is the change to the API backwards compatible?
- [ ] Should this result in any changes to our documentation, either updating existing docs or adding new ones?

#### Reference Issues/PRs
Fixes https://github.com/microsoft/DiskANN/issues/1003.

#### What does this implement/fix? Briefly explain your changes.
This change removes the legacy ANNError::log_* constructors and migrates all call sites to the newer ANNError::new(...) or ANNError::from_display(...) paths.
The main goal is to avoid the old eager string-formatting style and preserve more structured error handling where possible. Typed errors are now passed through ANNError::new(ANNErrorKind::..., err), while display-only legacy paths use ANNError::from_display(...) as a transitional wrapper. The legacy API documentation and direct tests for the removed log_* constructors were also cleaned up.

PR #1282, opened 2026-07-28
Slim down `ANNError`

hildebrandmw:
Remove `ANNErrorKind` from `ANNError`, removing all the old `log_` constructors and slimming the constructors to `new`, `message`, and `context.

Since `diskann-disk` leans pretty heavily to using the kinds in its tests, a new `diskann_disk::error::Error` type is introduces along with a `diskann_error!` macro to largely preserve test structure and error kind tagging. See the review notes for more information.

# Why
The error kind enum is largely unused at this point, is not applied consistently, and complicates the API for `ANNError`. It's time to remove it and bring the size of `ANNError` down to 8-bytes.

# Review Order

**`diskann`**
- `error/ann_error.rs`: Removal of the `ANNErrorKind` enum and the `log_*` style constructors. As a bonus, I added a `convert_error!` macro (happy to bike-shed) to stamp out simple conversion cases while preserving `track_caller`.

  The rest of the changes in `diskann` are removing the `ANNErrorKind` and switching over to the `convert_error!` macro.

**`diskann-tools`/`diskann-providers`/`diskann-garnet`/`diskann-label-filter`/`diskann-bftree`/`diskann-benchmark-core`/`diskann-benchmark`**
These mainly consist of updating the constructors and conversions. Note that where possible (and easily achievable), I use the [moving form of `lazy_format!`](https://github.com/microsoft/DiskANN/blob/e197b434912f7b13c64f1d89785f58ba7d5de22d/diskann-utils/src/lazystring.rs#L56-L74) to avoid eager string formatting.

**`diskann-disk`
This is the perhaps a more controversial change. `diskann-disk`'s tests lean heavily into the error kind and `log*` style constructors. I tried to preserve as much of the original behavior as possible by introducing an error type `diskann_disk::error::Error` consisting of a string-message an a `diskann_disk::error::ErrorKind` enum. A macro `diskann_disk::error::diskann_error!` helps embed this new `Error` inside of an `ANNError`. While this goes against the ethos of "don't have crate level errors", here is my rationale:

* By using the `log_*` constructors, `diskann_disk`'s error types are already fundamentally tagged string messages.
* Introducing an `Error` type means we can still check the `Kind` enum by downcasting the resulting `ANNError` down to the original `Error`.
* Since the existing `diskann_disk` code is already doing eager string formatting/allocating at its error construction callsites, this isn't introduction a regression. Indeed, I've added some tricks to avoid allocating when constructing `diskann_disk::error::Error` from string literals, which is fairly common.
* Moving the kind tag to `diskann_disk` allows the disk index to add error types without disrupting  `ANNError`.

# Alternatives
* This is an alternative to #1278 that I was unfortunately working on in parallel.
* One functionality we lost with this is the ability to recover the inner error types from the tag.  Theoretically, the tag type could be used by third-party crates to guide the proper type for a downcast. In practice, no one is doing this.

### DIFF
--- a/diskann-benchmark-core/src/build/graph/multi.rs
+++ b/diskann-benchmark-core/src/build/graph/multi.rs
@@ -6,7 +6,7 @@
 use std::{ops::Range, sync::Arc};
 
 use diskann::{
-    ANNError, ANNErrorKind, ANNResult,
+    ANNError, ANNResult,
     graph::{self, glue},
     provider,
 };
@@ -96,7 +96,6 @@ where
                 }
 
                 ANNError::message(
-                    ANNErrorKind::Opaque,
                     OutOfBounds {
                         max: self.data.nrows(),
                         start: range.start,
--- a/diskann-benchmark-core/src/build/graph/multi.rs
+++ b/diskann-benchmark-core/src/build/graph/multi.rs
@@ -95,13 +95,11 @@ where
                     }
                 }
 
-                ANNError::message(
-                    OutOfBounds {
-                        max: self.data.nrows(),
-                        start: range.start,
-                        end: range.end,
-                    },
-                )
+                ANNError::message(OutOfBounds {
+                    max: self.data.nrows(),
+                    start: range.start,
+                    end: range.end,
+                })
             })?
             .to_owned();
 
--- a/diskann-benchmark-core/src/build/ids.rs
+++ b/diskann-benchmark-core/src/build/ids.rs
@@ -7,8 +7,8 @@
 
 use std::marker::PhantomData;
 
-use diskann::{ANNError, ANNErrorKind, ANNResult};
-use diskann_utils::future::AsyncFriendly;
+use diskann::{ANNError, ANNResult};
+use diskann_utils::{lazy_format, future::AsyncFriendly};
 
 /// Convert an implicit data index to an external ID.
 ///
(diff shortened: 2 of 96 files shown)


harsha-simhadri (approved):

partychen:
how about impl_ann_error_from!? `impl_ann_error_from!` is more accurate because the macro generates a `From<T> for ANNError` implementation rather than converting an error value.

partychen:
Could we document the intended usage pattern for `ANNError`, perhaps in the Error Handling section of `agents.md` and the `ANNError` rustdoc? In particular, it would be helpful to clarify when to use `new`, `message`, and the conversion macro, and that `diskann_error!` is a compatibility mechanism for `diskann-disk` rather than a pattern new code should copy. Without this guidance, the transitional `ErrorKind` wrapper may become the de facto pattern.

hildebrandmw:
> Could we document the intended usage pattern for `ANNError`, perhaps in the Error Handling section of `agents.md` and the `ANNError` rustdoc? In particular, it would be helpful to clarify when to use `new`, `message`, and the conversion macro, and that `diskann_error!` is a compatibility mechanism for `diskann-disk` rather than a pattern new code should copy. Without this guidance, the transitional `ErrorKind` wrapper may become the de facto pattern.

Good points. I revamped the documentation for `ANNError` to describe its usage, hopefully also disambiguating when `new` vs `message` should be used (tldr: use `new` when you have an `Error` because it faithfully follows the error source - otherwise use `message`). For `diskann-disk` - I added module level documentation providing historical context and a brief trade-off. When possible, I prefer such documentation to live with the relevant code rather than being stuffed in an `agents.md` file.

On the naming, I'm going to stand by `convert_error!` for the following reasons: The `diskann` crate has one canonical error type, so use like `diskann::convert_error!` is IMO rather unambiguous. Further, implementing `From` **is** a conversion, so I do feel that the name is accurate.

arkrishn94 (approved):

partychen (approved):

PR #1284, opened 2026-07-28
Switch `SearchOutputBuffer` to accepting `Neighbor`s instead of tuples

hildebrandmw:
With #1273, the structural `Eq` bound on `Neighbor` has been removed. This PR is a follow-up that changes over `SearchOutputBuffer` to accept `Neighbor` instead of raw tuples. Outside of being symmetric, this has the nice property of allowing full-precision reranking methods to use `diskann::neighbor::ord::fast_distance` as their sorting callback instead of hand-rolling the implementation every time.

This PR does make the distance member of `Neighbor` generic, but defaults it to `f32` to keep syntactic compatibility.

Key files to review:
* `diskann/src/neighbors.rs`: Additional (defaulted) generic distance parameter to `Neighbor`.
* `diskann/src/graph/search_output_buffer.rs`: Updated trait definition.


diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -349,10 +349,10 @@ where
         let provider = accessor.provider;
 
         let mut uncached_ids = Vec::new();
-        let mut reranked = {
+        let mut reranked: Vec<_> = {
             let mut process = |n: u32| {
                 if let Some(entry) = accessor.scratch.distance_cache.get(&n) {
-                    Some(Ok::<((u32, _), f32), ANNError>(((n, entry.1), entry.0)))
+                    Some(Neighbor::new((n, entry.1), entry.0))
```
hildebrandmw, line 356: Note: The original code was unnecessarily wrapping the intermediate value in a `Result` but didn't call any fallible methods.

harsha-simhadri (approved):

JordanMaples (approved):

harsha-simhadri (approved):

PR #1287, opened 2026-07-29
PiPNN 2/6: add numerical kernels

SeliMeli:
## Summary

  Add GEMM-backed distance computation and SIMD top-k selection for PiPNN behind the opt-in diskann/pipnn feature.

  The numerical layer supports two operations:

  - Partition assignment: rank sampled leaders for each point and return their local IDs.
  - Leaf neighbor selection: compute each unordered point pair once and update both endpoints’ neighbor lists, returning local IDs and distances.

  This is part 2 of the PiPNN stack. Recursive graph construction integrates these kernels in #1290.

  ## Design

  Separate metric computation from neighbor selection. Metric implementations handle L2, cosine, normalized cosine, and inner product. Partition leaders retain metric-specific norms for reuse across point stripes. Kernels choose which distance slices
  to process; the shared TopK implementation handles SIMD loading, threshold filtering, scalar tails, and sorted insertion.

  Reuse each leaf distance for both endpoints. New diskann-linalg helpers compute or accumulate the lower triangle of a Gram matrix using faer. Leaf selection scans the strict lower triangle, excludes self-neighbors, and shares each loaded distance
  block between both endpoint updates.

  Specialize common top-k capacities. Capacities 1–10 use compile-time specialization to enable insertion-loop unrolling; larger capacities use the runtime path. Typical partition fanouts are 10 and 3, and typical leaf-neighbor counts are 2 or 3.
  Architecture-specific selection runs through diskann-wide at the TopK entry points.

  Keep storage reusable. TopK borrows caller-owned candidate buffers. Kernels manage distance and ranking scratch without exposing SIMD block types to their callers.

  The linear-algebra helpers validate matrix shapes and size-product overflow.

PR #1290, opened 2026-07-29
PiPNN 3/6: add core graph construction

### LINKED ISSUE #878: Support PiPNN builds

### LINKED ISSUE #1425: Implement PiPNN core graph construction
Build provider-independent PiPNN graphs from a MatrixView, graph policy, and Rayon pool. Cover recursive partitioning, leaf candidate accumulation, global ID mapping, and degree enforcement with shared RobustPrune.

Parent: #878.
Implementation: #1290.
Depends on the numerical kernels in #1422 / #1287.

SeliMeli:
## Purpose

This PR adds provider-independent PiPNN graph construction under `diskann::graph::pipnn`.

The core accepts a `MatrixView`, graph policy, and Rayon pool. It returns one adjacency list for each data point.

## Main changes

- `PiPNNConfig` defines partition and leaf policy.
- `build_graph` selects one architecture and one metric marker.
- `partitioning` owns recursion, leader sampling, row gathering, and assignment scatter.
- Partition workers pass packed point matrices and prepared leaders to the numerical stage.
- `leaf_build` gathers packed leaf matrices and maps local neighbors to graph IDs.
- Scatter and leaf mapping omit unassigned slots, so non-rankable scores cannot become graph IDs.
- Numerical kernels own GEMM, norm preparation, ranking, and reusable scratch.
- `finalization` applies shared RobustPrune only to overfull rows.
- Production callers remove the three temporary PR2 dead-code attributes.

The core does not load providers. It does not select start points. It does not serialize indexes.

## Required invariants

- Point IDs fit in `u32`.
- `0 < c_min <= c_max`.
- Each fanout value is positive.
- `leaf_k` is positive. Each leaf clamps it to its non-self point count.
- Leaf IDs are sorted and unique.
- Each worker reads only the active buffer prefix.

## Review order

1. Review `PiPNNConfig`, `PiPNNBuildContext`, and `build_graph` in `mod.rs`.
2. Review recursive partition flow in `partitioning.rs`.
3. Review leaf construction in `leaf_build.rs`.
4. Review final graph pruning in `finalization.rs`.
5. Confirm that PR2 dead-code attributes disappear in this layer.

## Validation

- Partition tests cover fanout boundaries, recursion, replicas, ordering, and buffer reuse.
- Leaf tests cover all source types, symmetry, duplicate removal, fixed insertion, and runtime insertion.
- Graph tests cover all metrics, deterministic builds, valid IDs, degree bounds, and non-rankable candidates.
- All-target Clippy passes with `-Dwarnings`.

## Stack

Stack 3/6. Depends on #1287. #1291 adds disk-index integration.


diskann/src/graph/pipnn/mod.rs
```diff
@@ -5,5 +5,200 @@
...
+    /// Maximum number of points in a leaf.
+    pub c_max: usize,
+    /// Minimum leaf size used by global small-leaf merging.
+    pub c_min: usize,
+    /// Fraction of a cluster sampled as partition leaders.
+    pub p_samp: f64,
+    /// Number of nearest leaders retained at each overlapping partition level.
+    pub fanout: Vec<usize>,
+    /// Number of nearest neighbors selected within each leaf.
+    pub k: usize,
```
magdalendobson: Request to use something like `partition_k` to not overload `k`

diskann/src/graph/pipnn/mod.rs
```diff
@@ -5,5 +5,200 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub struct PiPNNConfig {
+    /// Maximum number of points in a leaf.
+    pub c_max: usize,
+    /// Minimum leaf size used by global small-leaf merging.
+    pub c_min: usize,
+    /// Fraction of a cluster sampled as partition leaders.
+    pub p_samp: f64,
+    /// Number of nearest leaders retained at each overlapping partition level.
+    pub fanout: Vec<usize>,
```
magdalendobson, line 190: Could you explain why this is a vector? What would it mean if I passed in an arbitrary vector here?

diskann/src/graph/pipnn/mod.rs
```diff
@@ -5,5 +5,200 @@
...
+        if self.c_min == 0 {
+            return Err(config_error("c_min must be greater than zero"));
+        }
+        if self.c_min > self.c_max {
+            return Err(config_error(format!(
+                "c_min ({}) must not exceed c_max ({})",
+                self.c_min, self.c_max
+            )));
+        }
+        if !self.p_samp.is_finite() || !(0.0..=1.0).contains(&self.p_samp) || self.p_samp == 0.0 {
```
magdalendobson: Isn't the middle statement sufficient? Why do we need to explicitly check finite and nonzero?

diskann/src/graph/pipnn/mod.rs
```diff
@@ -5,5 +5,200 @@
...
+impl<'a> PiPNNBuildContext<'a> {
+    /// Validate and combine PiPNN configuration with outer graph policy.
+    pub fn new(
+        config: PiPNNConfig,
+        graph: &'a Config,
+        metric: Metric,
+        pool: &'a ThreadPool,
+    ) -> ANNResult<Self> {
+        config.validate()?;
+        if !graph.alpha().is_finite() || graph.alpha() < 1.0 {
```
magdalendobson: 0.0 < Alpha < 1.0 should not be disallowed. It is meaningful. Also, the `Config` should check for these issues already. If it doesn't, the errors should be moved into the validation of the config itself.

diskann-pipnn/src/partitioning.rs
```diff
@@ -0,0 +1,735 @@
...
+impl From<&PiPNNConfig> for PartitionConfig {
+    fn from(config: &PiPNNConfig) -> Self {
+        Self {
+            c_max: config.c_max,
+            c_min: config.c_min,
+            p_samp: config.p_samp,
+            fanout: config.fanout.clone(),
+            replicas: config.replicas,
+        }
+    }
```
magdalendobson: Why can't we just use `PiPNNConfig`? What would go wrong if we did?

SeliMeli:
Follow-up simplification in `2afc91b9`:

- `PiPNNConfig` validates `1 <= k <= 3`;
- partitioning borrows `PiPNNConfig`, removing the fanout `Vec` clone;
- leaf construction requires strictly increasing unique IDs from partitioning and removes its unordered `HashSet` fallback/test;
- top-level dataset area revalidation was removed because `MatrixView` now owns that invariant.

PiPNN production tests and all-target Clippy pass.

diskann/src/graph/pipnn/mod.rs
SeliMeli: Fixed. The core field is now `leaf_k`, which distinguishes local leaf-neighbor width from partition fanout and graph degree.

diskann/src/graph/pipnn/mod.rs
SeliMeli: `fanout[level]` is the number of nearest sampled centers assigned at that recursive partition level. The vector is a level schedule; after it is exhausted, later levels use fanout 1. The field documentation now states both rules.

diskann/src/graph/pipnn/mod.rs
SeliMeli: Fixed. Validation now uses only `0.0 < p_samp && p_samp <= 1.0`; that expression also rejects zero, infinities, and NaN.

diskann/src/graph/pipnn/partitioning.rs
```diff
@@ -0,0 +1,651 @@
...
+        .collect()
+}
+
+fn scatter_serial(
+    points: &[u32],
+    assignments: &[u32],
+    fanout: usize,
+    leaders: usize,
+) -> ANNResult<Vec<Vec<u32>>> {
+    let mut sizes = filled_vec(leaders, 0usize)?;
```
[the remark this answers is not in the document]
SeliMeli: Fixed in 109eede27. `scatter_serial` now checks `assignments.len() == points.len() * fanout` before counting or chunking, and a private-seam test covers a truncated assignment buffer.

diskann/src/graph/pipnn/mod.rs
SeliMeli: Fixed. PiPNN no longer adds alpha validation. `PiPNNBuildContext` accepts the shared graph `Config` policy unchanged; the regression test covers alpha below 1.

diskann/src/graph/pipnn/partitioning.rs
```diff
@@ -0,0 +1,803 @@
...
+
+fn assignment_stripe_point_count(leader_count: usize) -> usize {
+    let point_count = ASSIGNMENT_CACHE_TARGET_BYTES / (leader_count.max(1) * size_of::<f32>());
+    let point_count = if point_count.is_power_of_two() {
+        point_count
+    } else {
+        point_count.next_power_of_two() / 2
+    };
+    point_count.clamp(MIN_ASSIGNMENT_STRIPE_POINTS, MAX_ASSIGNMENT_STRIPE_POINTS)
+}
```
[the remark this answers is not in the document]
SeliMeli: `size_of` is imported from `std::mem` at module scope in the current file. The PiPNN test target and all-target Clippy both compile this path.

diskann-pipnn/src/partitioning.rs
SeliMeli: Fixed by the module move. Partitioning now borrows the single `PiPNNConfig`; the separate partition configuration type no longer exists.

SeliMeli:
Restored the intended stack boundary: PR3 now inherits the numerical kernels from #1287, adds their production partition/leaf callers, and removes all three temporary PR2 `allow(dead_code)` attributes. The cumulative PR3 and top-of-stack source trees are unchanged.

SeliMeli:
Updated the production callers for the split metric contracts. Partition code now builds `PartitionNorms` and requires `PartitionKernelMetric`; leaf code requires `LeafKernelMetric`; build orchestration requires both plus `MetricTag`. Formulas, architecture propagation, graph output, and the PR2/PR3 temporary suppression boundary remain unchanged.

diskann/src/graph/pipnn/mod.rs
```diff
@@ -3,29 +3,668 @@
...
+    /// Number of independent partition passes over the dataset.
+    pub replicas: usize,
+}
+
+impl PiPNNConfig {
+    /// Validate the algorithm-specific partition and leaf-build parameters.
+    pub fn validate(&self) -> ANNResult<()> {
+        if self.c_max == 0 {
+            return Err(config_error("c_max must be greater than zero"));
+        }
```
magdalendobson, line 202: Can you define an error type instead, similar to https://github.com/microsoft/DiskANN/blob/d8d97672177c9b34ec8cbedaca4cd0e08e231242/diskann/src/graph/search/knn_search.rs#L28?

diskann/src/graph/pipnn/mod.rs
```diff
@@ -3,29 +3,668 @@
...
+    Cosine, CosineNormalized, InnerProduct, L2, LeafKernelMetric, PartitionKernelMetric,
+};
+
+/// PiPNN partition and leaf-selection policy.
+///
+/// DiskANN graph policy separately supplies degree, alpha, and prune metric.
+#[derive(Clone, Debug, PartialEq)]
+pub struct PiPNNConfig {
+    /// Maximum number of points in a leaf.
+    pub c_max: usize,
```
magdalendobson, line 183: Can we make all the fields that aren't allowed to be zero `NonZeroUsize` instead? Including the elements in `fanout`.

diskann/src/graph/pipnn/mod.rs
```diff
@@ -3,29 +3,668 @@
...
+impl<'a> PiPNNBuildContext<'a> {
+    /// Check and combine PiPNN configuration with DiskANN graph policy.
+    pub fn new(
+        config: PiPNNConfig,
+        graph: &'a Config,
+        metric: Metric,
+        pool: &'a ThreadPool,
+    ) -> ANNResult<Self> {
+        config.validate()?;
+        if graph.prune_kind() != metric.into() {
```
magdalendobson, line 252: Would this check conflict with https://github.com/microsoft/DiskANN/pull/1298?

diskann/src/graph/pipnn/partitioning.rs
```diff
@@ -0,0 +1,1271 @@
...
+    A::f32x16: std::ops::Div<Output = A::f32x16>,
+    <A::f32x16 as SIMDVector>::Mask: SIMDSelect<A::f32x16>,
+    u64: From<<<<A::f32x16 as SIMDVector>::Mask as SIMDMask>::BitMask as SIMDMask>::Underlying>,
+    M: PartitionKernelMetric,
+    T: VectorRepr + Send + Sync,
+{
+    let initial_indices = point_ids(data.nrows())?;
+    if data.nrows() <= config.c_max {
+        let mut leaves = Vec::new();
+        leaves.try_reserve_exact(1).map_err(ANNError::new)?;
```
magdalendobson: Does it really help to call a `try_reserve_exact` in this case?

diskann/src/graph/pipnn/partitioning.rs
```diff
@@ -0,0 +1,1271 @@
...
+        leaves.push(initial_indices);
+        return Ok(leaves);
+    }
+
+    let mut leaves = Vec::new();
+    let mut work = Vec::new();
+    work.try_reserve_exact(1).map_err(ANNError::new)?;
```
magdalendobson: Again here, it looks like we create the vector and then reserve exactly one spot. Does this really help?

diskann/src/graph/pipnn/partitioning.rs
```diff
@@ -0,0 +1,1271 @@
...
+
+    for _ in 0..MAX_PARTITION_ITERATIONS {
+        if work.is_empty() {
+            return merge_undersized_leaves(leaves, config.c_min, config.c_max);
+        }
+
+        let mut results = Vec::new();
+        results
+            .try_reserve_exact(work.len())
+            .map_err(ANNError::new)?;
```
magdalendobson: Same question here I guess. Why all the `try_reserve_exact` with corresponding error handling? Are we expecting memory to be extremely constrained or something?

diskann/src/graph/pipnn/partitioning.rs
```diff
@@ -0,0 +1,1271 @@
...
+/// Return the number of centers to sample from one cluster.
+///
+/// The count is `ceil(points * sampling_fraction)`, limited by `LEADER_CAP` and
+/// the number of available points. A cluster with at least two points uses at
+/// least two centers.
+fn sampled_leader_count(points: usize, sampling_fraction: f64) -> usize {
+    ((points as f64 * sampling_fraction).ceil() as usize)
+        .clamp(2, LEADER_CAP)
+        .min(points)
+}
```
magdalendobson, line 242: Do we really need to cap the number of centers? It seems like this may be just introducing too much complication for little benefit.

diskann/src/graph/pipnn/mod.rs
```diff
@@ -3,29 +3,668 @@
...
+///
+/// DiskANN graph policy separately supplies degree, alpha, and prune metric.
+#[derive(Clone, Debug, PartialEq)]
+pub struct PiPNNConfig {
+    /// Maximum number of points in a leaf.
+    pub c_max: usize,
+    /// Minimum leaf size used by global small-leaf merging.
+    pub c_min: usize,
+    /// Fraction of a cluster sampled as child-partition centers.
+    pub p_samp: f64,
```
magdalendobson, line 187: Could it work better if we framed this parameter as `num_leaders: NonZeroUsize` instead of `p_samp`? Then we could avoid putting a hard cap on it (since it's in the hands of the user now) and simplify the code a little bit.

diskann/src/graph/pipnn/partitioning.rs
```diff
@@ -0,0 +1,1271 @@
...
+        clusters.push(cluster);
+    }
+    Ok(clusters)
+}
+
+/// Merge leaves smaller than `c_min` without exceeding `c_max`.
+///
+/// A `HashSet` removes duplicate point IDs across merged leaves. The function
+/// sorts each merged result before it returns.
+fn merge_undersized_leaves(
```
magdalendobson, line 456: I've contacted the authors of PiPNN to discuss this, but I wonder if we could skip the step of merging undersized leaves. Here is the reasoning:

Clusters below a certain minimum size are disallowed on the grounds that a too-small cluster would mean not enough potential neighbors for the prune algorithm. I was wondering if this observation is potentially leftover from when the authors were using data structures other than a 2-nearest neighbor graph, since now each cluster just supplies 4-6 neighbor. I would have thought that with sufficient replication, a few underfull clusters would not meaningfully impact the resultant graph quality. It also intuitively seems like merging underfull clusters randomly would be unlikely to impact the 2-3 nearest neighbors of most points, since their centroids might be far apart (though maybe this is not so true with high-dimensional data?).

Would this be something you could experiment with? It would be nice to reduce complexity in the partitioning step.

diskann/src/graph/pipnn/partitioning.rs
```diff
@@ -0,0 +1,1271 @@
...
+}
+
+fn replica_seed(replica: usize) -> u64 {
+    PARTITION_SEED.wrapping_add((replica as u64).wrapping_mul(REPLICA_SEED_STEP))
+}
+
+// This LCG derives child seeds. Wrapping arithmetic gives the same mapping in
+// debug and release builds on all supported platforms.
+fn mix_seed(seed: u64, salt: u64) -> u64 {
+    seed.wrapping_mul(6_364_136_223_846_793_005)
```
magdalendobson, line 251: Why was this number chosen?

diskann/src/graph/pipnn/partitioning.rs
```diff
@@ -0,0 +1,1271 @@
...
+    let mut leaves = Vec::new();
+    let mut work = Vec::new();
+    work.try_reserve_exact(1).map_err(ANNError::new)?;
+    work.push(WorkItem {
+        indices: initial_indices,
+        level: 0,
+        seed,
+    });
+
+    for _ in 0..MAX_PARTITION_ITERATIONS {
```
magdalendobson, line 150: Why do we need a max iterations? Can't this just be framed as a while loop until there are no overfull/underfull partitions?

diskann/src/graph/pipnn/partitioning.rs
```diff
@@ -0,0 +1,1271 @@
...
+    PiPNNConfig,
+    kernel_metric::PartitionKernelMetric,
+    partition_kernel::{PartitionInput, PartitionKernelWorkspace, PartitionNorms, nearest_leaders},
+};
+
+// These constants control internal batching and deterministic seed generation.
+const PARTITION_SEED: u64 = 1_000;
+const REPLICA_SEED_STEP: u64 = 7_919;
+const LEADER_CAP: usize = 1_000;
+const ASSIGNMENT_CACHE_TARGET_BYTES: usize = 524_288;
```
magdalendobson, line 40: Is this going to be a sensible value across most machines?

magdalendobson:
General question--what is the largest number of threads you have tested construction speed on? Can you share some (num_threads, build_time) statistics?

diskann/src/graph/pipnn/leaf_build.rs
```diff
@@ -0,0 +1,890 @@
...
+use super::{
+    kernel_metric::LeafKernelMetric,
+    leaf_kernel::{
+        LeafKernelError, LeafKernelWorkspace, LeafNeighbor, leaf_neighbor_count, nearest_neighbors,
+    },
+};
+
+/// Failure while converting leaves into direct graph candidates.
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum LeafBuildError {
```
magdalendobson, line 34: A lot of these errors seem like overkill to me. If we aren't exposing these functions publicly, we can have the contract be that the functions higher up in the stack will ensure things like nonzero dimension, u32 point ids, empty leaves, etc. Can you do a pass of what invariants are enforced higher up in the stack and get rid of the corresponding error types?

diskann/src/graph/pipnn/finalization.rs
```diff
@@ -0,0 +1,293 @@
...
+                    .try_reserve(
+                        workspace
+                            .candidate_slots
+                            .len()
+                            .saturating_sub(workspace.prune_states.len()),
+                    )
+                    .map_err(ANNError::new)?;
+                workspace
+                    .prune_states
+                    .resize(workspace.candidate_slots.len(), prune::State::default());
```
xinyuwen2: prune_states is thread-local scratch reused across source points. Vec::resize initializes only newly appended elements, so entries retained from the previous candidate list preserve last_checked, neighbor, and occlude_factor. When the next list has the same or a smaller size, robust_prune can use a stale last_checked and index beyond the current candidate list.

SeliMeli:
Structural follow-up: `partitioning` and `leaf_build` now own norm reduction and squared-norm conversion. Partitioning passes active point and dot matrix views. Kernels only validate prepared slice lengths and evaluate formulas. Fixed-seed graphs remain byte-identical for all four metrics.

diskann/src/graph/pipnn/finalization.rs
SeliMeli: Fixed. Finalization now resets every active `prune_states` entry after `resize`. A one-worker regression test compares reused scratch with fresh pruning for two consecutive source rows.

diskann/src/graph/pipnn/partitioning.rs
```diff
@@ -0,0 +1,1261 @@
...
+        })
+    })?;
+    let dots = &mut dot_buffer[..dots_len];
+    gather_vectors(data, point_ids, point_matrix.as_mut_slice())?;
+    diskann_linalg::sgemm(
+        Transpose::None,
+        Transpose::Ordinary,
+        point_count,
+        leader_count,
+        dimensions,
```
xinyuwen2, line 358: Maybe I'm misunderstanding something, but it seems that this part of the computation scales linearly with the vector dimension. Given that PiPNN is mainly targeting faster index construction, do we expect this to become a bottleneck on high-dimensional datasets, or has that not been an issue in practice?

diskann/src/graph/pipnn/leaf_build.rs
```diff
@@ -0,0 +1,898 @@
...
+                leaf,
+                point,
+                source: source.into(),
+            }
+        })?;
+    }
+
+    diskann_linalg::sgemm_aat_lower(
+        point_ids.len(),
+        data.ncols(),
```
xinyuwen2: Similar to my understanding that the partitioning step is affected by dimensionality, the leaf-building stage here also seems to have computational complexity that scales roughly linearly with the dimension.

diskann/src/graph/pipnn/partitioning.rs
xinyuwen2: I evaluated five datasets, and they all showed a very similar pattern: there was no significant improvement in build time, and recall remained largely unchanged. Based on the results, this step doesn't seem to have much impact on graph construction. My suggestion is we can remove it to make the code cleaner.

All experiments were conducted with leaf_k = 2. The following table summarizes the BigANN Recall@100 results.  @SeliMeli, @arkrishn94

Image

Image

PR #1298, opened 2026-07-30
Fix normalized cosine distance consistency for PQ graphs

partychen:
## Problem

For Product-PQ with `CosineNormalized`, graph-search queries used raw squared L2 while graph pruning used cosine distance.

Vamana pruning compares `distance_ik / distance_jk`. For normalized vectors, cosine distance is half squared L2, so mixing the two approximately doubled this ratio. This made `alpha = 1.2` behave like `alpha = 0.6`, causing over-pruning and poor recall.

## Fix

Use raw squared L2 for Product-PQ Hybrid and Quantized pruning, matching the existing PQ graph-search query approximation.

The change is scoped to Product-PQ pruning. Search behavior, public PQ distance APIs, other metrics, and other quantization strategies are unchanged.

## Validation

Regression tests cover all Hybrid operand combinations and the Quantized PQ/PQ pruning path.

The final commit was benchmarked on the first 100,000 BigANN SIFT base vectors and first 1,000 queries, converted to unit-normalized `f32` as required by `CosineNormalized`. Exact ground truth was recomputed for this subset. Configuration: 50 PQ chunks, `max_fp_vecs_per_prune = 48`, `max_degree = 64`, `l_build = 100`, `alpha = 1.2`, and `search_l = 100`.

### Search results

| Build | Recall@10 | Mean comparisons | Mean hops | Comparisons / hop |
|---|---:|---:|---:|---:|
| Before | **0.0455** | 1,206.963 | 119.609 | 10.09 |
| After | **0.9999** | 3,205.906 | 103.618 | 30.94 |

### Graph structure

| Build | Graph size | Out-edges | Mean degree | P50 | P90 | P99 | Degree <=4 | Degree >=33 |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| Before | 8.99 MiB | 2,256,586 | **22.565** | 22 | 30 | 40 | 0.002% | 6.161% |
| After | 22.02 MiB | 5,672,071 | **56.720** | 57 | 77 | 83 | 0.001% | 94.166% |

Checks:

- `cargo test -p diskann-providers --lib`
- `cargo clippy --workspace --all-targets -- -D warnings`

### DIFF
--- a/diskann-providers/src/model/graph/provider/async_/distances.rs
+++ b/diskann-providers/src/model/graph/provider/async_/distances.rs
@@ -66,7 +66,7 @@ pub mod pq {
 
     use diskann::utils::VectorRepr;
     use diskann_utils::Reborrow;
-    use diskann_vector::{DistanceFunction, distance::Metric};
+    use diskann_vector::DistanceFunction;
 
     use crate::model::pq;
 
@@ -112,33 +112,23 @@ pub mod pq {
 
     /// Distance computer that handles mixed full-precision and PQ-compressed operands.
     ///
-    /// For `CosineNormalized`, every operand combination uses half squared L2 so pruning
-    /// compares compatible values. For other metrics, full-precision pairs use the native
-    /// distance function and pairs involving a quantized operand use the PQ distance table.
-    /// Mixed pairs convert the full-precision side to `f32` for the PQ lookup.
+    /// When both operands are full-precision, the native distance function is used. When
+    /// at least one is quantized, the PQ distance table is used instead. Mixed pairs
+    /// (full vs quant) convert the full-precision side to `f32` for the PQ lookup.
     pub struct HybridComputer<'a, T>
     where
         T: VectorRepr,
     {
         quant: pq::distance::DistanceComputer<'a>,
         full: T::Distance,
-        full_scale: f32,
     }
 
     impl<'a, T> HybridComputer<'a, T>
     where
         T: VectorRepr,
     {
-        pub fn new(quant: pq::distance::DistanceComputer<'a>, dim: Option<usize>) -> Self {
-            let (full_metric, full_scale) = match quant.metric() {
-                Metric::CosineNormalized => (Metric::L2, pq::COSINE_NORMALIZED_L2_SCALE),
-                metric => (metric, 1.0),
-            };
-            Self {
-                quant,
-                full: T::distance(full_metric, dim),
-                full_scale,
-            }
+        pub fn new(quant: pq::distance::DistanceComputer<'a>, full: T::Distance) -> Self {
+            Self { quant, full }
         }
     }
 
@@ -151,7 +141,7 @@ pub mod pq {
         fn evaluate_similarity(&self, x: Hybrid<&[T], &[u8]>, y: Hybrid<&[T], &[u8]>) -> f32 {
             match x {
                 Hybrid::Full(x) => match y {
-                    Hybrid::Full(y) => self.full_scale * self.full.evaluate_similarity(x, y),
+                    Hybrid::Full(y) => self.full.evaluate_similarity(x, y),
                     Hybrid::Quant(y) => {
                         // SAFETY: This can only panic when T = `MinMaxElement` and the underlying slice is ill-defined.
                         // we are ok with panicking in distance functions for now.
@@ -169,74 +159,6 @@ pub mod pq {
                     Hybrid::Quant(y) => self.quant.evaluate_similarity(x, y),
                 },
             }
-        }
-    }
-
-    #[cfg(test)]
-    mod tests {
-        use diskann_vector::{DistanceFunction, distance::Metric};
-
-        use super::{Hybrid, HybridComputer};
-        use crate::model::pq::{FixedChunkPQTable, distance::DistanceComputer};
-
-        fn assert_all_hybrid_pair_distances(
-            computer: &HybridComputer<'_, u8>,
-            full0: &[u8],
-            full1: &[u8],
-            code0: &[u8],
-            code1: &[u8],
-            expected: f32,
-        ) {
-            for (left, right) in [
-                (Hybrid::Full(full0), Hybrid::Full(full1)),
-                (Hybrid::Full(full0), Hybrid::Quant(code1)),
-                (Hybrid::Quant(code0), Hybrid::Full(full1)),
-                (Hybrid::Quant(code0), Hybrid::Quant(code1)),
-            ] {
-                assert_eq!(computer.evaluate_similarity(left, right), expected);
-            }
-        }
-
-        #[test]
-        fn hybrid_cosine_normalized_matches_cosine_for_unit_vectors() {
-            let table = FixedChunkPQTable::new(
-                4,
-                vec![1.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0].into(),
-                vec![0, 2, 4].into(),
-            )
-            .unwrap();
-            let computer = HybridComputer::<u8>::new(
-                DistanceComputer::new(&table, Metric::CosineNormalized),
-                Some(4),
-            );
-            let full0 = [1, 0, 0, 0];
-            let full1 = [0, 1, 0, 0];
-            let code0 = [0, 0];
-            let code1 = [1, 1];
-
-            assert_all_hybrid_pair_distances(&computer, &full0, &full1, &code0, &code1, 1.0);
-        }
-
-        #[test]
-        fn hybrid_cosine_normalized_uses_scaled_l2_for_integer_vectors() {
-            let table = FixedChunkPQTable::new(
-                4,
-                vec![1.0, 0.0, 0.0, 1.0, 2.0, 0.0, 0.0, 2.0].into(),
-                vec![0, 2, 4].into(),
-            )
-            .unwrap();
-            let computer = HybridComputer::<u8>::new(
-                DistanceComputer::new(&table, Metric::CosineNormalized),
-                Some(4),
-            );
-            let full0 = [1, 0, 0, 2];
-            let full1 = [2, 0, 0, 1];
-            let code0 = [0, 1];
-            let code1 = [1, 0];
-
-            // Integer CosineNormalized normally dispatches to Cosine. Non-unit vectors make
-            // that differ from scaled L2, so this verifies Hybrid pruning keeps one definition.
-            assert_all_hybrid_pair_distances(&computer, &full0, &full1, &code0, &code1, 1.0);
         }
     }
 }
--- a/diskann-providers/src/model/graph/provider/async_/distances.rs
+++ b/diskann-providers/src/model/graph/provider/async_/distances.rs
@@ -112,9 +112,10 @@ pub mod pq {
 
     /// Distance computer that handles mixed full-precision and PQ-compressed operands.
     ///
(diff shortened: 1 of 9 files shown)


diskann-providers/src/model/pq/distance/l2.rs
```diff
@@ -48,9 +48,22 @@ impl<'a> TableL2<'a> {
...
+        pool: Option<Arc<ObjectPool<Vec<f32>>>>,
+        scale: f32,
     ) -> ANNResult<Self> {
         let mut object = Self::new_unpopulated(parent, pool);
         object.populate(query)?;
+        object
+            .lookup_table
+            .iter_mut()
+            .for_each(|distance| *distance *= scale);
         Ok(object)
```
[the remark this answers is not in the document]
partychen: Updated `TableL2::new` to retain its original populate-only path. Only `new_scaled`, used by `CosineNormalized`, performs the scaling pass now.

diskann-providers/src/model/graph/provider/async_/distances.rs
```diff
@@ -161,4 +161,53 @@ pub mod pq {
...
+            assert_relative_eq!(full_quant, expected_full_quant, max_relative = 1.0e-7);
+
+            let quant_quant = computer.evaluate_similarity(
+                Hybrid::Quant(code0.as_slice()),
+                Hybrid::Quant(code1.as_slice()),
+            );
+            let squared_l2: f32 =
+                SquaredL2::evaluate(reconstructed0.as_slice(), reconstructed1.as_slice());
+            let expected_quant_quant = 0.5 * squared_l2;
+            assert_relative_eq!(quant_quant, expected_quant_quant, max_relative = 1.0e-7);
```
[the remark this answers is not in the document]
partychen: The hybrid regression test now reuses `COSINE_NORMALIZED_L2_SCALE` instead of hard-coding `0.5`.

diskann-providers/src/model/pq/fixed_chunk_pq_table.rs
```diff
@@ -242,11 +244,14 @@ impl FixedChunkPQTable {
         )
     }
 
-    /// Calculate the distance between query and given centroid by cosine distance
+    /// Approximate normalized cosine distance between a query and a PQ vector.
+    ///
+    /// This uses half the squared L2 distance so that every PQ-involved
+    /// `CosineNormalized` comparison uses the same scale.
     /// * `query_vec` - query vector: 1 * dim
     /// * `base_vec` - given centroid array: 1 * num_pq_chunks
```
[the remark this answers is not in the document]
partychen: Clarified that `base_vec` is a PQ code containing one centroid index per chunk.

partychen:
> ## Pull request overview
> Copilot reviewed 8 out of 8 changed files in this pull request and generated no new comments.
> 
> Comments suppressed due to low confidence (2)
> **diskann-providers/src/model/graph/provider/async_/distances.rs:187**
> 
> * The hybrid regression test uses non-unit full vectors (`[1, 0, 0, 2]`, `[2, 0, 0, 1]`), but `Metric::CosineNormalized` is documented/implemented under a unit-norm assumption. Using unit vectors here would make the test’s intent clearer (CosineNormalized == 0.5*squared-L2) and avoid locking in behavior that only matches scaled-L2 for non-normalized inputs.
> 
> ```
>             let table = FixedChunkPQTable::new(
>                 4,
>                 vec![1.0, 0.0, 0.0, 1.0, 2.0, 0.0, 0.0, 2.0].into(),
>                 vec![0, 2, 4].into(),
>             )
> ```
> 
> **diskann-providers/src/model/graph/provider/async_/distances.rs:141**
> 
> * `HybridComputer::new` maps `Metric::CosineNormalized` full/full comparisons onto `Metric::L2` with a 0.5 scale. This contradicts the PR description’s stated scope that full/full comparisons remain on the native `CosineNormalized` implementation, and it can also change behavior when full vectors are not perfectly unit-norm. Consider keeping the full-precision path on `Metric::CosineNormalized` (scale 1.0) and relying on the PQ side’s scaled-L2 approximation for compatibility.
> 
> This issue also appears on line 183 of the same file.
> 
> ```
>         pub fn new(quant: pq::distance::DistanceComputer<'a>, dim: Option<usize>) -> Self {
>             let (full_metric, full_scale) = match quant.metric() {
>                 Metric::CosineNormalized => (Metric::L2, pq::COSINE_NORMALIZED_L2_SCALE),
>                 metric => (metric, 1.0),
>             };
> ```

Kept full/full scaled-L2 intentionally because Hybrid pruning mixes full/full, full/PQ, and PQ/PQ comparisons. Restoring native CosineNormalized only for full/full would reintroduce incompatible distance definitions, especially for u8/i8. The tests now separate and document unit-vector semantics and integer consistency.

diskann-providers/src/model/pq/fixed_chunk_pq_table.rs
```diff
@@ -242,11 +249,14 @@ impl FixedChunkPQTable {
...
 
-    /// Calculate the distance between query and given centroid by cosine distance
+    /// Approximate normalized cosine distance between a query and a PQ vector.
+    ///
+    /// This uses half the squared L2 distance so that every PQ-involved
+    /// `CosineNormalized` comparison uses the same scale.
     /// * `query_vec` - query vector: 1 * dim
-    /// * `base_vec` - given centroid array: 1 * num_pq_chunks
+    /// * `base_vec` - PQ code containing one centroid index per chunk
     pub fn cosine_normalized_distance(&self, query_vec: &[f32], base_vec: &[u8]) -> f32 {
```
wuw92, line 251: Note that this public method previously computed cosine distance against the reconstructed PQ vector, while this change makes it return a scaled-L2 approximation. Callers using FixedChunkPQTable directly would silently receive different results, independently of the Hybrid pruning fix.

Could you evaluate whether this behavior change is acceptable for non-Hybrid callers?

diskann-providers/src/model/graph/provider/async_/distances.rs
```diff
@@ -112,23 +112,33 @@ pub mod pq {
...
     }
 
     impl<'a, T> HybridComputer<'a, T>
     where
         T: VectorRepr,
     {
-        pub fn new(quant: pq::distance::DistanceComputer<'a>, full: T::Distance) -> Self {
-            Self { quant, full }
+        pub fn new(quant: pq::distance::DistanceComputer<'a>, dim: Option<usize>) -> Self {
+            let (full_metric, full_scale) = match quant.metric() {
```
wuw92: This routing appears to change the correct full/full behavior to match the scaled-L2 PQ paths. That equivalence holds for normalized floating-point vectors, but not for `u8`/`i8`, where `CosineNormalized` intentionally routes to cosine because the stored vectors are not generally unit-normalized.

Could we keep the existing full/full cosine behavior and change `u8`/`i8` PQ paths to use a compatible cosine distance instead of scaled-L2 
```

File: diskann-providers/src/model/pq/fixed_chunk_pq_table.rs (hunk 4 above)
Comment (partychen) [reply diskann-providers/src/model/pq/fixed_chunk_pq_table.rs:251]: Good point. The previous implementation was broader than necessary. I reverted the changes to `FixedChunkPQTable` and the public PQ distance behavior, keeping the existing search behavior unchanged. The fix is now scoped to Product-PQ pruning, matching the squared-L2 approximation used by PQ graph-search queries

File: diskann-providers/src/model/graph/provider/async_/inmem/product.rs [hunk 6]
```diff
@@ -42,6 +42,20 @@ use crate::model::{
 /// The default quant provider.
 pub type DefaultQuant = FastMemoryQuantVectorProviderAsync;
 
+fn quant_pruning_distance_computer(
```
wuw92: Using raw squared L2 consistently may be valid while these values are used only for internal ordering and pruning, However, the resulting values no longer follow the standard `CosineNormalized` contract if they are returned to or otherwise consumed by a caller as distances.

The underlying issue seems to be that the Product-PQ `CosineNormalized` paths already have different contracts:

- query/PQ returns raw squared L2;
- full/full uses the native implementation;
- full/PQ and PQ/PQ use cosine over reconstructed PQ vectors.

Overriding `CosineNormalized` with `L2` in the pruning strategy hides the existing contract mismatch and also changes `u8`/`i8` from angular to Euclidean behavior.

@arkrishn94 any thoughts on these contracts?

partychen: Agreed on the contract distinction. The remapped computer here is module-private and used only by Vamana pruning; it is not returned as a `CosineNormalized` distance, and the public PQ distance behavior remains unchanged.


The narrow intent is to make `distance_jk` use the same approximation as the search-pool `distance_ik`, since they participate in the same pruning ratio. I agree this does not resolve the broader Product-PQ contract inconsistency, particularly for `u8`/`i8`. If those inputs should preserve angular semantics, the query/PQ path and pruning path would need to be addressed together. I’m happy to follow the contract decision here.

magdalendobson:
Thank you for finding the bug and contributing this! I was wondering if you could add an integration test that would have caught this, as it seems like a fault in our testing that this wasn't found earlier?

partychen:
@magdalendobson Added in f140b00a. I extended the existing SIFT build-and-search coverage with normalized `CosineNormalized` cases for both the Hybrid Product-PQ path and the quant-only `Quantized` path. These tests build real indexes and validate searches against exact ground truth. I also verified that the new cases fail against the pre-fix code and pass with this change. Thanks for calling this out!

magdalendobson (approved):
Thanks for adding testing. Looks good to me now. Request that you take a second look at all the quantizer x metric combinations and make sure all of them are tested properly inside the index test, but not blocking.

xinyuwen2 (approved):

diskann-providers/src/model/graph/provider/async_/inmem/product.rs
arkrishn94: Hmm, I'm a little worried about adding this kind of bespoke hook to solve the problem here for the following reasons - 
- It is not clear for a caller of this function (or user of the underlying distance computer created), what the output `metric` can or cannot be used for. A caller could mistakenly assume that the underlying metric is `L2` when the metric used to create the computer is `CosineNormalized` for e.g and end up doing incorrect stuff.
- Another implementor of PQ-based prune has to know to call this bespoke function before creating the computer. This kind of implicit knowledge is almost certainly likely to lead to bugs.

It seems to me the problem is that the behavior of the `VTable` and the `QueryComputer` dispatch to different underlying estimators for `CosineNormalized`: the latter seems to do `cosine` and the former seems to do `l2`?

Is it possible to fix this by dispatching correctly there? @hildebrandmw to correct me if I'm wrong here.

arkrishn94 (changes requested):

diskann-providers/src/model/graph/provider/async_/inmem/product.rs
partychen: Good point. However, changing only the VTable dispatch would not fully resolve the inconsistency: Hybrid full/full comparisons bypass VTable and would still use native CosineNormalized, which is half the squared-L2 scale.
I’ll instead remap the metric in FixedChunkPQTable::create, so the entire Product-PQ graph-build path uses L2 consistently without a pruning-specific hook.

hildebrandmw (approved):
Thanks!

arkrishn94 (approved):

Issue #1016, opened 2026-05-05
QuantCompressor.new_at_stage(): investigate if generation of PQ pivots is needed there


arrayka:
QuantCompressor.new_at_stage() quantizes a dataset. It expects a file with PQ pivots to exist (a previous step generates it).

However, QuantCompressor.new_at_stage() contains a recovery logic which generates the PQ pivots if the pivots file is missing:

```
if !pivots_exists {
generate_pq_pivots(..);
}

Investigate if this fallback logic is useful.

xwj-ox:
I investigated this path. The missing-pivots branch appears to be necessary.

The earlier build-quantizer step writes the build-time PQ artifact, `<prefix>_build_pq_pivots.bin`. However, `PQGeneration` uses the disk-search PQ artifact, `<prefix>_pq_pivots.bin`, together with `<prefix>_pq_compressed.bin`.

In the disk-index build flow, `QuantDataGenerator` is explicitly instantiated with `PQGeneration`, but I could not find an earlier step that writes `<prefix>_pq_pivots.bin` before that. So for a fresh build, this branch appears to be what creates the disk-search PQ pivots.

Therefore, this logic should not be removed as dead fallback code. A possible cleanup would be to make this generation step explicit before constructing `PQGeneration`.

PR #1299, opened 2026-07-30
Train fresh disk-search PQ codebooks for each index build

xwj-ox:
- [x] Does this PR modify any existing APIs? Existing compressor and generator signatures are preserved. PQGeneration::new and generate_pq_pivots now always train; the latter only requires StorageWriteProvider.
- [ ] Is the change to the API backwards compatible? Existing call signatures remain usable, but callers that relied on implicit pretrained-codebook reuse, including CLI build_pq, must account for retraining.
- [x] Should this result in any changes to our documentation, either updating existing docs or adding new ones? Rustdoc documents constructor training and unconditional pivot generation, including CLI behavior.

#### Reference Issues/PRs

Refs #1016. Follows the checkpoint/continuation removal in #1254.

Previously, an existing pivot file could silently skip PQ training. This PR makes generate_pq_pivots always train and save a new codebook, for both disk-index construction and CLI build_pq. Each disk-index build uses its newly generated codebook for every compression block.

- Validate the dataset header, nonempty input, and compression block size before constructing the disk-search quantizer. Validate PQ chunk and center counts before sampling and training.
- Keep training in PQGeneration::new and preserve the existing QuantCompressor<T>::new and QuantDataGenerator<T, Q> signatures and public fields. No preparation-stage trait or additional compressor type is introduced.
- Use one generation entry point, generate_pq_pivots, with no existing-file training skip. CLI build_pq remains available; repeated runs using the same output prefix retrain and replace the pivots before generating compressed data.
- Preserve the training algorithm, random-source handling, L2 centering, and file format. Validate the saved codebook against the actual training dimension, including expanded vector representations.

Behavioral compatibility: PQGeneration::new and generate_pq_pivots require valid training data even when a pivot file already exists, and replace that file after successful training. Repeated CLI build_pq runs also retrain. Callers relying on implicit pretrained-codebook reuse must account for this behavior change. Validation and training errors before the write leave existing pivots unchanged; this does not add transactional rollback for storage or subsequent compression failures.

#### Any other comments?

Validated locally on Windows with Rust 1.97.1:

- cargo fmt --all and formatting verification passed.
- cargo test --locked --offline -p diskann-disk: 258 unit tests and 1 documentation test passed.
- cargo test --locked --offline -p diskann-providers --lib model::pq::pq_construction::pq_test: 47 tests passed.
- cargo clippy --locked --offline --workspace --all-targets -- -D warnings passed.

Regression coverage checks regeneration despite existing pivots, complete codebook and compression byte equality, and invalid inputs preserving existing outputs. Cross-platform execution and coverage collection remain for CI.


xwj-ox:

xwj-ox:
@copilot resolve the merge conflicts in this pull request

partychen:
Suggesting a different direction on this one

The problem this PR is going after is real: PQGeneration::new() is nominally a constructor, but it silently runs k-means and writes pivots to disk. A method called new shouldn't do that.

That said, hoisting the training to the call site trades one problem for another — it introduces an ordering contract that the compiler doesn't check. Callers of QuantDataGenerator must now call generate_pivots before new. That line returns (), produces no value anything downstream consumes, and still compiles if you delete it — it only fails at runtime. It also breaks up what used to be a clean new(...) → generate_data(...) flow in build.rs.

The deeper point: new shouldn't do this work, but not because the caller should — because it should be deferred to generate_data. new builds an object from config; generate_data is what actually does the work.

Concretely:

Have QuantDataGenerator hold &Q::CompressorContext instead of a Q, so new is pure config and doesn't need to return Result
Rename the trait method QuantCompressor::new to prepare, and call it from inside generate_data
PQGeneration::prepare calls generate_pivots first (keep that extracted named function — it's the genuinely useful part of this PR), then does the existing load
build.rs goes back to two clean lines, and the long turbofish disappears along with the line above it.

This also fixes a real defect along the way: the max_block_size == 0 and empty-dataset checks live at the top of generate_data, but pivots have already been trained and loaded back in QuantDataGenerator::new by then. So a bad parameter costs you a full wasted k-means run before it reports. Deferring construction puts validation first and drops that cost.

Two more points, independent of which approach you pick:

The num_chunks > dim check and the pivot_data_exist call are now duplicated across generate_pivots and new — introduced by the refactor. Under the approach above, the copy in prepare can just go away.

The old test_create_and_load_pivots_file had a step that called generate_pq_pivots independently and compared the result byte-for-byte against what the compressor wrote — that was verifying codebook contents are correct. The new test only asserts the file exists and the dimensions match, with nothing replacing that check. It's unrelated to this restructuring, so I'd restore it.

diskann-disk/src/build/builder/build.rs
```diff
@@ -157,6 +157,8 @@ where
             metric: self.index_configuration.dist_metric,
         };
 
+        PQGeneration::<Data::VectorDataType, StorageProvider>::generate_pivots(&quantizer_context)?;
```
partychen: We should not add an explicit `generate_pivots` call here. The `generator.generate_data(...)` call below already generates the pivots when they do not exist. The previous behavior was correct; this PR should only refactor the generation logic out of the constructor, without changing the existing workflow.

partychen:
`new` should only construct the quantizer and should not contain pivot-generation logic. Generation is an explicit operation, just like `compress`, and should be invoked by the caller when needed. Please extract it into a separate `generate` method and call it at the appropriate call site.

partychen (changes requested):

diskann-disk/src/storage/quant/compressor.rs
```diff
@@ -29,7 +30,8 @@ where
 {
     type CompressorContext;
 
-    fn new(context: &Self::CompressorContext) -> ANNResult<Self>;
+    fn new(context: Self::CompressorContext) -> Self;
```
xwj-ox: AI: This changes a public API even though the PR is scoped as a no-behavior-change refactor. `QuantCompressor` and `QuantDataGenerator` are re-exported from `diskann_disk::storage::quant`, but this changes the constructor from `new(&Context) -> ANNResult<Self>` to `new(Context) -> Self` plus a required `generate(&self)` step, and changes when initialization failures surface. Either keep a compatibility path / make these abstractions crate-private if they are internal, or mark and document the breaking API/migration impact in the PR/release notes.

diskann-disk/src/storage/quant/generator.rs
```diff
@@ -93,6 +93,7 @@ where
             ));
         }
 
+        self.quantizer.generate()?;
```
xwj-ox: AI: The generator tests no longer prove that this new lifecycle hook is actually invoked before compression. The test `DummyCompressor::generate()` is a no-op and `compress()` succeeds without any generated state, so the tests would still pass if this line were removed or bypassed, while real `PQGeneration` requires generated/loaded pivots. Please make the dummy compressor record that `generate()` ran, have `compress()` fail/assert if it did not, and assert the success path calls it exactly once while validation-failure paths do not call it.

diskann-disk/src/storage/quant/compressor.rs
```diff
@@ -29,7 +30,8 @@ where
...
+    fn generate(&self) -> ANNResult<()>;
```
xwj-ox: AI: The new public two-step lifecycle needs a stronger rustdoc contract. The current docs list `generate`, but they do not say that it must complete before `compress`, what state it prepares, or which errors can occur. Please document the required call order and prepared-state invariant, with `# Errors` notes for `generate`/`compress` as appropriate, so external implementers do not accidentally compile a no-op or skipped preparation path.

diskann-disk/src/storage/quant/pq/pq_generation.rs
```diff
@@ -109,6 +109,34 @@ where
...
+        }
+    }
+
+    fn generate(&self) -> diskann::ANNResult<()> {
+        if self.table.get().is_some() {
+            return Ok(());
+        }
+
+        let context = &self.context;
+        Self::generate_pivots(context)?;
```
xwj-ox: AI: `generate()` is now callable through `&self`, but only the final table publication is protected by `OnceLock`. Two concurrent callers can both see an empty table, both run `generate_pivots` / load the same pivot path, and then one returns `PQ compressor was generated concurrently` even though the compressor is usable afterward. Please guard the whole generation/load sequence with `OnceLock::get_or_try_init` or a mutex-backed equivalent so concurrent callers wait for and share the same initialization result.

partychen (approved):

wuw92:
Going back to #1016, `new_at_stage()` managed the checkpoint lifecycle. On a fresh start, it could generate missing pivots. On resume, it required existing pivots and rejected missing ones, ensuring that newly compressed vectors used the same codebook as previously written codes.

With checkpoint/resume removed in #1254, that lifecycle no longer applies. We also have no requirement to reuse pretrained codebooks. The code should therefore evolve toward a simpler, fresh-build-only flow:

**Validate inputs → train disk-search PQ pivots → save pivots → compress the dataset using that codebook.**

The current PR retains the old “load if present, otherwise train” policy behind a new `prepare()` abstraction. I would prefer to use this PR to finish simplifying the lifecycle after #1254, rather than introduce additional public traits and types to preserve behavior we no longer need.

xwj-ox:
> Going back to #1016, `new_at_stage()` managed the checkpoint lifecycle. On a fresh start, it could generate missing pivots. On resume, it required existing pivots and rejected missing ones, ensuring that newly compressed vectors used the same codebook as previously written codes.
> 
> With checkpoint/resume removed in #1254, that lifecycle no longer applies. We also have no requirement to reuse pretrained codebooks. The code should therefore evolve toward a simpler, fresh-build-only flow:
> 
> **Validate inputs → train disk-search PQ pivots → save pivots → compress the dataset using that codebook.**
> 
> The current PR retains the old “load if present, otherwise train” policy behind a new `prepare()` abstraction. I would prefer to use this PR to finish simplifying the lifecycle after #1254, rather than introduce additional public traits and types to preserve behavior we no longer need.

@wuw92 Thank you for your review and comments! Agreed for the disk-index build path; this PR now always trains a fresh disk-search PQ codebook.

There is still an existing caller to account for: CLI build_pq uses generate_pq_pivots, which skips training when an existing pivot file has matching dimensions and center count. I preserved that reuse policy to keep the change scoped to disk-index construction.

That is why I added `generate_pq_pivots_fresh`: disk builds explicitly request a fresh codebook, while CLI callers retain the existing behavior. Both entry points share the same training implementation.

diskann-providers/src/model/pq/pq_construction.rs
```diff
@@ -67,6 +67,10 @@ where
 /// file pq_pivots_path as a s num_centers*dim floating point binary file
 /// PQ pivot table layout: {pivot offsets data: METADATA_SIZE}{pivot vector:[dim; num_centroid]}{centroid vector:[dim; 1]}{chunk offsets:[chunk_num+1; 1]}
 ///
+/// Preserves the legacy reuse behavior: an existing pivot file with matching dimensions and
+/// center count is reused, without checking its chunk count. Use [`generate_pq_pivots_fresh`]
+/// when the caller requires a newly trained codebook.
```
hildebrandmw: I understand this is left because of `build_pq`? I vote we either update `build_pq` to handle this itself, change the behavior of `build_pq` to always retrain, or get rid of it and no longer support the magic training skip.

hildebrandmw (commented):
I agree with @wuw92 that we should go all the way with simplifying. I have no argument about changing the behavior of `build_pq` if it means the core logic is simpler. No one on our side uses `build_pq`.

diskann-providers/src/model/pq/pq_construction.rs
xwj-ox: Thank you for your comments! I consolidated the two entry points into generate_pq_pivots, which now always trains and saves a new codebook. generate_pq_pivots_fresh is no longer needed.

build_pq remains available, but it now retrains even when a pivot file already exists. The regression tests now verify replacement of existing codebooks and preservation of existing files on invalid training input.

hildebrandmw (approved):
Looks good from `diskann-providers/diskann-tools`. Thanks!

PR #1303, opened 2026-07-30
fix: prefetch compilation on x86_64 without AVX2

Akvear:

The fallback `prefetch` implementation uses not(any(...)), which disables it on x86_64 when AVX2 is not enabled. This leaves no prefetch implementation available and causes compilation to fail. It is changed to not(all(...)).

This allows diskann-inmem to compile on x86_64 targets without AVX2 enabled.`


hildebrandmw (approved):
Thanks!

magdalendobson (approved):

PR #1305, opened 2026-07-31
Tweak distribution of `diskann-wide` float tests.

hildebrandmw:
The `Finite` distribution for generating floating point numbers in the `diskann-wide` random tests is slightly biased: the same value is used to select the kind of floating-point number *and* the value of that number. Practically speaking, the bias is relatively small - but it does mean that some valid floating-point bit patterns cannot be created.

This PR fixes that by splitting the kind generation from the value generation. As a side-effect, it also removes the tiny bias introduced by the modulo. This does mean that normal numbers have a slightly higher fraction of the generated values, but not significantly.

### DIFF
--- a/diskann-wide/src/test_utils/distribution.rs
+++ b/diskann-wide/src/test_utils/distribution.rs
@@ -93,7 +93,7 @@ macro_rules! finite {
                     //
                     // The mantissa is allowed to be all zeros.
                     (<$T>::EXPONENT_MASK | <$T>::MANTISSA_MASK, false, true)
-                } else if kind < 123 {
+                } else if kind < 122 {
                     // Generate a subnormal floating point number.
                     //
                     // The exponent must be all zero and the mantissa cannot be zero.
@@ -242,8 +242,8 @@ mod tests {
         T: TestDistribution,
     {
         let normal_weight = 116;
-        let subnormal_weight = 7;
-        let zero_weight = 5;
+        let subnormal_weight = 6;
+        let zero_weight = 6;
         let total_weight = normal_weight + subnormal_weight + zero_weight;
 
         let num_trials: i64 = 1_000_000;
--- a/diskann-wide/src/test_utils/distribution.rs
+++ b/diskann-wide/src/test_utils/distribution.rs
@@ -53,7 +53,7 @@ impl Layout for f32 {
 pub struct Finite;
 
 macro_rules! finite {
-    ($T:ty, $bits:ty) => {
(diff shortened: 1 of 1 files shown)


harsha-simhadri (approved):

harsha-simhadri (approved):

diskann-wide/src/test_utils/distribution.rs
```diff
@@ -235,9 +241,9 @@ mod tests {
     where
         T: TestDistribution,
     {
-        let normal_weight = 90;
-        let subnormal_weight = 5;
-        let zero_weight = 5;
+        let normal_weight = 116;
```
metajack: Should these be constants since you use them elsewhere as well?

metajack (approved):

diskann-wide/src/test_utils/distribution.rs
hildebrandmw: That's fair: done.

diskann-wide/src/test_utils/distribution.rs
```diff
@@ -68,26 +79,32 @@ macro_rules! finite {
...
+                //
+                // * the value that will be used to make the final floating point number
+                //   (the lower bits).
+                //
+                // * a selector for the kind of floating point number we are going to
+                //   generate (the upper bits).
+                //
+                // Generating a number twice as big allows us to perform just a single sample
+                // from the random number generator without biasing the result.
+                let twice: $twice = StandardUniform {}.sample(rng);
```
partychen, line 94: The aggregate frequency checks do not directly cover the regression this change fixes: selector/payload independence. An implementation that accidentally reused the lower bits for both could restore correlated or unreachable float bit patterns while preserving the tested category totals. Please add a deterministic RNG regression test for both f16 and f32 that controls the upper selector bits and lower payload bits independently, verifying that each half can vary without influencing the other.

hildebrandmw: Thanks, but I'd rather not. Such a test would overfit the implementation and not test the distribution within each category, which really requires statistical methods.

If this were a publicly exported distribution, such statistical tests would be warranted. However, the goal of this trait is to generate floating point numbers with sufficient distribution to catch high level numerical differences between implementations, for which even the previous biased implementation was sufficient. I don't believe the engineering ROI is there, particularly since the integers used to compose the final result are drawn from a uniform distribution already.

partychen (approved):

PR #1312, opened 2026-08-05
[diskann-wide] Add u64x8

hildebrandmw:
Another step towards better 1-bit ops on AVX512.

### DIFF
--- a/diskann-wide/src/emulated.rs
+++ b/diskann-wide/src/emulated.rs
@@ -544,7 +544,7 @@ macro_rules! impl_sumtree {
 impl_sumtree!(f32, 1, 2, 4, 8, 16);
 impl_sumtree!(i32, 4, 8, 16);
 impl_sumtree!(u32, 4, 8, 16);
-impl_sumtree!(i64, 2, 4);
+impl_sumtree!(i64, 2, 4, 8);
 impl_sumtree!(u64, 2, 4, 8);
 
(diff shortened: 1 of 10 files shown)


diskann-wide/src/emulated.rs
```diff
@@ -692,6 +692,7 @@ impl_splitjoin!(u8, 64 => 32);
 impl_splitjoin!(u32, 8 => 4);
 impl_splitjoin!(u32, 16 => 8);
 impl_splitjoin!(u64, 4 => 2);
+impl_splitjoin!(u64, 8 => 4);
```
partychen, line 706: The new Scalar/Emulated `u64x8` split/join path is not directly covered, while the V3, V4, and AArch64 implementations all have split/join tests. A Scalar-specific half-order or lane-order regression could therefore remain green. Please add `test_utils::ops::test_splitjoin!(Emulated<u64, 8> => Emulated<u64, 4>, <seed>, SC);` alongside the new emulated `u64x8` tests.

hildebrandmw: Sure. These tests were omitted because `Emulated`'s impl macro basically just called the implementation for fixed sized arrays. That said, it's easy to add tests, so might as well.

partychen (approved):

magdalendobson (approved):

PR #1315, opened 2026-08-06
PiPNN 1/6: extract shared RobustPrune

### LINKED ISSUE #878: Support PiPNN builds

SeliMeli:
## Purpose

This PR extracts the private Vamana RobustPrune selection loop. PiPNN can reuse this loop in a later PR.

The change adds no public API. It does not change Vamana pruning behavior.

## Main changes

- `graph::internal::prune::robust_prune` now owns the alpha-round selection loop.
- The function accepts one nearest-first cache of source distances and candidate vectors.
- A debug assertion checks cache order with the canonical neighbor distance ordering.
- `DiskANNIndex::occlude_list` still sorts candidates, loads vectors, maps IDs, and applies saturation.
- State updates, alpha rounds, and unsafe checks retain their original logic.

## Preserved behavior

- Candidate order remains unchanged.
- Excluded candidates remain positional `None` entries.
- Alpha progression and prune-kind behavior remain unchanged.
- Saturation order and duplicate handling remain unchanged.
- Provider errors remain in `occlude_list`.

## Review order

1. Review `diskann/src/graph/internal/prune.rs`.
2. Compare the extracted loop with the old loop in `index.rs`.
3. Review the call site in `DiskANNIndex::occlude_list`.
4. Confirm that preparation, ID mapping, and saturation remain outside the extracted function.

## Validation

- All `diskann` library tests pass.
- All-target Clippy passes with `-Dwarnings`.
- Existing Vamana prune scenarios pass.

## Stack

Stack 1/6. #1287 adds PiPNN numerical kernels.

### DIFF
--- a/diskann/src/graph/internal/prune.rs
+++ b/diskann/src/graph/internal/prune.rs
@@ -97,16 +97,14 @@ pub(crate) struct State {
 
 /// Select a degree-bounded neighbor set with Vamana RobustPrune.
 ///
-/// `pool` contains candidates in nearest-first order from the source point.
-/// `cache` contains the vector for the candidate at the same position. `None`
-/// excludes that candidate without changing positional alignment. `states` has
-/// one entry for each candidate position.
+/// `sorted_cache` stores each source distance and candidate vector in ascending
+/// source-distance order. `None` excludes that candidate without changing
+/// positional alignment. `states` has one entry for each candidate position.
 ///
 /// The function writes selected candidate indexes to `states[..result]` and
 /// returns `result`. The caller converts those indexes to graph IDs.
-pub(in crate::graph) fn robust_prune<I, V, D>(
-    pool: &SortedNeighbors<'_, I>,
-    cache: &[(f32, Option<V>)],
+pub(in crate::graph) fn robust_prune<V, D>(
+    sorted_cache: &[(f32, Option<V>)],
     states: &mut [State],
     degree: usize,
     alpha: f32,
@@ -114,9 +112,13 @@ pub(in crate::graph) fn robust_prune<I, V, D>(
     mut compute_distance: D,
 ) -> usize
 where
-    I: Eq,
     D: FnMut(&V, &V) -> f32,
 {
+    debug_assert!(
+        sorted_cache.is_sorted_by_key(|(distance, _)| distance),
+        "candidate cache must be sorted by source distance"
+    );
+
     let mut current_alpha = 1.0f32;
     let increment_factor = alpha.min(1.2);
 
@@ -136,11 +138,11 @@ where
     //
     // On the implementation side, we use `states` in the following way:
     //
-    // * `states[n].neighbor` is the **index** in `pool` of the `n`th **neighbor**.
+    // * `states[n].neighbor` is the **index** in `sorted_cache` of the `n`th **neighbor**.
     //   Note that a "neighbor" is a candidate that passes pruning.
     //
     //   Very important: to get the index `j` in the above description, we need to
-    //   check `pool[states[n].neighbor]`.
+    //   check `sorted_cache[states[n].neighbor]`.
     //
     //   This indexing naturally skips candidates `j` that have not been promoted to
     //   neighbors.
@@ -150,19 +152,17 @@ where
     //   excludes it from future consideration.
     //
     // * `states[i].last_checked` is the highest value of `n` against which the
-    //   occlude factor for `j = pool[states[n].neighbor]` has been checked.
+    //   occlude factor for `j = sorted_cache[states[n].neighbor]` has been checked.
     //
     //   The maximum value this should reach is `i`.
     //
     // Note that we use `states` for both "candidate" and "neighbor" tracking.
     let mut found = 0;
     while found < degree {
-        for (i, (_, neighbor)) in cache.iter().enumerate() {
+        for (i, (neighbor_distance, neighbor)) in sorted_cache.iter().enumerate() {
             if found >= degree {
                 break;
             }
-
-            let neighbor_distance = pool[i].distance();
 
             // The tracking states for candidate `i`.
             let State {
@@ -198,8 +198,8 @@ where
                 let result_position = states[last_checked as usize].neighbor.into_usize();
                 last_checked += 1;
 
-                // If the position of this result in `pool` is greater than or equal
-                // the current working position, then skip this candidate.
+                // If the position of this result in `sorted_cache` is greater than or equal
+                // to the current working position, then skip this candidate.
                 if result_position >= i {
                     debug_assert!(states.get(i).is_some(), "index {i} is out of bounds");
                     // SAFETY: We've already checked `states[i]`.
@@ -209,7 +209,7 @@ where
 
                 // Otherwise, compute the distance between the result and this neighbor
                 // and update the occlude factor.
-                let distance = match &cache[result_position] {
+                let distance = match &sorted_cache[result_position] {
                     (_, Some(v)) => compute_distance(neighbor, v),
                     (_, None) => f32::MAX,
                 };
--- a/diskann/src/graph/internal/prune.rs
+++ b/diskann/src/graph/internal/prune.rs
@@ -7,7 +7,12 @@ use thiserror::Error;
 
 use super::SortedNeighbors;
 
-use crate::{ANNError, error, graph::AdjacencyList, neighbor::Neighbor, utils::VectorId};
+use crate::{
+    ANNError, error,
+    graph::{AdjacencyList, config::PruneKind},
+    neighbor::Neighbor,
+    utils::{IntoUsize, VectorId},
+};
 
 /// Options provided to prune. See the field-level documentation for more details.
 ///
@@ -78,8 +83,8 @@ where
 
 /// Position-wise state tracking.
 ///
-/// Refer to the inline documentation in [`DiskANNIndex::occlude_list`] for documentation
-/// on the use of these fields.
+/// Refer to the inline documentation in [`robust_prune`] for documentation on the use
+/// of these fields.
 #[derive(Debug, Clone, Copy, Default)]
 pub(crate) struct State {
     /// The occlude factor for the pool item at the corresponding index.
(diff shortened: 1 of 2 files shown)


diskann/src/graph/robust_prune.rs
```diff
@@ -90,6 +91,117 @@ pub(crate) struct State {
...
+
+            let candidate = match candidate {
+                Some(candidate) => candidate,
+                None => {
+                    debug_assert!(
+                        states.get(index).is_some(),
+                        "index {index} is out of bounds"
+                    );
+                    // SAFETY: `index` comes from iterating `candidates`, which has
+                    // the same length as `states` by the caller contract.
```
hildebrandmw: This comment is not a valid safety argument as is. While the contract on the function states that `candidates` and `states` must have the same length, the function itself is not marked as unsafe and therefore unsafe code cannot rely on that contract being upheld. Nor is this invariant checked in the function preamble (and it need not be with the original justification).

The original comment gave a correct (albeit terse) reason why the indexing was safe: `states` was already accessed at the given index earlier and would have panicked if `index` was out of bounds.

Same for the other occurances.

If someone later turns the first destructure into `get_unchecked`, the cited justification still reads true while soundness silently evaporates.

diskann/src/graph/robust_prune.rs
```diff
@@ -90,6 +91,117 @@ pub(crate) struct State {
     pub(in crate::graph) neighbor: u16,
 }
 
+/// Select candidate positions using Vamana's existing alpha-round state machine.
+///
+/// The caller retains ownership of candidate preparation, allocation, ID translation,
+/// and saturation. `candidates` and `states` must have equal lengths, and the caller
+/// must enforce the existing `u16` candidate-position bound before entering.
```
hildebrandmw: Write docs as if they are the current contract ... because they are. If this merged all references to "existing" and "retains" are immediately stale and lose all relevance. This is not the only occurrence of this pattern.

diskann/src/graph/robust_prune.rs
```diff
@@ -90,6 +91,117 @@ pub(crate) struct State {
...
+{
+    let mut current_alpha = 1.0f32;
+    let increment_factor = alpha.min(1.2);
+    let mut selected = 0;
+
+    // `states` serves two roles, matching the original in-place implementation:
+    // `states[i]` tracks candidate `i`, while `states[..selected]` stores the
+    // positions of candidates already promoted to neighbors. `last_checked`
+    // indexes that selected prefix, allowing later alpha rounds to resume rather
+    // than repeat distance comparisons.
```
hildebrandmw: Please restore the deleted comments, relocating the loop-level ones alongside the code they describe. The `State` doc link needs repointing at the extracted function rather than restoring verbatim.

diskann/src/graph/robust_prune.rs
```diff
@@ -90,6 +91,117 @@ pub(crate) struct State {
     pub(in crate::graph) neighbor: u16,
 }
 
+/// Select candidate positions using Vamana's existing alpha-round state machine.
+///
+/// The caller retains ownership of candidate preparation, allocation, ID translation,
+/// and saturation. `candidates` and `states` must have equal lengths, and the caller
+/// must enforce the existing `u16` candidate-position bound before entering.
+pub(in crate::graph) fn robust_prune<V, D>(
+    candidates: &[(f32, Option<V>)],
```
hildebrandmw: This adds the real requirement that `candidates` must be sorted by increasing distance, but that is not documented. This is a new function where-as before the invariant was satisfied by construction. Please document the additional requirements at the very least.

diskann/src/graph/mod.rs
```diff
@@ -34,6 +34,7 @@ pub mod ext;
 pub use search::{KnnSearchError, RangeSearchError, Search};
 
 mod internal;
+mod robust_prune;
```
hildebrandmw: Call this module `prune`. It's one user already does that via aliasing. Also please keep it in the `internal` module.

hildebrandmw (changes requested):
This PR would have been easy to review, but at nontrivially regresses internal documentation and comments. For such important internal algorithms, please avoid changing comments unnecessarily, particularly when unsafe is involved.

Please restore the deleted or altered comments, making only the necessary updates for the new location. Also, please avoid renaming variables when just moving code around. A simpler diff stands on its own.

diskann/src/graph/internal/robust_prune.rs
```diff
@@ -0,0 +1,328 @@
...
+
+    let mut current_alpha = 1.0f32;
+    let increment_factor = alpha.min(1.2);
+    let mut selected = 0;
+
+    while selected < degree {
+        for (index, candidate) in candidates.iter().enumerate() {
+            if selected >= degree {
+                break;
+            }
```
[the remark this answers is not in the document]
SeliMeli: Intentionally unchanged in this refactor. Every `states[...]` access remains bounds-checked, so malformed/non-finite alpha can panic but cannot write out of bounds. Capping the selection count or changing alpha handling would alter pre-existing Vamana behavior and is outside this behavior-preserving extraction.

diskann/src/graph/robust_prune.rs
SeliMeli: Fixed in c2e6be62. `robust_prune` checks pool/cache/state length equality before the loop, and each unsafe comment again relies on the preceding checked `states[i]` access exactly as the original code did.

diskann/src/graph/robust_prune.rs
SeliMeli: Fixed. Source documentation now states the current contract in present tense. Existing prune types and their documentation remain in `graph/internal/prune.rs`.

diskann/src/graph/robust_prune.rs
SeliMeli: Fixed. Loop-level algorithm/state comments, inline comments, and variable terminology are restored beside the extracted loop. `State` now links to `robust_prune`.

diskann/src/graph/robust_prune.rs
SeliMeli: Fixed by making order part of the accepted type: `robust_prune` takes `&SortedNeighbors`, whose constructor sorts and caps by source distance. A raw unsorted candidate slice cannot be passed through this interface; the function docs state the ordering contract.

diskann/src/graph/mod.rs
SeliMeli: Fixed. The helper now remains `graph::internal::prune::robust_prune`; no new module or public API is introduced.

SeliMeli:
Addressed review in `c2e6be62`:

- kept module and existing types in `graph::internal::prune`;
- restored original loop comments, variable terminology, and checked-index unsafe justifications;
- changed the helper boundary to `&SortedNeighbors` so source-distance ordering is enforced by the existing type;
- added local pool/cache/state length checks before unchecked mutation;
- retained Vamana preparation, ID mapping, saturation, errors, and alpha behavior unchanged.

Validation: 311 `diskann` library tests, all-target Clippy, LSP diagnostics, and `git diff --check` pass.

SeliMeli:
Local IAI-Callgrind check against `main` (`16c7451f`):

- triangle 750→64: instructions +0.27259%, estimated cycles +0.29260%;
- occluding+saturation 750→64: instructions -0.01482%, estimated cycles +0.01124%.

Measured under WSL2 with the archived local regression fixture; no submitted benchmark target. Full record: `experiments/pipnn_local_regression/2026-08-07-wrapper-simplification-callgrind.md` in the local analysis workspace.

diskann/src/graph/robust_prune.rs
SeliMeli: Follow-up: the two new pool/cache/state `assert_eq!` checks have been removed. PR1 now preserves the pre-refactor `debug_assert!` / `get_unchecked_mut` sequence and its comments unchanged. `occlude_list` still constructs `cache` and `states` from the same `SortedNeighbors` length, so this extraction adds no new assertion or failure mode.

SeliMeli:
All review threads now point to superseded lines. The final extraction keeps the helper in `graph::internal::prune`, accepts `SortedNeighbors`, restores the original loop/state/unsafe comments and existing `debug_assert!` sequence, and adds no assertions or pruning behavior changes. I am resolving the outdated threads.

diskann/src/graph/internal/prune.rs
```diff
@@ -90,6 +95,169 @@ pub(crate) struct State {
...
+    //
+    // Note that we use `states` for both "candidate" and "neighbor" tracking.
+    let mut found = 0;
+    while found < degree {
+        for (i, (_, neighbor)) in cache.iter().enumerate() {
+            if found >= degree {
+                break;
+            }
+
+            let neighbor_distance = pool[i].distance();
```
wuw92: Could we keep this extraction identical to the old implementation by destructuring (neighbor_distance, neighbor) from `cache`, as the original loop did?

Once the distance comes from `cache`, pool has no remaining use inside `robust_prune`, so it can also be removed from the function signature. This avoids having two sources for the same candidate distance that could disagree.

wuw92 (approved):
LGTM. Since it is intended to be a focused extraction, I would keep the implementation as close to the original as possible and simplify the PR description.

diskann/src/graph/internal/prune.rs
hildebrandmw: +1 to this. Drop the `pool` argument in favor of a documented, sorted `cache`.

SeliMeli: Updated `robust_prune` to read the source distance from `cache` and removed the `pool` parameter. This restores the pre-extraction distance source.

SeliMeli: Follow-up: `cache` is now the only distance source. I retained `SortedNeighbors` as the sortedness witness required by this private API and added an alignment assertion between the sorted pool and cache. RobustPrune does not read a distance from the pool.

wuw92: Since `pool` is only used for a debug-only length check, it does not actually establish that the separately supplied `cache` has the same ordering. Could we drop `pool`, document that `cache` must be sorted by source distance, and add a debug assertion using the canonical distance ordering? This keeps the helper’s contract attached to the data it actually consumes.

SeliMeli: Fixed in c3fd81827. `robust_prune` now accepts only the nearest-first cache. The caller retains `pool` for cache construction and ID mapping. A debug assertion checks adjacent cache distances with the same canonical ordering used by `SortedNeighbors`.

diskann/src/graph/internal/prune.rs
```diff
@@ -90,6 +95,171 @@ pub(crate) struct State {
...
+    degree: usize,
+    alpha: f32,
+    prune_kind: PruneKind,
+    mut compute_distance: D,
+) -> usize
+where
+    D: FnMut(&V, &V) -> f32,
+{
+    debug_assert!(
+        sorted_cache
```
hildebrandmw: Use
```rust
.is_sorted_by_key(|(dist, _)| dist)
```

hildebrandmw (approved):
One more small suggestion. Thanks!

diskann/src/graph/internal/prune.rs
SeliMeli: Fixed in 6bf2139a6. The debug assertion now uses `.is_sorted_by_key(|(distance, _)| distance)`, and the extra distance-order helper is removed.

wuw92 (approved):

PR #1331, opened 2026-08-13
Supporting spherical quantization builds for the disk index

partychen:
## Summary

- add spherical quantization as a disk index build option
- train and wire the 1-bit spherical quantizer into one-shot and sharded in-memory builds
- account for spherical vector storage in build RAM estimation
- validate the currently supported 1-bit configuration
- add disk build and search coverage for L2, inner product, and cosine metrics

## Validation

- `cargo test -p diskann-disk` (250 unit tests and 2 doc tests passed)
- `cargo test -p diskann-disk test_spherical_disk_index_builder_with_metric` (2 passed)
- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/diskann-disk/src/build/builder/quantizer.rs
+++ b/diskann-disk/src/build/builder/quantizer.rs
@@ -4,7 +4,7 @@
  */
 //! Disk index quantizer implementation.
 use crate::data_model::GraphDataType;
-use diskann::ANNResult;
+use diskann::{ANNError, ANNResult};
 use diskann_providers::storage::{StorageReadProvider, StorageWriteProvider};
 use diskann_providers::{
     index::diskann_async::train_pq,
@@ -13,12 +13,11 @@ use diskann_providers::{
         FixedChunkPQTable, IndexConfiguration, MAX_PQ_TRAINING_SET_SIZE,
     },
     storage::{PQStorage, SQStorage},
-    utils::{create_thread_pool, BridgeErr, PQPathNames},
+    utils::{create_thread_pool, gen_random_slice, BridgeErr, PQPathNames},
 };
 use diskann_quantization::{
     algorithms::transforms::{TargetDim, TransformKind},
     alloc::GlobalAllocator,
-    error::Format,
     scalar::train::ScalarQuantizationParameters,
     spherical::{PreScale, SphericalQuantizer, SupportedMetric},
 };
@@ -44,7 +43,7 @@ impl BuildQuantizer {
         build_quantization_type: &QuantizationType,
         index_path_prefix: &str,
         index_configuration: &IndexConfiguration,
-        pq_storage: &PQStorage,
+        data_path: &str,
         storage_provider: &StorageProvider,
     ) -> ANNResult<Self>
     where
@@ -62,8 +61,9 @@ impl BuildQuantizer {
                     let mut rnd =
                         diskann_providers::utils::create_rnd_provider_from_optional_seed(seed)
                             .create_rnd();
-                    let (train_data, train_size, train_dim) = pq_storage
-                        .get_random_train_data_slice::<Data::VectorDataType, _>(
+                    let (train_data, train_size, train_dim) =
+                        gen_random_slice::<Data::VectorDataType, _>(
+                            data_path,
                             p_val,
                             storage_provider,
                             &mut rnd,
@@ -103,8 +103,9 @@ impl BuildQuantizer {
                 let rng = diskann_providers::utils::create_rnd_provider_from_optional_seed(
                     index_configuration.random_seed,
                 );
-                let (train_data_vector, train_size, train_dim) = pq_storage
-                    .get_random_train_data_slice::<Data::VectorDataType, _>(
+                let (train_data_vector, train_size, train_dim) =
+                    gen_random_slice::<Data::VectorDataType, _>(
+                        data_path,
                         p_val,
                         storage_provider,
                         &mut rng.create_rnd(),
@@ -127,20 +128,21 @@ impl BuildQuantizer {
                 Ok(Self::Scalar1Bit(WithBits::<1>::new(quantizer)))
             }
             QuantizationType::Spherical(SphericalBits::One) => {
+                let metric: SupportedMetric =
+                    index_configuration.dist_metric.try_into().bridge_err()?;
                 let rng = diskann_providers::utils::create_rnd_provider_from_optional_seed(
                     index_configuration.random_seed,
                 );
                 let mut rnd = rng.create_rnd();
-                let (train_data, train_size, train_dim) = pq_storage
-                    .get_random_train_data_slice::<Data::VectorDataType, _>(
+                let (train_data, train_size, train_dim) =
+                    gen_random_slice::<Data::VectorDataType, _>(
+                        data_path,
                         p_val,
                         storage_provider,
                         &mut rnd,
                     )?;
                 let train_data =
                     MatrixView::try_from(&train_data, train_size, train_dim).bridge_err()?;
-                let metric: SupportedMetric =
-                    index_configuration.dist_metric.try_into().bridge_err()?;
                 let quantizer = SphericalQuantizer::train(
                     train_data,
                     TransformKind::DoubleHadamard {
@@ -151,13 +153,7 @@ impl BuildQuantizer {
                     &mut rnd,
                     GlobalAllocator,
                 )
-                .map_err(|err| {
-                    diskann_error!(
-                        ErrorKind::IndexError,
-                        "Failed to train spherical quantizer: {}",
-                        Format(err)
-                    )
-                })?;
+                .map_err(|err| ANNError::new(err).context("Failed to train spherical quantizer"))?;
 
                 Ok(Self::Spherical1Bit(quantizer))
             }
--- a/diskann-disk/src/build/builder/quantizer.rs
+++ b/diskann-disk/src/build/builder/quantizer.rs
@@ -3,8 +3,6 @@
  * Licensed under the MIT license.
  */
 //! Disk index quantizer implementation.
-use std::sync::Arc;
-
 use crate::data_model::GraphDataType;
 use diskann::ANNResult;
 use diskann_providers::storage::{StorageReadProvider, StorageWriteProvider};
@@ -20,6 +18,7 @@ use diskann_providers::{
 use diskann_quantization::{
     algorithms::transforms::{TargetDim, TransformKind},
     alloc::GlobalAllocator,
+    error::Format,
     scalar::train::ScalarQuantizationParameters,
     spherical::{PreScale, SphericalQuantizer, SupportedMetric},
 };
@@ -28,15 +27,14 @@ use tracing::info;
 
 use crate::{
     error::{diskann_error, ErrorKind},
-    QuantizationType,
+    QuantizationType, SphericalBits,
 };
 
 /// Quantizer types used specifically for async disk index building.
-#[derive(Clone)]
 pub enum BuildQuantizer {
     NoQuant(NoStore),
     Scalar1Bit(WithBits<1>),
-    Spherical1Bit(Arc<SphericalQuantizer>),
+    Spherical1Bit(SphericalQuantizer),
     PQ(FixedChunkPQTable),
 }
 
@@ -128,14 +126,7 @@ impl BuildQuantizer {
 
                 Ok(Self::Scalar1Bit(WithBits::<1>::new(quantizer)))
             }
-            QuantizationType::Spherical(nbits) => {
-                if nbits != 1 {
-                    return Err(diskann_error!(
-                        ErrorKind::IndexConfigError("build_quantization_type"),
-                        "Spherical quantization is only supported for 1 bit",
-                    ));
-                }
-
+            QuantizationType::Spherical(SphericalBits::One) => {
                 let rng = diskann_providers::utils::create_rnd_provider_from_optional_seed(
                     index_configuration.random_seed,
                 );
@@ -152,7 +143,7 @@ impl BuildQuantizer {
                     index_configuration.dist_metric.try_into().bridge_err()?;
                 let quantizer = SphericalQuantizer::train(
                     train_data,
-                    TransformKind::PaddingHadamard {
+                    TransformKind::DoubleHadamard {
                         target_dim: TargetDim::Natural,
                     },
                     metric,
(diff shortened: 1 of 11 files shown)


diskann-quantization/src/spherical/quantizer.rs
```diff
@@ -156,6 +156,11 @@ where
         self.shift.allocator()
     }
 
+    /// Return an independently allocated copy of this quantizer.
+    pub fn try_clone(&self) -> Result<Self, AllocatorError> {
+        <Self as TryClone>::try_clone(self)
+    }
```
hildebrandmw, line 162: Please leave this just as the trait method. If you need `try_clone`, either import the trait or use fully qualified syntax.

diskann-disk/src/build/configuration/quantization_types.rs
```diff
@@ -38,6 +38,9 @@ pub enum QuantizationType {
         /// be greater than 1.0.
         standard_deviation: Option<Positive<f64>>,
     },
+
+    /// Spherical quantization bit width. Disk-index builds currently support only 1 bit.
+    Spherical(usize),
```
hildebrandmw: This is a perfect example of where you can use types to make this code far more robust. (please excuse typos and brevity here, I am typing with one hand and everything takes like 8x longer). The problem with a raw `usize` is that parsing needs to know that only 1 bit is supported, users need to read the docs, and all uses of this field need to revalidate that it is just "1". We can do better like this:
```rust
enum SphericalBits {
    One,
}

impl SphericalBits {
    fn as_usize(&self) -> usize {
        match self {
            Self::One => 1,
        }
    }
}

impl FromStr for SphericalBits {
    // Reject unsupported values and parse errors.
}
```
This does a lot for you:
* Structurally encodes the restriction instead of spreading in across all users and via documentation.
* Matching on `SphericalBits` means if you ever add support for 2 or 4 bits, the compiler helpfully shows you all the parts you need to update.
* Uses of `SphericalBits` don't need to validate. That work is already done.

diskann-disk/src/build/builder/quantizer.rs
```diff
@@ -120,6 +128,48 @@ impl BuildQuantizer {
...
+                    )?;
+                let train_data =
+                    MatrixView::try_from(&train_data, train_size, train_dim).bridge_err()?;
+                let metric: SupportedMetric =
+                    index_configuration.dist_metric.try_into().bridge_err()?;
+                let quantizer = SphericalQuantizer::train(
+                    train_data,
+                    TransformKind::PaddingHadamard {
+                        target_dim: TargetDim::Natural,
+                    },
```
hildebrandmw, line 150: Use `DoubleHadamard`. It's better across all dimensions.

diskann-disk/src/build/builder/quantizer.rs
```diff
@@ -120,6 +128,48 @@ impl BuildQuantizer {
...
+                    metric,
+                    PreScale::ReciprocalMeanNorm,
+                    &mut rnd,
+                    GlobalAllocator,
+                )
+                .map_err(|err| {
+                    diskann_error!(
+                        ErrorKind::IndexError,
+                        "Failed to train spherical quantizer: {}",
+                        err
```
hildebrandmw: Might want to wrap in `diskann_quantization::error::Format` to render the full source chain.

diskann-disk/src/build/builder/quantizer.rs
```diff
@@ -120,6 +128,48 @@ impl BuildQuantizer {
...
+                let (train_data, train_size, train_dim) = pq_storage
+                    .get_random_train_data_slice::<Data::VectorDataType, _>(
+                        p_val,
+                        storage_provider,
+                        &mut rnd,
+                    )?;
+                let train_data =
+                    MatrixView::try_from(&train_data, train_size, train_dim).bridge_err()?;
+                let metric: SupportedMetric =
+                    index_configuration.dist_metric.try_into().bridge_err()?;
```
hildebrandmw: I wonder if a misconfiguration here should be caught early? Alternatively, if `CosineNormalized`is used, it can be remapped to `SupportedMetric::Cosine` without performance penalty.

diskann-disk/src/build/builder/quantizer.rs
hildebrandmw: The code seems to compile fine without `Clone`. It could be removed to avoid the need for `Arc`, but I will let the crate maintainers weigh in.

hildebrandmw (commented):
Added some comments on the use of spherical quantization and some maintenance suggestions. Please get a review from the maintainers of `diskann-disk` regarding how this feature fits in architecturally.

diskann-disk/src/build/builder/quantizer.rs
```diff
@@ -120,6 +134,42 @@ impl BuildQuantizer {
 
                 Ok(Self::Scalar1Bit(WithBits::<1>::new(quantizer)))
             }
+            QuantizationType::Spherical(SphericalBits::One) => {
+                let rng = diskann_providers::utils::create_rnd_provider_from_optional_seed(
+                    index_configuration.random_seed,
+                );
+                let mut rnd = rng.create_rnd();
+                let (train_data, train_size, train_dim) = pq_storage
+                    .get_random_train_data_slice::<Data::VectorDataType, _>(
```
wuw92: `BuildQuantizer::train` only uses `PQStorage` to recover the source `data_path`, and `get_random_train_data_slice` immediately delegates to the quantizer-neutral `gen_random_slice`. Could we pass `data_path: &str` instead of `pq_storage: &PQStorage` here and sample directly, so SQ, spherical, and future build quantizers do not depend on PQ output storage?

diskann-disk/src/build/builder/quantizer.rs
```diff
@@ -120,6 +134,42 @@ impl BuildQuantizer {
...
+                    &mut rnd,
+                    GlobalAllocator,
+                )
+                .map_err(|err| {
+                    diskann_error!(
+                        ErrorKind::IndexError,
+                        "Failed to train spherical quantizer: {}",
+                        Format(err)
+                    )
+                })?;
```
wuw92: nit: Could we preserve `TrainError` as the structured source and attach the operation as context here?

```suggestion
                .map_err(|err| {
                    diskann::ANNError::new(err)
                        .context("Failed to train spherical quantizer")
                })?;
```

The current `Format(err)` eagerly flattens the source chain into the tagged error message. Keeping `TrainError` inside `ANNError` preserves downcasting and its source chain while retaining the useful training context.

diskann-disk/src/build/builder/tests.rs
```diff
@@ -7,33 +7,59 @@
...
         build_type: BuildType,
     ) {
         let index_path_prefix = "/disk_index_build/test_disk_index_build".to_string();
 
         run_disk_index_builder_test(index_path_prefix, use_sharded_build, build_type);
     }
 
+    #[rstest]
+    fn test_spherical_disk_index_builder_with_metric(
+        #[values(Metric::InnerProduct, Metric::Cosine, Metric::CosineNormalized)] metric: Metric,
```
wuw92: The SIFT fixture vectors are not normalized; `Metric::CosineNormalized` requires normalized data and queries.

wuw92 (commented):
Thanks for your change to support spherical in disk path. Is spherical disk-index support intentionally limited to 1-bit quantization, or is support for other bit widths planned?

diskann-quantization/src/spherical/quantizer.rs
partychen: `TryClone` is currently crate-private, so `diskann-disk` cannot import the trait or use fully qualified syntax across the crate boundary.

The trait also explicitly documents why it should not be exposed:
```rust
/// Keep this `pub(crate)` for now because we do not want general users of the crate
/// relying on the current implementations for [`Poly`]. In particular, the base case should
/// be `Poly<T> where T: TryClone` instead of `Poly<T> where T: Clone`.

```
Making `TryClone` public here would contradict that intent and unnecessarily expand the public API. I propose keeping the narrow inherent `SphericalQuantizer::try_clone()` wrapper instead. Does that sound reasonable?

diskann-disk/src/build/configuration/quantization_types.rs
partychen: Accepted. `SphericalBits` now structurally limits disk builds to the supported one-bit mode, parsing rejects other widths, and exhaustive matches will identify the call sites that need updates when more widths are added.

partychen:
> Thanks for your change to support spherical in disk path. Is spherical disk-index support intentionally limited to 1-bit quantization, or is support for other bit widths planned?

The 1-bit limitation is intentional for this PR. The immediate goal is to provide a spherical replacement for the 1-bit scalar-quantized disk builds used in production. The underlying async spherical provider already supports 1, 2, and 4 bits, and the typed `SphericalBits` configuration leaves room to add those widths later, but enabling them would expand the build configuration and test matrix beyond the current request.

diskann-disk/src/build/builder/quantizer.rs
partychen: We chose the early-validation path. The metric conversion now happens before sampling or training, so `CosineNormalized` is rejected immediately as unsupported for spherical builds.

diskann-disk/src/build/builder/quantizer.rs
partychen: Nice to get it. Thanks!

diskann-disk/src/build/builder/quantizer.rs
partychen: Thanks. I initially used `Format`, but switched to `ANNError::new(err).context(...)` following the updated `ANNError` guidance. This still renders the full source chain while preserving `TrainError` for downcasting instead of flattening it into a string.

diskann-disk/src/build/builder/quantizer.rs
partychen: Sure. Updated!

diskann-disk/src/build/builder/quantizer.rs
partychen: Yes. Changed!

diskann-disk/src/build/builder/quantizer.rs
partychen: Fix it!

diskann-quantization/src/spherical/quantizer.rs
hildebrandmw: Oops, you're right. Yes, making this an inherent method is the move. One small request: Can you make the actual implementation live in the inherent method and direct the `TryClone` impl to that instead of the other way around?

diskann-disk/src/build/builder/quantizer.rs
hildebrandmw: Just be a little careful about keeping consistent error styles within `diskann-disk`. Not that I think anyone is actually downcasting error types though.

diskann-disk/src/build/builder/quantizer.rs
hildebrandmw: Fair enough. Please coordinate with Wei though on keeping consistent error styles.

hildebrandmw (approved):

diskann-disk/src/build/builder/tests.rs
partychen: removed

diskann-quantization/src/spherical/quantizer.rs
partychen: resolved

wuw92 (approved):

diskann-disk/src/build/builder/quantizer.rs
```diff
@@ -97,8 +103,9 @@ impl BuildQuantizer {
                 let rng = diskann_providers::utils::create_rnd_provider_from_optional_seed(
                     index_configuration.random_seed,
                 );
-                let (train_data_vector, train_size, train_dim) = pq_storage
-                    .get_random_train_data_slice::<Data::VectorDataType, _>(
+                let (train_data_vector, train_size, train_dim) =
+                    gen_random_slice::<Data::VectorDataType, _>(
+                        data_path,
```
arkrishn94, line 108: Can you confirm that there is no difference in the edge case behavior of this random slice generator?

diskann-disk/src/build/builder/quantizer.rs
```diff
@@ -120,6 +127,36 @@ impl BuildQuantizer {
 
                 Ok(Self::Scalar1Bit(WithBits::<1>::new(quantizer)))
             }
+            QuantizationType::Spherical(SphericalBits::One) => {
+                let metric: SupportedMetric =
+                    index_configuration.dist_metric.try_into().bridge_err()?;
```
arkrishn94, line 132: Just for my knowledge, remind me what this metric check is for?

arkrishn94 (commented):
Thanks Junkui, it mostly looks good to me. Just had a couple of clarifying questions for my understanding.

suri-kumkaran (approved):

diskann-disk/src/build/builder/quantizer.rs
partychen: Yes. I traced both paths: `PQStorage::get_random_train_data_slice` was a thin wrapper around `gen_random_slice`, passing its stored source `data_path` together with the same sampling probability, storage provider, and RNG. The new call invokes that same function with the same source path and arguments. Therefore the sampling behavior, RNG consumption, error propagation, and edge-case handling are unchanged; this only removes the unnecessary dependency on PQ output storage.

diskann-disk/src/build/builder/quantizer.rs
partychen: This is an early compatibility check and type conversion. The disk configuration uses the broader `Metric` type, while `SphericalQuantizer::train` accepts only `SupportedMetric`. The conversion allows L2, inner product, and cosine, and rejects unsupported metrics such as `CosineNormalized` before sampling or training begins. It does not change the supported metric’s behavior; it validates and narrows the type.

diskann-disk/src/build/builder/quantizer.rs
arkrishn94: Awesome! Do we need `PQStorage::get_random_train_data_slice` anymore then?

diskann-disk/src/build/builder/quantizer.rs
arkrishn94: Ah yes, I forgot that this is a check needed for training the quantizer.

arkrishn94 (approved):

diskann-disk/src/build/builder/quantizer.rs
partychen: Yes, it is still used by the actual PQ generation path, `diskann-tools`, and PQ-specific tests, where sampling through a `PQStorage` that owns the source path remains coherent. This change only removes it from the generic build-quantizer path, so I would keep the method for now rather than broaden this PR into a PQ API cleanup.

PR #1333, opened 2026-08-13
Redesign the multi-vector kernels to decouple layout, micro-kernel and reduction

### LINKED ISSUE #989: Add `Reduce` abstraction to support quantizations and modify `tiled_reduce` accordingly
# Add `Reduce` abstraction to support quantizations and modify `tiled_reduce` accordingly

## Description
Building on the foundational `tiled_reduce` pipeline, this issue serves two main architectural purposes. First, it extracts the initially hardcoded reduction logic into a dedicated `Reduce` abstraction, which is strictly necessary to support multi-vector quantizations (handling scaling factors, offsets, and lookup tables).

Second, it improves the `tiled_reduce` traversal and reduction flow to better support quantized execution paths and extensible reduction strategies. The new abstraction boundary enables reduction implementations to specialize accumulation behavior, metadata handling, and intermediate storage layouts without coupling those concerns directly to the core kernel compute pipeline.

## Tasks
- [ ] Design the `Reduce` abstraction to decouple the similarity reduction step from the core `Kernel` compute, ensuring it can handle quantization metadata efficiently.
- [ ] Refactor `tiled_reduce` to route reduction and accumulation logic through the new `Reduce` abstraction.
- [ ] Implement reduction interfaces capable of supporting quantized accumulation paths, including scaling factors, offsets, and lookup-table based reductions where required.
- [ ] Refactor the existing `f32` and `f16` implementations to utilize the new `Reduce` abstraction.
- [ ] Validate that the abstraction boundary does not introduce a performance regression against the baseline AVX2 metrics.

suri-kumkaran:
Rebuilds `diskann-quantization/src/multi_vector/distance/kernels/` around a contract instead
of a monolith. Same results, same public API, same tests. Only the internals change.

Big thanks to @hildebrandmw for the valuable design insights and constant support in shaping and refining this design.

## Why

The reduction was baked into the kernel. `tiled_reduce.rs` was 806 lines in which the cache
tiling, the inner FMA loop and the max-reduction were one function body, written out once per
element type per instruction set. To feed it quantized vectors, or compute anything other than
MaxSim, there was no seam to cut along. You copied the file.

So the redesign cuts it into four pieces, each with an obvious owner:

- a walk turns a matrix into blocks small enough to sit in cache
- a leaf is the SIMD micro-kernel that multiplies one pair of those blocks
- a drain decides what the finished accumulator means (for MaxSim, the max down each query row)
- one `drive` function owns the loop over blocks, written once for everybody

A new element type is now just a new walk, a new reduction just a new drain, and neither
touches a leaf or the loop. f16 already shows this: it has no kernel code left at all, just a
walk that widens to f32 on the way in. It used to be a separate path.

The other reason is safety. Raw pointers ran through 5 of the 8 old files, and `unsafe`
through 6 of them, because the layout helpers, the tiling loop and the leaves each did their
own address arithmetic. Now `unsafe` appears in two files only, the two leaves, as three blocks
each, every one wrapping a single load, read or store.

## What

Gone: `tiled_reduce.rs`, `layouts.rs`, `reduce.rs`, `f32/{mod,scalar,v3}.rs`. In their place,
roughly in dependency order:

- `mod.rs`, the four traits above plus the cache planner and `drive`
- `tiles.rs`, turning a matrix into blocks and panels
- `strip.rs`, the accumulator that leaves write into and drains read out of
- `leaves/v3.rs` and `leaves/scalar.rs`, the two micro-kernels
- `float.rs`, the f32 instantiation and its tests
- `f16.rs`, the widening walks

Worth knowing going in: `drive` only hands out ordinals, like "A-panel 3, B-panels 8..12",
never a stride or an address. That's what lets a drain whose panels are a different width
reuse the same loop.

One behaviour change to flag. Budgets and panel geometry are unchanged, but the B-panel count
now charges the accumulator strip against L1, which the old planner never counted. More
accurate, though it shrinks the B tile at small dims: 31 panels to 24 at dim 64, 13 to 12 at
dim 128, no change from 256 up.

Six files under `distance/` appear in the diff with nothing but a licence header change. The
module was using `//` headers where the rest of the repo uses `/* */`. Fixed while in here.

## Performance

Against `main` on a shared machine, 30 alternating rounds per side, taking the best round of
each cell. Median +1.25%, p90 +4.45%.

Eight of the nine shapes land between +0.44% and +2.09%, at or near the noise floor.
The ninth, the smallest, does not. At 8 queries x 32 docs x dim 128:

- f16 v4, 1160us to 1251us, +7.8%
- f32 scalar, 661us to 701us, +6.1%
- f16 v3, 1175us to 1237us, +5.3%
- f32 v3, 809us to 845us, +4.5%

That is the shape with the least work to amortise a fixed cost, and there is a new fixed
cost. The old kernel fused the multiply and the reduction, so the accumulator lived in
vector registers and folded into the caller's scratch on the way out; it was never
materialised. Splitting accumulate from drain means it has to exist somewhere the drain can
read it, which is one `vec![0.0f32; plan.strip_len()]` per call. That is the price of the
seam, paid once per call however much work follows.

Fixable, but not here: the strip is a fixed-size buffer with a known bound, so it can be
hoisted to the caller or held on the stack. Left as a follow-up to keep this PR a
restructure.

### Running it

The benchmark is behind a feature flag. Run each side, then compare.

    cargo run --release -p diskann-benchmark --features multi-vector -- run --input-file diskann-benchmark/perf_test_inputs/multi-vector.json --output-file before.json

Switch revision, rebuild, run again into `after.json`, then:

    cargo run --release -p diskann-benchmark --features multi-vector -- check run --tolerances diskann-benchmark/perf_test_inputs/multi-vector-tolerance.json --input-file diskann-benchmark/perf_test_inputs/multi-vector.json --before before.json --after after.json

### Reading it

One run per side is not enough. A shared machine drifts between faster and slower states and
holds each for longer than a measurement takes, so a single pair of runs will breach the
tolerance in places with no code change at all. The drift is one-sided, it can only make
things slower, so repeat both sides and take the minimum, and do enough runs that the numbers
stop moving. A bigger `num_measurements` or a looser tolerance is not a substitute.

The input carries `reference` rows, the same code on both sides. Their spread is your error
bar; anything smaller is unmeasured.

## Review order

1. `kernels/mod.rs` first. It's the contract; everything else implements it.
2. `kernels/tiles.rs`
3. `kernels/strip.rs`. Short, but mind the axes: a doc is a row of the input and a column here.
4. `kernels/leaves/`. All the unsafe in the PR, ~120 lines each. Worth reading the SAFETY
   comments properly.
5. `kernels/float.rs`, where it comes together, plus the test suite.
6. `kernels/f16.rs`, the proof it composes: a second element type, no new leaf.
7. `factory.rs`, the wiring and the only caller-facing diff.


diskann-quantization/src/multi_vector/distance/kernels/float.rs
```diff
@@ -0,0 +1,516 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+//! f32 MaxSim: the accumulator is already the score, so the drain is a bare reduction.
+
+use core::ops::Range;
```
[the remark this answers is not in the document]
suri-kumkaran: Not requried.

diskann-quantization/src/multi_vector/distance/kernels/leaves/v3.rs
```diff
@@ -0,0 +1,124 @@
...
+/// Panics if `b` does not hold exactly `UNROLL` rows of `a`'s contraction length — the
+/// extent every unchecked access below relies on.
+#[inline(always)]
+pub(crate) fn f32_store_microkernel<const UNROLL: usize, L: Length>(
+    arch: V3,
+    a: QueryPanel<'_, f32, A_PANEL>,
+    b: DocPanel<'_, f32, B_PANEL, L>,
+    mut out: Slot<'_, f32, A_PANEL, B_PANEL>,
+) {
+    const { assert!(UNROLL >= 1 && UNROLL <= B_PANEL) }
```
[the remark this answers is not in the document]
suri-kumkaran: Not required.

diskann-quantization/src/multi_vector/distance/kernels/leaves/v3.rs
```diff
@@ -0,0 +1,124 @@
...
+        for (acc_jr, dst) in acc_j.iter().zip(col.chunks_exact_mut(f32s::LANES)) {
+            // SAFETY: `chunks_exact_mut` yields exactly `LANES` elements.
+            unsafe { acc_jr.store_simd(dst.as_mut_ptr()) };
+        }
+    }
+}
+
+/// Fold `acc`'s columns into the running per-A-row maxima in `state`, one A-panel wide.
+#[inline(always)]
+pub(crate) fn fold_columns(arch: V3, acc: &[f32], state: &mut [f32]) {
```
hildebrandmw: This spills pretty heavily to the stack (all 8 vector registers get spilled). Generally, trying to rely on LLVM to correctly unroll loops like this where things are in arrays is a recipe for stack spilling. I had better luck just manually unrolling the implementation (shaving 256 bytes of stack space).

Also a nit: `fold_columns` is a pretty generic name for a rather specific operation.

diskann-quantization/src/multi_vector/distance/kernels/tiles.rs
```diff
@@ -0,0 +1,341 @@
...
+        self.0
+    }
+
+    /// The contraction length.
+    pub(super) fn k(&self) -> usize {
+        self.0.len() / AR
+    }
+}
+
+pub(super) struct QueryPanels<'a, T, const AR: usize>(ChunksExact<'a, T>);
```
hildebrandmw: So, one thing to consider that came up while I was inspecting the generated code. These iterators (while better than raw pointer soup) kind of introduce a death by a thousand cuts when it comes to code generation.

* `ChunksExact` has a stack footprint of 40 bytes. But for `QueryPanels`, we know the remainder is not used. We could easily drop this to 24 bytes (and maybe even 16 bytes if `k` were supplied externally).
* `DocPanels` has it worse since it has a `ChunksExact` and a `k` when the `TailIterator` protocol makes it possible to derive the remainder naturally rather than eagerly materializing it.
* We could shave 8 bytes off `Cursor` by not making it resettable and instead creating a new `Cursor` on demand. Tile walkers need not be resettable if we can cheaply remake them.
* `Slots` has a slice when its size is statically known (shaving another 8 bytes)
* `Slots` suffers the same size bloat caused by `ChunksExact`.

Now, LLVM may be able to elide some of these fields. But all of this adds up to >700 bytes of stack space for the V3 kernel (or around 450 bytes after fixing `fold_columns`) with lots of messing around with the stack.

This isn't necessarily a blocker for this PR, but it is something to consider for the overall flow. The combination of extra stack space + panicking code-paths can affect overall code generation. While the code for the innermost micro-kernel looks good, it's the surrounding bit that worries me.

Usually stuff like this doesn't matter so much, but for a matrix/matrix style kernel, these optimizations can add up.

diskann-quantization/src/multi_vector/distance/kernels/tiles.rs
```diff
@@ -0,0 +1,341 @@
...
+
+use super::{NoTail, Paneled, TailIterator, TileAt, TileWalk};
+use crate::bits::{Dynamic, Length, Static};
+use crate::multi_vector::{BlockTransposedRef, MatRef, Standard};
+
+// ── Query side ───────────────────────────────────────────────────
+
+/// One block-transposed block: `AR` rows × `k` columns, column-major within the block.
+#[derive(Clone, Copy)]
+pub(super) struct QueryPanel<'a, T, const AR: usize>(&'a [T]);
```
hildebrandmw: Can we use more descriptive names? A `QueryPanel` is how this happens to be used, but not what it is. Maybe `BlockTransposedPanel` instead?

diskann-quantization/src/multi_vector/distance/kernels/tiles.rs
```diff
@@ -0,0 +1,341 @@
...
+
+impl<T, const AR: usize> TailIterator for QueryPanels<'_, T, AR> {
+    type Tail = NoTail;
+
+    fn tail(self) -> Option<NoTail> {
+        None
+    }
+}
+
+/// A run of whole blocks — block-transposed storage is padded to `AR`, hence [`NoTail`].
```
hildebrandmw: Nit: the `NoTail` comment probably belongs elsewhere?

diskann-quantization/src/multi_vector/distance/kernels/tiles.rs
```diff
@@ -0,0 +1,341 @@
...
+    k: usize,
+}
+
+impl<'a, T, const BR: usize> DocWalk<'a, T, BR> {
+    /// # Panics
+    ///
+    /// Panics if `docs` has zero-length rows.
+    pub(super) fn new(docs: MatRef<'a, Standard<T>>, b_panels: usize) -> Self {
+        let k = docs.vector_dim();
+        assert!(k > 0, "DocWalk requires a non-empty contraction");
```
hildebrandmw: Suggestion: Rust has the wonderful `NonZeroUsize` type. Use it internally. This removes needing to constantly recheck it - and Rust/LLVM know that it cannot be zero, so divides in the future won't emit code to handle the divide-by-zero case.

diskann-quantization/src/multi_vector/distance/kernels/strip.rs
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
```
hildebrandmw: Nit: why is the header style in the `distances` module different from the rest of the repo?

diskann-quantization/src/multi_vector/distance/kernels/strip.rs
```diff
@@ -0,0 +1,104 @@
...
+
+//! The accumulator a fill writes and a drain reads.
+//!
+//! Rows are query rows and one column is one doc — note that a doc arrives as a *row* of
+//! the input and lands as a *column* here. Column-major over the whole strip: column `c`
+//! occupies `[c * AR, (c + 1) * AR)`, and slot `p` covers columns `p * BR ..`, so a drain
+//! can address a run of columns without knowing which slot produced them.
```
hildebrandmw: A picture would make this infinitely clearer. And why not call it row major with the axes flipped?

diskann-quantization/src/multi_vector/distance/kernels/strip.rs
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+//! The accumulator a fill writes and a drain reads.
```
hildebrandmw, line 6: This is only for one particular `Fill` and `Drain` combination though, right? The documentation here makes it sound like a global property.

hildebrandmw (commented):
Thanks - I took one pass over this. This is a step in the right direction, but there is one large theme that I think would really help a future review cycle, and that is documentation. More concise and self-contained documentation would go a long way toward making this easier to review and maintain.

Left hand arguments are referred to as `query` and `A` while the right is referred to as `document` or `B`, and these are conflated throughout the stack. I'd recommend sticking to just `A` and `B` for everything but the uppermost layers.

As an example, the documentation for `QueryTile` is
```
A run of whole blocks — block-transposed storage is padded to `AR`, hence [`NoTail`].
```
But rewriting as
```
A view over consecutive blocks from a [`BlockTransposedRef<T, AR, 1>`].

Its [`Paneled`] implementation yields one [`QueryPanel`] per block.
```
with the following on `QueryPanel`:
```
A single sub-block of a [`BlockTranpose`] containing `AR` rows
in a **column-major** layout.
```
ties the implementation to the logical data structure its operating over (and it's corresponding documentation), while providing enough breadcrumbs that someone reading the code here can piece things together a little more.

diskann-quantization/src/multi_vector/distance/kernels/leaves/v3.rs
suri-kumkaran: Manually unrolled it. The accumulators are named locals spread across a fixed number of independent max chains now, instead of an array walk I was hoping LLVM would unroll. Renamed to `max_into_rows`.

diskann-quantization/src/multi_vector/distance/kernels/tiles.rs
suri-kumkaran: Fixed.

diskann-quantization/src/multi_vector/distance/kernels/tiles.rs
suri-kumkaran: Nice idea, tried to adapt wherever I can.

diskann-quantization/src/multi_vector/distance/kernels/tiles.rs
suri-kumkaran: Make sense, done.

diskann-quantization/src/multi_vector/distance/kernels/strip.rs
suri-kumkaran: Right, tried my hand at the art!

diskann-quantization/src/multi_vector/distance/kernels/strip.rs
suri-kumkaran: Fixed.

diskann-quantization/src/multi_vector/distance/kernels/strip.rs
suri-kumkaran: Fixed everything.

diskann-quantization/src/multi_vector/distance/kernels/tiles.rs
suri-kumkaran: Sizes now, none of them carrying a `ChunksExact` (40 bytes):

- `BlockTransposedPanels` 24, slice + stride. Not 16, that would put the stride in the leaf signature.
- `RowMajorPanels` 24, slice + `k`, tail derived in `tail()` rather than materialised.
- `Slots` 16, via `split_first_chunk_mut`. Not 8, the slot count is runtime.
- `Cursor` unchanged, per your reply on the `Widen` buffer.

suri-kumkaran:
> Thanks - I took one pass over this. This is a step in the right direction, but there is one large theme that I think would really help a future review cycle, and that is documentation. More concise and self-contained documentation would go a long way toward making this easier to review and maintain.
> 
> Left hand arguments are referred to as `query` and `A` while the right is referred to as `document` or `B`, and these are conflated throughout the stack. I'd recommend sticking to just `A` and `B` for everything but the uppermost layers.
> 
> As an example, the documentation for `QueryTile` is
> 
> ```
> A run of whole blocks — block-transposed storage is padded to `AR`, hence [`NoTail`].
> ```
> 
> But rewriting as
> 
> ```
> A view over consecutive blocks from a [`BlockTransposedRef<T, AR, 1>`].
> 
> Its [`Paneled`] implementation yields one [`QueryPanel`] per block.
> ```
> 
> with the following on `QueryPanel`:
> 
> ```
> A single sub-block of a [`BlockTranpose`] containing `AR` rows
> in a **column-major** layout.
> ```
> 
> ties the implementation to the logical data structure its operating over (and it's corresponding documentation), while providing enough breadcrumbs that someone reading the code here can piece things together a little more.

Thanks @hildebrandmw for the guidance on the documentation, tried to fix all.

diskann-quantization/src/multi_vector/distance/kernels/leaves/v3.rs
```diff
@@ -0,0 +1,118 @@
...
+            // SAFETY: `as_chunks_mut` yields exactly `LANES` elements.
+            unsafe { acc_jr.store_simd(dst.as_mut_ptr()) };
+        }
+    }
+}
+
+/// Merge each column of `acc` into the running per-A-row maxima in `state`.
+///
+/// The maxima are held in registers for the whole sweep, so `state` is read once on entry
+/// and written once on exit, not once per column.
```
hildebrandmw, line 84: For documentation of internal methods like this, please state what it actually does, not how it's used. For example:
```
Computes `state[a] = max(state[a], acc[0][a], acc[1][a], ...)`
```
is much clearer.

diskann-quantization/src/multi_vector/distance/kernels/leaves/v3.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    let (rows, _) = state.as_chunks_mut::<{ f32s::LANES }>();
+    for (m, row) in merged.iter().zip(rows) {
+        *row = m.max_simd(f32s::from_array(arch, *row)).to_array();
+    }
+}
+
+/// Merge up to [`WAYS`] consecutive columns, one per chain.
+#[inline(always)]
+fn max_into_chains(arch: V3, chains: &mut [[f32s; REGS]; WAYS], src: &[[f32; A_PANEL]]) {
```
hildebrandmw, line 111: You could probably use [`as_flattened`](https://doc.rust-lang.org/std/primitive.slice.html#method.as_flattened) to simplify this considerably.

diskann-quantization/src/multi_vector/distance/kernels/leaves/scalar.rs
```diff
@@ -0,0 +1,123 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
```
hildebrandmw, line 4: This seems copied *almost* verbatim from `v3.rs` with small tweaks: different panel sizes and no FMA. Is that really the best way to structure it?

diskann-quantization/src/multi_vector/distance/kernels/leaves/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+//! shallow, so it runs several independent max chains rather than serializing on the
+//! latency of a single one.
+
+pub(super) mod scalar;
+#[cfg(target_arch = "x86_64")]
+pub(super) mod v3;
+
+/// Independent max chains in `max_into_rows`: enough to keep a multi-cycle max off its own
+/// critical path, few enough that the chains stay in registers.
+const WAYS: usize = 4;
```
hildebrandmw, line 23: Maybe move this closer to its use?

diskann-quantization/src/multi_vector/distance/kernels/leaves/mod.rs
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+//! Per-ISA micro-kernels.
+//!
+//! Panel geometry is fixed per ISA, not derived from a lane count. The scalar leaf is
+//! deliberately narrower than `2 × LANES` would suggest, because its "lanes" are a
+//! loop, not silicon.
```
hildebrandmw, line 10: This is misleading. Panel geometry **is** derived from a lane count (literally: `pub(crate) const A_PANEL: usize = 2 * f32s::LANES`). While yes this is informed formed by the ISA, I don't see why the negative is needed. Would it not be clearer to say "Panel geometry is optimized for each ISA"?

The comment about scalar is also strange. It does not use `2 x LANES`.

While I may be harping on documentation, realize that as a review, I'm reading the documentation to get an understanding for a change. If the documentation leaves me more confused than before, the review gets considerably harder.

diskann-quantization/src/multi_vector/distance/kernels/leaves/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+//!
+//! Every leaf here **stores** its products into a slot and leaves the reduction to
+//! `max_into_rows`. That reduction is not negligible work when the strip is wide and
```
hildebrandmw, line 13: Question: this PR claims that it regresses performance because of this choice. However ... we have all the pieces to fuse the `max` directly into the accumulator, right? Why not do that and *not* have a performance regression?

diskann-quantization/src/multi_vector/distance/kernels/float.rs
```diff
@@ -0,0 +1,494 @@
...
+    }
+}
+
+///////////
+// Entry //
+///////////
+
+/// Plan, allocate the strip, and drive.
+///
+/// `k` is the *physical* row length both walks stride by, so it must be A's padded column
```
hildebrandmw, line 185: Why mention column padding when this kernel doesn't accept block-transposes with `PACK > 1`?

diskann-quantization/src/multi_vector/distance/kernels/float.rs
```diff
@@ -0,0 +1,494 @@
...
+pub(super) struct Kernel;
+
+/// Reduces a strip into the running per-A-row maxima, one entry per A row.
+///
+/// This is the module's single [`Drain`], which makes it the one place the accumulator's
+/// axes acquire meaning. A strip column holds one B row's inner products against every A
+/// row. A B row therefore arrives as a *row* of the input and lands as a *column* of the
+/// accumulator.
+///
+/// Carries `nd` because a strip's trailing columns belong to B rows past the end of the
```
hildebrandmw, line 40: "Belong to `B` rows past the end of the matrix". How is it belonging to rows that don't exist?

hildebrandmw (commented):
Thanks Suryansh. I've been staring at this for a while trying to puzzle though how everything works and I'm still struggling.

X buckets of concerns:

1. The performance loss outlined in the PR description seems almost entirely self-inflicted. This redesign introduces the `Strip` for the accumulator (which aggregates partial products like a normal matrix multiplication) and `drains` it out to the final result. But it seems like this design is entirely compatible with accumulating into the result like the old one? Or at least eagerly compressing the maximum elements when writing back to the accumulator. Doing so would considerably reduce the space needed for the accumulator and probably get back some or all of the performance. It would also resolve the semantics of "rows" and "columns" in `Strip` as it would just become one-dimensional.
2. Lots of code is duplicated between the v3 and the scalar kernel. It's not the end of the world but does stand out to me.
3. There is a non-locality that is hard to follow. It's better than dancing around with raw pointers everywhere, but still feels like you need to know how everything works to know anything works.

  I understand that this last point is particularly vague and it could be that everything will click tomorrow when I come back fresh. And it could be that this is just an intrinsic property of writing dense kernels.

  While not necessarily a blocker, this property in the code and particularly the documentation (which often focuses on how something fits into some global picture rather than describing how a struct works and the invariants it needs/maintains) it making it difficult for me to conceptualize how everything fits together.

diskann-quantization/src/multi_vector/distance/kernels/f16.rs
```diff
@@ -1,52 +1,214 @@
...
+impl<Arch: Architecture, const BR: usize> TileWalk for RowMajorWiden<'_, Arch, BR>
+where
+    SliceCast<f32, half::f16>: for<'x> Target2<Arch, (), &'x mut [f32], &'x [half::f16]>,
+{
+    fn next(&mut self) -> Option<RowMajorTile<'_, f32, BR>> {
+        let k = self.0.k;
+        self.0.next().map(|data| RowMajorTile::new(data, k))
+    }
+
+    fn reset(&mut self) {
```
partychen, line 133: The refactor removes the old asymmetric multi-tile f16 coverage, while the replacement tiny-budget tests only exercise the f32 walks. This leaves the new reusable widening buffer and `RowMajorWiden::reset` path untested. Please add forced multi-tile f16 Scalar and V3 cases against the naive reference, including asymmetric A/B tile counts.

diskann-quantization/src/multi_vector/distance/factory.rs
```diff
@@ -59,14 +60,11 @@ where
...
-            F32Kernel::<GROUP>,
-            self.prepared.reborrow(),
-            doc,
-            &mut scratch,
-        );
-        for (dst, &src) in scores.iter_mut().zip(&scratch[..self.prepared.nrows()]) {
+        // `run` seeds the max itself, so the fill value here is arbitrary.
+        let mut state = vec![0.0; self.prepared.padded_nrows()];
+        self.arch
+            .run3(MaxIp, self.prepared.reborrow(), doc, &mut state);
```
partychen, line 66: This call no longer performs the old `query.ncols() == doc.vector_dim()` boundary check. For non-empty contractions, a mismatch reaches the opaque `"B panel extent"` assertion; for zero-dimensional queries or empty documents, the early returns accept the mismatch and return scores. Please validate dimensions before either early return for both f32 and f16, and report both values.

diskann-quantization/src/multi_vector/distance/kernels/mod.rs
```diff
@@ -48,63 +59,315 @@ impl Default for TileBudget {
...
+/// each disjoint from the last.
+trait Scratch: for<'s> SlotsAt<'s> {
+    fn slots(&mut self) -> <Self as SlotsAt<'_>>::Slots;
+}
+
+//////////////////
+// Compute side //
+//////////////////
+
+/// One A-panel × one B-panel → one accumulator slot.
```
partychen, line 190: Because `Scratch` is reused, an `Accumulate` implementation must overwrite every live output cell represented by `B`; leaving stale slot data would silently corrupt the following drain. Please document this full-overwrite requirement and clarify that unused tail cells may remain untouched.

diskann-quantization/src/multi_vector/distance/kernels/float.rs
```diff
@@ -0,0 +1,494 @@
...
+/// guard runs before any walk exists. A zero-length row would give a walk a zero stride.
+///
+/// On return `state` holds the per-A-row maximum inner product, one entry per padded A
+/// row. Its incoming contents are ignored: seeding the max is this function's job,
+/// not the caller's, because a caller that got it wrong would silently clamp the result
+/// instead of failing.
+///
+/// # Panics
+///
+/// Panics if `state` is shorter than A's padded row count: the drain indexes it one
```
partychen, line 198: The `# Panics` section omits incompatible A/B contraction lengths. The generated walks can carry different `k` values, which reaches the leaf's `"B panel extent"` assertion. Please document this requirement here; caller-facing validation should reject it with the actual dimensions.

suri-kumkaran:
This work has been merged with a improvised design in #1368

PR #1337, opened 2026-08-14
Reform Range Search

magdalendobson:
Range search had some problems, and the new filtered range search inherited many of them. 

*Problem 1:* Range search added some extra elements to `scratch` for convenience. However, given that the result size of range search can vary wildly, it doesn't make too much sense to reuse them, and it clogged up the scratch space. This PR pulls them out of scratch and manages them directly inside the search algorithm.

*Problem 2:* Range search has a lot of parameters, and their meaning and use was kind of opaque to people who aren't me. This PR adds a nice long docstring that actually explains them.

*Problem 3*: Two parameters--the initial search slack and the range search slack--weren't really tested. This PR adds some tests that ensure that they behave as intended. 

*Problem 4*: There was a lot of weirdness in handling of `max_results`. With start point filtering, if start points fell within the radius, they could be excluded during post-process and the search could return fewer than `max_results` points even if satisfying points existed. This PR sets an "effective" max results that includes the number of start points. Furthermore, since inner radius filtering wasn't done until the end, if the inner radius ended up excluding a lot of points at the end, we could again return way fewer than `max_results` points even when they exist. This PR fixes that by including inner radius filtering during the second round of search. 

To do this, it includes a new `InRange` struct which automatically handles enforcing the radius constraints, max results, and filtering if applicable. This removes the need for the `DistanceFiltered` buffer that was previously handed to the post-processor. 

NB: this PR also removes one test from both the filtered and unfiltered range search. The test didn't make a lot of sense and it is now subsumed by the testing of `initial_slack`.


hildebrandmw (approved):

hildebrandmw:
Sorry, accidentally hit the "approve" button.

diskann/src/graph/search/range_search.rs
```diff
@@ -182,6 +196,25 @@ impl Range {
 }
 
 /// Builder for [`Range`] search parameters.
+///
```
hildebrandmw, line 204: Request: can you please wrap the comments at column 100 or so? Otherwise, they get quite difficult to read and edit in editors.

diskann/src/graph/search/range_search.rs
```diff
@@ -182,6 +196,25 @@ impl Range {
...
+/// `max_returned`: If specified, the search will stop and return results once this number of points has been found within both the inner and outer radii.
+/// Note that due to adding some extra slack for start points, occasionally slightly more than this number of points may be returned. Since the initial
+/// search phase does not respect `max_returned`, this parameter may not be set lower than `starting_l`.
+///
+/// `starting_l`: the L_search parameter for the initial search phase. Must be greater than zero.
```
hildebrandmw, line 211: Maybe cross reference with `KNN::search_l`?

diskann/src/graph/search/range_search.rs
```diff
@@ -182,6 +196,25 @@ impl Range {
...
+///
+/// `beam_width`: the beam width for parallel graph exploration. If not specified, defaults to 1. Must be greater than zero if specified.
+///
+/// `radius`: the outer radius for the range search. Points within this distance from the query are candidates for inclusion in the results.
+///
+/// `inner_radius`: the inner radius for the range search. Points closer than this distance from the query are excluded from the results. Must be less than or equal to `radius` if specified.
```
hildebrandmw: I believe radii must be positive and finite as well?

diskann/src/graph/search/range_search.rs
```diff
@@ -182,6 +196,25 @@ impl Range {
 }
 
 /// Builder for [`Range`] search parameters.
+///
+/// `max_returned`: If specified, the search will stop and return results once this number of points has been found within both the inner and outer radii.
+/// Note that due to adding some extra slack for start points, occasionally slightly more than this number of points may be returned. Since the initial
```
harsha-simhadri: in this corner case, why cant we truncate the list of starting points and just honor the max_returned list?

magdalendobson: I pushed an update that does this but it is a little awkward. The problem is that filtering out start points happens during `post_process`, so the truncation needs to happen after that point, and there isn't a built-in way to truncate an output buffer. The workaround is to introduce a new output buffer that handles the truncation after filtering, which I do here, with a corresponding integration test that checks that the desired behavior happens (e.g. if more than the maximum number of points are returned, but none of them are start points, truncation happens). 

I do think this adds a lot of code for a relatively small gain, so would like opinions on whether to 1) stick with behavior in the PR as-is, 2) revert to sometimes over-returning, or 3) allow under-returning, thus going back to behavior on current main. CC @hildebrandmw

hildebrandmw: We already have some precedent for over returning in the normal KNN search because we don't know precisely whether or not start points will be filtered in post processing. For KNN, this is usually solved by the user supplying a fixed size buffer to terminate post-processing. I don't think it's unreasonable to document `max_returned` as more of an early termination hint rather than hard guarantee.

That said, I've been operating under the assumption that start points could be either valid or invalid depending on the caller. This gives us the flexibility in the future to seed searches with valid points if we need to.

diskann/src/graph/search/range_search.rs
magdalendobson: Radii do NOT have to be positive because of inner product search. I guess radii should be finite, but we don't explicitly check for it at the moment, is that something you want to change?

diskann/src/graph/search/range_search.rs
magdalendobson: Yep, done

diskann/src/graph/search/range_search.rs
magdalendobson: Done

diskann/src/graph/search/range_search.rs
magdalendobson: After our side conversation, I added is_finite checks for all float values and also a note on negative radius for inner product.

diskann/src/graph/search/range_search.rs
magdalendobson: Restored to the previous functionality after the buffer proof-of-concept turned out to be messy.

diskann/src/graph/test/cases/range_search.rs
```diff
@@ -405,34 +405,96 @@ fn max_results_respected_means_no_second_round() {
...
     assert_range_invariants(&results, radius, None);
     assert_no_duplicates(&results);
 }
 
 #[test]
-fn max_results_respected_and_second_round_triggered() {
-    let description = "Two round search test to validate that max_results > \
-    starting_l means a second round is triggered.";
+fn initial_slack_low_triggers_second_round() {
+    let _description = "Test that low initial_slack triggers second round. \
```
hildebrandmw, line 418: Why is this now ignored? In addition, there seems to be a copy+paste issue with this description being duplicated in a few places?

diskann/src/graph/test/cases/range_search.rs
```diff
@@ -405,34 +405,96 @@ fn max_results_respected_means_no_second_round() {
...
-    let radius = 1.0e9; // every point will be in range with this radius
-    let starting_l = 4; // small set to trigger multiple rounds
-    let max_results = 5; // max_returned greater than starting_l, so second round should be triggered
+    let radius = 50.0;
+    let starting_l = 4;
+    let low_slack = 0.5;
 
     let range_search = Range::builder(starting_l, radius)
-        .max_returned(Some(max_results))
+        .initial_slack(low_slack)
```
hildebrandmw, line 434: I'm not so sure these tests are accurate. Removing the `low_slack` argument does not cause the test to fail.

diskann/src/graph/test/cases/range_search.rs
```diff
@@ -405,34 +405,96 @@ fn max_results_respected_means_no_second_round() {
...
+    let name = path.push("initial_slack_high_avoids_second_round");
+
+    let grid_size = 5;
+    let (index, query) = setup_grid_index_and_default_query(grid_size, Grid::Three);
+    let radius = 50.0;
+    let starting_l = 4;
+    let high_slack = 1.0;
+
+    let range_search = Range::builder(starting_l, radius)
+        .initial_slack(high_slack)
```
hildebrandmw, line 495: As with the test above, I'm not sure what this is measuring. Setting `initial_slack = 0.0` doesn't cause the test to fail.

hildebrandmw (commented):
Thanks Magdalen - Implementation looks good, but can you please take another look at the tests? At least in range search, the tests don't seem to break with changes in `initial_slack`.

diskann/src/graph/test/cases/range_search.rs
magdalendobson: Thanks, this should be fixed now.

diskann/src/graph/test/cases/range_search.rs
magdalendobson: You're right, I fixed this and the previous test and made sure that they do what's intended.

diskann/src/graph/test/cases/range_search.rs
```diff
@@ -463,18 +471,155 @@ fn max_results_respected_and_second_round_triggered() {
...
+    let rt = current_thread_runtime();
+    let mut test_root = root();
+    let mut path = test_root.path();
+    let name = path.push("initial_slack_high_avoids_second_round");
+
+    let grid_size = 5;
+    let (index, query) = setup_grid_index_and_default_query(grid_size, Grid::Three);
+    let radius = 3.0;
+    let starting_l = 4;
+    let high_slack = 0.51;
```
hildebrandmw: I'm going to credit the agent with finding this. If you look at the baseline generated for this test, the second round of search is still triggered. This is because check in the algorithm converts (`starting_l * initial_slack`) to an integer (truncating) before doing the check. So `floor(4 * 0.51)` is still 2.

Probably increasing slack will do the trick, since I assume usually L will be fairly large so this kind of rounding doesn't impact much.

hildebrandmw (commented):
Thanks - found one small issue with one of the range search tests, otherwise looks like a good improvement.

diskann/src/graph/test/cases/range_search.rs
magdalendobson: Thanks, I tested this explicitly and switched the two values to ensure that one triggers expansion and one doesn't.

hildebrandmw (approved):

harsha-simhadri (approved):

PR #1341, opened 2026-08-18
Migrate disk PQ flat scan to flat API

partychen:
- [x] Does this PR have a descriptive title that could go in our release notes? **Yes.**
- [ ] Does this PR add any new dependencies? **No.**
- [ ] Does this PR modify any existing APIs? **No.** The query-aware flat-search API was introduced separately in #1359; this PR adopts it in `diskann-disk`.
- [x] Is the change to the API backwards compatible? **Yes.** Existing disk search modes and result semantics are preserved.
- [x] Should this result in any changes to our documentation, either updating existing docs or adding new ones? **Yes.** The affected implementation rustdoc is updated.

#### Reference Issues/PRs

Built on the query-aware flat-search API merged in #1359.

- Replaces the disk-specific manual PQ flat-scan pipeline with the shared flat k-NN API.
- Implements `DistancesUnordered` on `DiskAccessor` to expose complete, batched PQ-distance scanning.
- Initializes query-dependent PQ state after every pooled scratch checkout.
- Preserves scan-time filtering before approximate top-k selection and full-precision reranking afterward.
- Preserves the existing pooled scratch and indexed-vector result behavior across graph and flat search modes.

The disk backend constructs a query-aware `DiskAccessor` and passes it directly to `flat::knn_search`. The generic flat layer now owns top-k selection, comparison accounting, error escalation, and post-processing. `DiskAccessor` continues to own disk-specific PQ preprocessing, batching, filtering, data access, and distance computation.

#### Any other comments?

This PR has been rebased onto `main` after #1359 merged. Its diff is limited to the two `diskann-disk` implementation files.

#### Architecture simplification

Before this change, disk flat search manually coordinated filtering, batching, PQ-distance collection, top-k selection, comparison accounting, and post-processing inside `DiskANNIndex::flat_search`. Graph and flat search already used the same `DiskAccessor` and scratch pool, but the flat algorithm duplicated orchestration now provided by the shared flat API.

```mermaid
flowchart TB
    subgraph Before["Before: disk-specific flat orchestration"]
        direction LR
        F1["FlatScan"] --> M["DiskANNIndex::flat_search"]
        M --> FI["filter IDs"]
        FI --> B["manual batch loop"]
        B --> PQ1["DiskAccessor::pq_distances"]
        PQ1 --> K1["local NeighborPriorityQueue"]
        K1 --> PP1["disk post-processor"]
    end

    subgraph After["After: shared flat orchestration"]
        direction LR
        F2["FlatScan"] --> K2["flat::knn_search"]
        K2 --> DU["DiskAccessor<br/>DistancesUnordered"]
        DU --> PQ2["filtered, batched PQ scan"]
        K2 --> TK["shared top-k · stats · errors"]
        TK --> PP2["RerankAndFilter"]
    end

    G["Graph search"] --> SA["DiskAccessor<br/>SearchAccessor"]
    DU --> S["pooled DiskSearchScratch<br/>per-query PQ preparation"]
    SA --> S
```

`DiskAccessor` now exposes the disk scan through `DistancesUnordered`, allowing `flat::knn_search` to drive the common k-NN workflow while graph traversal continues to use the existing `SearchAccessor` implementation. Both paths preserve their distinct filtering stages and share the same pooled query-state lifecycle.

### DIFF
--- a/diskann-disk/src/search/pq/quantizer_preprocess.rs
+++ b/diskann-disk/src/search/pq/quantizer_preprocess.rs
@@ -6,7 +6,6 @@
 use diskann::ANNResult;
 use diskann_vector::distance::Metric;
 
-use diskann_providers::model::compute_pq_distance;
 use diskann_providers::utils::BridgeErr;
 
 use super::{PQData, PQScratch};
@@ -33,11 +32,7 @@ impl PQScratch {
         .bridge_err()?;
 
         match metric {
-            // Prior to the introduction of the `quantizer_preprocess` method, the
-            // disk index was hard-coded to use L2 distance for comparisons.
-            //
-            // We're keeping that behavior here - treating `Cosine` and `CosineNormalized`
-            // as L2 until a more thorough evaluation can be made.
+            // Disk search historically treats cosine metrics as L2 here.
             Metric::L2 | Metric::Cosine | Metric::CosineNormalized => {
                 table.process_into::<diskann_quantization::distances::SquaredL2>(
                     &self.query_scratch,
@@ -54,28 +49,4 @@ impl PQScratch {
 
         Ok(())
     }
-}
-
-/// Preprocesses the query stored in `pq_scratch` and computes PQ distances
-/// for `id_to_calculate_pq_distance`.
-///
-/// The resulting distances are written to `pq_scratch.aligned_dist_scratch`.
-pub fn quantizer_preprocess(
-    pq_scratch: &mut PQScratch,
-    pq_data: &PQData,
-    metric: Metric,
-    id_to_calculate_pq_distance: &[u32],
-) -> ANNResult<()> {
-    pq_scratch.preprocess_query(pq_data, metric)?;
-
-    compute_pq_distance(
-        id_to_calculate_pq_distance,
-        pq_data.get_num_chunks(),
-        &pq_scratch.aligned_pqtable_dist_scratch,
-        pq_data.pq_compressed_data().as_slice(),
-        &mut pq_scratch.aligned_pq_coord_scratch,
-        &mut pq_scratch.aligned_dist_scratch,
-    )?;
-
-    Ok(())
 }
--- a/diskann-disk/src/search/pq/quantizer_preprocess.rs
+++ b/diskann-disk/src/search/pq/quantizer_preprocess.rs
@@ -11,46 +11,63 @@ use diskann_providers::utils::BridgeErr;
 
 use super::{PQData, PQScratch};
 
-/// Preprocesses the query vector for PQ distance calculations.
-/// This function rotates the query vector and prepares the PQ table distances
-/// for efficient computation during search operations.
+impl PQScratch {
+    pub(crate) fn prepare_query(
+        &mut self,
+        pq_data: &PQData,
+        metric: Metric,
+        query: &[f32],
+    ) -> ANNResult<()> {
+        self.set(query)?;
+        self.preprocess_query(pq_data, metric)
+    }
+
+    fn preprocess_query(&mut self, pq_data: &PQData, metric: Metric) -> ANNResult<()> {
+        let table = pq_data.pq_table();
+        let expected_len = table.ncenters() * table.nchunks();
+        let dst = diskann_utils::views::MutMatrixView::try_from(
+            &mut self.aligned_pqtable_dist_scratch[..expected_len],
+            table.nchunks(),
+            table.ncenters(),
+        )
+        .bridge_err()?;
+
+        match metric {
+            // Prior to the introduction of the `quantizer_preprocess` method, the
+            // disk index was hard-coded to use L2 distance for comparisons.
+            //
+            // We're keeping that behavior here - treating `Cosine` and `CosineNormalized`
+            // as L2 until a more thorough evaluation can be made.
+            Metric::L2 | Metric::Cosine | Metric::CosineNormalized => {
+                table.process_into::<diskann_quantization::distances::SquaredL2>(
+                    &self.query_scratch,
+                    dst,
+                );
+            }
+            Metric::InnerProduct => {
+                table.process_into::<diskann_quantization::distances::InnerProduct>(
+                    &self.query_scratch,
+                    dst,
+                );
+            }
+        }
+
+        Ok(())
+    }
+}
+
+/// Preprocesses the query stored in `pq_scratch` and computes PQ distances
+/// for `id_to_calculate_pq_distance`.
+///
+/// The resulting distances are written to `pq_scratch.aligned_dist_scratch`.
 pub fn quantizer_preprocess(
     pq_scratch: &mut PQScratch,
     pq_data: &PQData,
     metric: Metric,
     id_to_calculate_pq_distance: &[u32],
 ) -> ANNResult<()> {
-    let table = pq_data.pq_table();
-    let expected_len = table.ncenters() * table.nchunks();
-    let dst = diskann_utils::views::MutMatrixView::try_from(
-        &mut (*pq_scratch.aligned_pqtable_dist_scratch)[..expected_len],
-        table.nchunks(),
-        table.ncenters(),
-    )
-    .bridge_err()?;
+    pq_scratch.preprocess_query(pq_data, metric)?;
 
-    match metric {
-        // Prior to the introduction of the `quantizer_preprocess` method, the
-        // disk index was hard-coded to use L2 distance for comparisons.
-        //
-        // We're keeping that behavior here - treating `Cosine` and `CosineNormalized`
-        // as L2 until a more thorough evaluation can be made.
-        Metric::L2 | Metric::Cosine | Metric::CosineNormalized => {
-            table.process_into::<diskann_quantization::distances::SquaredL2>(
-                &pq_scratch.query_scratch,
-                dst,
-            );
-        }
-        Metric::InnerProduct => {
-            table.process_into::<diskann_quantization::distances::InnerProduct>(
-                &pq_scratch.query_scratch,
-                dst,
-            );
-        }
-    }
-
-    // Compute the pq distance between query vector to all the vertex in the pq
-    // calculation id scratch.
     compute_pq_distance(
         id_to_calculate_pq_distance,
         pq_data.get_num_chunks(),
--- a/diskann-disk/src/search/pq/quantizer_preprocess.rs
+++ b/diskann-disk/src/search/pq/quantizer_preprocess.rs
@@ -32,7 +32,11 @@ impl PQScratch {
         .bridge_err()?;
 
         match metric {
-            // Disk search historically treats cosine metrics as L2 here.
+            // Prior to moving query preprocessing onto `PQScratch`, the disk index
+            // was hard-coded to use L2 distance for comparisons.
+            //
+            // We're keeping that behavior here - treating `Cosine` and `CosineNormalized`
+            // as L2 until a more thorough evaluation can be made.
             Metric::L2 | Metric::Cosine | Metric::CosineNormalized => {
                 table.process_into::<diskann_quantization::distances::SquaredL2>(
                     &self.query_scratch,
--- a/diskann-disk/src/search/provider/disk_provider.rs
+++ b/diskann-disk/src/search/provider/disk_provider.rs
@@ -670,16 +670,7 @@ where
         _context: &DefaultContext,
         query: &'this [Data::VectorDataType],
     ) -> Result<Self::SearchAccessor, Self::SearchAccessorError> {
-        DiskAccessor::new(
-            provider,
-            self.io_tracker,
-            query,
-            self.vertex_provider_factory,
-            self.scratch_pool,
-            self.cache_indexed_vectors,
-            // Graph filtering is handled by post-processing or `FilteredAccessor`.
-            None,
-        )
+        DiskAccessor::new(provider, query, self)
     }
 }
 
@@ -968,17 +959,13 @@ where
     Data: GraphDataType<VectorIdType = u32>,
     VP: VertexProvider<Data>,
 {
-    fn new<VPF>(
+    fn new<ProviderFactory>(
         provider: &'a DiskProvider<Data>,
-        io_tracker: &'a IOTracker,
         query: &'a [Data::VectorDataType],
-        vertex_provider_factory: &'a VPF,
-        scratch_pool: &'a Arc<ObjectPool<DiskSearchScratch<Data, VP>>>,
-        cache_indexed_vectors: bool,
-        flat_scan_filter: Option<PostprocessFilter<'a>>,
+        strategy: &'a DiskSearchStrategy<'a, Data, ProviderFactory>,
     ) -> ANNResult<Self>
     where
-        VPF: VertexProviderFactory<Data, VertexProviderType = VP>,
+        ProviderFactory: VertexProviderFactory<Data, VertexProviderType = VP>,
     {
         let pq_points = provider.pq_data.pq_compressed_data().nrows();
         if pq_points != provider.num_points {
@@ -990,21 +977,26 @@ where
         }
 
         let scratch = DiskSearchScratch::pooled_for_query(
-            scratch_pool,
+            strategy.scratch_pool,
             provider,
-            io_tracker,
+            strategy.io_tracker,
             query,
-            vertex_provider_factory,
+            strategy.vertex_provider_factory,
         )?;
 
         Ok(Self {
             provider,
-            io_tracker,
+            io_tracker: strategy.io_tracker,
             scratch,
             query,
-            cache_indexed_vectors,
-            flat_scan_filter,
+            cache_indexed_vectors: strategy.cache_indexed_vectors,
+            flat_scan_filter: None,
         })
+    }
+
+    fn with_flat_scan_filter(mut self, filter: Option<PostprocessFilter<'a>>) -> Self {
+        self.flat_scan_filter = filter;
+        self
     }
 
     fn ensure_loaded(&mut self, ids: &[u32]) -> Result<(), ANNError> {
@@ -1225,15 +1217,8 @@ where
             PostprocessStrategy::AcceptAll => None,
             PostprocessStrategy::Apply(filter) => Some(filter),
         };
-        let mut visitor = DiskAccessor::new(
-            self.index.provider(),
-            strategy.io_tracker,
-            query,
-            strategy.vertex_provider_factory,
-            strategy.scratch_pool,
-            strategy.cache_indexed_vectors,
-            flat_scan_filter,
-        )?;
+        let mut visitor = DiskAccessor::new(self.index.provider(), query, strategy)?
+            .with_flat_scan_filter(flat_scan_filter);
         let FlatSearchStats { cmps, result_count } = flat_knn_search(
             &mut visitor,
             k,
--- a/diskann-disk/src/search/provider/disk_provider.rs
+++ b/diskann-disk/src/search/provider/disk_provider.rs
@@ -16,15 +16,18 @@ use std::{
 use crate::data_model::GraphDataType;
 use diskann::{
     error::IntoANNResult,
+    flat::{knn_search as flat_knn_search, DistancesUnordered, SearchStats as FlatSearchStats},
     graph::{
         self,
         ext::labeled::{self, QueryLabelProvider},
-        glue::{self, DefaultPostProcessor, SearchPostProcess, SearchStrategy},
+        glue::{
+            self, DefaultPostProcessor, SearchPostProcess, SearchStrategy as GraphSearchStrategy,
+        },
         search::{AdaptiveL, InlineFilterSearch, Knn},
         search_output_buffer::{self, BufferState, IdDistanceAssociatedData},
         DiskANNIndex,
     },
-    neighbor::{self, Neighbor, NeighborPriorityQueue},
+    neighbor::{self, Neighbor},
     provider::{DataProvider, DefaultContext, HasId, NoopGuard},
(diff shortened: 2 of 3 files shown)


partychen:
@arkrishn94  I ended up making a few design changes beyond the `Visitor` implementation, and I’d appreciate a sanity check on whether these are the right tradeoffs:

1. **Borrowed-provider flat search**
`DiskProvider` is already owned by `DiskANNIndex`, so I added a borrowed-provider `flat::knn_search` entry point instead of creating another `FlatIndex`, cloning/wrapping the provider, or introducing a `BorrowedFlatIndex` type. The existing `FlatIndex::knn_search` delegates to it. Does this seem like the right API shape?

2. **Shared rerank implementation**
`RerankAndFilter` needs to work with both `DiskAccessor` and `FlatVisitor`. I extracted the common implementation into `rerank_and_filter`, leaving two thin `SearchPostProcess` adapters. The alternative would be a shared accessor trait exposing provider/scratch through associated types. I felt that trait would be more abstraction than the two callers justify, but I’d like your opinion.

3. **PQ query-computer construction and pooling**
I separated the query-to-centroid lookup state from `PQScratch`. Both graph and flat search now obtain a `PQQueryComputer` from a dedicated object pool, while `DiskSearchScratch` only retains the batch distance and coordinate buffers. This also preserves the existing `SearchStrategy::build_query_computer(query)` API: `DiskSearchStrategy` borrows the PQ schema, metric, and query-computer pool from `DiskIndexSearcher` instead of passing the provider into the trait method. Does this separation and pooling boundary seem appropriate?

One related detail: filtering happens in `FlatVisitor` before candidates enter the top-k queue, so reranking uses `AcceptAll` to avoid evaluating the predicate twice.

These were the main areas where the migration required broader architectural choices, so feedback on them would be helpful before finalizing the approach.

hildebrandmw (commented):
As usual, I will defer to the maintainers of `diskann-disk` to make the judgement calls here, but what immediately stands out to me is that trying to fit the flat scan into the `diskann` flat-scan API is essentially recreating the custom flat-scan implementation but with significantly more code. That is, this appears to be working hard to fit the API (and indeed changing the API in `diskann`) without materially benefiting from doing so.

To me, this indicates two things:

1. There is an ergonomic gap in the flat API that needs to be fixed. For example - it requires a `QueryComputer` which is causing some of the churn in this PR [1]. I don't think that's a good direction since it separates the compute engine from the internal of the `FlatAccessor`, when closer coupling (e.g. how `SearchAccessor` works now for the graph index) allows for safer optimization.
2. We're missing even lower-level infrastructure (e.g. generic batch PQ computation independent of `diskann-disk`) that would help with reusability. Think: a more generally reuseable version of `compute_pq_distance`.

There are parts that look good. Extracting `rerank_and_filter` to a synchronous function (instead of the current unfortunate bounce through async) is a good improvement. Simplifying PQ scratch initialization is good - though I might suggest keeping it in `DiskSearchScratch` fusing it with the `DiskSearchScratch`'s pooled API to avoid the multi-stage initialization that is currently done.

[1] The graph portion of `diskann` [used to work](https://github.com/microsoft/DiskANN/pull/1067) this way and it turns out to be way better for a huge number of reasons to not.

partychen:
I agree with your assessment. This migration exposed a limitation in the current flat API: separating the visitor from `QueryComputer` works for simple element-wise scans, but forces a backend such as disk PQ to split one query across multiple objects, pools, and initialization stages. Continuing to adapt around that API would recreate the original custom flat scan without gaining much from the shared abstraction.

I also agree that the better direction is to improve the flat API itself. Following the principle established in PR #1067, I propose making flat visitors query-aware and responsible for producing distances.

### Proposed API

```rust
pub trait DistancesUnordered: HasId + Send + Sync {
    type Error: ToRanked + Debug + Send + Sync + 'static;

    fn distances_unordered<F>(
        &mut self,
        f: F,
    ) -> impl SendFuture<Result<(), Self::Error>>
    where
        F: Send + FnMut(Self::Id, f32);
}

pub trait SearchStrategy<'a, P, T>: Send + Sync
where
    P: DataProvider,
{
    type Visitor: DistancesUnordered<Id = P::InternalId>;
    type Error: StandardError;

    fn create_visitor(
        &'a self,
        provider: &'a P,
        context: &'a P::Context,
        query: T,
    ) -> Result<Self::Visitor, Self::Error>;
}
```

The generic flat-search flow becomes:

```rust
let mut visitor = strategy.create_visitor(provider, context, query)?;
visitor.distances_unordered(callback).await?;
processor.post_process(&mut visitor, query, candidates, output).await?;
```

The responsibility boundary would be:

- The generic flat layer owns top-k selection, comparison accounting, scan-error escalation, and post-processing.
- The backend visitor owns query preprocessing, filtering, batching, data access, and distance computation.

For disk PQ, graph and flat search can then use one pooled `DiskSearchScratch` containing the query buffer, PQ lookup table, batch buffers, vertex provider, and reranking cache. This removes the separate `PQQueryComputer`, its object pool, and the additional initialization stage.

The main advantages are:

- It follows the accessor-ownership model from PR #1067.
- It gives backends a coarse boundary for fusing access and computation.
- It supports both simple element-wise visitors and optimized batch implementations.
- It simplifies disk PQ query state and resource management.

The main trade-off is a public flat-trait change. To limit migration cost, the existing trait and method names remain. I also searched for visible consumers and did not find an independent public implementation outside DiskANN itself, forks, and vendored copies.

I have tried this proposal in the latest revision of the PR so that the design can be reviewed through a concrete implementation:

- `create_visitor` now receives the query.
- The visitor produces distances without a separate `QueryComputer`.
- Graph and flat disk PQ search share one pooled `DiskSearchScratch`.
- Flat filtering still happens before candidates enter top-k, so reranking uses `AcceptAll` and does not evaluate the predicate twice.
- RFC 00983 has been updated to describe the proposed design.

I also agree that a generic batch PQ primitive independent of `diskann-disk` would improve reuse. I have kept that as separate follow-up work so this proposal stays focused on the flat API boundary.

I would appreciate your review of both the proposed API direction and the implementation in this revision. Does this align with what you had in mind? @hildebrandmw @arkrishn94

rfcs/00983-flat-search.md
```diff
@@ -21,7 +21,7 @@ The problem-statement here is simple: provide first-class support for sequential
 
 ### 1.3 Goals
 
-1. Define a fused iterate-and-score primitive — `flat::DistancesUnordered<C>` — that
+1. Define a fused iterate-and-score primitive — `flat::DistancesUnordered` — that
```
hildebrandmw: Please don't retroactively update an RFC. They represent points in time and should not be updated.

diskann/src/flat/index.rs
```diff
@@ -35,6 +35,60 @@ pub struct FlatIndex<P: DataProvider> {
...
+    query: T,
+    output: &'a mut OB,
+) -> impl SendFuture<ANNResult<SearchStats>> + 'a
+where
+    P: DataProvider,
+    S: SearchStrategy<'a, P, T> + 'a,
+    T: Copy + Send + Sync + 'a,
+    O: Send + 'a,
+    PP: SearchPostProcess<S::Visitor, T, O> + Send + Sync + 'a,
+    OB: SearchOutputBuffer<O> + Send + ?Sized + 'a,
```
hildebrandmw: Most of these lifetimes are unnecessary. This is sufficient:
```rust
pub fn knn_search<'a, P, S, T, O, PP, OB>(
    provider: &'a P,
    k: NonZeroUsize,
    strategy: &'a S,
    processor: PP,
    context: &'a P::Context,
    query: T,
    output: &mut OB,
) -> impl SendFuture<ANNResult<SearchStats>>
where
    P: DataProvider,
    S: SearchStrategy<'a, P, T>,
    T: Copy + Send + Sync,
    O: Send,
    PP: SearchPostProcess<S::Visitor, T, O> + Send + Sync,
    OB: SearchOutputBuffer<O> + Send + ?Sized,
{
```

hildebrandmw (commented):
Thanks - reworking the flat API to resemble the graph API and merging the compute into `DiskSearchScratch` is much cleaner.

However, my larger concern still applies. I do not see what the flat API is enabling in `diskann-disk` to justify the increased complexity. @arkrishn94 - can you weigh in?

diskann/src/flat/index.rs
```diff
@@ -35,6 +35,60 @@ pub struct FlatIndex<P: DataProvider> {
...
+/// Streams every distance produced by the strategy's accessor, keeps the best `k`
+/// candidates in a [`NeighborPriorityQueue`], then runs `processor` over the survivors
+/// to populate `output`.
+///
+/// # Errors
+///
+/// Returns an error if accessor construction, distance scanning, or result
+/// post-processing fails. Distance-scan errors are escalated because a
+/// partial flat scan cannot produce correct k-nearest-neighbor results.
+pub fn knn_search<'a, P, S, T, O, PP, OB>(
```
arkrishn94: I think by making the `knn_search` a free-function that is public, we are creating another entry point for consumers to implement invoke the flat scan in addition to the associated method for the `FlatIndex`. 

I'm not necessarily opposed to having just the free function (in fact, I might be for it) and getting rid of the `FlatIndex` entirely...

diskann-disk/src/search/pq/quantizer_preprocess.rs
```diff
@@ -36,21 +36,44 @@ pub fn quantizer_preprocess(
...
+    metric: Metric,
+    id_to_calculate_pq_distance: &[u32],
+) -> ANNResult<()> {
+    preprocess_query(
+        &pq_scratch.query_scratch,
+        &mut pq_scratch.aligned_pqtable_dist_scratch,
+        pq_data,
+        metric,
+    )?;
+
```
arkrishn94: nit: Can we get rid of these free functions if we call `pq_scratch.set(query)` in the constructor for `DiskSearchScratch` that calls `prepare_query`?

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -757,33 +882,161 @@ where
...
+pub struct FlatVisitor<'a, Data, VP>
+where
+    Data: GraphDataType<VectorIdType = u32>,
+    VP: VertexProvider<Data>,
+{
+    provider: &'a DiskProvider<Data>,
+    filter: Option<PostprocessFilter<'a>>,
+    scratch: PoolOption<DiskSearchScratch<Data, VP>>,
+    cache_indexed_vectors: bool,
+}
```
arkrishn94: Maybe I'm missing something, but is it possible to just get rid of the `FlatVisitor` and implement `DistancesUnordered` over the `DiskAccessor`?

arkrishn94 (commented):
Thanks Junkui. Firstly, I apologize for the severely delayed review for this PR. 

I'm largely on board with the direction of these changes. I like the simplification of the flat search API and might even suggest getting rid of the `FlatIndex` entirely. I guess if we go down this path, it might make sense to open a pre-cursor PR to this with just the changes to the flat API.

The simplification to `DiskSearchScratch` looks good, although as Mark said it would be nice to consolidate the initialization for it. 

The one comment I had about the complexity of introducing the flat API here is- can we get rid of the `FlatVisitor` struct entirely and just work over the `DiskAccessor`?

diskann-disk/src/search/pq/quantizer_preprocess.rs
partychen: Done. I moved `prepare_query` and the lookup-table preprocessing onto `PQScratch`, so the new free helpers are gone. `DiskSearchScratch::pooled_for_query` now checks out the pooled scratch and calls `pq_scratch.prepare_query(...)`, which performs `set(query)` followed by preprocessing.

I kept this initialization after checkout rather than in `TryAsPooled::try_create`, because `try_create` runs only when a pool entry is first allocated; reused entries must be reinitialized for every query. The remaining public `quantizer_preprocess` function is the existing API and is retained for compatibility.

partychen:
@hildebrandmw @arkrishn94 Thanks for the detailed feedback. I reorganized the implementation around the following design.

### Disk PQ flat-scan design

Disk PQ flat search constructs a query-initialized `DiskAccessor` and passes it directly to `flat::knn_search`.

`DiskAccessor` implements both `SearchAccessor` for graph traversal and `DistancesUnordered` for exhaustive flat scanning. There is no separate `FlatVisitor`. Both search modes use the same accessor, pooled `DiskSearchScratch`, PQ preprocessing, vertex loading, reranking cache, and indexed-vector handling.

For flat search, `DistancesUnordered` enumerates every eligible vector ID, applies the optional filter before top-k selection, computes approximate PQ distances in scratch-sized batches, and emits each `(id, distance)` pair to `flat::knn_search`.

`flat::knn_search` owns the backend-independent workflow: bounded top-k selection, comparison accounting, scan-error handling, post-processing, and output statistics. The surviving candidates are then reranked using their full-precision indexed vectors.

#### Why use the flat API here?

The flat API removes algorithm orchestration from `diskann-disk`. Without it, the disk backend must maintain its own priority queue, comparison counter, batch loop, error handling, and post-processing invocation.

The abstraction does not attempt to hide disk PQ details. `DiskAccessor` retains control over filtering, batching, PQ computation, disk access, and scratch state. The shared layer only owns behavior common to exhaustive k-NN search. This provides reuse without preventing backend-specific optimizations.

#### Scratch initialization

`DiskAccessor::new` checks out a `DiskSearchScratch` through `DiskSearchScratch::pooled_for_query`. After every checkout, `PQScratch::prepare_query` copies and preprocesses the current query. This must happen per checkout rather than only when a pool entry is first allocated, because the same scratch object is reused by subsequent queries.

Graph and flat search therefore share one initialization path while keeping query state isolated between pool uses.

#### Filtering and reranking

Flat filtering occurs during the scan, before candidates enter the approximate top-k queue. Consequently, reranking uses `AcceptAll`; applying the predicate again would be redundant.

Graph search retains its existing filtering stages because traversal and flat scan have different candidate-selection semantics.

diskann-disk/src/search/pq/quantizer_preprocess.rs
```diff
@@ -11,46 +11,63 @@ use diskann_providers::utils::BridgeErr;
...
+        Ok(())
+    }
+}
+
+/// Preprocesses the query stored in `pq_scratch` and computes PQ distances
+/// for `id_to_calculate_pq_distance`.
+///
+/// The resulting distances are written to `pq_scratch.aligned_dist_scratch`.
 pub fn quantizer_preprocess(
```
wuw92: Remove `quantizer_preprocess` and its re-export, as there are no callers after refactor.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -444,6 +442,63 @@ where
     Ok(count)
 }
 
+fn rerank_and_filter<Data, VP, I, B>(
```
wuw92: Nit: `rerank_and_filter` appears code movement.  Extracting it alongside the flat-search migration makes it harder to see whether the reranking semantics changed,  so reviewers have to re-validate the moved logic alongside the behavioral changes.

diskann-disk/src/search/provider/disk_provider.rs
wuw92, line 1164: Could we pass the flat-scan filter explicitly instead of routing it through `postprocess_filter`? The current flow converts `postprocess_filter` into a scan-time filter and then hard-codes `AcceptAll` for post-processing, which makes it difficult to tell where filtering actually belongs. Passing `AcceptAll` to the strategy and a separate filter argument to `flat_search` would make the stage ownership explicit.

diskann-disk/src/search/pq/quantizer_preprocess.rs
partychen: Good catch. This was left over from the intermediate implementation. Once the graph and flat paths were consolidated onto the same DiskAccessor, there were no remaining callers, and I missed cleaning up the old API. I removed quantizer_preprocess and its re-export in 0b0e333b.

diskann-disk/src/search/provider/disk_provider.rs
partychen: You're right. This extraction was needed when the graph and flat paths used separate accessor types. After both paths were consolidated onto DiskAccessor, it no longer served a purpose, and I missed folding the code back. I inlined the reranking logic into the existing post-processor in 0b0e333b to keep the diff focused.

diskann-disk/src/search/provider/disk_provider.rs
partychen: Agreed. Routing the filter through postprocess_filter was another leftover from the earlier design with separate accessors, and I missed restoring the explicit boundary after consolidating on DiskAccessor. In 0b0e333b, the flat-scan filter is passed directly to flat_search while the strategy uses AcceptAll, so the scan-time and post-processing responsibilities are explicit.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -849,55 +925,46 @@ where
...
+            io_tracker: strategy.io_tracker,
             scratch,
             query,
-            cache_indexed_vectors,
+            cache_indexed_vectors: strategy.cache_indexed_vectors,
+            flat_scan_filter: None,
         })
     }
 
+    fn with_flat_scan_filter(mut self, filter: Option<PostprocessFilter<'a>>) -> Self {
```
wuw92: nit: consider renaming `PostprocessFilter`, the current name suggests it is only valid during post-processor.

wuw92 (approved):

diskann-disk/src/search/provider/disk_provider.rs
partychen: thanks，updated!

xwj-ox (approved):

hildebrandmw (commented):
I again question what is being gained from this. We're trading a relatively self contains 50ish line function for an implementation spread between `diskann-disk` and `diskann`. This couples `diskann-disk` with an API in `diskann` that does not have a second consumer nor materially adds functionality - the orchestration performed by the flat API in `diskann` is trivial.

I won't block on this concern - please feel free to merge if there is consensus. I just ask the increased maintenance and coupling is considered against the benefits.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -229,7 +231,7 @@ pub enum PostprocessStrategy<'a> {
     /// (filtered at scan time) and `InlineFilter` (filtered at visit time).
     AcceptAll,
     /// Apply the given predicate; non-matching candidates are dropped.
-    Apply(PostprocessFilter<'a>),
+    Apply(IdFilter<'a>),
```
arkrishn94, line 234: Is this naming change intentional?

diskann-disk/src/search/pq/quantizer_preprocess.rs
```diff
@@ -6,59 +6,51 @@
 use diskann::ANNResult;
 use diskann_vector::distance::Metric;
 
-use diskann_providers::model::compute_pq_distance;
 use diskann_providers::utils::BridgeErr;
 
 use super::{PQData, PQScratch};
 
-/// Preprocesses the query vector for PQ distance calculations.
```
arkrishn94, line 14: nit: for some reason the diff here looks longer than it is p

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -2742,6 +2780,136 @@ mod disk_provider_tests {
...
+            assert_eq!(
+                actual.vertex_id, expected_id,
+                "flat baseline ID mismatch at result {index}",
+            );
+            assert!(
+                (actual.distance - expected_distance).abs() <= 0.02,
+                "flat baseline distance mismatch at result {index}: expected \
+                 {expected_distance}, got {}",
+                actual.distance,
+            );
```
arkrishn94, line 2827: nit: `0.02` is quite a large error bound. I think `1e-4` is reasonable if you expect near exact results in floating point.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -784,6 +820,46 @@ where
...
+
+            loop {
+                ids.clear();
+                ids.extend(remaining.by_ref().take(batch_size));
+                if ids.is_empty() {
+                    break;
+                }
+
+                self.pq_distances(&ids, |distance, id| f(id, distance))?;
+            }
```
arkrishn94, line 856: nit: I'm curious, why implement the loop this way? Why not just iterate over the filtered ids using [`std::slice::Chunks`](https://doc.rust-lang.org/std/slice/struct.Chunks.html)?

arkrishn94: I see now that this is how it was done before too... Interesting

arkrishn94 (approved):
Thanks Junkui, I don't have any major concerns happy to approve. I agree with Mark that there might not be much value in this refactor - apologies for leading us down this path. Glad we got some consolidation in the flat API and the DiskSearchScratch out of it.

partychen:
@arkrishn94 @hildebrandmw Thanks for the feedback.

- The `IdFilter` rename is intentional: the predicate is used for both scan-time and post-processing filtering.
- These squared-L2 distances are around 256,000, where one f32 ULP is 0.015625. An absolute tolerance of 1e-4 would effectively require identical results. I've prepared a named constant and a comment explaining why 0.02 is retained.
- `remaining` is a lazy filtered iterator, not a slice. Using slice chunks would require materializing the matching IDs; the current loop keeps the buffer bounded to one batch.
- The preprocessing diff mainly reflects moving the existing logic onto `PQScratch` and consolidating query initialization.

I agree the benefit is modest: shared flat-search orchestration and consolidated scratch initialization, not new functionality. The additional coupling is a real trade-off.

xinyuwen2 (approved):

PR #1345, opened 2026-08-19
Return native indexed vectors from disk search

### LINKED ISSUE #1339: Add support for passing full-precision vectors along with results in disk-index search.
Some downstream tasks require the underlying full-precision vectors in addition to their IDs returned in the top-k search. We need to add support for this correctly in the `DiskIndexSearcher`

yjiez:
## Summary

Adds an opt-in disk search API that returns each valid ANN result with its canonical native indexed vector:

```rust
search_with_indexed_vectors(...)
```

The existing `search()` API and its padded result behavior remain unchanged.

## Public contract

```rust
pub struct SearchResultItemWithIndexedVector<A, V> {
    pub vertex_id: u32,
    pub data: A,
    pub distance: f32,
    pub indexed_vector: Box<[V]>,
}
```

- The vector is the native representation stored in the graph and used for exact scoring; it is not a PQ code and may differ from the original input vector.
- The indexed-vector API returns valid results only: `results.len() == stats.result_count`.
- A missing vector is returned as `ANNError`, not a panic.

## For reviewers

Recommended review order:

1. **Public API:** `SearchResultWithIndexedVectors`, `SearchResultItemWithIndexedVector`, and `search_with_indexed_vectors`.
2. **Output compatibility:** `SearchPayload` and `SearchOutput` extend the existing slice-backed output without changing legacy `search()`.
3. **Traversal reuse:** `DiskAccessor::ensure_loaded` optionally stores the native vector beside the existing exact-distance cache entry.
4. **Final output:** `extend_output` moves traversal-cached winner vectors into results or copies winners from the post-process batch already loaded by the existing rerank path.
5. **Scratch cleanup:** `DiskAccessor::drop` immediately releases vectors left in the request-local cache.
6. **Coverage:** the existing 128-dimensional disk-search test checks result parity and vector values with both no cache and static cache.

The intended invariants are:

- A returned vector always belongs to the same `vertex_id` as its result.
- Enabling vector output does not change IDs or distances.
- Legacy `search()` does not allocate or copy indexed vectors.
- Returning vectors does not add a final disk-read round.

### Internal search payload

```rust
type SearchPayload<A, V> = (u32, A, Option<Box<[V]>>);
```

The `Option` is internal: legacy search writes `None`; indexed-vector search writes `Some(vector)`. The public indexed-vector result contains a non-optional `Box<[V]>`.

### Output adapter

```rust
struct SearchOutput<'a, A, V> {
    output: IdDistanceAssociatedData<'a, u32, A>,
    indexed_vectors: Option<&'a mut [Option<Box<[V]>>]>,
}
```

`SearchOutput` keeps the existing ID/distance/associated-data buffers and adds an optional vector lane. One `push` writes all fields at the same position, preserving ID/vector alignment without replacing the existing search pipeline.

### Request-local traversal cache

```rust
distance_cache: HashMap<
    u32,
    (f32, AssociatedData, Option<Box<[VectorData]>>),
>
```

The map already existed for exact distance and associated data. When indexed vectors are requested, the same entry also owns the native vector loaded during traversal. There is no collector, mutex, or second hash map. The map is cleared when the query accessor is dropped so loser vectors are not retained in the scratch pool between requests.

## Data flow

```text
search_with_indexed_vectors
    -> enable vector capture for this query
    -> traversal load: distance_cache stores distance + data + vector
    -> existing rerank path loads uncached candidates as one batch
    -> extend_output:
         cached winner   -> move Box from distance_cache
         uncached winner -> copy from current in-memory post-process batch
    -> return only stats.result_count valid results
```

`extend_output` never calls `load_vertices`, so it does not introduce a final fallback I/O pass.

## Benchmark

[GitHub Actions run with immediate cleanup](https://github.com/microsoft/DiskANN/actions/runs/32388396932), using K=1000, L=2000, recall@100, and four search threads:

| Dataset | Mean API latency vs legacy | QPS vs legacy | Mean I/O count | Indexed-vector peak delta |
|---|---:|---:|---:|---:|
| Wikipedia-100K, 768d | +1.43% | -1.41% | unchanged | +23.25 MiB  (+26.78%) | 
| OpenAI-100K, 1536d | +1.42% | -2.09% | unchanged | +46.75 MiB  (+13.79%)  | 

  Expected Peak-Memory Increase

  The indexed-vector API retains the full-precision vectors needed by active searches. Its expected vector-payload increase can therefore be approximated as:
  Expected peak delta ≈ C × L × D × S

  where:
  - C is the number of concurrent searches;
  - L is the search-list size;
  - D is the vector dimension;
  - S is the size of each vector element (4 bytes for f32).

  For Wikipedia-100K:
  4 × 2000 × 768 × 4 bytes ≈ 23.44 MiB
  This closely matches the measured increase of 23.25 MiB.

Closes #1339

### DIFF
--- a/diskann-disk/src/search/provider/disk_provider.rs
+++ b/diskann-disk/src/search/provider/disk_provider.rs
@@ -238,6 +238,7 @@ where
 {
     // Borrowed from `search_internal` so the strategy can be passed by value
     io_tracker: &'a IOTracker,
+    cache_indexed_vectors: bool,
     /// Consumed only by `default_post_processor()` → `RerankAndFilter`.
     /// `FlatScan` and `InlineFilter` filter earlier in their pipelines and
     /// pass `AcceptAll` here to avoid a redundant second pass.
@@ -319,14 +320,135 @@ impl<'a> DeterminantDiversityAndFilter<'a> {
     }
 }
 
+type IndexedVector<VectorData> = Option<Box<[VectorData]>>;
+type DistanceCacheEntry<AssociatedData, VectorData> =
+    (f32, AssociatedData, IndexedVector<VectorData>);
+type IndexedVectorOutput<'a, VectorData> = Option<&'a mut [IndexedVector<VectorData>]>;
+type SearchPayload<AssociatedData, VectorData> = (u32, AssociatedData, IndexedVector<VectorData>);
+
+struct SearchOutput<'a, A, V> {
+    output: search_output_buffer::IdDistanceAssociatedData<'a, u32, A>,
+    indexed_vectors: IndexedVectorOutput<'a, V>,
+}
+
+impl<'a, A, V> SearchOutput<'a, A, V> {
+    fn new(
+        ids: &'a mut [u32],
+        distances: &'a mut [f32],
+        associated_data: &'a mut [A],
+        indexed_vectors: IndexedVectorOutput<'a, V>,
+    ) -> Self {
+        if let Some(vectors) = &indexed_vectors {
+            assert_eq!(ids.len(), vectors.len());
+        }
+        Self {
+            output: search_output_buffer::IdDistanceAssociatedData::new(
+                ids,
+                distances,
+                associated_data,
+            ),
+            indexed_vectors,
+        }
+    }
+}
+
+impl<A: Clone + Send, V> search_output_buffer::SearchOutputBuffer<SearchPayload<A, V>>
+    for SearchOutput<'_, A, V>
+{
+    fn size_hint(&self) -> Option<usize> {
+        search_output_buffer::SearchOutputBuffer::size_hint(&self.output)
+    }
+
+    fn push(
+        &mut self,
+        neighbor: Neighbor<SearchPayload<A, V>>,
+    ) -> search_output_buffer::BufferState {
+        let position = search_output_buffer::SearchOutputBuffer::current_len(&self.output);
+        let ((id, data, vector), distance) = neighbor.as_tuple();
+        let state = search_output_buffer::SearchOutputBuffer::push(
+            &mut self.output,
+            Neighbor::new((id, data), distance),
+        );
+        if search_output_buffer::SearchOutputBuffer::current_len(&self.output) != position {
+            if let Some(vectors) = &mut self.indexed_vectors {
+                vectors[position] = vector;
+            }
+        }
+        state
+    }
+
+    fn current_len(&self) -> usize {
+        search_output_buffer::SearchOutputBuffer::current_len(&self.output)
+    }
+
+    fn extend<Iter>(&mut self, iter: Iter) -> usize
+    where
+        Iter: IntoIterator<Item = Neighbor<SearchPayload<A, V>>>,
+    {
+        let start = self.current_len();
+        for neighbor in iter {
+            let previous = self.current_len();
+            let state = self.push(neighbor);
+            if self.current_len() == previous || state.is_full() {
+                break;
+            }
+        }
+        self.current_len() - start
+    }
+}
+
+fn extend_output<Data, VP, B, I>(
+    accessor: &mut DiskAccessor<'_, Data, VP>,
+    reranked: I,
+    output: &mut B,
+) -> ANNResult<usize>
+where
+    Data: GraphDataType<VectorIdType = u32>,
+    VP: VertexProvider<Data>,
+    I: IntoIterator<Item = Neighbor<(u32, Data::AssociatedDataType)>>,
+    B: search_output_buffer::SearchOutputBuffer<
+            SearchPayload<Data::AssociatedDataType, Data::VectorDataType>,
+        > + Send
+        + ?Sized,
+{
+    if !accessor.cache_indexed_vectors {
+        return Ok(output.extend(reranked.into_iter().map(|candidate| {
+            let ((id, data), distance) = candidate.as_tuple();
+            Neighbor::new((id, data, None), distance)
+        })));
+    }
+
+    let mut count = 0;
+    for candidate in reranked {
+        if output.size_hint() == Some(0) {
+            break;
+        }
+        let ((id, data), distance) = candidate.as_tuple();
+        let vector = match accessor
+            .scratch
+            .distance_cache
+            .remove(&id)
+            .and_then(|(_, _, vector)| vector)
+        {
+            Some(vector) => vector,
+            None => Box::from(accessor.scratch.vertex_provider.get_vector(&id)?),
+        };
+        count += 1;
+        if output
+            .push(Neighbor::new((id, data, Some(vector)), distance))
+            .is_full()
+        {
+            break;
+        }
+    }
+    Ok(count)
+}
+
 impl<Data, VP>
     SearchPostProcess<
         DiskAccessor<'_, Data, VP>,
         &[Data::VectorDataType],
-        (
-            <DiskProvider<Data> as DataProvider>::InternalId,
-            Data::AssociatedDataType,
-        ),
+        SearchPayload<Data::AssociatedDataType, Data::VectorDataType>,
     > for RerankAndFilter<'_>
 where
     Data: GraphDataType<VectorIdType = u32>,
@@ -342,8 +464,9 @@ where
     ) -> Result<usize, Self::Error>
     where
         I: Iterator<Item = Neighbor<u32>> + Send,
-        B: search_output_buffer::SearchOutputBuffer<(u32, Data::AssociatedDataType)>
-            + Send
+        B: search_output_buffer::SearchOutputBuffer<
+                SearchPayload<Data::AssociatedDataType, Data::VectorDataType>,
+            > + Send
             + ?Sized,
     {
         let provider = accessor.provider;
@@ -384,7 +507,7 @@ where
         reranked.sort_unstable_by(neighbor::ord::fast_distance);
 
         // Store the reranked results.
-        Ok(output.extend(reranked))
+        extend_output(accessor, reranked, output)
     }
 }
 
@@ -392,10 +515,7 @@ impl<Data, VP>
     SearchPostProcess<
         DiskAccessor<'_, Data, VP>,
         &[Data::VectorDataType],
-        (
-            <DiskProvider<Data> as DataProvider>::InternalId,
-            Data::AssociatedDataType,
-        ),
+        SearchPayload<Data::AssociatedDataType, Data::VectorDataType>,
     > for DeterminantDiversityAndFilter<'_>
 where
     Data: GraphDataType<VectorIdType = u32>,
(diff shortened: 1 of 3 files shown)


diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -383,6 +518,45 @@ where
...
+                    "Fetching missing indexed vectors after rerank"
+                );
+                ensure_vertex_loaded(&mut accessor.scratch.vertex_provider, &final_fetch_ids)?;
+                for id in final_fetch_ids {
+                    let vector = accessor.scratch.vertex_provider.get_vector(&id)?;
+                    collector.capture_required(id, vector)?;
+                }
+            }
+        }
+
```
arkrishn94: Thanks for working on this Yujie. I must admit, I am bit confused by the complexity of the approach here. 

There's a few problems from what I'm seeing in the current design- 
1. The collector seems to be an additional abstraction just to store full precision vectors, if needed, during the traversal but the `CachedDiskVertexProvider` is already doing this. 
2. With the way the collector is implemented (i.e. `MutexLock`, hash-lookup and hash-map on insertion), this seems to be pretty expensive operations in the hot-path (i.e. in `ensure_loaded`). 
3. We're threading the collector through the strategy and the accessor and collecting results there when the `glue::SearchPostProcess` architecture is explicitly designed to be generic over the shape of the output. So it feels weird to me that you need to have a new structure to collect results in the accessor. 

I'm curious, have we tried doing the simple thing of just loading the vertices (by calling `load_vertices` on the vertex provider) from the result set after the regular search path? Why can't we just do that in `search_with_indexed_vectors`? It seems like that should allow you to get rid of the collector completely  

Let me know if I'm missing something here

partychen:
The API and its intended behavior make sense, but I’m concerned about the complexity and performance cost of the collector design.


`ensure_loaded` copies every loaded vector into a request-local `Mutex<HashMap>`, so memory usage, allocations, locking, and hash lookups scale with all visited vertices rather than the final result count. Could we instead complete the regular search first, then batch-load only the final result vectors? That would keep this logic out of the traversal hot path and eliminate most of the additional plumbing.


I also think the `expect()` and `assert_ne!()` on public search paths should return `ANNError` rather than panic, and the per-query `info!` log should probably be lowered to `debug!`.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -319,14 +320,135 @@ impl<'a> DeterminantDiversityAndFilter<'a> {
...
+{
+    fn size_hint(&self) -> Option<usize> {
+        search_output_buffer::SearchOutputBuffer::size_hint(&self.output)
+    }
+
+    fn push(
+        &mut self,
+        neighbor: Neighbor<SearchPayload<A, V>>,
+    ) -> search_output_buffer::BufferState {
+        let position = search_output_buffer::SearchOutputBuffer::current_len(&self.output);
```
arkrishn94: nit: `self.current_len()` is easier to read here?

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -319,14 +320,135 @@ impl<'a> DeterminantDiversityAndFilter<'a> {
...
+        let ((id, data, vector), distance) = neighbor.as_tuple();
+        let state = search_output_buffer::SearchOutputBuffer::push(
+            &mut self.output,
+            Neighbor::new((id, data), distance),
+        );
+        if search_output_buffer::SearchOutputBuffer::current_len(&self.output) != position {
```
arkrishn94: Hmm, this is exploiting an implicit contract of `self.output` that the `current_len` only increases when the push to the buffer succeeds.  

Small suggestion (maybe for follow up): why rely on the implementation of `SearchOutputBuffer<IdDistanceAssociatedData>`? Just implement it directly as it is done for `IdDistanceAssociatedData`?

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -319,14 +320,135 @@ impl<'a> DeterminantDiversityAndFilter<'a> {
...
+
+    fn extend<Iter>(&mut self, iter: Iter) -> usize
+    where
+        Iter: IntoIterator<Item = Neighbor<SearchPayload<A, V>>>,
+    {
+        let start = self.current_len();
+        for neighbor in iter {
+            let previous = self.current_len();
+            let state = self.push(neighbor);
+            if self.current_len() == previous || state.is_full() {
```
arkrishn94: nit: When will you have `self.current_len() == previous` and `state.is_full() == false`?

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -850,6 +987,18 @@ pub struct SearchResultItem<AssociatedData> {
     pub distance: f32,
 }
 
+pub struct SearchResultWithIndexedVectors<AssociatedData, VectorData> {
```
arkrishn94: nit: This is a bit verbose, maybe `ResultWithVectors`?

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -786,19 +911,31 @@ where
...
             scratch
                 .distance_cache
-                .insert(*id, (distance, associated_data));
+                .insert(*id, (distance, associated_data, indexed_vector));
         }
         Ok(())
     }
 }
 
+impl<Data, VP> Drop for DiskAccessor<'_, Data, VP>
```
arkrishn94, line 929: Why does the accessor now need a `Drop` implementation?

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -319,14 +320,135 @@ impl<'a> DeterminantDiversityAndFilter<'a> {
...
+impl<'a, A, V> SearchOutput<'a, A, V> {
+    fn new(
+        ids: &'a mut [u32],
+        distances: &'a mut [f32],
+        associated_data: &'a mut [A],
+        indexed_vectors: IndexedVectorOutput<'a, V>,
+    ) -> Self {
+        if let Some(vectors) = &indexed_vectors {
+            assert_eq!(ids.len(), vectors.len());
+        }
```
arkrishn94: Let's throw an error here?

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -1372,6 +1621,30 @@ mod disk_provider_tests {
...
+            .zip(&indexed.results)
+            .all(|(left, right)| (left.vertex_id, left.distance)
+                == (right.vertex_id, right.distance)));
+        assert_eq!(indexed.results.len(), indexed.stats.result_count as usize);
+        for item in &indexed.results {
+            assert_eq!(
+                item.indexed_vector.as_ref(),
+                source.row(item.vertex_id as usize)
+            );
+        }
```
arkrishn94: Can we test this through the `test_disk_search` path on multiple queries?  From what I can tell this is the only path exercising your change...

arkrishn94 (approved):
Thanks Yujie, I have left some comments but they can mostly come in follow-up since I know this is time sensitive. I am approving but would be good to beef-up the testing a bit before merging. From what I can tell there is only one test with a single query testing this new path. Would also be good to test the error cases.

diskann-disk/src/search/provider/disk_provider.rs
yjiez: Without this Drop, non-winning vectors remain owned by distance_cache until that particular scratch is checked out again and try_modify() clears it. When running multiple queries concurrently, the non-winning O(L-K) vectors from pooled scratches can overlap with active query caches and other allocations, which will increase the peak memory.
This is also consistent with benchmark result:
## Benchmark

[GitHub Actions run with immediate cleanup](https://github.com/microsoft/DiskANN/actions/runs/32388396932), compared with the [same run without immediate cleanup](https://github.com/microsoft/DiskANN/actions/runs/32385752047), using K=1000, L=2000, recall@100, and four search threads:

| Dataset | Mean API latency vs legacy | QPS vs legacy | Mean I/O count | Indexed-vector peak delta | Delta without immediate cleanup |
|---|---:|---:|---:|---:|---:|
| Wikipedia-100K, 768d | +1.43% | -1.41% | unchanged | +23.25 MiB | +32.88 MiB |
| OpenAI-100K, 1536d | +1.42% | -2.09% | unchanged | +46.75 MiB | +64.38 MiB |

Immediate cache cleanup reduced the feature-specific peak-memory increase by 27-29%. The remaining increase closely matches the live traversal-vector payload expected from L=2000 across four concurrent searches.

diskann-disk/src/search/provider/disk_provider.rs
yjiez: Thanks. Since mimir vertex provider doesn’t keep cache info consistent; it drops the old cache while loading the new vertex. So, we need to cache vector when traversing to avoid disk i/o in final rerank.

diskann-disk/src/search/provider/disk_provider.rs
yjiez: Thanks, fixed

diskann-disk/src/search/provider/disk_provider.rs
yjiez: Good point. I updated the existing test_disk_search helper so that every query now runs both the legacy path and
 search_with_indexed_vectors. For each query, the test verifies identical result count, IDs, and distances and each returned vector matches the source row for its vertex_id.
 I also added one representative indexed-vector check for each separate search-mode path: FlatScan, partial-result Graph filtering, selective InlineFilter, and DiverseGraph.

diskann-disk/src/search/provider/disk_provider.rs
yjiez: Updated. Thanks

diskann-disk/src/search/provider/disk_provider.rs
yjiez: Renamed it to SearchResultWithVectors to keep it consistent with the existing SearchResult type while shortening the name.

diskann-disk/src/search/provider/disk_provider.rs
yjiez: Good point. I kept SearchOutput as a wrapper because its role is to add an optional indexed-vector lane on top of IdDistanceAssociatedData, rather than duplicate the complete ID/distance/associated-data buffer implementation.
 I updated push() to explicitly check current_len() == capacity() before delegating.
 A follow-up can generalize the shared output buffer with an optional extra lane and remove this local adapter entirely.

diskann-disk/src/search/provider/disk_provider.rs
yjiez: That combination cannot occur for this output implementation. push() returns Full when the buffer was already full,
 or after successfully inserting the final item. It returns Available only after an insertion when capacity remains.
 I removed the redundant current_len() == previous check, so extend() now stops solely onself.push(neighbor).is_full().

SeliMeli (approved):

PR #1347, opened 2026-08-20
Return indexed vectors after disk search

yjiez:
## Summary

- add an optional disk-search API that returns each final hit with its canonical indexed vector
- run the existing search unchanged, then batch-load only the final result IDs
- reuse the existing scratch pool and cap final reads at 128 IDs per batch
- keep traversal, post-processing, caches, and search statistics untouched

This supersedes #1345 with the post-search approach requested in review.

## Validation

- `cargo check -p diskann-disk --tests`
- `cargo test -p diskann-disk --lib indexed_vectors`
- CI format, clippy, docs, CodeQL, SDE, and Linux workspace tests

## Benchmark

K=1000, L=2000, four order-balanced paired runs (AB, BA, BA, AB):

| Dataset | QPS delta | Mean API latency | P95 API latency | Peak memory delta |
|---|---:|---:|---:|---:|
| Wikipedia-100K | -10.71% | +12.00% | +12.37% | +14.9 MiB |
| OpenAI-100K | -15.43% | +18.55% | +18.44% | +29.4 MiB |

Recall is unchanged. Returned payload is 3,072,000 bytes/query for Wikipedia and 6,144,000 bytes/query for OpenAI.

The final-only approach uses less peak memory than traversal caching at the same K/L (23.3 MiB and 46.7 MiB), at the cost of rereading the final K nodes.

Runs: [AB 1](https://github.com/microsoft/DiskANN/actions/runs/32393525336), [BA 1](https://github.com/microsoft/DiskANN/actions/runs/32395793064), [BA 2](https://github.com/microsoft/DiskANN/actions/runs/32398069939), [AB 2](https://github.com/microsoft/DiskANN/actions/runs/32400368014). Cached comparison: [32388396932](https://github.com/microsoft/DiskANN/actions/runs/32388396932).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/diskann-disk/src/search/provider/disk_provider.rs
+++ b/diskann-disk/src/search/provider/disk_provider.rs
@@ -1127,28 +1127,44 @@ where
             });
         }
 
-        let ids = results
-            .iter()
-            .map(|result| result.vertex_id)
-            .collect::<Vec<_>>();
-        let graph_header = &self.index.provider().graph_header;
-        let mut vertex_provider = self
-            .vertex_provider_factory
-            .create_vertex_provider(ids.len(), graph_header)?;
-        ensure_vertex_loaded(&mut vertex_provider, &ids)?;
+        let provider = self.index.provider();
+        let mut scratch = PoolOption::try_pooled(
+            &self.scratch_pool,
+            &DiskSearchScratchArgs {
+                graph_degree: provider.graph_header.max_degree::<Data::VectorDataType>()?,
+                pq_dim: provider.pq_data.get_dim(),
+                num_pq_chunks: provider.pq_data.get_num_chunks(),
+                num_pq_centers: provider.pq_data.get_num_centers(),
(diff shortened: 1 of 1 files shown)


diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -1535,6 +1608,76 @@ mod disk_provider_tests {
         result.into_inner().into_vec()
     }
 
+    #[rstest]
+    #[case(CachingStrategy::None)]
+    #[case(CachingStrategy::StaticCacheWithBfsNodes(32))]
+    fn test_search_with_indexed_vectors(#[case] caching_strategy: CachingStrategy) {
```
SeliMeli, line 1614: seems like we are asserting/test multiple behavior in one tests, can we split it so one thing to test per ut?

PR #1349, opened 2026-08-21
[vectorset] Runbook support

metajack:
Adds big-ann-benchmarks style runbook support to vectorset.


vectorset/src/driver.rs
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+use std::time::Duration;
```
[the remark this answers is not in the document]
metajack: Future is in the prelude.

vectorset/test_data/f32_good.bin
magdalendobson, line 1: Can we move these to the higher-level test data folder? Also, they are small, but since they're not human-readable it might be better to use LFS in keeping with the other binary files, even small ones.

vectorset/datasets/wikipedia-100K.yaml
magdalendobson, line 1: Let's maybe not commit these as files since they won't work without set paths, and might get out of date. Instead, can we create one for a dataset in test_data, and then have a test that actually takes it in and validates it?

magdalendobson:
In general, it feels like a ton of what is set up here is very very close to what already exists in `diskann-benchmark` and `diskann-benchmark-core`, including some pretty close redefining of structs. Could this be reworked to reuse as much as possible from there? Or if not could you explain the differences?

Another question/wonder: should this just be in `diskann-benchmark` where all the other runbooks are run?

PR #1350, opened 2026-08-24
Deduplicate virtual start point edges during disk serialization

partychen:
- [ ] Does this PR modify any existing APIs? No.
- [x] Is the change to the API backwards compatible? Yes.
- [ ] Should this result in any documentation changes? No.

#### Reference Issues/PRs

N/A

The in-memory graph uses a virtual start point ID, while the serialized disk graph must replace it with the actual medoid ID.

Previously, this replacement happened after adjacency-list uniqueness had already been established. If a list contained both the virtual start point and the actual medoid, the remapping produced a duplicate medoid edge. On the medoid's own adjacency list, it could also produce a self-loop.

This PR updates `DiskAdaptor` to:

- remap the virtual start point to the actual medoid;
- preserve the original neighbor order;
- retain only the first resulting medoid edge;
- remove self-loops after remapping.

It does not backfill removed entries with arbitrary neighbors, so affected adjacency lists may contain one fewer unique edge.

The regression test covers both a remapping collision and the medoid self-loop case. The expected disk-index fixture is updated for the corrected serialized graph.

#### Any other comments?

The in-memory adjacency list is already unique before serialization, so only the alias introduced by virtual-start-point remapping needs post-remap deduplication. This keeps the change limited to disk serialization and does not alter graph construction or pruning behavior.

### DIFF
--- a/diskann-providers/src/model/graph/provider/async_/simple_neighbor_provider.rs
+++ b/diskann-providers/src/model/graph/provider/async_/simple_neighbor_provider.rs
@@ -301,7 +301,7 @@ impl storage::bin::GetAdjacencyList for SimpleNeighborProviderAsync {
 /// Key differences between the formats:
 /// 1. Disk format requires a valid vector ID as start point, while async index uses a
 ///    virtual ID (max_points + 1) that exceeds the valid dataset range
-/// 2. In-memory index appends the virtual start point at the end of adjacency lists
+/// 2. In-memory adjacency lists may contain the virtual start point ID
 /// 3. Disk format expects additional_points = 0, while async index uses additional_points = 1
 ///
 /// This adaptor handles these differences by:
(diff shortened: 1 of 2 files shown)


diskann-providers/src/model/graph/provider/async_/simple_neighbor_provider.rs
```diff
@@ -301,7 +301,7 @@ impl storage::bin::GetAdjacencyList for SimpleNeighborProviderAsync {
 /// Key differences between the formats:
 /// 1. Disk format requires a valid vector ID as start point, while async index uses a
 ///    virtual ID (max_points + 1) that exceeds the valid dataset range
-/// 2. In-memory index appends the virtual start point at the end of adjacency lists
+/// 2. In-memory adjacency lists may contain the virtual start point ID
```
hildebrandmw: Maybe "the disk format does not support the virtual start-point used in the in-memory index" or something like that?

diskann-providers/src/model/graph/provider/async_/simple_neighbor_provider.rs
```diff
@@ -327,15 +327,26 @@ impl storage::bin::GetAdjacencyList for DiskAdaptor<'_> {
...
         self.provider.get_neighbors_sync(i, &mut list)?;
 
-        // Need to change to a `Vec` because remapping the start point can cause duplicates,
-        // and changing the logic to not have duplicates changes the exact nature of the
-        // graph and breaks integration tests for the disk index builder.
         let mut list: Vec<_> = list.into();
-        for i in list.iter_mut() {
-            if *i == self.inmem_start_point {
-                *i = self.actual_start_point;
+        let node_id = u32::try_from(i).ok();
```
hildebrandmw: Should we error instead of silently swallowing?

hildebrandmw (approved):

diskann-providers/src/model/graph/provider/async_/simple_neighbor_provider.rs
partychen: Good point. Valid graph IDs are expected to fit in `u32`, but silently ignoring a failed conversion would skip self-loop removal. I’ll propagate the conversion error instead.

diskann-providers/src/model/graph/provider/async_/simple_neighbor_provider.rs
partychen: refined

xinyuwen2 (approved):

PR #1352, opened 2026-08-24
[diskann-inmem] Prepare code for quantization and beyond

hildebrandmw:
Prepare `diskann-inmem` for quantization and beyond.

Note that this is **infrastructure** work to get the code ready. Quantization is not yet integrated.

# Goals

The requirements to support quantization are annoyingly orthogonal:

1. Collections like PQ, scalar, and spherical quantization should be able to run in "quant-only" mode and "quant + full-precision" mode, where the full-precision store is used for reranking. Internally, I would like these to be the same type to cut down on unnecessary monomorphization. This means we need support for at least two collections managed by the same epoch protected `Store` (one quant, the other full-precision). In this case, the full-precision store can do without the invasive tags used for the primary store, but also needs to be optional and ideally support any of `f32`, `f16`, `u8`, `i8` and beyond.

2. For testing purposes, we probably want to retain the ability for PQ to do hybrid pruning (part full-precision, part quantized). This completely breaks the current model used by inmem of managing raw `&[u8]` slices. While we could technically make it work, if we needed to do something like multi-vector operations, a `&[u8]` is just not the right approach anyways.

3. We also want to be able to support multi-vectors and other kinds of non-uniform data in an epoch guarded `Store`, which gets rid of the uniform assumption of the current invasive store.

Supporting all of these required a pretty drastic reorganization of `diskann-inmem`.

# Architecture

This PR is all about moving things up and down. The architecture went from this:
```text
Provider

                   Prune
          +======= ExpandBeam  ===+
          |                       |
   +------|-----------------------|------+
   |      |                       |      |
   |  +-------+      +----------------+  |
   |  | layer |      | invasive-store |  |
   |  +-------+      +----------------+  |
   |                                     |
   +-------------------------------------+
```
where the provider had a separate `Layer` and `Store` and combined the two to make a `SearchAccessor` and `PruneAccessor` to this
```text
Provider ------ diskann::graph::glue ----+
   +--- drives --+                       |
                 |     SearchAccessor <--+
                 |     PruneAccessor
                 |             ^
                 V             |
+------------------------------|--------+
| Repr  --- bypasses ---+      |        |
|   +--- drives --+     |      |        |
|                 |     +-- ExpandBeam  |
|                 v     |   Prune       |
|  +--------------------|-----+         |
|  | Store - drives -+  |     |         |
|  |                 |  |     |         |
|  |                 v  v     |         |
|  |               +--------+ |         |
|  |               | Slots  | |         |
|  |               +--------+ |         |
|  +--------------------------+         |
+---------------------------------------+
```
From top to bottom:
* The `Provider` now contains a `Representation` instead of a `Layer` and a `Store`. Instead, the `Store` has been moved directly into the `Representation`. Instead of the `Provider` being responsible for building a `SearchAccessor` from pieces exposed by the old `Layer`, the `Provider` completely delegates `SearchAccessor` and `PruneAccessor` construction to its `Representation`.

  The rest of the `Provider`'s job is interfacing the `diskann::graph::glue` API to the simplified `Layer` API.

* `Representation` now gains the `Store`. The `Representation` family of traits is extended to include logical operations like insert and retire. As mentions above, it is also responsible for building `SearchAccessor`s and `PruneAccessors`.

* `Store` receives minor changes - its internal `Buffer` where it used to manage the invasive data store directly has now been moved to a `Slots` trait. In this architecture, the `Store` is just responsible for driving the `Slots` trait's lifecycle API and is completely uninvolved with the mechanics of raw data reading and writing.

* `Slots` is a new trait for a slot-based store whose slots are driven by `Store`. The old invasive store is an example of such a trait.

In addition to this hierarchical layering, a `Representation` is allowed (and indeed, expected) to bypass its immediate `Store` to the underlying `Slot` directly to build the various accessors. For the `Invasive` store, this works because concurrency tags are embedded directly in the `Slots` - readers do not need anything to do with the parent store.

## Why This Mostly (Probably) Works

The `Slots` trait provides an extension point for managing different types of data. For the quantized case, we reuse `Invasive` for the quantized data and a simpler `Buffer`-based one for the full-precision reranking data. Then we can create `Slots` consisting of both layers. The short-cut from a `Representation` to its `Slots` means a quantized provider can create its `SearchAccessor/PruneAccessor`s with knowledge of both. We could even have an `Invasive` store for both the quantized data and the full-precision data and reuse the existing full-precision infrastructure to support both quantized and full-precision searches over the same `Representation`.

Additionally, `Slots` makes no requirements on the kind of data store. This allows us to store un-even sized allocations in a `Slots`. It's the `Representation`'s job to make sense of everything.

Finally, the `Representation` knowing the details of its `Slots` means we can (with some creativity) still support hybrid pruning.

# Suggested Reviewing Order
This is a *large* PR, but I tried very hard to keep things structured. The reviewing order outlined here is a suggested bottom-up order. Understanding how the lower levels work is important for understanding how the higher ones come together.

* `num.rs`: A quick warm-up. This PR introduces some strongly typed integers with specific semantics.
* `prefetch.rs`: Another warm-up. I wasn't satisfied with the safety/flexibility of prefetching in the current in-mem provider. This PR exacerbated the situation, so I introduced a bit more structure on prefetchers.
* `store/slots.rs`: This defines the `Slots` trait and it's expected lifecycle. I captured the nuances in healthy module level and trait level documentation. This is probably the most nuanced change in this PR.
* `store/checked.rs`: An implementation of `Slots` that aggressively checks that the invariants required for the `Slots` API are upheld by `Store`. Again, there is healthy module-level documentation  describing the logic.
* `store/invasive.rs`: The old invasive data store moved to implement the `Slots` trait. This largely preserves what was already in the old `Store` and is conceptually much simpler than `store/checked.rs`.
* `store/mod.rs`: Modifying `Store` to work against a `Slots` generic instead of directly managing the invasive store. Note that there are some changes to initialization. `Store::new` now takes three distinct arguments:
  - `Layout`: Description for capacity, number of frozen points, and maximum degree.
  - `Config`: Configuration state dedicated directly to management of the internal concurrency data structures.
  - `SlotsConfig`: Configuration of the internal plugin. The `XConfig` traits are used in this PR to perform deferred initialization of large data structure.

* **Detour**: With the introduction of a modularized `Store`, changes were made to the following integration-test related files to enable concurrent stress-test of different `Store` implemntations.
  - `src/integration/store/*`: Shared boiler plate and implementations for exposing different `Store` to the integration test framework. This PR exposes wrappers for the `Invasive` and `Checked` stores in the `invasive` and `checked` modules respectively.
  - `integration/store/*`: Integration test exposure for the different stores as different jobs in the integration test suite.
  For these, the overall structure of the exposed stores is extremely similar. Most of churn in these files is moving things around so keep the amount of repeated **code** to a minimum.

* `repr/mod.rs`: Reworked `Representation`/`Set`/`Search`/`Insert` traits for the new architecture.
  - `Representation` (used to be `Layer`): Gains a few life-cycle related items.
  - `Set`: Is now responsible for also obtaining an internal slot for the inserted element.
  - `Search`/`Insert`: Reworked to return `SearchAccessor` and `PruneAccessor`s directly. This is mainly to allow us to keep the internal `Store`/`Slots` details hidden from the public interface.

  In addition:
  - `ExpandBeam`: Moves from its old location in `provider.rs`. Otherwise, is mostly unchanged.
  - `Prune`: A new trait for pruning. Prune implementation now internally "buffer" items in the prune set. This effectively makes the **type** of the elements being pruned hidden, allowing for hybrid pruning in the future.

* `repr/full.rs`: This is where everything comes together. The `Config` is used to group together full-precision related constructor arguments and `Full` now gains a `Store`. Importantly, now that `Full` knows the full details of its store, we can more aggressively optimize `ExpandBeam` with fewer bounds and length checks. Note that start point initialization is now managed by `Full`'s constructor.

  The implementation of `ExpandBeam` is taken pretty much directly from the old `provider.rs` code. Additionally, the tests have gotten more robust with Miri having more coverage of the `ExpandBeam` implementation and correctness tests for the various distance specializations.

I highly recommend looking at the public docs
```
RUSTDOCFLAGS="-D rustdoc::all" cargo doc --locked --no-deps --package diskann-inmem
```
and the private docs
```
RUSTDOCFLAGS="-D rustdoc::all" cargo doc --locked --no-deps --package diskann-inmem --all-features --document-private-items
```
to get a feeling for the public API and internal documentation.

AI Disclosure: An agent was used review changes and implementation details, help brainstorm, and make focused edits to documentation.


diskann-inmem/src/provider.rs
```diff
@@ -446,22 +370,25 @@ impl glue::SearchAccessor for SearchAccessor<'_> {
...
-                    self.expand_beam
-                        .expand_beam(&self.ids, &self.reader, &mut self.buffer)
-                }?;
+                // SAFETY:
+                //
+                // 1. We've verified that each entry in `self.ids` is in-bounds using
+                //   the `IdLimit` derived from this `ExpandBeam` object (enforced in the
+                //   constructor).
+                //
+                // 2. `self.buffer` is long enough to hold all the IDs.
```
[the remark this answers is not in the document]
hildebrandmw: The safety comment is fine as-is. The safety only relies on the `id_limit` from `ExpandBeam`. While it would be *nice* for this to be consistent across `Neighbors` and the `Store`. That's not required for safety.

diskann-inmem/src/num.rs
```diff
@@ -161,6 +155,82 @@ impl std::fmt::Display for Align {
...
+
+        impl std::fmt::Display for $name {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                write!(f, concat!(stringify!($name), "({})"), self.value())
+            }
+        }
+    };
+}
+
+// TODO: Provide a linkable reference for "immutable" points.
```
metajack: Leaving this TODO in the code on purpose? Or did you forget to add this?

diskann-inmem/src/prefetch.rs
```diff
@@ -0,0 +1,263 @@
...
+    /// * The slice must point to valid memory within a single allocation. There are no
+    ///   aliasing requirements.
+    ///
+    /// * [`Self::check`] must return `Ok(())` for `len`.
+    unsafe fn prefetch(self, ptr: *const u8, len: Bytes);
+}
+
+/// A call to [`Prefetch::check`] failed.
+#[derive(Debug)]
+pub(crate) struct InvalidPrefetch(());
```
metajack, line 88: Why do you need the interior?

diskann-inmem/src/prefetch.rs
```diff
@@ -0,0 +1,263 @@
...
+    }
+}
+
+impl std::error::Error for InvalidPrefetch {}
+
+diskann::convert_error!(InvalidPrefetch);
+
+/// Prefetch data using a simple `for` loop.
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct Loop(());
```
metajack, line 108: I don't not understand these prefetchers. They don't seem to have anything to them other than their struct names. What is the purpose of them?

diskann-inmem/src/store/mod.rs
```diff
@@ -0,0 +1,1028 @@
...
+                    }
+                }
+            }
+
+            if acquired.is_some() {
+                return acquired;
+            }
+
+            chunks_since_freelist_check += 1;
+            if chunks_since_freelist_check == 4 {
```
metajack: A comment here would be nice especially to document the hardcoded constant. I see the point of this is spelled out in the function level docs, but some tiny reference to that may be helpful here.

diskann-inmem/src/store/mod.rs
```diff
@@ -0,0 +1,1028 @@
...
+    #[cfg(test)]
+    fn writable(&self) -> std::ops::Range<u32> {
+        0..self.unfrozen.value() as u32
+    }
+}
+
+/// Errors occurring during [`Store::new`].
+#[derive(Debug, Error)]
+#[error(transparent)]
+pub(crate) struct StoreError(StoreErrorInner);
```
metajack, line 604: Why the error indirection?

diskann-inmem/src/repr/full.rs
```diff
@@ -0,0 +1,1558 @@
...
+}
+
+impl<T> repr::Set<&[T]> for Full<T>
+where
+    T: FullPrecision,
+{
+    type Guard<'a> = Guard<'a>;
+
+    fn set(&self, v: &[T]) -> ANNResult<Guard<'_>> {
+        if v.len() != self.dim() {
```
metajack: Slightly odd to write a check_dim and then not use it. Though I see you return a different error.

metajack (approved):

diskann-inmem/src/num.rs
hildebrandmw: This is still left around. I imagined adding documentation about "immutable" points to the crate-level documentation eventually. I guess this kind of hanging out here is weird though. I removed it.

diskann-inmem/src/prefetch.rs
hildebrandmw: I'm going through a weird phase where I don't want error types like this accidentally getting created by other parts of the code, which is what the private inner type gives you. Is it really necessary? No. Is it over-engineering? Probably. Does it make me happy? Kinda.

diskann-inmem/src/prefetch.rs
hildebrandmw: This came as a side-effect of a particularly cursed side-quest I went on trying to optimize prefetching (I was seeing if I could build a better dynamic prefetch function by using a jump table into a bunch of prefetch instructions. The answer is that it's possible to do, but the surrounding codegen is compromised). My longer-term thought is that I want the ability to customize prefetch more directly in the future. For example, prefetch the last cache line (with the tag) and then just the first up-to 8 or 16 cache lines. Or prefetch into L2. Or skip the tag altogether for stores that don't need it.

Using structs like this is a nice way to enable that flexibility when `ExpandBeam` implementations are getting created, so I kept it.

diskann-inmem/src/store/mod.rs
hildebrandmw: See my other comment about going through a weird phase with respect trying to prevent errors from being created non-locally. I.E., dumb over engineering.

diskann-inmem/src/repr/full.rs
hildebrandmw: Good find. Present me forgot what past me did apparently. The bespoke check is now gone. And I've learned that providing an otherwise unused struct to a macro (in this case, `SetError` to `diskann::convert_error!`) apparently counts as a use. At least, clippy didn't complain about `SetError` being unused after doing the switch. I'm a bit surprised by that. In any case, `SetError` is now gone.

diskann-inmem/src/store/mod.rs
hildebrandmw: Yeah, good point. I promoted this to a named constant (next to `RETRY_LIMIT`) gave it a commentary.

diskann-inmem/src/store/slots.rs
```diff
@@ -0,0 +1,159 @@
...
+//!   +-----------+                               +----------+
+//!   | Available |<---+                          | Retiring |
+//!   +-----------+    |                          +----------+
+//!         |          |                               ^
+//!         |          |                               |
+//!         |       `abort`                         `retire`
+//!     `acquire`      |                               |
+//!         |          |                               |
+//!         |      +----------+                   +-----------+
+//!         +----->| Slot<'_> |---- `publish` --->| Published |
```
magdalendobson: In the office hours meeting a week (or two?) ago we discussed naming the state something like `Writing` or `Acquired` rather than `Slot<'_>` since it's confusing that `Slot` is a state that a slot can be in.

diskann-inmem/src/store/slots.rs
```diff
@@ -0,0 +1,159 @@
...
+//! ```
+//!
+//! ## Readable States
+//!
+//! * `published`: **New** references to slots may be given out in the "published" state. It is
+//!   possible for a transition to go from "published" to "retiring" while references are lent
+//!   out. This is fine as long as the lifetime of these references is bounded by a
+//!   [`crate::epoch::Guard`]. Using [`super::Store::guard`] will provide such a guard.
+//!
+//! * `frozen`: Since "frozen" is a terminal state, it is safe to give out references to
```
magdalendobson, line 46: Maybe mention in the documentation that the intent is for frozen points. I know there is also some possibility for optimization around read-only graphs, but it would just be useful to have some written down intent for the state.

diskann-inmem/src/store/slots.rs
hildebrandmw: Right - updated to `Exclusive`, which has a pretty strong meaning in Rust and is what's intended by this state.

diskann-inmem/src/store/slots.rs
hildebrandmw: Added a bit more docs on the intention behind frozen points. After thinking about it more, all they really do now are protect these slots from accidentally being reclaimed through the `Store` API.

harsha-simhadri (approved):

PR #1354, opened 2026-08-25
Tweak behavior of adaptive L

magdalendobson:
Adaptive L search had a bit of strange behavior at the margins with the interaction of `sample_count` and resizing. If no matching samples were found when a resize was attempted, it would default to the maximum multiplier, meaning that if `sample_count` is relatively low, and selectivity is low but not nearly low enough to prompt resizing to the maximum multiplier, you could see an unnecessary, very large increase in `L_search`. This PR fixes that issue by resizing to `1/sample_count` if no matching elements were found, and then allowing more than one opportunity for the queue to resize, allowing a resize attempt each time the number of comparisons doubled until at least one matching point was found. This allows for more graceful behavior, and makes it harder to enact a pathological mismatch between sampling and resizing.

Along the way, after Mark noticed some issues with nondeterminism in `log10`, it adds some tolerance to the unit tests that depend on that function.


diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -253,21 +253,22 @@ where
         scratch.cmps += one_hop_neighbors.len() as u32;
         scratch.hops += scratch.beam_nodes.len() as u32;
 
-        // Adaptive L: after enough samples, estimate specificity and scale L.
+        // Adaptive L: estimate specificity at N samples, then at 2N, 4N, and so on.
```
hildebrandmw, line 256: Silly nit: May want to add one more: `2N, 4N, 8N` since this plausibly looks like the sequence could be `2N, 4N, 6N`.

diskann/src/graph/search/inline_filter_search.rs
```diff
@@ -253,21 +253,22 @@ where
...
-            l_adjusted = true;
             let new_l = compute_adaptive_l(
-                l_search,
+                original_l_search,
                 sample_visited,
                 sample_matched,
                 adaptive_l.scale_factor,
             );
-            if new_l > l_search {
+            if new_l > scratch.best.capacity() {
```
hildebrandmw, line 267: I'll agree with the bot - we probably want `scratch.best.search_l()` here.

diskann/src/graph/test/cases/inline.rs
```diff
@@ -188,8 +188,8 @@ impl Setup1D {
     ///
     /// With a sample count of 10, the two seeded IDs will result in a 20% hit-rate.
     ///
-    /// This will boost `l` to 10. The additional point 44 requires this larger `l` to hit.
-    /// We do not expect `43` to be hit.
+    /// This will boost `l` to 10. Re-evaluating at each 2x sample threshold allows the
+    /// search to reach the additional points 44 and 43.
```
hildebrandmw, line 192: Including "43" in the test results defeats the purpose of the test. Since all matching points are now returned, we are no longer testing that the search window is upper-bounded.

While this can be fixed by changing the point to "42", I have another concern. The behavior of the algorithm becomes much harder to predict once multiple sample points are introduced.  Before when there was a single sample point and change to `L`, it was easier to understand and test. Now that there are multiple rounds, this gets harder. For this test case, the behavior looks something like:

* Sample at 10 visited with 2 matches. Specificity = 0.2 so we double the original L-search (5) to 10.
* Sample at 20 visited with 2 matches. Specificity = 0.1, so we double the original L-search (5) to 10.
* Sample at 40 visited with 2 matches. Specificity = 0.05, so we change the original L-search (5) to 12 and thus pick up the extra point that used to be on the boundary.

I also suspect that changing L-search later into a search yields diminishing returns since nodes that have already fallen off the back of the queue cannot be recovered. This can kind of be seen in `graph/test/cases/inline/inline_adaptive_l_max` which (1) is no longer testing the maximum search case, so is now misnamed and (2) goes from finding 3 of the 4 matching points to just 1 (albeit, with less exploration).

Can we find an alternate way to avoid this complexity? For example, resample only when `matched == 0` and then stop resampling once at least one matched point is found? Alternatively, if this is the best approach, can the tests be adjusted so corner cases and bounds are still exercised and checked?

diskann/src/graph/test/cases/inline.rs
```diff
@@ -218,8 +218,8 @@ impl Setup1D {
         }
     }
 
-    /// No matching items are found durihng the sample window. Adaptive will boost the
-    /// window size to the max.
+    /// No matching items are found during the initial sample window. With 2x sample
+    /// thresholds, adaptive L grows incrementally and reaches the nearest match.
```
hildebrandmw, line 222: See my other comment: this case is now either misnamed or is not set up incorrectly.

hildebrandmw (commented):
Thanks Magdalen - I have some concerns about the increased complexity and loss of test accuracy. Please see the inline comments.

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: Haha sure, done

diskann/src/graph/search/inline_filter_search.rs
magdalendobson: thanks, done

diskann/src/graph/test/cases/inline.rs
magdalendobson: After some testing I agree with you that we can restrict this fix to the zero-match case, since the main pathological behavior happens when the number of initial samples is relatively small compared to the maximum multiplier. After making this change, some of the tests could be reverted, and others had to be tweaked to preserve their intent.

diskann/src/graph/test/cases/inline.rs
```diff
@@ -488,17 +490,17 @@ fn inline_search_three_level_no_adaptive_l_with_l1_finds_no_matches() {
 }
 
 #[test]
-fn inline_search_three_level_adaptive_l_with_l1_finds_matches() {
+fn inline_search_three_level_adaptive_l_with_l7_finds_matches() {
     let rt = current_thread_runtime();
     let mut test_root = root();
     let mut path = test_root.path();
-    let name = path.push("inline_search_three_level_adaptive_l_with_l1_finds_matches");
+    let name = path.push("inline_search_three_level_adaptive_l_with_l7_finds_matches");
```
hildebrandmw: Is there any adaptive L adjustment when `l = 7`? We'd find 9 naturally right?

magdalendobson: You're right, it turns out that with the changes to the adaptive L behavior, the current three-level example didn't work in the sense that there was no L value that yielded zero matches without adaptive L and at least one match with adaptive L. This was fixed by widening the tree a bit.

hildebrandmw (approved):

diskann/src/graph/test/cases/inline.rs
```diff
@@ -74,35 +69,34 @@ fn build_three_level_labeled_provider() -> test_provider::Provider {
...
+    for level_one_id in 1..=10 {
+        let first_child = 11 + (level_one_id - 1) * 2;
+        points.push((
+            level_one_id,
+            vec![0.0],
+            AdjacencyList::from_iter_untrusted([start_id, first_child, first_child + 1]),
+        ));
+    }
+    for level_two_id in 11..=30 {
+        let parent = 1 + (level_two_id - 11) / 2;
```
metajack, line 84: on the first one, level_two_id is 11, so this is 1/2. Does that make sense? This is the id of the parent node right? So it should be a number between 1 and 10?

metajack (approved):

diskann/src/graph/test/cases/inline.rs
magdalendobson: Isn't it 1? 1 + (11-11)/2=1

PR #1357, opened 2026-08-26
[diskann-garnet] Implement `continue_search()`

metajack:
Implements `continue_search()` in the Garnet FFI.

The buffers for ids that Garnet passes may be insufficient since external IDs are user-provided byte strings of arbitrary length. The distances buffer will always be correctly sized. In the case the id buffer is too small, a `Continuation` is boxed and returned, which can be used by potentially repeated calls to `continue_search()` to retrieve the rest of the results.

This set up is used all the `search_X` FFI methods. It is not used in `random_members()` since due to how other things are handled with that it can just use multiple calls to get more random members if needed.

### DIFF
--- a/diskann-garnet/src/lib.rs
+++ b/diskann-garnet/src/lib.rs
@@ -117,27 +117,42 @@ pub enum VectorQuantType {
 
 /// Helper struct to manage the FFI buffers for handling search results
 ///
-/// NOTE: The ids will be 4-byte length prefixed, and external IDs are arbitrary length
-/// byte strings.
+/// If the supplied ID buffer from Garnet isn't large enough, `overflow_ids` and
+/// `overflow_dists` will allocate enough space to store the remaining entries
+/// so they can be fetched by Garnet later.
+///
+/// Having overflow allocations means we don't allocate extra until actually
+/// needed. Short result sets can happen without allocation.
+///
+/// NOTE: The ids will be 4-byte length prefixed, and external IDs are arbitrary
+/// length byte strings.
 struct SearchResults<'a> {
+    k: usize,
     ids: &'a mut [u8],
     dists: &'a mut [f32],
     index: usize,
     id_index: usize,
+    overflow_ids: Vec<u8>,
+    overflow_dists: Vec<f32>,
 }
 
 impl SearchResults<'_> {
     /// Construct from the raw pointers
-    fn new(ids: *mut u8, ids_len: usize, dists: *mut f32, dists_len: usize) -> Self {
+    fn new(k: usize, ids: *mut u8, ids_len: usize, dists: *mut f32, dists_len: usize) -> Self {
         let ids = unsafe { slice::from_raw_parts_mut(ids, ids_len) };
         let dists = unsafe { slice::from_raw_parts_mut(dists, dists_len) };
         let index = 0;
         let id_index = 0;
+        let overflow_ids = Vec::new();
+        let overflow_dists = Vec::new();
         Self {
+            k,
             ids,
             dists,
             index,
             id_index,
+            overflow_ids,
+            overflow_dists,
         }
     }
 
@@ -146,20 +161,43 @@ impl SearchResults<'_> {
     fn push_id(&mut self, id: GarnetId) -> diskann::graph::BufferState {
         self.push(Neighbor::new(id, 0.0))
     }
+
+    fn overflowing(&self) -> bool {
+        !self.overflow_ids.is_empty()
+    }
+
+    fn into_overflows(self) -> (Vec<u8>, Vec<f32>) {
+        (self.overflow_ids, self.overflow_dists)
+    }
+
+    fn is_full(&self) -> bool {
+        self.index + self.overflow_dists.len() >= self.k
+    }
 }
 
 impl SearchOutputBuffer<GarnetId> for SearchResults<'_> {
     fn size_hint(&self) -> Option<usize> {
-        Some(self.dists.len() - self.index)
+        Some(self.k - self.index - self.overflow_dists.len())
     }
 
     fn push(&mut self, neighbor: Neighbor<GarnetId>) -> diskann::graph::BufferState {
         let (id, distance) = neighbor.as_tuple();
 
-        if self.index >= self.dists.len()
+        if self.is_full() {
+            return diskann::graph::BufferState::Full;
+        } else if self.overflowing()
+            || self.index >= self.dists.len()
             || self.id_index + mem::size_of::<u32>() + id.len() > self.ids.len()
         {
-            return diskann::graph::BufferState::Full;
+            self.overflow_ids
+                .extend_from_slice(id.as_prefixed_key_bytes());
+            self.overflow_dists.push(distance);
+
+            if self.is_full() {
+                return diskann::graph::BufferState::Full;
+            } else {
+                return diskann::graph::BufferState::Available;
+            }
         }
 
         let id_len = id.len() as u32;
@@ -172,7 +210,7 @@ impl SearchOutputBuffer<GarnetId> for SearchResults<'_> {
         self.index += 1;
         self.id_index += id.len();
 
-        if self.index >= self.dists.len() || self.id_index >= self.ids.len() {
+        if self.is_full() {
             diskann::graph::BufferState::Full
         } else {
             diskann::graph::BufferState::Available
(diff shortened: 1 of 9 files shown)


diskann-garnet/src/lib.rs
```diff
@@ -837,14 +1007,28 @@ pub unsafe extern "C" fn search_element(
 pub unsafe extern "C" fn continue_search(
     _context: u64,
     _index_ptr: *const c_void,
-    _continuation: *mut c_void,
-    _output_ids: *mut u8,
-    _output_ids_len: usize,
-    _output_distances: *mut f32,
-    _output_distances_len: usize,
-    _new_continuation: *mut c_void,
+    continuation: *mut c_void,
```
hildebrandmw, line 1022: Why not modify the existing continuation in place? This effectively moves out of the old continuation and into `new_continuation`, but doesn't take steps to make `continuation` null. The caller then has to know to not attempt to free `continuation`.

diskann-garnet/src/lib.rs
```diff
@@ -837,14 +1007,28 @@ pub unsafe extern "C" fn search_element(
...
+    output_ids: *mut u8,
+    output_ids_len: usize,
+    output_distances: *mut f32,
+    output_distances_len: usize,
+    new_continuation: *mut *mut c_void,
 ) -> i32 {
-    -1
+    if continuation.is_null() || new_continuation.is_null() {
```
hildebrandmw: You may want to set `new_continuation` to `null` if `continuation` is `null` but `new_continuation` is not. Otherwise, if a caller provides a non-null pointer for some reason, they might assume that `new_continuation` is initialized when it's not.

diskann-garnet/src/lib.rs
```diff
@@ -837,14 +1007,28 @@ pub unsafe extern "C" fn search_element(
...
+
+    let output_ids = unsafe { slice::from_raw_parts_mut(output_ids, output_ids_len) };
+    let output_distances =
+        unsafe { slice::from_raw_parts_mut(output_distances, output_distances_len) };
+    let mut continuation = unsafe { Continuation::from_ptr(continuation) };
+    let count = continuation.drain(output_ids, output_distances);
+
+    if !continuation.is_empty() {
+        unsafe { new_continuation.write(continuation.into_ptr()) };
+    }
```
hildebrandmw, line 1055: Same as above - is it worth writing `null` to `new_continuation` to make absolutely sure it's not dangling?

diskann-garnet/src/lib.rs
```diff
@@ -695,7 +843,7 @@ pub unsafe extern "C" fn search_vector(
     output_distances: *mut f32,
     output_distances_len: usize,
     beam_width: u32,
-    _continuation: *mut c_void,
+    continuation: *mut *mut c_void,
```
hildebrandmw, line 846: Do we want to support `null` continuation pointers (implying no continuation). Or is it expected that all calls need to be able to handle (and properly free) a continuation?

This optionality might help prevent `random_members` from yeeting IDs into the void if the IDs buffer was incorrectly sized and instead terminate a little earlier.

hildebrandmw (commented):
Thanks Jack - just a few small API questions!

diskann-garnet/src/lib.rs
```diff
@@ -674,8 +712,118 @@ pub unsafe extern "C" fn set_attribute(
...
+                .copy_from_slice(&self.id_buffer[self.id_index..self.id_index + prefix_len]);
+
+            // Copy length prefix
+            ids[id_index..id_index + prefix_len]
+                .copy_from_slice(&self.id_buffer[self.id_index..self.id_index + prefix_len]);
+
+            // We check there is room before advancing the indices
+            if id_index + prefix_len + len as usize > ids.len() {
+                break;
+            }
```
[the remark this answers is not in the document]
metajack: Not checking self.id_index is intention. It is constructed correctly and shouldn't be possible to have a corrupt entry. The second issue is fixed.

diskann-garnet/src/lib.rs
```diff
@@ -743,7 +892,17 @@
         if stats.result_count > i32::MAX as u32 {
             -1
         } else {
-            stats.result_count as i32
+            let count = output.current_len();
+
+            if output.overflowing() {
+                let (id_buffer, dist_buffer) = output.into_overflows();
+                let cont = Continuation::new(id_buffer, dist_buffer);
+                unsafe { continuation.write(cont.into_ptr()) };
```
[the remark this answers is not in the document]
metajack: Fixed.

diskann-garnet/src/lib.rs
```diff
@@ -743,7 +892,17 @@
...
+            } else {
+                unsafe { continuation.write(ptr::null_mut()) };
```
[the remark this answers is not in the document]
metajack: Fixed.

diskann-garnet/src/lib.rs
```diff
@@ -817,7 +979,15 @@
         if stats.result_count > i32::MAX as u32 {
             -1
         } else {
-            stats.result_count as i32
+            let count = output.current_len();
+
+            if output.overflowing() {
+                let (id_buffer, dist_buffer) = output.into_overflows();
+                let cont = Continuation::new(id_buffer, dist_buffer);
+                unsafe { continuation.write(cont.into_ptr()) };
```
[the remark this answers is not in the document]
metajack: Fixed.

diskann-garnet/src/lib.rs
```diff
@@ -1006,7 +1192,16 @@
 
     output.extend(neighbors);
 
-    output.current_len() as i32
+    let count = output.current_len();
+    if output.overflowing() {
+        let (id_buffer, dist_buffer) = output.into_overflows();
+        let cont = Continuation::new(id_buffer, dist_buffer);
+        unsafe { continuation.write(cont.into_ptr()) };
```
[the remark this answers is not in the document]
metajack: Fixed.

diskann-garnet/src/lib.rs
```diff
@@ -1006,7 +1192,16 @@
...
+    } else {
+        unsafe { continuation.write(ptr::null_mut()) };
```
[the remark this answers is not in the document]
metajack: Fixed.

diskann-garnet/src/lib.rs
metajack: Fixed.

diskann-garnet/src/lib.rs
metajack: Fixed.

diskann-garnet/src/lib.rs
metajack: It is expected that it is non null and that Garnet provides a possible continuation.

`random_members` doesn't have one because if it is short, it can just call the function again. That's perhaps not the cheapest path, but since the function is for debugging I think the Garnet team though it was just simpler that way.

diskann-garnet/src/lib.rs
metajack: It doesn't actually move out. The pointer actually remains the same.

That said, I did discuss this with the Garnet team, and they prefer to keep it as is, but added notes to their docs taht the continuation parameter should not be used after being passed in.

hildebrandmw (approved):

diskann-garnet/src/lib.rs
```diff
@@ -117,27 +117,42 @@ pub enum VectorQuantType {
 
 /// Helper struct to manage the FFI buffers for handling search results
 ///
-/// NOTE: The ids will be 4-byte length prefixed, and external IDs are arbitrary length
-/// byte strings.
+/// If the supplied buffers from Garnet aren't large enough, `overflow_ids` and
+/// `overflow_dists` will allocate enough space to store the remainining entries
```
magdalendobson: typo :)

magdalendobson: I don't understand why we would be in a situation where distances overflow. Large id sizes might mean we don't allocate enough space for the ids, but it seems like we should always know how many distances we're requesting, right?

magdalendobson:
Low stakes feedback but I think that `continue_search` and `Continuation` are potentially confusing names. They may make people think that DiskANN is doing more search in a paged search manner. Could we workshop a bit on naming, maybe something like `continue_buffering` and `BufferContinuation`?

diskann-garnet/src/lib.rs
metajack: I changed the doc to hopefully make this more clear.

metajack: Fixed.

metajack:
Good suggestion. I've renamed it to `overflow_results` and `Overflow`.

magdalendobson (approved):

PR #1359, opened 2026-08-27
Make flat search visitors query-aware

partychen:

#### Reference Issues/PRs

Prerequisite API refactor requested during review of #1341.

Makes flat search visitors query-aware, moves the search algorithm to the free `flat::knn_search` entry point, and removes the unnecessary `FlatIndex` wrapper. It also updates the generic flat tests, test providers, benchmark integration, and API rustdoc for the redesigned public API.

The redesigned interface has several benefits:

- A visitor is constructed for a specific query, so it can own or borrow query preprocessing results and combine them with backend-specific I/O, batching, filtering, and distance-computation state. This is particularly useful for streamed and quantized backends such as the disk PQ scan in #1341.
- `DistancesUnordered` now emits `(id, distance)` pairs directly. Backends can fuse scanning and distance computation instead of exposing every stored element through a common `ElementRef` and external `QueryComputer` abstraction.
- Implementations have a smaller and less brittle type surface. The redesign removes the `ElementRef`, `QueryComputer`, and `QueryComputerError` associated types, the visitor GAT, and their HRTB/lifetime constraints.
- The free `flat::knn_search(&provider, ...)` function borrows the provider directly, removing a stateless ownership wrapper and making shared providers and concurrent searches more natural.
- Responsibilities are clearer: the generic algorithm manages top-k selection and post-processing, while the query-aware visitor owns the backend-specific complete scan.

#### Any other comments?

This intentionally changes the existing public flat-search API and is not backwards compatible. The trade-off is a breaking migration for current callers in exchange for an interface that can naturally represent query-aware, streaming, and quantized backends. #1341 will remain open and be rebased onto `main` after this prerequisite merges.

### DIFF
--- a/diskann/src/flat/index.rs
+++ b/diskann/src/flat/index.rs
@@ -3,7 +3,7 @@
  * Licensed under the MIT license.
  */
 
-//! Brute-force k-nearest-neighbor search over a [`DataProvider`].
+//! Brute-force k-nearest-neighbor search over a [`DistancesUnordered`] visitor.
 use std::num::NonZeroUsize;
 
 use diskann_utils::future::SendFuture;
@@ -11,10 +11,9 @@ use diskann_utils::future::SendFuture;
 use crate::{
     ANNResult,
     error::{ErrorExt, IntoANNResult},
-    flat::{DistancesUnordered, SearchStrategy},
+    flat::DistancesUnordered,
     graph::{SearchOutputBuffer, glue::SearchPostProcess},
     neighbor::{Neighbor, NeighborPriorityQueue},
-    provider::DataProvider,
 };
 
 /// Statistics collected during a flat search.
@@ -27,43 +26,35 @@ pub struct SearchStats {
     pub result_count: u32,
 }
 
-/// Brute-force k-nearest-neighbor search over a borrowed provider.
+/// Brute-force k-nearest-neighbor search over an initialized visitor.
 ///
-/// Streams every distance produced by the strategy's query-aware visitor, keeps the best `k`
-/// candidates in a [`NeighborPriorityQueue`], then runs `processor` over the survivors
-/// to populate `output`.
+/// Borrows `visitor` for the duration of the search, streams every distance it produces,
+/// keeps the best `k` candidates in a [`NeighborPriorityQueue`], then runs `processor`
+/// over the same visitor and the surviving candidates to populate `output`.
 ///
-/// The provider, strategy, and context are borrowed for the shared lifetime used to
-/// construct the visitor. This permits the visitor to retain references to their search
-/// state until scanning and post-processing are complete.
+/// The visitor remains owned by the caller and becomes available again after the returned
+/// future completes. Whether it supports another scan is determined by its implementation.
 ///
 /// # Errors
 ///
-/// Returns an error if visitor construction, distance scanning, or result
-/// post-processing fails. Distance-scan errors are escalated because a
-/// partial flat scan cannot produce correct k-nearest-neighbor results.
-pub fn knn_search<'a, P, S, T, O, PP, OB>(
-    provider: &'a P,
+/// Returns an error if distance scanning or result post-processing fails. Distance-scan
+/// errors are escalated because a partial flat scan cannot produce correct k-nearest-neighbor
+/// results.
+pub fn knn_search<V, T, O, PP, OB>(
+    visitor: &mut V,
     k: NonZeroUsize,
-    strategy: &'a S,
     processor: PP,
-    context: &'a P::Context,
     query: T,
     output: &mut OB,
 ) -> impl SendFuture<ANNResult<SearchStats>>
 where
-    P: DataProvider,
-    S: SearchStrategy<'a, P, T>,
+    V: DistancesUnordered,
     T: Copy + Send + Sync,
     O: Send,
-    PP: SearchPostProcess<S::Visitor, T, O> + Send + Sync,
+    PP: SearchPostProcess<V, T, O> + Send + Sync,
     OB: SearchOutputBuffer<O> + Send + ?Sized,
 {
     async move {
-        let mut visitor = strategy
-            .create_visitor(provider, context, query)
-            .into_ann_result()?;
-
         let k = k.get();
         let mut queue = NeighborPriorityQueue::new(k);
         let mut cmps: u32 = 0;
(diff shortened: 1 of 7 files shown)


diskann/src/flat/test/cases/flat_knn_search.rs
```diff
@@ -3,20 +3,17 @@
  * Licensed under the MIT license.
  */
 
-//! Baseline-cached regression sweep for [`crate::flat::FlatIndex::knn_search`].
+//! Baseline-cached regression sweep for [`crate::flat::knn_search`].
 //!
 //! Bbuilds a fresh index per parameter combination, runs `knn_search` through the
 //! [`crate::flat::test::harness`], snapshots the result + statistics into
```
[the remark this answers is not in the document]
partychen: Fixed both stale references: the module docs now say a fresh provider is built, and `run_row` now documents the shared provider. Also corrected the `Bbuilds` typo.

diskann/src/flat/strategy.rs
```diff
@@ -8,96 +8,58 @@
...
-pub trait DistancesUnordered<C>: HasId + Send + Sync
-where
-    C: for<'a> PreprocessedDistanceFunction<Self::ElementRef<'a>, f32>,
-{
-    /// Lifetime is intentionally unconstrained so it can appear under HRTB without
-    /// inducing a `'static` bound on `Self`.
-    type ElementRef<'a>;
-
+/// The accessor owns the query-specific computation state so implementations can fuse
+/// query preprocessing, data access, batching, filtering, and distance computation.
```
arkrishn94: Can we expand this documentation to be a bit more descriptive and remove the stuff about why the computer is owned (which is context specific to this PR, rather than to a future user). 

Something like - 
```rust
/// Implementations drive an entire scan over the underlying data, scoring each element 
/// and invoking `f` with the resulting `(id, distance)` pair.
```

diskann/src/flat/strategy.rs
```diff
@@ -8,96 +8,58 @@
...
-    fn distances_unordered<F>(
-        &mut self,
-        computer: &C,
-        f: F,
-    ) -> impl SendFuture<Result<(), Self::Error>>
+    /// Drive the entire scan, invoking `f` with each `(id, distance)` pair.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error when the backend cannot complete the scan.
```
arkrishn94: There could be other errors? Maybe just omit the errors section?

diskann/src/flat/strategy.rs
```diff
@@ -8,96 +8,58 @@
...
-            Id = P::InternalId,
-        >
-    where
-        Self: 'a,
-        P: 'a;
+    /// The query-aware visitor used to execute the scan.
+    type Visitor: DistancesUnordered<Id = P::InternalId>;
 
-    /// The error type for [`Self::create_visitor`].
+    /// An error that can occur while constructing [`Self::Visitor`].
```
arkrishn94: nit: why change documentation?

diskann/src/flat/strategy.rs
```diff
@@ -8,96 +8,58 @@
...
     type Error: StandardError;
 
-    /// Construct a fresh visitor over `provider` for the given request `context`.
-    fn create_visitor<'a>(
+    /// Construct a fresh visitor for `query`.
```
arkrishn94: nit: there is no `query` in the argument here, maybe leave the documentation?

diskann/src/flat/index.rs
```diff
@@ -28,74 +27,57 @@ pub struct SearchStats {
...
+/// Streams every distance produced by the strategy's query-aware visitor, keeps the best `k`
+/// candidates in a [`NeighborPriorityQueue`], then runs `processor` over the survivors
+/// to populate `output`.
+///
+/// # Errors
+///
+/// Returns an error if visitor construction, distance scanning, or result
+/// post-processing fails. Distance-scan errors are escalated because a
+/// partial flat scan cannot produce correct k-nearest-neighbor results.
+pub fn knn_search<'a, P, S, T, O, PP, OB>(
```
arkrishn94: I'm curious, why do we need to add a lifetime to this function? It seems like it is introducing unnecessary coupling between the provider, strategy and context?

arkrishn94 (commented):
Thanks Junkui, I like the simplification here. Only left some minor comments, but apart from that happy to approve.

diskann/src/flat/strategy.rs
partychen: Updated as suggested. The trait documentation now describes the durable scan-and-score contract: implementations scan the underlying data, score each element, and emit each `(id, distance)` pair through the callback.

diskann/src/flat/strategy.rs
partychen: Updated this to cover the general contract rather than naming one backend failure mode: implementations return their implementation-defined error when the complete scan cannot be produced. I kept the `# Errors` section because this public method returns a `Result`, and the repository rustdoc conventions expect its failure contract to be documented.

diskann/src/flat/strategy.rs
partychen: Restored the concise wording: `The error type for Self::create_visitor`.

diskann/src/flat/strategy.rs
partychen: Clarified this as constructing a visitor over the provider for the given request context and query, matching all four method arguments.

diskann/src/flat/index.rs
partychen: The named lifetime selects the `SearchStrategy<'a, P, T>` implementation and its associated `Visitor`, which may borrow from the strategy, provider, context, and a borrowed query. It does not require those values to have identical original lifetimes; Rust reborrows them to a common shorter lifetime for this call. Without a named lifetime here, we would need to reintroduce a `Visitor<'a>` GAT or use a stronger HRTB that does not model borrowed query types well. This follows the same trait-level lifetime shape as the graph-search strategy API, so I kept the signature unchanged.

hildebrandmw (approved):
Thanks! I've been wanting to move the graph index to a free-function style interface (taking the `SearchAccessor`/`PruneAccessor`s directly) for some time now, though that will have to happen after the inmem2 transition is completely.

One thing to consider is if it makes sense here to take the `Visitor` directly rather than the `Strategy` + `Provider` + `Query` combination.

arkrishn94 (approved):

xwj-ox (approved):

PR #1365, opened 2026-09-01
Fuse MaxSim reduction at panel boundaries

partychen:
## Context

This draft builds directly on #1333 at `4683fd8732988cbc9833186d55f1530b4c3b53b0` and targets `users/suryangupta/multi_vector_kernel_new_design`. It preserves that PR's separation of layout, tiling, and leaf micro-kernels while exploring a narrower reduction boundary for MaxSim.

Mark observed that materializing every raw A-panel x B-panel product into `Strip` and draining it later adds avoidable traffic even though the leaf already knows when a complete panel result is available. This follow-up keeps the parent architecture and refines only that handoff.

## Problem

The current `Scratch` / `Slots` / `Accumulate` / `Drain` contract writes tile-sized raw products and then makes a second pass to reduce them. Besides the extra memory traffic, the contract spreads MaxSim-specific behavior across non-local plumbing, and Scalar and V3 duplicate substantial storage/reduction machinery.

## Design

- Replace the strip contract with a typed `PanelOp` invoked after each full A-panel x B-panel contraction.
- Let the f32 Scalar and V3 leaves fold completed document lanes directly into operation-owned running MaxSim state.
- Keep f16 widening its tiles and reusing the f32 path.
- Remove `kernels/strip.rs` and the tile-sized intermediate allocation.
- Keep panel geometry, tile traversal, layout types, and micro-kernel selection unchanged.

This is an architecture/performance remediation of the parent refactor, not a new kernel implementation.

## Correctness

The panel operation receives the same ordinal A/B panel positions as the previous drive contract, preserving full panels, B tails, A/B tile ordinal behavior, empty dimensions/documents, and padded query rows.

The reduction retains Rust `f32::max` behavior around NaNs. In V3, finite running state is intentionally the right operand to the hardware max instruction; this preserves a valid prior maximum when one document lane produces NaN without introducing a generic standard-max comparison. A regression test covers this case.

## Performance methodology

The primary evidence is the original strict three-way comparison of:

- base/main: `3218478b5f7d1721840c29309e22d9fd74b1b4bc`
- #1333 HEAD: `4683fd8732988cbc9833186d55f1530b4c3b53b0`
- this streaming panel-reduction implementation

It used the official nine shapes, release builds, 50 measurements per shape, minimum ns/IP, and rotated implementation order to reduce ordering bias. Lower percentages are better. Reference is excluded from conclusions because this patch does not alter it. Auto duplicates V3 on this machine and is omitted below.

### Aggregate results

| Implementation | HEAD vs base median | HEAD vs base P90 | This PR vs base median | This PR vs base P90 | This PR vs HEAD median | This PR vs HEAD P90 |
|---|---:|---:|---:|---:|---:|---:|
| f32 Scalar | +1.96% | +4.68% | +0.35% | +0.84% | -1.27% | -0.31% |
| f32 V3 | +0.73% | +4.05% | -0.20% | 0.00% | -0.93% | -0.45% |
| f16 V3 | +1.88% | +60.60% | 0.00% | +0.37% | -2.46% | +0.45% |

The f16 V3 P90 for HEAD is dominated by one noisy/sensitive shape; the complete rows are included rather than hiding that variance.

### Complete per-shape results

| Type / ISA | Q | D | Dim | Base ns/IP | HEAD ns/IP | This PR ns/IP | HEAD vs base | This PR vs base | This PR vs HEAD |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| f32 Scalar | 16 | 64 | 256 | 6.8359 | 6.9238 | 6.8359 | +1.29% | 0.00% | -1.27% |
| f32 Scalar | 32 | 1250 | 128 | 3.0725 | 3.1500 | 3.0950 | +2.52% | +0.73% | -1.75% |
| f32 Scalar | 32 | 128 | 384 | 10.3760 | 10.4614 | 10.4126 | +0.82% | +0.35% | -0.47% |
| f32 Scalar | 32 | 16 | 256 | 6.8457 | 6.9922 | 6.8848 | +2.14% | +0.57% | -1.54% |
| f32 Scalar | 32 | 32 | 512 | 13.9648 | 14.0820 | 14.0039 | +0.84% | +0.28% | -0.55% |
| f32 Scalar | 64 | 1250 | 512 | 14.1312 | 14.1438 | 14.1000 | +0.09% | -0.22% | -0.31% |
| f32 Scalar | 64 | 32 | 128 | 3.0737 | 3.1494 | 3.0811 | +2.46% | +0.24% | -2.17% |
| f32 Scalar | 64 | 32 | 264 | 6.9922 | 7.1289 | 7.0508 | +1.96% | +0.84% | -1.10% |
| f32 Scalar | 8 | 32 | 128 | 3.1719 | 3.3203 | 3.1875 | +4.68% | +0.49% | -4.00% |
| f32 V3 | 16 | 64 | 256 | 4.1504 | 4.1992 | 4.1504 | +1.18% | 0.00% | -1.16% |
| f32 V3 | 32 | 1250 | 128 | 2.0650 | 2.0800 | 2.0625 | +0.73% | -0.12% | -0.84% |
| f32 V3 | 32 | 128 | 384 | 6.2012 | 6.2378 | 6.1890 | +0.59% | -0.20% | -0.78% |
| f32 V3 | 32 | 16 | 256 | 4.1992 | 4.2773 | 4.1895 | +1.86% | -0.23% | -2.05% |
| f32 V3 | 32 | 32 | 512 | 8.3008 | 8.3594 | 8.2812 | +0.71% | -0.24% | -0.93% |
| f32 V3 | 64 | 1250 | 512 | 8.3250 | 8.3500 | 8.3125 | +0.30% | -0.15% | -0.45% |
| f32 V3 | 64 | 32 | 128 | 2.0850 | 2.1240 | 2.0801 | +1.87% | -0.23% | -2.07% |
| f32 V3 | 64 | 32 | 264 | 4.2773 | 4.3066 | 4.2676 | +0.68% | -0.23% | -0.91% |
| f32 V3 | 8 | 32 | 128 | 4.2422 | 4.4141 | 4.2422 | +4.05% | 0.00% | -3.89% |
| f16 V3 | 16 | 64 | 256 | 5.3711 | 5.4297 | 5.3613 | +1.09% | -0.18% | -1.26% |
| f16 V3 | 32 | 1250 | 128 | 2.2425 | 2.2350 | 2.2450 | -0.33% | +0.11% | +0.45% |
| f16 V3 | 32 | 128 | 384 | 6.7871 | 6.9580 | 6.7871 | +2.52% | 0.00% | -2.46% |
| f16 V3 | 32 | 16 | 256 | 6.7578 | 6.8848 | 6.6309 | +1.88% | -1.88% | -3.69% |
| f16 V3 | 32 | 32 | 512 | 10.5469 | 12.3438 | 10.5859 | +17.04% | +0.37% | -14.24% |
| f16 V3 | 64 | 1250 | 512 | 8.5562 | 8.6312 | 8.5812 | +0.88% | +0.29% | -0.58% |
| f16 V3 | 64 | 32 | 128 | 2.8027 | 2.7930 | 2.7368 | -0.35% | -2.35% | -2.01% |
| f16 V3 | 64 | 32 | 264 | 7.1387 | 11.4648 | 7.0020 | +60.60% | -1.92% | -38.93% |
| f16 V3 | 8 | 32 | 128 | 6.1250 | 6.3125 | 6.1406 | +3.06% | +0.26% | -2.72% |

## Scope / non-goals

This PR intentionally does not add V4 or Neon kernels, change MinMax or Spherical behavior, modify `ScopedAllocator`, add cache detection, or introduce any other feature. It changes only the six MaxSim kernel paths needed to move reduction to the completed-panel boundary.

## Validation

- `cargo fmt --all --check`
- `cargo test -p diskann-quantization multi_vector` (82 passed)
- `cargo clippy -p diskann-quantization --all-targets -- -D warnings`
- Full six-job benchmark input completed all nine shapes per job on the exact committed code

## Limitations

This machine does not support AVX-512, so no V4 execution claim is made. Standalone later runs, including the exact-final completion run, experienced machine-frequency drift; they are used only to confirm that the exact code completes every benchmark case, not as cross-version percentage evidence. The strict rotated three-way table above remains the primary performance comparison.


diskann-quantization/src/multi_vector/distance/kernels/mod.rs
```diff
@@ -77,43 +73,34 @@ fn at_least_one_panel(panels: usize) -> NonZeroUsize {
...
+    /// only feeds a walk stride, which the cursor already bounds by the data it holds.
     ///
     /// Row sizes are [`NonZeroUsize`] because they reach a divisor. The entries reject an
     /// empty contraction before planning.
     fn new(
         a_row_bytes: NonZeroUsize,
         b_row_bytes: NonZeroUsize,
         b_rows: usize,
-        acc_bytes: usize,
+        work_bytes_per_b_row: usize,
```
partychen, line 85: AI: consider: `work_bytes_per_b_row` sounds like the complete working-set cost for one B row, but `Plan::new` multiplies it by `AR`. That makes the actual unit “bytes per A-row per B-row”; a future operation supplying an already-total per-B-row cost would over-reserve L1 by `AR`. Could we encode that unit in the parameter name (for example, `work_bytes_per_a_row_per_b_row`) and update the test wording? (`names-match-vocabulary`)

diskann-quantization/src/multi_vector/distance/kernels/leaves/v3.rs
```diff
@@ -69,50 +67,17 @@ pub(crate) fn f32_store_microkernel<const UNROLL: usize, L: Length>(
...
-        max_into_chains(arch, &mut chains, group);
-    }
-    // Fewer than WAYS columns remain, so each still lands on its own chain.
-    max_into_chains(arch, &mut chains, groups.remainder());
-
-    let mut merged = chains[0];
-    for chain in &chains[1..] {
-        for (m, c) in merged.iter_mut().zip(chain) {
-            *m = m.max_simd(*c);
+    // `run` seeds state with a finite identity, and every update preserves that invariant.
```
partychen, line 70: AI: nit: The invariant here is “non-NaN,” not “finite”: valid inputs or overflow can produce `+∞`, which remains a valid running maximum. Could this say that `run` seeds a non-NaN identity and that placing the prior non-NaN state on the right preserves it when a candidate is NaN?

partychen:
AI: blocking: The benchmark section identifies the compared commits, shapes, measurement count, and ordering, but not the host hardware. For these ISA-sensitive results, please also record the CPU model, core topology, OS, and relevant frequency/power configuration so the measurements are reproducible and comparable (`perf-claims-need-numbers`).

PR #1372, opened 2026-09-03
Add Neon inner product U8xU1, U8xU2 and U8xU4 kernels

pfoxARM:
#### What does this implement
Adds inner product 8-bit x 4-bit unsigned vectors using neon
Adds inner product 8-bit x 2-bit unsigned vectors using neon
Adds inner product 8-bit x 1-bit unsigned vectors using neon
Adds Zip and Unzip intrinsics into diskann-wide

### DIFF
--- a/diskann-quantization/src/bits/distances.rs
+++ b/diskann-quantization/src/bits/distances.rs
@@ -73,9 +73,9 @@
 //! | `USlice<7>`   | `USlice<7>`   | `MV<u32>` | Fallback  | No            | Uses V3   | Fallback  |
 //! | `USlice<8>`   | `USlice<8>`   | `MV<u32>` | Yes       | Yes           | Yes       | Fallback  |
 //! |               |               |           |           |               |           |           |
-//! | `USlice<8>`   | `USlice<4>`   | `MV<u32>` | Fallback  | Yes           | Uses V3   | Fallback  |
-//! | `USlice<8>`   | `USlice<2>`   | `MV<u32>` | Fallback  | Yes           | Uses V3   | Fallback  |
-//! | `USlice<8>`   | `USlice<1>`   | `MV<u32>` | Fallback  | Yes           | Uses V3   | Fallback  |
+//! | `USlice<8>`   | `USlice<4>`   | `MV<u32>` | Fallback  | Yes           | Uses V3   | Optimized |
+//! | `USlice<8>`   | `USlice<2>`   | `MV<u32>` | Fallback  | Yes           | Uses V3   | Optimized |
+//! | `USlice<8>`   | `USlice<1>`   | `MV<u32>` | Fallback  | Yes           | Uses V3   | Optimized |
 //! |               |               | `       ` |           |               |           |           |
 //! | `TSlice<4>`   | `USlice<1>`   | `MV<u32>` | Optimized | Optimized     | Optimized | Optimized |
 //! |               |               | `       ` |           |               |           |           |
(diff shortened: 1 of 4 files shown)


diskann-wide/src/arch/aarch64/double.rs
```diff
@@ -90,6 +90,37 @@ super::macros::aarch64_zipunzip!(u8x16, vzip1q_u8, vzip2q_u8, vuzp1q_u8, vuzp2q_
...
+            let halves = self.split();
+            let lo_raw = halves.lo.to_underlying();
+            let hi_raw = halves.hi.to_underlying();
+            LoHi::new(
+                u8x8::from_underlying(self.arch(), vuzp1_u8(lo_raw, hi_raw)),
+                u8x8::from_underlying(self.arch(), vuzp2_u8(lo_raw, hi_raw)),
+            )
+        }
+    }
+}
```
hildebrandmw, line 122: Can you move this to the `u8x16` module and ensure coverage with `test_utils::ops::test_zipunzip!`?

I'm working on a change adding this more broadly to `diskann-wide` using the `aarch64_zipunzip` macro, but that's turning out to be a little more involved.

diskann-quantization/src/bits/distances.rs
```diff
@@ -2314,17 +2314,429 @@ impl Target2<diskann_wide::arch::x86_64::V3, MathematicalResult<u32>, USlice<'_,
...
+    fn run(
+        self,
+        arch: diskann_wide::arch::aarch64::Neon,
+        x: USlice<'_, 8>,
+        y: USlice<'_, 4>,
+    ) -> MathematicalResult<u32> {
+        // returns number of quantized vectors
+        let len = check_lengths!(x, y)?;
+
+        diskann_wide::alias!(u8s_16 = <diskann_wide::arch::aarch64::Neon>::u8x16);
```
hildebrandmw, line 2335: We can apply this patch to `diskann-wide`:
```diff
diff --git a/diskann-wide/src/lib.rs b/diskann-wide/src/lib.rs
index 4f6ca5c9..442babe9 100644
--- a/diskann-wide/src/lib.rs
+++ b/diskann-wide/src/lib.rs
@@ -195,6 +195,9 @@ pub const ARCH: arch::Current = arch::current();
 ///    ```text
 ///    use of generic parameter from outer item
 ///    ```
+/// 5. `diskann_wide::alias!(f32s = type a::b::f32x4) => type f32s<A> = a::b::f32x4`:
+///    Use a direct path to the type rather than reaching the type as an associated type of
+///    [`Architecture`].
 #[macro_export]
 macro_rules! alias {
     ($var:ident) => {
@@ -211,6 +214,10 @@ macro_rules! alias {
         #[allow(non_camel_case_types)]
         type $var<$arch> = <$arch as $crate::Architecture>::$type;
     };
+    ($var:ident = type $type:path) => {
+        #[allow(non_camel_case_types)]
+        type $var = $type;
+    };
 }

 //////////////
```
to have a specific arm of the `diskann_wide::alias!` macro for providing direct paths to types. That's probably useful for neon in general.

diskann-quantization/src/bits/distances.rs
```diff
@@ -2314,17 +2314,429 @@ impl Target2<diskann_wide::arch::x86_64::V3, MathematicalResult<u32>, USlice<'_,
...
+                let lo = input;
+                let hi = input >> 4;
+                let halves = LoHi::new(lo, hi);
+                let zipped_halves = u8s_16::zip(halves);
+                zipped_halves & u8s_16::splat(arch, 0x0f)
+            }
+
+            while i + 8 <= y_bytes {
+                // SAFETY: `i + 8 <= y_bytes` guarantees 8 readable bytes at `py_u8.add(i)`.
+                let y_vec = split_and_zip(unsafe { u8s_8::load_simd(arch, py_u8.add(i)) }, arch);
```
hildebrandmw, line 2363: Curious: did you try something like
```rust
                let y = unsafe { u8x16::load_simd(arch, py_u8.add(i)) };

                let LoHi {
                    lo: xodd,
                    hi: xeven,
                } = unsafe { u8x32::load_simd(arch, px_u8.add(2 * i)) }.unzip();

                s0 = s0.dot_simd(y & mask, xodd).dot_simd((y >> 4) & mask, xeven);
```
(disclaimer: didn't quite test it due to our apparent testing gap :cry:).

This would be about 3 loads, 3 integer ops, 2 shuffles, and 2 dots for 32 elements.

The current approach does 4 loads, 4 integer ops, 4 shuffles, and 2 dots for 32 elements.

Since vectors usually have dimensions that are a multiple of 2, I wouldn't worry too much about the extra remainder.

diskann-quantization/src/bits/distances.rs
```diff
@@ -2314,17 +2314,429 @@ impl Target2<diskann_wide::arch::x86_64::V3, MathematicalResult<u32>, USlice<'_,
...
+        diskann_wide::alias!(u32s = <diskann_wide::arch::aarch64::Neon>::u32x4);
+
+        let px_u8: *const u8 = x.as_ptr().cast();
+        let py_u8: *const u8 = y.as_ptr().cast();
+
+        let mut i = 0;
+        let mut s: u32 = 0;
+
+        #[inline(always)]
+        fn split_and_zip_four_bit(input: u8s_8, arch: diskann_wide::arch::aarch64::Neon) -> u8s_16 {
```
hildebrandmw, line 2442: Note that `arch` can come from `input.arch()`, so doesn't necessarily need to be passed separately.

diskann-quantization/src/bits/distances.rs
```diff
@@ -2314,17 +2314,429 @@ impl Target2<diskann_wide::arch::x86_64::V3, MathematicalResult<u32>, USlice<'_,
...
+        let y_bytes = len / 8; // number of y bytes over the underlying slice
+        if i < y_bytes {
+            let mut s0 = u32s::default(arch);
+            let mut s1 = u32s::default(arch);
+            let mut s2 = u32s::default(arch);
+            let mut s3 = u32s::default(arch);
+
+            while i + 8 <= y_bytes {
+                // SAFETY: `i + 8 <= y_bytes` guarantees that 8 bytes from `py_u8` are readable at offset `i`.
+                let (y_vec1, y_vec2, y_vec3, y_vec4) =
```
hildebrandmw, line 2642: Question: did you try something like
```rust
pub fn test_function(arch: Neon, x: &u8, y: &u8) -> u8x16 {
    let combined: u8x16 = LoHi::new(u8x8::splat(arch, *x), u8x8::splat(arch, *y)).join();
    let shifts = u8x16::from_array(arch, [0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2, 3, 4, 5, 6, 7]);
    (combined >> shifts) & u8x16::splat(arch, 1)
}
```
for the bit-unpacking? With `shifts` and the mask amortized, this costs 5 instructions for 16 elements. I counted around 21 instructions per 64-elements in the approach taken in the PR.

The approach above does hit the load units more, but we might be able to introduce indexing operations for `SIMDVector` to amortize the loading of the 1-bit values.

hildebrandmw (commented):
Thanks - I have a bit of feedback on this on. Mainly questions on different approaches.

I also realized that the tests for heterogeneous kernels are woefully under powered when it comes to testing the neon kernels (as in, they aren't tested at all). Some work will be needed to refactor our tests. I'll try to get around to that today or so.

MustafaIdrisArm:
> Thanks - I have a bit of feedback on this on. Mainly questions on different approaches.
> 
> I also realized that the tests for heterogeneous kernels are woefully under powered when it comes to testing the neon kernels (as in, they aren't tested at all). Some work will be needed to refactor our tests. I'll try to get around to that today or so.

Thanks for the feedback. With regards to the testing, I added the following code to `all_ip_dispatches()` in the `heterogeneous_ip_tests_8xM` macro for internal testing

```rs
#[cfg(target_arch = "aarch64")]
                    if let Some(arch) = diskann_wide::arch::aarch64::Neon::new_checked() {
    fuzz_heterogeneous_ip::<$M>(
        MAX_DIM,
        TRIALS_PER_DIM,
        $max_val,
        &|x, y| arch.run2(InnerProduct, x, y),
        "neon",
        &mut rng,
    );
}
```

With MAX_DIM set to 256 and TRIALS_PER_DIM set to 20, the coverage should be sufficient for the general use cases of the kernel. `heterogeneous_ip_tests_8xM` itself is expanded for all of our cases already (M = 4,2,1).

Note: There are also some corner case tests such as `max_values()` and `edge_cases()` where we could add neon dispatch so we test for those cases as well. The implementations are just as simple if not simpler than the snippet above, just let me know what fits best and I'll go for that.

diskann-quantization/src/bits/distances.rs
MustafaIdrisArm: Indeed, it would also be of great help to the `f32*Uslice` kernels

pfoxARM: Done :)

diskann-wide/src/arch/aarch64/double.rs
pfoxARM: Makes sense, done.

hildebrandmw (approved):
Thanks @MustafaIdrisArm - looks good to me. Regarding the added test - those look great! My earlier concern was more that our test infrastructure as a whole was not set up as well for dedicated testing as the other kernels. What you did works well within the current constraints! If you think the coverage is sufficient, that's good enough for me!

diskann-quantization/src/bits/distances.rs
pfoxARM: Will take a look at this more closely! Did briefly check this and didn't see an obviously uplift compared to this branch.

diskann-quantization/src/bits/distances.rs
pfoxARM: Appreciate this Mark, will also benchmark this further and get back to you.

diskann-quantization/src/bits/distances.rs
pfoxARM: Gotcha, will rewrite for this.

pfoxARM:
Thanks Mark, happy to rewrite for `input.arch()` before merging if you want, as well as testing your suggestions. Otherwise will follow up next week. We also have squared l2 (USlice2 and USlice4) kernels coming next.

harsha-simhadri (approved):

PR #1376, opened 2026-09-04
[diskann-utils] Simplify `StridedView`.

hildebrandmw:
We apparently don't need owned or mutable `StridedViews`. This struct was getting in the way of simplifying the `Matrix` type, so I decided to do the following:

1. Collapse everything into a single `Strided<'a, T>` struct.
2. Ensure that the layout for a `Strided` is well formed.
3. Shrink the memory representation from 40 bytes to 32 bytes.
4. Implement a custom `Rows` iterator so we no longer panic if the column-stride is zero.

Note that currently, `From<MatrixView<'a, T>>` can panic. This is because `MatrixView` and friends don't actually guarantee that `nrows * ncols` does not overflow `usize::MAX`. I have another PR in the pipeline to fix that, but ran into a chicken and egg problem with `Strided` preventing that PR from making progress.

### DIFF
--- a/diskann-utils/src/internal.rs
+++ b/diskann-utils/src/internal.rs
@@ -4,6 +4,6 @@
  */
 
 pub(crate) fn slice_to_nonnull<T>(s: &[T]) -> std::ptr::NonNull<T> {
-    // SAFETY: sliced are guaranteed to have non-null base pointers.
+    // SAFETY: slices are guaranteed to have non-null base pointers.
     unsafe { std::ptr::NonNull::new_unchecked(s.as_ptr().cast_mut()) }
 }
--- a/diskann-utils/src/internal.rs
+++ b/diskann-utils/src/internal.rs
@@ -1,0 +1,9 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+pub(crate) fn slice_to_nonnull<T>(s: &[T]) -> std::ptr::NonNull<T> {
+    // SAFETY: sliced are guaranteed to have non-null base pointers.
+    unsafe { std::ptr::NonNull::new_unchecked(s.as_ptr().cast_mut()) }
+}
--- a/diskann-utils/src/strided.rs
+++ b/diskann-utils/src/strided.rs
@@ -214,7 +214,7 @@ impl<'a, T> Strided<'a, T> {
     pub fn as_slice(&self) -> &[T] {
         let layout = self.layout();
 
-        // SAFETY: Constructors verify that the backing memory has a length as least
+        // SAFETY: Constructors verify that the backing memory has a length at least
         // `layout.linear_length()`.
         unsafe { std::slice::from_raw_parts(self.as_ptr(), layout.linear_length()) }
     }
@@ -325,7 +325,7 @@ impl<'a, T> Strided<'a, T> {
     }
 }
 
-/// Errors for [`Strided::new`].
+/// Errors for [`Strided::try_from_data`].
 #[derive(Debug, Error)]
 pub enum TryFromError {
     #[error(transparent)]
--- a/diskann-utils/src/strided.rs
+++ b/diskann-utils/src/strided.rs
@@ -5,11 +5,17 @@
 
 use std::{
     fmt,
+    marker::PhantomData,
+    num::NonZeroUsize,
     ops::{Index, IndexMut},
+    ptr::NonNull,
 };
 use thiserror::Error;
 
-use crate::views::{self, DenseData, MutDenseData};
+use crate::{
+    views::{self, DenseData, MutDenseData},
+    Reborrow, ReborrowMut,
+};
 
 /// A row-major strided matrix.
 ///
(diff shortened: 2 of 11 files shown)


diskann-utils/src/internal.rs
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+pub(crate) fn slice_to_nonnull<T>(s: &[T]) -> std::ptr::NonNull<T> {
+    // SAFETY: sliced are guaranteed to have non-null base pointers.
+    unsafe { std::ptr::NonNull::new_unchecked(s.as_ptr().cast_mut()) }
+}
```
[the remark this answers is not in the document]
hildebrandmw: Thanks.

diskann-utils/src/strided.rs
```diff
@@ -3,592 +3,671 @@
...
+    views::{self},
+    Reborrow,
+};
 
-/// A row-major strided matrix.
+/// The layout for [`Strided`].
 ///
-/// This is a generalization of the `MatrixBase` class as it does not mandate a dense
-/// layout in memory.
+/// This struct ensures that the [`Self::cstride`] is greater than [`Self::ncols`] and that
```
suri-kumkaran: typo: "greater than" -> "greater than or equal to"

diskann-utils/src/strided.rs
```diff
@@ -3,592 +3,671 @@
...
-/// slice 1 -> | b0 b1 b2 b3 | b4 b5 b6     |
-/// slice 2 -> | c0 c1 c2 c3 | c4 c5 c6   nrows
-/// slice 3 -> | d0 d1 d2 d3 | d4 d5 d6     |
-/// slice 4 -> | e0 e1 e2 e3 | e4 e5 e6     |
-/// slicf 5 -> | f0 f1 f2 f3 | f4 f5 f6     v
-///            +-------------+
-///                  ^
-///                  |
-///             StridedView
+/// self.nrows.saturating_sub(1) * self.cstride + self.nrows.min(1) * self.ncols
```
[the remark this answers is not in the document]
suri-kumkaran: This need to be addressed right?

diskann-utils/src/strided.rs
```diff
@@ -3,592 +3,671 @@
...
+/// ```text
+///            |<------ cstride ----->|
+///            |<-- ncols -->|
+///            +-------------+
+/// slice 0 -> | a0 a1 a2 a3 | a4 a5 a6     ^
+/// slice 1 -> | b0 b1 b2 b3 | b4 b5 b6     |
+/// slice 2 -> | c0 c1 c2 c3 | c4 c5 c6   nrows
+/// slice 3 -> | d0 d1 d2 d3 | d4 d5 d6     |
+/// slice 4 -> | e0 e1 e2 e3 | e4 e5 e6     |
+/// slicf 5 -> | f0 f1 f2 f3 | f4 f5 f6     v
```
suri-kumkaran: typo: "slicf" -> "slice"

diskann-utils/src/strided.rs
```diff
@@ -3,592 +3,671 @@
...
-    {
-        assert!(self.ncols() != 0);
-
-        let ncols = self.ncols();
-        let cstride = self.cstride();
-        self.data
-            .as_mut_slice()
-            .chunks_mut(cstride)
-            .map(move |i| &mut i[..ncols])
+    pub fn rows(&self) -> Rows<'_, T> {
```
suri-kumkaran, line 323: Would it be better for `rows()` to return `Rows<'a, T>` rather than `Rows<'_, T>`?

Elision binds `'_` to the `&self` borrow, so rows can't outlive the handle. But
`Rows` only stores a raw pointer and three integers, and `Rows::new` already takes `Strided<'a, T>` and returns `Rows<'a, T>`, so we have the lifetime, we just drop it at the boundary.

Where it bites is returning rows from a helper:

```rust
let strided = Strided::try_from_data(data, n, c, s).unwrap();
strided.rows()  // E0515, though the data lives for 'a
```

Same question for `row`, `row_or_panic`, `row_unchecked` and the `element` family returning `&'a [T]` / `&'a T`. Receivers can stay `&self`.

suri-kumkaran (approved):

harsha-simhadri (approved):

diskann-utils/src/strided.rs
hildebrandmw: Fixed - thanks.

diskann-utils/src/strided.rs
hildebrandmw: Thanks!

diskann-utils/src/strided.rs
hildebrandmw: I'm going to keep them with the standard elision bound for now. This type isn't used that heavily - going to `'a` in the future can be done if it becomes necessary. Ideally, the interface was designed to be behind a trait (for implementation by both reference and mutable reference variants). If implemented by a `StridedMut`, we cannot soundly give out a `'a` is sketchier.

suri-kumkaran (approved):

PR #1388, opened 2026-09-09
[disk] Remove the owned runtime and synchronous DiskIndexSearcher API

### LINKED ISSUE #1330: Remove the owned Tokio runtime and synchronous  `DiskIndexSearcher`  API
`DiskIndexSearcher`  currently owns a Tokio runtime to provide a synchronous wrapper around its asynchronous search implementation. Internal consumers no longer require the synchronous API, so retaining the runtime adds unnecessary ownership and lifecycle complexity.

Proposed change
- Remove the owned Tokio runtime from  DiskIndexSearcher .
- Remove the synchronous search wrapper.
- Keep the asynchronous search API as the supported interface.

wuw92:
- [ ] Is the change to the API backwards compatible? No; callers must remove the runtime constructor argument and await searches or drive them with their own executor.

#### Reference Issues/PRs

Fixes #1330.

Remove the owned Tokio runtime and synchronous search interface from `DiskIndexSearcher`.

- **Runtime ownership:** Construction no longer accepts or creates a runtime. Both `search` and `search_with_indexed_vectors` are asynchronous, and all search modes await the underlying operations directly. Synchronous benchmark execution manages its runtime at the caller.
- **Internal structure:** Consolidate search dispatch into one private helper. Builder ground-truth checks and search behavior tests exercise the public interface; filtered-search tests no longer repeat the same assertions through the internal helper.

Search modes, result layouts, filtering, and parameter validation remain unchanged.

#### Any other comments?

None.

### DIFF
--- a/diskann-disk/src/search/provider/disk_provider.rs
+++ b/diskann-disk/src/search/provider/disk_provider.rs
@@ -1228,118 +1228,124 @@ where
 
     /// Perform a search on the disk index.
     /// return the list of nearest neighbors and associated data.
-    pub async fn search(
-        &self,
-        query: &[Data::VectorDataType],
+    pub fn search<'a>(
+        &'a self,
+        query: &'a [Data::VectorDataType],
         return_list_size: u32,
         search_list_size: u32,
         beam_width: Option<usize>,
-        mode: SearchMode<'_>,
-    ) -> ANNResult<SearchResult<Data::AssociatedDataType>> {
-        let mut query_stats = QueryStatistics::default();
-        let mut indices = vec![0u32; return_list_size as usize];
-        let mut distances = vec![0f32; return_list_size as usize];
-        let mut associated_data =
-            vec![Data::AssociatedDataType::default(); return_list_size as usize];
+        mode: SearchMode<'a>,
+    ) -> impl SendFuture<ANNResult<SearchResult<Data::AssociatedDataType>>> + 'a {
+        async move {
+            let mut query_stats = QueryStatistics::default();
+            let mut indices = vec![0u32; return_list_size as usize];
+            let mut distances = vec![0f32; return_list_size as usize];
+            let mut associated_data =
+                vec![Data::AssociatedDataType::default(); return_list_size as usize];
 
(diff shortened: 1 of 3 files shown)


diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -1938,26 +1904,20 @@ mod disk_provider_tests {
             .map(|path| load_source_data(params.storage_provider, path));
 
         let pool = create_thread_pool(params.thread_num.into_usize()).unwrap();
+        let runtime = tokio::runtime::Builder::new_current_thread()
+            .build()
+            .unwrap();
```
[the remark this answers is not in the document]
wuw92: [Tokio documents that `Runtime::block_on` can be called concurrently from multiple threads](https://docs.rs/tokio/1.53.1/tokio/runtime/struct.Runtime.html#method.block_on), even with a current-thread runtime. Each query future runs on its calling Rayon worker, so sharing the runtime does not serialize these queries.

diskann-disk/src/search/provider/disk_provider.rs
```diff
@@ -1238,7 +1228,7 @@ where
 
     /// Perform a search on the disk index.
     /// return the list of nearest neighbors and associated data.
-    pub fn search(
+    pub async fn search(
```
hildebrandmw: One thing to be careful of is that making large functions `async` can cause the compiler to lose [`Send`](https://emschwartz.me/pinning-down-future-is-not-send-errors/) bounds. This may not be a concern for the disk index since it always works synchronously, but is a footgun to be aware of.

partychen (approved):

diskann-disk/src/search/provider/disk_provider.rs
wuw92: Thanks for calling this out. Both `search` and `search_with_indexed_vectors` now explicitly return `impl SendFuture<...> + 'a`, so the compiler checks the `Send` guarantee at the public API boundary.

PR #1389, opened 2026-09-10
Fix inplace deletion for providers that reclaim vector data

### LINKED ISSUE #1153: inplace_delete shouldn't access any data associated with a vector after `DataProvider::delete()` is called
The current semantics of delete/release aren't really used. Instead, only delete is ever called, so DataProvider implementation implement this as a hard delete. However, inplace_delete calls `on_neighbors` for the deleted ID, which means DataProviders must leave the neighbor data intact.

We should instead treat delete as removing all data associated with the ID, and inplace_delete should copy any data it needs before the delete so it doesn't access them after.

xwj-ox:

#### Reference Issues/PRs

Fixes #1153.

`inplace_delete` previously read and cleared a target's adjacency after calling the provider's delete method, so providers could not immediately reclaim all of its data. Prepare candidate searches and repair edges while the targets are still accessible, then clear target adjacency before deletion and repair only surviving nodes. Batched deletion waits for all preparation tasks, excludes the whole batch before candidate truncation, and deduplicates IDs within each chunk.

If a provider rejects deletion while the target remains valid, restore its adjacency. Finish repairs for completed deletions before returning a later batch error. Enable Garnet neighbor-record cleanup, including the case where graph insertion has not yet created a neighbor record, and clarify the provider deletion contract. Public signatures and dependencies are unchanged.

#### Any other comments?

Validation on Windows with the repository-pinned Rust 1.97.1:

- `cargo fmt --all --check` passed.
- `cargo clippy --workspace --all-targets -- -D warnings` passed.
- Coverage-instrumented library suites passed: 331 `diskann`, 48 `diskann-garnet`, and 89 `diskann-inmem` tests (468 total).
- Local LCOV reports 404/415 changed executable lines covered (97.35%); core deletion implementation coverage is 121/131 changed executable lines (92.37%). CI remains authoritative for Codecov.
- New regressions cover all three deletion methods, physical vector/adjacency removal, real multi-ID batches, duplicate IDs, limited replacement budgets, pruning, rejected deletions, partial batch failures, and Garnet search/reinsertion after deletion. The initial hard-delete regressions fail on the original implementation.

Batch deletion remains non-atomic when a provider operation fails.

### DIFF
--- a/diskann-garnet/src/ffi_tests.rs
+++ b/diskann-garnet/src/ffi_tests.rs
@@ -176,6 +176,51 @@ mod tests {
     }
 
     #[test]
+    fn remove_reclaims_data_and_preserves_graph() {
+        let store = Store::new();
+        let (index_ptr, ctx) = create_test_index(&store, VectorQuantType::NoQuant);
+
+        for (id, vector) in [(10, [1.0, 0.0]), (20, [0.0, 1.0]), (30, [1.0, 1.0])] {
+            assert_eq!(
+                insert_f32_vector(&ctx, index_ptr, id, &vector),
+                InsertResult::Success
+            );
+        }
+
+        let id = 20u32;
+        let id_bytes = bytemuck::bytes_of(&id);
+        let iid_bytes = store
+            .get(ctx.term(Term::IntMap).get(), id_bytes)
+            .expect("missing internal ID mapping");
+        assert!(
+            store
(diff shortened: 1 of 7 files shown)


hildebrandmw:
Thanks for your interest - but please hold off. A longer-term fix for this is in the works with inmem2 and the deprecation of soft-deletes entirely. This particular issue is not very pressing, and I'd rather we fix it properly and wholistically.

PR #1394, opened 2026-09-12
[multi vector] Add tiled MinMax MaxSim kernels

### LINKED ISSUE #990: Add MicroKernels to support f32, f16, u8, i8 and MinMax quantizations.
# Add MicroKernels for required architecture-type/quantization pairs (Blanket Issue)

## Description
Implementation of the hardware-specific microkernels for multi-vector reductions.

## Tasks
- [ ] Implement microkernel for different types and quantizations for multi-vector distances (Chamfer/MaxSim).
- [ ] Ensure the microkernel logic is decoupled enough to be tested via the standalone API.
- [ ] *Blanket Task:* Add subsequent sub-issues for ARM NEON, AVX-512, etc., as needed.

partychen:
## Summary

Adds prepared, tiled MaxSim kernels for MinMax8 queries against packed MinMax4 documents, with Scalar, x86-64 V3/V4, and ARM64 Neon implementations.

Query preparation packs the query vectors and their compensation metadata once. Each computation borrows canonical MinMax4 documents, decodes document tiles into a shared byte layout, and computes one MaxSim distance per query vector.

### Query storage

Queries use `BlockTransposed<u8, MR, PACK>` storage. `MR` is the number of query rows in a panel, and `PACK` is the number of consecutive bytes from each query consumed by one integer dot-product step. The packed views and panels carry `PACK` explicitly.

`EvenOdd64Layout` defines the dimension order shared by query preparation and document decoding. Within each block of 64 dimensions, even dimensions precede odd dimensions:

```text
[d0, d2, d4, ..., d62, d1, d3, d5, ..., d63]
```

The original dimension `D` is rounded up to a padded dimension `K`, a multiple of 64. Dimension padding and unused query rows contain zero. Zero-dimensional inputs are handled separately. MinMax compensation uses the original dimension `D`.

`PACK` controls physical query grouping independently of the logical dimension permutation. The contraction length counts padded scalar dimensions, with `K / PACK` dot-product steps.

### Document decoding

The decoder consumes four-bit codes without canonical headers or compensation metadata. Each complete block contains 32 packed bytes and produces 64 byte-valued codes: the 32 low nibbles followed by the 32 high nibbles.

- **V3:** A 32-byte SIMD load followed by vector masks and shifts produces the two output halves.
- **V4:** Uses the AVX2-capable decoder and VNNI for integer dot products.
- **Neon:** Two 16-byte SIMD loads produce the same 64-byte output layout.
- **Scalar:** Implements the same layout with portable scalar operations.

Partial blocks use a bounded, zero-filled local buffer. When `D` is odd, the unused high nibble of the final packed byte is cleared. The decoder overwrites every output byte, including padding.

## Components and execution

| Component | Responsibility |
|---|---|
| `Prepared` | Owns packed query values, query compensation, and the selected architecture capability |
| `layout.rs` | Defines the permutation, padding, and controlled packed-query storage/view construction |
| `decode.rs` | Expands packed code bytes into the shared layout using the selected architecture |
| `reader.rs` | Adapts canonical MinMax4 rows, preserves metadata order, and invokes the decoder |
| `Driver` | Validates shapes, manages scratch and cache tiling, and coordinates traversal |
| `PanelKernel` / `MicroKernel` | Perform integer contraction, MinMax compensation, and MaxSim reduction |

The driver uses B-first traversal so decoded document data is reused across query tiles and panels:

```text
Visit B tile
  Read canonical rows and decode values into scratch
  Visit A cache subviews
    Visit A panels
      Visit decoded B panels and remainder
        Integer contraction -> compensation -> MaxSim reduction
```

Traversal uses the library's packed/unpacked views, panel visitors, remainder handling, byte-based cache-blocking parameters, and bounds-tracked pointers. Cache working-set estimates include query and document compensation metadata.

The driver receives a `PackedQueryView` and `MinMax4Rows`, which retain their layout and representation contracts. Integer contraction consumes byte panels and the padded dimension; compensation separately uses metadata and the original dimension. Register representations remain opaque behind `ExtraWide`.

Two per-call scratch buffers hold decoded B values and B metadata for one tile and are reused across the computation. Persistent query state remains read-only. Documents are not modified, and no full intermediate distance matrix is materialized.

## Architecture configurations

| ISA | MR x NR | PACK | Integer computation |
|---|---:|---:|---|
| Scalar | 8 x 6 | 4 | Portable u32 accumulation |
| x86-64 V3 | 16 x 8 | 4 | AVX2 `maddubs` and `madd` |
| x86-64 V4 | 16 x 8 | 8 | AVX-512 VNNI `dpbusd`, with adjacent accumulator lanes folded during reduction |
| ARM64 Neon + dotprod | 8 x 8 | 4 | `UDOT` |

Architecture-specific operations execute within their target-feature context. Logical query-row counts allow a backend to omit an unused upper half of a query panel.

## API

```rust
let kernel = build_minmax_max_sim(MaxSimIsa::Auto, query_minmax8, BoxErase)?;
let mut scores = vec![0.0; kernel.nrows()];
kernel.compute_max_sim(doc_minmax4, &mut scores)?;
```

The factory supports automatic architecture selection and explicit Scalar, V3, V4, and Neon requests. Unavailable requested architectures return `NotSupported`. `MaxSimIsa::Reference` is not supported by this factory.

### Behavior

- Produces one MaxSim distance per query vector and overwrites scores on each call.
- Rejects mismatched output lengths or document dimensions before modifying scores.
- Returns `f32::MAX` for empty documents and zero for nonempty zero-dimensional documents.
- Ignores NaN pairwise distances while preserving the best valid score.
- Keeps document scratch local to each call so prepared query state can be shared.
- Adds no dependencies; public entry points are additive.

## Test coverage

The PR includes focused tests for:

- Explicit-PACK views and panels, padded columns, group spans, subviews, and tails.
- Query permutation, controlled row writes, and row/dimension padding.
- Scalar and SIMD decoding, unaligned input, nonzero unused nibbles, output sentinels, and invalid buffer lengths.
- Canonical row/metadata alignment and decoded scratch extents.
- Integer contraction against independently constructed inputs and expected results.
- Register accumulation, compensation, cache-tile variations, and single/multiple query panels.
- Factory equivalence, shape errors, empty inputs, repeated computation, prepared-query ownership, and NaN handling.


hildebrandmw (commented):
There's a lot going on here. It does add a minmax8x4 kernel, but basically builds a second matrix-kernel decomposition architecture rather than extending the existing abstractions introduced in 1368.

The point of the machinery introduced there was to use enable better higher-level experimentation of cache tilings and ordering without fully relying on Miri or checked accesses for correctness debugging. All of these pieces could then be tested, reviewed, and reused independently.

Here, packing and traversal are all reimplemented manually with inconsistent bounds tracking applied. This is hard to test, validate, or reuse.

My suggestion would be to follow the design philosophy of 1368 and build the traversal mechanism out of lower-level parts (which we can then reuse for other quantization kernels). In addition, matrix kernels should not own any non-scratch state because that also makes them much more difficult to reuse.

I also have some other design concerns:

* The microkernel is currently doing a lot, and I feel that there has to be a more efficient unpacking than doing this much bit-twiddling in general purpose registers. One design worth considering incrementally unpacking B and reusing that for the whole A traversal (flipping the cache tiling order - which again is easier to do via paneled views). This can perhaps be coupled with a different permutation strategy of the A-side query for more efficiency.
* `ExtraWide` should as much as possible keep the associated types opaque to the caller (like `ExtraWide` in the f32 kernel). This makes it **significantly** easier to express architecture specific optimizations of coarser grained kernels without the profoundly heaver `SIMDVector` and such constraints.
* How many of the Miri exceptions are really needed? Miri can emulate a good number of Neon and AVX2 intrinsics.
* Directed testing of the various steps is pretty sparse. Especially given its ad-hoc mix of checked, unchecked, and `Bound`s based indexing. Using `Bound`s based checking in a more disciplined way at least provides higher confidence of integration-based tests.

Have you considered making the A-side packing even more aggressive, splitting into groups of even and odd indices? The idea there is that it would naturally fit the order of unpacked nibbles much better and avoid a lot of the interleaving logic. Basically, a lot the micro-kernels are concerned about restoring a dimension ordering that we anyways control.

partychen:
> There's a lot going on here. It does add a minmax8x4 kernel, but basically builds a second matrix-kernel decomposition architecture rather than extending the existing abstractions introduced in 1368.
> 
> The point of the machinery introduced there was to use enable better higher-level experimentation of cache tilings and ordering without fully relying on Miri or checked accesses for correctness debugging. All of these pieces could then be tested, reviewed, and reused independently.
> 
> Here, packing and traversal are all reimplemented manually with inconsistent bounds tracking applied. This is hard to test, validate, or reuse.
> 
> My suggestion would be to follow the design philosophy of 1368 and build the traversal mechanism out of lower-level parts (which we can then reuse for other quantization kernels). In addition, matrix kernels should not own any non-scratch state because that also makes them much more difficult to reuse.
> 
> I also have some other design concerns:
> 
> * The microkernel is currently doing a lot, and I feel that there has to be a more efficient unpacking than doing this much bit-twiddling in general purpose registers. One design worth considering incrementally unpacking B and reusing that for the whole A traversal (flipping the cache tiling order - which again is easier to do via paneled views). This can perhaps be coupled with a different permutation strategy of the A-side query for more efficiency.
> * `ExtraWide` should as much as possible keep the associated types opaque to the caller (like `ExtraWide` in the f32 kernel). This makes it **significantly** easier to express architecture specific optimizations of coarser grained kernels without the profoundly heaver `SIMDVector` and such constraints.
> * How many of the Miri exceptions are really needed? Miri can emulate a good number of Neon and AVX2 intrinsics.
> * Directed testing of the various steps is pretty sparse. Especially given its ad-hoc mix of checked, unchecked, and `Bound`s based indexing. Using `Bound`s based checking in a more disciplined way at least provides higher confidence of integration-based tests.
> 
> Have you considered making the A-side packing even more aggressive, splitting into groups of even and odd indices? The idea there is that it would naturally fit the order of unpacked nibbles much better and avoid a lot of the interleaving logic. Basically, a lot the micro-kernels are concerned about restoring a dimension ordering that we anyways control.

@hildebrandmw Thanks for the detailed review. I reworked the implementation around the abstractions from #1368 in `ed75906e`:

- There is now a single generic `Driver -> PanelKernel -> MicroKernel` traversal for Scalar, V3, V4, and Neon; the separate V4/direct traversal was removed.
- `Grouped<N>` owns both query packing and canonical MinMax4 group expansion. Scalar/V3/Neon use `Grouped<4>` with even/odd dimensions, while V4 uses contiguous `Grouped<8>` values.
- B remains in canonical packed MinMax4 form. The same micro-kernel pipeline expands only the current contraction group, so the matrix kernel has no persistent state or document scratch allocation.
- `ExtraWide` keeps its register types opaque; architecture backends only provide load/splat/dot/reduce behavior.
- The Miri exceptions are limited to operations Miri cannot execute natively: BMI2 expansion, the emulated V4 broadcast, and adjacent-lane reduction. The native V4 path and the Miri-emulated V4 test both pass.

I also added/ran directed packing, tail, bounds, driver/register, factory-equivalence, repeated-computation, empty-input, and NaN-compensation coverage. Native x86-64 V3/V4 tests, ARM64 cross-Clippy, and the V4 Miri test pass.

For the 1,000-document 16x16 workload on a Xeon Platinum 8370C, the latest release measurements are:

| Dim | Existing | V3 | V4 |
|---:|---:|---:|---:|
| 250 | 11.47 us/doc | 1.65 us/doc | 0.52 us/doc |
| 256 | 4.09 us/doc | 1.61 us/doc | 0.49 us/doc |

The 250-dimensional Existing path has a scalar remainder, so I am keeping the 250- and 256-dimensional results explicitly separate.

diskann-quantization/src/matrix_kernels/mod.rs
```diff
@@ -9,7 +9,7 @@
 //!
 //! * Proper run time cache-size detection.
 //! * GEMM-like kernels beyond "maxsim".
-//! * Quantization support.
+//! * Additional quantized kernels.
```
hildebrandmw: Why the change?

diskann-quantization/src/matrix_kernels/blocks/packed.rs
```diff
@@ -16,77 +16,103 @@ use crate::{
...
+/// Each block contains `SZ` bands of `k` padded scalar columns. Within a block,
+/// `PACK` adjacent elements of each band are contiguous before advancing to the next
+/// band. There are `k / PACK` groups of `SZ * PACK` elements, regardless of element type.
+/// See [`BlockTransposedRef`] for the physical layout. No block may be partially filled.
 ///
 /// # Class Invariants
 ///
 /// * The tracked length `ptr.len()` must be equal to `SZ * blocks * k`.
 /// * `SZ` may not be zero.
+/// * `PACK` may not be zero and must divide both `SZ` and `k`.
```
hildebrandmw: I don't think we can guarantee that `PACK` divides `k`. We should support any `k`.

diskann-quantization/src/matrix_kernels/blocks/packed.rs
```diff
@@ -16,77 +16,103 @@ use crate::{
...
         T: Copy,
     {
-        if SZ == 0 {
+        if SZ == 0 || PACK == 0 {
             return None;
         }
 
         let blocks = NonZeroUsize::new(v.num_blocks())?;
-        let k = DimK::new(NonZeroUsize::new(v.ncols())?);
+        let k = DimK::new(NonZeroUsize::new(v.padded_ncols())?);
```
hildebrandmw: Again, this should be `ncols` and we should correctly handle elsewhere, like in the micro-kernel loop or data unpacking stages.

diskann-quantization/src/matrix_kernels/blocks/packed.rs
```diff
@@ -268,15 +311,40 @@ impl<'a, T, const SZ: usize> Panel<'a, T, SZ> {
...
+    /// Return the span for dot-product group `group`.
+    ///
+    /// Group `g` covers columns `[g * PACK, (g + 1) * PACK)` of every band and therefore
+    /// occupies `SZ * PACK` contiguous elements starting at `group * SZ * PACK`.
+    ///
+    /// # Safety
+    ///
+    /// `group` must be strictly less than `k / PACK`, where `k` is the contraction
+    /// dimension tracked by [`Self::k`].
+    pub(in crate::matrix_kernels) unsafe fn group(&self, group: usize) -> Slice<'a, T> {
```
hildebrandmw: This still keeps the layout implicit. Maybe we could have something like
```rust
/// A fixed-size rowmajor view
struct Patch<'a, T, const ROWS: usize, const COLS: usize> {
    ptr: Slice<'a, T>,
}
```
which gets returned to keep the geometry explicit.

hildebrandmw: Can also add a quick
```rust
bounds::check_lt(group, self.k);
```
for a slightly better error message when out-of-bounds. The slice indexing will error already, but it will be less clear.

diskann-quantization/src/matrix_kernels/blocks/packed.rs
```diff
@@ -496,18 +597,122 @@ mod tests {
...
+                let view = View::<_, SZ, PACK>::from_block_transposed(matrix.as_view()).unwrap();
+
+                assert_eq!(view.blocks().get(), nrows.div_ceil(SZ));
+                assert_eq!(view.k().value(), ncols.next_multiple_of(PACK));
+
+                view.checked_visit_panels(|panel, block| {
+                    let packed = panel.checked_as_std_slice();
+                    for row in 0..SZ {
+                        for col in 0..ncols.next_multiple_of(PACK) {
+                            let index = (col / PACK) * SZ * PACK + row * PACK + col % PACK;
```
hildebrandmw: There really should be a logical to linear and linear to logical index conversion helper instead of repeating this logic all the time.

hildebrandmw (commented):
I got through `packed.rs`. I think there's a little we can do to make the `group` access a bit semantically tighter. I'm still working my way through the rest. Is there a concern about using the same blocking strategy across all architectures?

diskann-quantization/src/matrix_kernels/mod.rs
partychen: rollbacked

diskann-quantization/src/matrix_kernels/blocks/packed.rs
partychen: Agreed

diskann-quantization/src/matrix_kernels/blocks/packed.rs
partychen: Done

diskann-quantization/src/matrix_kernels/blocks/packed.rs
partychen: Added.  `Panel::group` now returns a `Patch<'a, T, SZ, PACK>`, an explicit `SZ x PACK` row-major view. `ExtraWide::load` takes the `Patch`, so it no longer needs to be `unsafe`.

partychen: Added

diskann-quantization/src/matrix_kernels/blocks/packed.rs
partychen: Done

partychen:
> I got through `packed.rs`. I think there's a little we can do to make the `group` access a bit semantically tighter. I'm still working my way through the rest. Is there a concern about using the same blocking strategy across all architectures?

Thanks! I addressed the `packed.rs` comments in 8dd1752e and babca7c2; replies are inline.

> Is there a concern about using the same blocking strategy across all architectures?

For V3, yes. I measured register and cache blocking on a Xeon Platinum 8370C (AVX-512) with release builds, across the multi-vector shapes from 16x16x256 to 64x1250x512. I didn't use `#[test]` timings: `Bound` tracking is enabled under `cfg(test)`, and it distorts the results.

- **V3:** 16x8 needs 16 accumulators, all 16 `ymm` registers, so it spills. d9d43795 switches V3 to 16x6, like the f32 V3 kernel. It is 4-9% faster for 16/64 queries at 250/256 dimensions. 16x4 was about the same as 16x6; 8x8 and 8x12 didn't help.
- **V4:** 16x8 is still the best overall. 16x12 is 3-16% slower on five of six shapes and 4% faster on the largest. 32x4 and 32x6 are about 1.7x slower at 16 queries, where half of each panel is padding. V4 stays at 16x8.
- **Cache blocking:** the shared byte-based parameters look fine for MinMax. The default L1 budget is within 2% of the best B-tile size on five of six shapes; not tiling B is much slower, up to 1.85x on V4. `a_panels_in_l2` has no measurable effect: the B-first driver visits all of A for each decoded B tile, so splitting A into L2-sized subviews doesn't change the access order.

PR #1395, opened 2026-09-14
Expose metric-aware quantizer centroid distance

cherryrhea:
## Summary

- expose a quantizer API that computes squared L2 distance from a candidate centroid to the training centroid
- align centroid-distance computation with the metric-specific transformation used by `SphericalQuantizer::preprocess`
- forward the API through the public `Quantizer` trait implementations
- validate dimensions and reject non-finite computed distances

## Motivation

LibVector needs to compare a current data centroid with the centroid persisted in a trained quantizer. Keeping this computation in `diskann-quantization` avoids reconstructing private quantizer state in an external FFI layer and preserves metric-specific preprocessing semantics, especially for cosine.

## Validation

- `cargo test -p diskann-quantization`
  - 412 unit and compile tests passed
  - 3 integration tests passed
  - 28 doc tests passed; 6 ignored
  - 0 failures
- centroid-distance invariants run through the existing L2, inner-product, and cosine test matrices
- finite-input overflow is rejected through output validation
- `cargo fmt --all -- --check` passes
- `git diff --check` passes


diskann-quantization/src/spherical/quantizer.rs
```diff
@@ -145,6 +154,56 @@ where
...
+        &self,
+        centroid: &[f32],
+    ) -> Result<f64, CentroidDistanceError> {
+        if centroid.len() != self.input_dim() {
+            return Err(CentroidDistanceError::DimensionMismatch {
+                expected: self.input_dim(),
+            });
+        }
+
+        if centroid.iter().any(|value| !value.is_finite()) {
```
hildebrandmw: Instead of checking the input like this, we can check the output instead. This will be more efficient (no need to loop multiple times over the input) and more correct (overflowed accumulators during the norm computation will be detected.

diskann-quantization/src/spherical/quantizer.rs
```diff
@@ -145,6 +154,56 @@ where
...
+            SupportedMetric::SquaredL2 | SupportedMetric::InnerProduct => {
+                self.pre_scale.into_inner()
+            }
+        };
+
+        if !correction.is_finite() {
+            return Err(CentroidDistanceError::NonFiniteValue);
+        }
+
+        let correction = f64::from(correction);
```
hildebrandmw: What's the motivation for using `f64` here? Are there observed accuracy issues with `f32`?

diskann-quantization/src/spherical/quantizer.rs
```diff
@@ -2534,6 +2593,63 @@ mod tests {
             "if this failed, the likely culprit is exceeding the value of the 16-bit correction terms"
         );
     }
+
+    #[test]
+    fn centroid_squared_distance_uses_metric_correction() {
```
hildebrandmw: I'm worried that a few spot-checks for behavior will miss the larger contract requirement that `preprocess` and `centroid_squared_distance` should agree.

I'd suggest hooking into the existing `test_l2`/`test_ip`/`test_cosine` functions and checking that `centroid_squared_distance` is within some small tolerance of `Preprocessed.shifted_norm`. This will provide more coverage (more tests cases) and ensure that if one path gets updated, the other needs to as well.

hildebrandmw (commented):
Thanks - a few maintainability requests on my end with respect to testing. It's okay that this is technically an API breaking change.

Also feel free to ignore Copilot's claims about divergence in the cosine correction. There appears to be an existing bug with pre-scales not being correctly rejected when `Cosine` is used. As long as the implementation follows `SphericalQuantizer::preprocess`, it's fine.

diskann-quantization/src/spherical/quantizer.rs
cherryrhea: Addressed in 2b5ffd21. The separate input scan is removed; the implementation now validates the final computed distance. This avoids an extra pass and catches non-finite results caused by intermediate overflow. I also added a finite-input overflow regression using f32::MAX.

diskann-quantization/src/spherical/quantizer.rs
cherryrhea: Addressed in 2b5ffd21. I did not have measured evidence requiring f64, so the method now follows the existing preprocessing numeric contract and computes/returns f32. The downstream C FFI can widen the result at its ABI boundary where it exposes a double.

diskann-quantization/src/spherical/quantizer.rs
cherryrhea: Addressed in 2b5ffd21. I removed the standalone metric spot test and added a shared invariant assertion to the existing test_l2, test_ip, and test_cosine helpers. Every exercised trial now checks centroid_squared_distance(v) against preprocess(v).shifted_norm squared within f32 tolerance, so future preprocessing changes must remain synchronized.

diskann-quantization/src/spherical/quantizer.rs
```diff
@@ -145,6 +154,52 @@ where
...
+    /// Compute the squared L2 distance from `centroid` to the training centroid.
+    ///
+    /// The result is expressed in the input vector's units. For cosine quantizers,
+    /// the candidate's norm replaces the otherwise irrelevant pre-scaling term,
+    /// matching the normalization applied during vector preprocessing.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error when `centroid` has the wrong dimension or the distance
+    /// computation produces a non-finite value.
```
hildebrandmw: Doc request:
```
Compute the effective squared L2 distance of `centroid` 
relative to the quantizer's internal shift.

Metric-specific adjustments are applied consistently with vector compression,
so this generally differs from the direct squared L2 distance between `centroid` and [`Self::shift`].

# Errors

Returns an error if `centroid.len()` is not equal to [`Self::input_dim`] or the computed distance is non-finite.
```

hildebrandmw (commented):
One request for documentation update, otherwise looks great to me. Thanks!

diskann-quantization/src/spherical/quantizer.rs
cherryrhea: Updated the Rustdoc as requested in f90a220e, including the effective-distance semantics, the distinction from direct distance to `Self::shift`, and the precise error conditions. Validation: `cargo fmt --all -- --check` passed; `cargo test -p diskann-quantization` passed (412 unit/compile tests, 3 integration tests, and 28 doc tests; 6 doc tests ignored).

hildebrandmw (approved):

PR #1396, opened 2026-09-14
[disk] Replace custom cached I/O with standard buffered readers and writers

wuw92:

#### Reference Issues/PRs

N/A.

Replace `CachedReader` and `CachedWriter` with `std::io::BufReader` and `BufWriter` in disk index construction, partitioning, and graph merging, removing the hand-maintained buffering implementation without adding dependencies.

- Preserve storage-provider abstraction and existing buffer capacities. A shared `open_buf_reader` helper caps reader capacity at file length, with a focused test for that policy. Buffer-size values use `usize`.
- Use `read_exact` and `write_all` for complete transfers. Reuse the disk-layout writer to rewind and fill its header instead of reopening the file; explicitly flush completed production outputs to propagate write errors.

#### Any other comments?

This removes the publicly re-exported `CachedReader`/`CachedWriter` types and changes `READ_WRITE_BLOCK_SIZE` from `u64` to `usize`, so downstream source compatibility is not preserved. Truncated reads now follow standard I/O error handling rather than the custom cached-reader EOF diagnostic; initial reads are lazy rather than performed during construction.


hildebrandmw (approved):
This looks like a great simplification to me! Thanks!

SeliMeli (approved):

PR #1398, opened 2026-09-14
Add Neon Squared L2 kernels for USlice2 and USlice4

pfoxARM:

- Adds U4 Squared L2 U2*U2 kernel and U4*U4 Squared L2 kernel
- Adds SIMDAbsDiff trait in diskann-wide with emulated paths and testing

#### Any other comments?
Tested for performance on C100 and C200 using `graph-index-build-sq`


diskann-quantization/src/bits/distances.rs
```diff
@@ -713,6 +713,246 @@ impl Target2<diskann_wide::arch::x86_64::V4, MathematicalResult<u32>, USlice<'_,
...
+                // from the offset `i`
+                let x_vec = unsafe { u8s::load_simd(arch, px_u8.add(i)) };
+                // SAFETY: y_vec meets the same condition as x_vec
+                // due to equal length and same offset
+                let y_vec = unsafe { u8s::load_simd(arch, py_u8.add(i)) };
+
+                // compute abs diff then dot product for lower 4 bits, result is stored as 32x4
+                let lower_x: u8s = x_vec & mask;
+                let lower_y: u8s = y_vec & mask;
+                let d = lower_x.abs_diff_simd(lower_y);
```
hildebrandmw: I'm not a big fan of introducing a whole `SIMDAbsDiff` trait for a single leaf intrinsic. Especially since I'm not fully convinced by the proposed signature of `SIMDAbsDiff`. It's not clear to me if this is the correct signature when the input types are signed, or if we want widening instead. And there's not a great x86 equivalent.

For now, please keep this as a local function like
```rust

#[inline(always)]
fn abs_diff(arch: Neon, lhs: u8s, rhs: u8s) -> u8s {
    if cfg!(miri) {
        let lhs = lhs.to_array();
        let rhs = rhs.to_array();
        u8s::from_array(
            arch,
            std::array::from_fn(|i| lhs[i].abs_diff(rhs[i])),
        )
    } else {
        u8s::from_underlying(arch, unsafe {
            std::arch::aarch64::vabdq_u8(
                lhs.to_underlying(),
                rhs.to_underlying(),
            )
        })
    }
}
```
Note that miri does not support this intrinsic, so we need to explicitly provide a fallback.

hildebrandmw (commented):
Thanks! The kernels themselves look fine, but I don't think the broadening of `diskann-wide`'s API is quite worth it. Please see the comment below.

diskann-quantization/src/bits/distances.rs
MustafaIdrisArm: Agreed, I think it's worth mentioning that we have also been looking at using absolute difference for some of the `distances/simd.rs` kernels in diskann-vector. But by no means do we depend on this implementation, and you've also raised a good point about the proposed signature (especially with signed inputs). For now we will happily use the inline function proposed to carry out the implementation of the kernels for this PR.

pfoxARM: Deleted the commits that added SIMDAbsDiff, kernels now use `vabdq_u8` directly

hildebrandmw: Hmm, yeah. I guess we don't have a really good place for putting like one-off intrinsics like this. I suppose we could add an inherent method to `Neon`:
```
impl Neon {
    ....
    fn vabdq_u8(self, x: u8x16, y: u8x16) -> u8x16 {
         if cfg!(miri) {
             // emulate
         } else {
             unsafe { vabdq_u8(x, y) }
         }
    }
}
```

This kind of approach is taken by libraries like `fearless_simd`. Testing seems like it could be a bit fiddly though.

diskann-quantization/src/bits/distances.rs
```diff
@@ -713,6 +713,308 @@ impl Target2<diskann_wide::arch::x86_64::V4, MathematicalResult<u32>, USlice<'_,
...
+                let y_vec = unsafe { u8s::load_simd_first(arch, py_u8.add(i), remaining_bytes) };
+
+                // compute abs diff then dot product for lower 4 bits, result is stored as 32x4
+                let lower_x: u8s = x_vec & mask;
+                let lower_y: u8s = y_vec & mask;
+                let d = u8s::from_underlying (
+                        arch,
+                        // SAFETY: the caller assumes target supports neon
+                        // instruction required by `vabdq_u8`.
+                        unsafe {  vabdq_u8(lower_x.to_underlying(), lower_y.to_underlying()) }
```
hildebrandmw: Thanks - one more ask. Can you move the intrinsic call into a helper method
```
#[inline(always)]
fn abs_diff(arch: Neon, lhs: u8s, rhs: u8s) -> u8s {
    if cfg!(miri) {
        let lhs = lhs.to_array();
        let rhs = rhs.to_array();
        u8s::from_array(
            arch,
            std::array::from_fn(|i| lhs[i].abs_diff(rhs[i])),
        )
    } else {
        u8s::from_underlying(arch, unsafe {
            std::arch::aarch64::vabdq_u8(
                lhs.to_underlying(),
                rhs.to_underlying(),
            )
        })
    }
}
```
At least for me, Miri is unable to emulate `vabdq_u8`, so we would lose Miri coverage of `diskann-quantization` without a fallback.

hildebrandmw (approved):

diskann-quantization/src/bits/distances.rs
pfoxARM: Yes I should use miri to check next time. I've written the helper in distances.rs hopefully it's not too ugly

metajack (approved):

Issue #1397, opened 2026-09-14
Reduce code review effort by automating mechanical checks


wuw92:
Inspired by #1318, I reviewed three months of historical DiskANN PR comments to identify rules that could be checked mechanically.

The goal is to reduce repetitive review effort by adding or improving CI/lint checks for suitable rules, so reviewers can focus on correctness and design. Some candidates may need a new check; others may only need better coverage from existing tools.

The checklist below tracks the candidate rules identified from that review.

- [ ] **Rustdoc links:** catch broken internal links in Rust documentation through CI. (#1400)
- [ ] **License headers:** catch missing or nonstandard license headers in Rust source files through CI. (#1401)
- [ ] **SAFETY comments:** consistent documentation-presence checks for unsafe code.
- [ ] **Dependency versions:** avoid unnecessary duplicate versions.

<details>
<summary>Supporting references</summary>

- **Rustdoc links:** [maintainer request](https://github.com/microsoft/DiskANN/pull/1318#discussion_r3770946930).
- **License headers:** [maintainer's scope clarification](https://github.com/microsoft/DiskANN/pull/1318#discussion_r3759006941), [related discussion](https://github.com/microsoft/DiskANN/pull/1338#discussion_r3787153897).
- **SAFETY comments:** [discussion](https://github.com/microsoft/DiskANN/pull/1318#discussion_r3761640292), [presence versus correctness](https://github.com/microsoft/DiskANN/pull/1315#discussion_r3730664750).
- **Dependency versions:** [maintainer request](https://github.com/microsoft/DiskANN/pull/1205#discussion_r3492804509), [transitive-dependency discussion](https://github.com/microsoft/DiskANN/pull/1205#discussion_r3492820848), [follow-up](https://github.com/microsoft/DiskANN/pull/1205#discussion_r3495015568).

</details>

PR #1401, opened 2026-09-15
[CI] Check Rust license headers

wuw92:
- [ ] Does this PR modify any existing APIs? No.
- [x] Is the change to the API backwards compatible? No API changes.
- [x] Should this result in any changes to our documentation, either updating existing docs or adding new ones? Updated the agent guidance.

#### Reference Issues/PRs

Refs #1397

I standardize Rust license headers using `.github/LICENSE_HEADER.txt` and add a read-only check to the `repo-policy` CI job, required by `basics` and `CI gate`. The checker requires an exact template prefix on all tracked and non-ignored untracked `.rs` files, including existing files. Other file extensions are outside its scope.

I also normalize existing Rust headers and point the agent guidance to the shared template. Manual `python .github/scripts/check_license_headers.py fix` prepends the template to nonmatching files without removing existing content, so it can leave an older header below the new one. CI only runs `check`.

#### Any other comments?

N/A


.github/scripts/check_license_headers.py
```diff
@@ -0,0 +1,56 @@
...
+
+import argparse
+import os
+from pathlib import Path
+import subprocess
+import sys
+
+
+ROOT = Path(__file__).resolve().parents[2]
+TEMPLATE = ROOT / ".github" / "LICENSE_HEADER.txt"
```
hildebrandmw: I have a preference for passing root as a command-line argument (maybe defaulting to "." when not given). Just to decouple us a little from the exact directory structure.

.github/scripts/check_license_headers.py
```diff
@@ -0,0 +1,56 @@
...
+            raise ValueError(f"Empty template: {TEMPLATE}")
+        result = subprocess.run(
+            ["git", "ls-files", "-z", "--cached", "--others", "--exclude-standard", "--", "*.rs"],
+            cwd=ROOT, capture_output=True, check=True,
+        )
+        paths = sorted(set(result.stdout.split(b"\0")) - {b""})
+        missing = 0
+        for name in paths:
+            path = ROOT / os.fsdecode(name)
+            content = path.read_bytes()
```
hildebrandmw: We can get a small perf improvement by just reading the first `len(header)` lines.

hildebrandmw (approved):
This is a great quality of life improvement. Thanks!

.github/scripts/check_license_headers.py
wuw92: Thanks, I added `--root`, defaulting to the current directory.

.github/scripts/check_license_headers.py
wuw92: I updated the check to read only `len(header)` bytes. The full file is read only when `fix` needs to prepend the template, preserving the existing content.

.github/scripts/check_license_headers.py
```diff
@@ -0,0 +1,56 @@
...
+            if content.startswith(header):
```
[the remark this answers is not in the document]
wuw92: There are currently no tracked symlinks in repo.

randybird (approved):

PR #1408, opened 2026-09-16
Adds aarch64 prefetching into diskann-vector and diskann-inmem

pfoxARM:
#### What does this implement/fix?
Adds prefetching into diskann-vector and diskann-inmem for aarch64 using `prfm pldl1keep, [{ptr}]`

#### Any other comments?
I did test both `prfm pldl1keep, [{ptr}]` and `prfm pldl1strm, [{ptr}]` (Streaming or non-temporal L1 prefetch) on C100 VM and saw l1keep was marginally better. Although it'd be nice to compare on C200
But `l1keep` is also closer to the other architecture paths that already exist.

More info on `prfm` here https://support.arm.com/documentation/ddi0602/2024-06/Base-Instructions/PRFM--immediate---Prefetch-memory--immediate--?lang=en


diskann-inmem/src/prefetch.rs
```diff
@@ -202,7 +202,41 @@ pub(crate) unsafe fn prefetch(ptr: *const u8, len: usize) {
 /// # Safety
 ///
 /// The memory range `[ptr, ptr.add(len))` must be valid.
-#[cfg(not(all(target_arch = "x86_64", target_feature = "avx2")))]
+#[cfg(target_arch = "aarch64")]
+#[inline(always)]
+pub(crate) unsafe fn prefetch(ptr: *const u8, len: usize) {
```
hildebrandmw, line 207: I would think these would be a bit easier to maintain if we did a decomposition like:

```rust
#[cfg(arch = "aarch64")]
unsafe fn prefetch_l1(ptr: *const i8) {
    // aarch64 stuff
}

#[cfg(arch = "x86_64")]
unsafe fn prefetch_l1(ptr: *const i8) {
    // x86-stuff
}

#[cfg(any(
    all(target_arch = "x86_64", target_feature = "avx2"),
    target_arch = "aarch64",
))]
#[inline(always)]
pub(crate) unsafe fn prefetch(ptr: *const u8, len: usize) {
    use std::arch::x86_64::*;

    // Fetch the final location (the one containing the tag) first.
    let stride = Bytes::CACHELINE.value();
    let ptr = ptr.cast::<i8>();
    let lines = len.div_ceil(stride);
    if lines == 0 {
        return;
    }

    // SAFETY: Inherited from caller.
    unsafe { prefetch_l1(ptr.add(stride * (lines - 1))) };
    for i in 0..(lines - 1) {
        // SAFETY: Inherited from caller.
        unsafe {
            prefetch_l1(ptr.add(stride * i));
        }
    }
}
```

Then the fiddly bits don't need to be duplicated across architectures.

hildebrandmw (commented):
Thanks - one small maintenance ask on my side.

Did you observe much of a performance difference when prefetching was enabled?

pfoxARM:
> Thanks - one small maintenance ask on my side.
> 
> Did you observe much of a performance difference when prefetching was enabled?

Using C100 and C200 16-thread VMs and looking at 8-thread results:

Comparing QPS diff against main = cdaf3b466a6dc9292ce (16th September)

  - C100, exhaustive u8x2:
  -- l1keep -0.31%, l2keep -0.25%.
  - C100, exhaustive u8x4:
  --  l1keep +0.67% , l2keep -0.77% .
  - C100, spherical u2: 
  -- l1keep +60.21%, l2keep +72.88%.
  - C100, spherical u4: 
  -- l1keep +23.34%, l2keep +20.69%.

  - C200, exhaustive u8x2: 
  -- l1keep -7.02%, l2keep +1.92%.
  - C200, exhaustive u8x4: 
  -- l1keep +6.93%, l2keep +5.13%.
  - C200, graph u2: 
  -- l1keep +45.62%, l2keep +50.98%.
  - C200, graph u4: 
  -- l1keep +32.91%, l2keep +9.29%.
 
I also looked at Keep vs Streamed (non-temporal) prefetch for L1 and L2 on int8 * int8, but saw mixed marginal gains/loss on C100 and C200.

I'll keep looking into this for now, check out cache PMU metrics to see how aggressive it's being.

PR #1420, opened 2026-09-18
Matrix overhaul 2 of N - `Matrix` invariants

hildebrandmw:
Currently, `MatrixBase` and friend do not enforce two very important constraints:

* `nrows * ncols` does not overflow :sad:
* `nrows * ncols * std::mem::size_of::<T>()` does not exceed Rust's addressable limit of `isize::MAX`.

Generally speaking, we don't hit these kinds of limits, but constructing Matrices from unvalidated inputs can easily overflow. This PR does the following:

1. Introduces `diskann_utils::views::Layout<T>`, which contains `nrows` and `ncols` and validates the above constraints.
2. Replaces the `nrows` and `ncols` fields in `MatrixBase` with `Layout`.
3. Changes `Matrix::new` to panic if creation of a valid `Layout` fails. This is the least disruptive change to make since `Matrix::new` is called around 287 times across DiskANN.
4. Adds two new constructors for matrix
  ```rust
  /// Attempt to create a new matrix, returning an error if a `Layout` cannot be constructed.
  Matrix::try_new<U>(generator: U, nrows: usize, ncols usize) -> Result<Self, LayoutError>;
  
  /// Create a new matrix with the given layout.
  Matrix::new_with_layout<U>(generator: U, layout: Layout) -> Self
  ```
  These provide fallible and "infallible" alternatives to the panicking constructor. Do note that `new_with_layout` is not truly infallible since allocation *can* still fail.
5. Updates `MatrixBase::map` to panic if the layout for the new matrix fails the allocation size requirement.
6. Adds `Matrixbase::try_map` as a fallible version of `MatrixBase::map`.

### Panicking Constructor
I don't love that `Matrix::new` can panic, but the other alternative is a very noisy update of the numerous callsites and the introduction of `Result` in functions that currently don't return `Result`s. This is similar to how, for example, `Vec` works when provided a too-large allocation size. So I think it is the correct call to keep the most common use the least visually disruptive.

### Additional Type Parameter on `MatrixBase`
It's desirable to have `MatrixView` be covariant with respect to its borrow. Unfortunately, projecting through the `DenseData::Elem` associated type for `Layout<D::Elem>` causes all of matrix view to be covariant with-respect to `&[T]`. Having a second generic parameter fixes the issue.

As mentioned in #1415, the end-point of this overhaul is moving `Matrix` and `MatrixMut` behind traits, at which point the variance won't be an issue. So consider this a short-term workaround.


harsha-simhadri (approved):

diskann-utils/src/views.rs
```diff
@@ -562,10 +782,15 @@ where
...
-                ncols: self.ncols(),
+            // SAFETY: The successful checked index into `self.as_slice()` attests that
+            // `rows.len()` is no greater than `self.nrows()`. So the `Layout` is valid.
+            //
+            // By construction, `data.len() == rows * self.ncols()`.
+            Some(unsafe {
+                MatrixView::from_data_unchecked(
+                    data,
+                    Layout::new_unchecked(rows.len(), self.ncols()),
+                )
```
suri-kumkaran, line 793: If `ncols` is zero, for example, with a `3 x 0` matrix, a `subview(100..200)` can still succeed. In that case, the safety argument that "`rows.len()` is no greater than `self.nrows()`" does not hold.

I think these issues mainly arise because `nrows` and `ncols` are allowed to be zero.

diskann-utils/src/views.rs
```diff
@@ -91,76 +92,192 @@ unsafe impl<T> MutDenseData for Box<[T]> {
...
-pub struct MatrixBase<T>
-where
-    T: DenseData,
-{
-    data: T,
+/// This type guarantees the following invariants:
+///
+/// * `self.nrows() * self.ncols()` does not exceed `usize::MAX`.
+/// * `self.nrows() * self.ncols() * std::mem::size_of::<T>()` does not exceed `isize::MAX`.
+#[derive(Debug, PartialEq, Eq)]
```
suri-kumkaran, line 105: Does it make sense to implement these manually to avoid the implicit `T: Debug` bound? The same question applies to `PartialEq` and `Eq`. Should we do something similar for `Strided` as well?

suri-kumkaran (approved):

diskann-utils/src/views.rs
```diff
@@ -91,76 +92,192 @@ unsafe impl<T> MutDenseData for Box<[T]> {
...
+pub struct Layout<T> {
     nrows: usize,
     ncols: usize,
```
suri-kumkaran, line 108: Do we plan to make these `NonZeroUsize` in a future PR to avoid having to handle a lot of unnecessary corner cases?

PR #1429, opened 2026-09-21
i8 x i8 MaxSim Kernels

suri-kumkaran:
Integer MaxSim matrix kernels for `i8 x i8`, alongside the existing f32 and f16 paths.
Scores stay in `i32` end to end instead of routing through f32, so results are exact.

## Status: work in progress

Not furnished for review or merge yet. Up early for visibility on the design.

* **V4 is work in progress** and currently retargets to V3. VNNI only exposes `u8 x i8` and `i8 x u8` (`vpdpbusd`), so a native signed by signed kernel needs the offset by 128 trick with a correction term.
* Scalar is slower than the reference. Its `ExtraWide<8, 2>` impl widens to `i32` before multiplying, which discards the packed layout.
* Factory tests only build `MaxSimIsa::Auto`, so the Scalar path is not covered end to end. Fix is to loop the shared helpers over available ISAs, which also helps f32 and f16.

## What is here

`BlockTransposed<T, GROUP, PACK>` gains a `PACK` parameter so contraction elements can be interleaved to match the width of the target dot product instruction.

| ISA | MR | NR | PACK | Instruction |
| --- | -- | -- | ---- | ----------- |
| Scalar | 8 | 2 | 2 | emulated |
| V3 | 16 | 6 | 2 | `vpmaddwd` |
| V4 | 16 | 6 | 2 | retargets to V3, WIP |
| Neon | 8 | 6 | 4 | `sdot` |

PACK differs by target because the instruction does. `vpmaddwd` is a 2 way dot per 32 bit lane, `sdot` is 4 way.

Also adds `max_sim_kernel_i8` in `fallback.rs` as the oracle and `reference` benchmark ISA (products bounded by 128 * 128, so `i32` is safe for `k <= 131_071`), and `int8` as a selectable element type in `multi-vector-op`.

## Performance

V3 on an x86-64-v3 host, best of 13 interleaved runs, reference held at 1.000 as control: **2.22x geomean over reference**, ranging 2.25x to 2.61x across shapes. 
The one weak shape is 8 x 32 x 128 at 1.25x, too small to amortize packing.

Neon on an Apple M5 Pro, best of 15 interleaved runs, reference held at 1.000 as control: **1.75x geomean over reference**, ranging 1.66x to 1.88x across shapes.


diskann-quantization/src/matrix_kernels/test_util.rs
```diff
@@ -64,6 +64,12 @@ impl Distribution<f32> for TestDistr {
     }
 }
 
+impl Distribution<i8> for TestDistr {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> i8 {
+        rng.random_range(i8::MIN..=i8::MAX)
```
hildebrandmw, line 69: `Uniform {}.sample(rng)`?

diskann-quantization/src/matrix_kernels/blocks/packed.rs
```diff
@@ -268,12 +288,25 @@ impl<'a, T, const SZ: usize> Panel<'a, T, SZ> {
     pub(in crate::matrix_kernels) const fn k(&self) -> Bound {
         self.k
     }
+
+    /// Return the number of elements spanned by one physical row.
+    pub(in crate::matrix_kernels) const fn row_stride(&self) -> Elements<T> {
+        Elements::new(SZ * PACK)
```
hildebrandmw, line 294: This isn't a row stride, right?

diskann-quantization/src/matrix_kernels/maxsim/packed_i8_x_unpacked_i8.rs
```diff
@@ -0,0 +1,1145 @@
...
+                    SIMDVector::load_simd(self, slice.add(Elements::new(16)).as_ptr()),
+                ]
+            };
+
+            bytes.map(Self::Splat::from)
+        }
+
+        #[inline(always)]
+        fn splat(self, group: [i8; 2]) -> Self::Splat {
+            u32x8::<V3>::splat(self, i16_pair(group)).reinterpret_simd()
```
hildebrandmw, line 624: Unpacking like this seems a little slow. The core sequence is
```
        movsx eax, byte ptr [rcx + 4*r9]
        movzx eax, ax
        vmovd xmm14, eax
        vpbroadcastd ymm14, xmm14
        vpmaddwd ymm15, ymm13, ymm14
        vpaddd ymm3, ymm15, ymm3
        vpmaddwd ymm14, ymm12, ymm14
        vpaddd ymm2, ymm14, ymm2
        lea rax, [r9 + 4*r9]
```
and I'd mainly be worried about the dependency on the GPR loads.

What happens if a B-tile is eagerly expanded to i16s (like how the f16 path does eager conversion). This can all be done in SIMD, and then we don't need to unpack like this in the micro-kernel.

diskann-quantization/src/matrix_kernels/test_util.rs
suri-kumkaran: Done

diskann-quantization/src/matrix_kernels/blocks/packed.rs
suri-kumkaran: Sorry about the unclear naming. I have updated the terminology to use “pack” instead.

